PR #10, opened 2023-12-22
improved init by providing defaults

brenkao:


mirascope/cli/cli.py
```diff
@@ -54,7 +54,8 @@ def main():
 
     # Adding 'init' command
     parser_init = subparsers.add_parser("init", help="Initialize mirascope project")
```
willbakst, line 56: should we make the help `Initialize Mirascope project` i.e. capitalize, here and everywhere else?

mirascope/cli/commands.py
```diff
@@ -197,19 +197,23 @@ def init(args) -> None:
...
     Args:
         args: The command line arguments for the `init` command, containing:
-            - `directory_name`: The name of the directory to create.
+            - `--mirascope_location`: The root mirascope directory to create.
+            - `--prompts_location`: The user's prompts directory.
     """
     destination_dir = Path.cwd()
-    directory_name = args.directory_name
-    versions_directory = os.path.join(directory_name, "versions")
+    mirascope_location = args.mirascope_location or "mirascope"
```
willbakst, line 204: should these be constants?

willbakst (approved):
LGTM, two minor comments

mirascope/cli/cli.py
brenkao: in the context of library everything is lowercase i think?

mirascope/cli/commands.py
brenkao: they could be, but since its only used in this file and function i opted to not. i could set the default in cli.py instead

mirascope/cli/cli.py
willbakst: @wcbudz what are your thoughts here?

mirascope/cli/commands.py
willbakst: thinking again, do we need this `or` at all? since we're defaulting `args.mirascope_location` shouldn't this always have a value (with the default "mirascope")?

mirascope/cli/cli.py
wcbudz: I think when referring specifically to the library it's fine for lowercase because that's how people will type it in code.

PR #13, opened 2023-12-23
Initial cli example

willbakst:


cookbook/movie_bot/recommend_movies.py
```diff
@@ -0,0 +1,21 @@
...
+from prompts import MovieRecommendationPrompt
+
+from mirascope import OpenAIChat
+
+os.environ["OPENAI_API_KEY"] = "YOUR_API_KEY"
+
+
+def recommend_movies(genre: str):
+    """Run the movie recommendation script."""
+    model = OpenAIChat()
```
brenkao, line 13: do you need to add api_key here?

brenkao (approved):

cookbook/movie_bot/recommend_movies.py
willbakst: not if you set `os.environ["OPENAI_API_KEY"] = "YOUR_API_KEY"` which I put at the top

PR #17, opened 2024-01-03
Added type hint for messages when using @messages decorator

brenkao:


mirascope/prompts.py
```diff
@@ -14,6 +14,11 @@
 class Prompt(BaseModel):
     """A Pydantic model for prompts."""
 
+    @property
+    def messages(self) -> list[tuple[str, str]]:
```
willbakst, line 18: put this after __str__ and before save

willbakst (approved):

PR #24, opened 2024-01-20
Make using tools easy with OpenAITool and Function -> OpenAITool automatic conversion

willbakst:


brenkao:
tools link missing in mkdocs.yml

mirascope/chat/tools.py
```diff
@@ -24,7 +31,7 @@ def tool_schema(cls) -> ChatCompletionToolParam:
             raise ValueError("Tool must have a dosctring description.")
```
brenkao, line 31: typo dosctring

mirascope/chat/tools.py
```diff
@@ -42,3 +49,38 @@ def tool_schema(cls) -> ChatCompletionToolParam:
...
+    """
+
+    def decorator(cls: Type[T]) -> Type[T]:
+        """A decorator for adding a function to a tool class."""
+
+        def fn_property(self) -> Callable:
+            """Returns the function that the tool describes."""
+            return fn
+
+        setattr(cls, "fn", property(fn_property))
```
brenkao, line 83: do you need to define fn_property. could you pass fn to setattr instead?

mirascope/chat/utils.py
```diff
@@ -15,3 +20,63 @@ def get_openai_chat_messages(
...
+        if not docstring_param.description:
+            raise ValueError("All parameters must have a description.")
+
+        field_info = {"description": docstring.params[i].description}
+        if parameter.default != Parameter.empty:
+            field_info["default"] = parameter.default
+
+        field_definitions[parameter.name] = (
+            parameter.annotation,
+            FieldInfo(**field_info),  # type: ignore
```
brenkao, line 74: this ignore does not seem necessary?

brenkao, line 74: nit: can we define field info as FieldInfo passing in description and default instead of unpacking dict

brenkao (commented):
tools link missing in mkdocs.yml

mirascope/chat/tools.py
willbakst: You need to define self, but we can further simplify using a lambda

PR #28, opened 2024-01-23
feat: added async openai client

brenkao:


mirascope/chat/models.py
```diff
@@ -90,3 +90,83 @@ def stream(
 
         for chunk in completion_stream:
             yield OpenAIChatCompletionChunk(chunk=chunk)
+
+
+class AsyncOpenAIChat:
```
willbakst, line 95: I would refactor this into the following structure:

```
chat/models/
    __init__.py
    async_openai_chat.py
    openai_chat.py
    utils.py
```

Make sure to then update the `__init__` everywhere so imports still work as desired.

Then I would consider pulling the `if tools:` block into a utility function. If you don't think that's worthwhile, don't include the `utils.py` addition.

mirascope/chat/models.py
```diff
@@ -90,3 +90,83 @@ def stream(
...
+            prompt: The `Prompt` to use for the call.
+            **kwargs: Additional keyword arguments to pass to the API call. You can
+                find available keyword arguments here:
+                https://platform.openai.com/docs/api-reference/chat/create
+
+        Yields:
+            A `OpenAIChatCompletionChunk` for each chunk of the response.
+
+        Raises:
+            Re-raises any exceptions thrown by the openai chat completions create call.
```
willbakst, line 163: please update this and the same one for stream so that it makes it clear that the exception is thrown when iterating through the generator, not on the initial call to it.

willbakst (approved):
LGTM, left one refactor comment and one doc comment

mirascope/chat/models.py
brenkao: you mean update this and also the non async one?

mirascope/chat/models.py
brenkao: i'll hold off on adding utils for now

cookbook/api_examples/async_api_example.py
```diff
@@ -0,0 +1,43 @@
+"""An asynchronous FastAPI app integrated with a multi-chain prompt for recommending books on a topic
```
willbakst, line 1: line too long. Make sure it fits cleanly within the ruler with a period for the first line. Then include a newline between that and any additional text.

cookbook/extract_examples/async_extract_book_info.py
```diff
@@ -0,0 +1,31 @@
+"""An example script for extracting the title of a book and author given some string
```
willbakst, line 1: see earlier comment about single line fitting with final period.

cookbook/tools_examples/requirements.txt
```diff
@@ -0,0 +1 @@
+mirascope>=0.2.0,<1.0.0
```
willbakst, line 1: can you do this for other examples as well please?

docs/cookbook/fastapi.md
```diff
@@ -0,0 +1,7 @@
+# FastAPI
+
+!!! note
+
+    Take a look at the [code](https://github.com/Mirascope/mirascope/blob/main/cookbook/api_examples/async_create_example.py) in our repo.
```
willbakst, line 5: 1. should we rename `api_examples` to `fastapi_examples`
2. should we link the directory here? both files? or just async like you have?

docs/cookbook/rag.md
```diff
@@ -0,0 +1,7 @@
+# RAG (Retrieval-Augmented Generation)
+
+!!! note
+
+    Take a look at the [code](https://github.com/Mirascope/mirascope/blob/main/cookbook/rag_examples/local_embeddings.py) in our repo.
```
willbakst, line 5: same thing here, should we only link to local_embeddings.py or pinecone or the directory?

mirascope/chat/models/async_openai_chat.py
```diff
@@ -0,0 +1,160 @@
...
+            yield OpenAIChatCompletionChunk(chunk=chunk)
+
+    async def extract(
+        self,
+        schema: Type[BaseModelT],
+        prompt: Optional[Union[Prompt, str]] = None,
+        retries: int = 0,
+        **kwargs,
+    ) -> BaseModelT:
+        """Asynchronously extracts the given schema from the response of a chat `create` call.
```
willbakst, line 125: line too long

mirascope/chat/models/async_openai_chat.py
```diff
@@ -0,0 +1,160 @@
...
+                https://platform.openai.com/docs/api-reference/chat/create
+
+        Yields:
+            A `OpenAIChatCompletionChunk` for each chunk of the response.
+
+        Raises:
+            Re-raises any exceptions thrown by the openai chat completions
+            when iterating through the generator.
+        """
+        if not prompt:
```
willbakst, line 101: please add a todo here to refactor all of the setup into a `setup_openai_kwargs` function or something when we add tools to streaming (since at that point all four of create/acreate/stream/astream will have exactly duplicated code.

mirascope/chat/models/openai_chat.py
```diff
@@ -96,7 +95,8 @@ def stream(
             A `OpenAIChatCompletionChunk` for each chunk of the response.
 
         Raises:
-            Re-raises any exceptions thrown by the openai chat completions create call.
+            Re-raises any exceptions thrown by the openai chat completions
```
willbakst, line 98: these sections I'm realizing are missing the `ValueError` thrown if neither `prompt` nor `messages` is not provided.

I'm thinking it might be worth creating our own MirascopeOpenAIError (like we discussed before) that wraps any error bubbled up? That way we can say something like:

```python
"""
Raises:
    ValueError: ...
    MirascopeOpenAIError: raises any OpenAI errors
"""
```

tests/chat/models/test_openai_chat.py
```diff
@@ -109,7 +109,7 @@ def test_openai_chat_tools(
     side_effect=Exception("base exception"),
 )
 def test_openai_chat_error(mock_create, fixture_foobar_prompt):
-    """Tests that `OpenAIChat` handles openai errors thrown during __call__."""
+    """Tests that `OpenAIChat` handles OpenAI errors thrown during __call__."""
```
willbakst, line 112: please update `__call__` to `create` that's my bad

tests/chat/models/test_async_openai_chat.py
```diff
@@ -0,0 +1,129 @@
+"""Tests for mirascope async openai chat API model classes."""
```
willbakst, line 1: you're missing some tests around `Prompt` vs. `messages` and `str` prompts etc.

I have those tests for the openai_chat so reference that

cookbook/tools_examples/requirements.txt
brenkao: do you want me to raise the min version for examples that havent changed? or only the new ones we added

docs/cookbook/fastapi.md
brenkao: 1. yes
2. since this is placeholder, doesn't really matter. i think just one will suffice until we upgrade this section

mirascope/chat/models/openai_chat.py
brenkao: dont think we need a MirascopeOpenAIError, when we can achieve the same thing with calling it OpenAIError and linking to their docs

brenkao: I added this line to all the openai class docstirngs
```
```

PR #29, opened 2024-01-24
feat: OpenAIChat Extraction + String Prompts

willbakst:
Adding the following functionality:
- `OpenAIChat.extract(...)` method that extracts the information defined by a user created `BaseModel` schema from the provided prompt. It also comes with automatic retries (basic) such that it will attempt `retries` more times in the event of a `ValidationError` (i.e. 5 retries is 6 attempts including the first call).
- Enable `OpenAIChat.create(...)` and `OpenAIChat.extract(...)` to accept `str` prompts so the user is not forced into using our prompts if they don't want to.
- Fixing some misc bugs found with tools during implementation of `extract`
- Updating cookbook to include example code
- Updating docs to include a writeup of extraction in the LLM Convenience Wrappers concepts section.


mirascope/chat/utils.py
```diff
@@ -51,27 +54,41 @@ def convert_function_to_openai_tool(fn: Callable) -> Type[OpenAITool]:
...
+                    f"Function parameter name {parameter.name} does not match docstring "
+                    f"parameter name {docstring_param.arg_name}. Make sure that the "
+                    "parameter names match exactly."
+                )
+            if not docstring_param.description:
+                raise ValueError("All parameters must have a description.")
+            param_description = docstring_param.description
+
+        field_info_kwargs = {"annotation": hints[parameter.name]}
+        if parameter.default != Parameter.empty:
```
brenkao, line 78: Add some comments, why does default need to check empty but description can check falsy

willbakst: Done.

mirascope/chat/models.py
```diff
@@ -1,13 +1,23 @@
...
-from .utils import convert_function_to_openai_tool, get_openai_chat_messages
+from .utils import (
+    convert_base_model_to_openai_tool,
+    convert_function_to_openai_tool,
+    get_openai_chat_messages,
+)
+
+logger = logging.getLogger("mirascope")
+
+Schema = TypeVar("Schema", bound=BaseModel)
```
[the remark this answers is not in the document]
willbakst: I'm just going to call this BaseModelT and then the param will still be called schema

docs/concepts/llm_convenience_wrappers.md
```diff
@@ -118,6 +130,40 @@ chunk.delta    # original.choices[0].delta
...
+assert isinstance(book_info, BookInfo)
+assert book_info.model_dump() == {
+    "title": "The Name of the Wind",
+    "author": "Patrick Rothfuss",
+}
+```
+
+#### Retries
+
+Often you will want to retry your query in the event of a `ValidationError` when the model fails to properly populate the schema you've provided. Just set the number of retries and `extract` will automatically retry up to that many times (by default `retries` is `0`):
```
brenkao, line 157: when the model fails to properly populate the schema you've provided

to

when extract fails to convert the string response into a pydantic model

brenkao: Just (https://www.linkedin.com/pulse/stop-using-word-just-david-niebach-crm-mba-cpa-cga)

to

You can simply

mirascope/chat/models.py
```diff
@@ -20,7 +30,7 @@ def __init__(self, model: str = "gpt-3.5-turbo", api_key: Optional[str] = None):
 
     def create(
         self,
-        prompt: Prompt,
+        prompt: Union[Prompt, str],
```
brenkao, line 33: What if instead of prompt, they want to pass in messages directly through kwargs. so `prompt: Union[Prompt, str, None]` and then if prompt is None check if messages is inside kwargs

mirascope/chat/utils.py
```diff
@@ -51,27 +54,42 @@ def convert_function_to_openai_tool(fn: Callable) -> Type[OpenAITool]:
...
+            field_info_kwargs["default"] = parameter.default
+        if docstring_description:  # we check falsy here because this comes from docstr
+            field_info_kwargs["description"] = docstring_description
+
+        param_name = parameter.name
+        if "model_" in param_name:
+            param_name = "aliased_" + param_name
+            field_info_kwargs["alias"] = parameter.name
```
brenkao, line 86: Were there issues using `FieldInfo`? Could we instantiate a FieldInfo object like `field_info = FieldInfo()` and then do something like `field_info.alias = parameter.name`? I feel like if possible, it would be less error prone than using strings.

mirascope/chat/utils.py
```diff
@@ -51,27 +54,42 @@ def convert_function_to_openai_tool(fn: Callable) -> Type[OpenAITool]:
...
+            docstring_description = docstring_param.description
+
+        field_info_kwargs = {"annotation": hints[parameter.name]}
+        if parameter.default != Parameter.empty:
+            field_info_kwargs["default"] = parameter.default
+        if docstring_description:  # we check falsy here because this comes from docstr
+            field_info_kwargs["description"] = docstring_description
+
+        param_name = parameter.name
+        if "model_" in param_name:
```
brenkao, line 84: What is this if statement checking? Could we either use a descriptive variable name or short comment explaining this

mirascope/chat/utils.py
```diff
@@ -51,27 +54,42 @@ def convert_function_to_openai_tool(fn: Callable) -> Type[OpenAITool]:
...
+            param_name = "aliased_" + param_name
+            field_info_kwargs["alias"] = parameter.name
+            field_info_kwargs["validation_alias"] = parameter.name
+            field_info_kwargs["serialization_alias"] = parameter.name
+
+        field_definitions[param_name] = (
+            hints[parameter.name],
+            FieldInfo(**field_info_kwargs),
```
brenkao, line 92: see above, if we make a typo via strings, we would get pydantic validation errors here

mirascope/chat/utils.py
```diff
@@ -80,3 +98,22 @@ def convert_function_to_openai_tool(fn: Callable) -> Type[OpenAITool]:
...
+        f"An `{schema.__name__}` instance with all correctly typed parameters "
+        "extracted from the completion. Must include required parameters and may "
+        "exclude optional parameters unless present in the text."
+    )
+    field_definitions = {
+        field_name: (field_info.annotation, field_info)
+        for field_name, field_info in schema.model_fields.items()
+    }
+    return create_model(
+        schema.__name__ + "Tool",
```
brenkao, line 115: nit: use f string

docs/concepts/llm_convenience_wrappers.md
willbakst: Hilarious that the author uses "just" in the first few sentences haha

willbakst: Done.

mirascope/chat/models.py
```diff
@@ -90,3 +103,40 @@ def stream(
...
+        completion = self.create(
+            prompt,
+            tools=[tool],
+            tool_choice={
+                "type": "function",
+                "function": {"name": tool.__name__},
+            },
+        )
+
+        try:
```
brenkao, line 133: BONUS: consider extracting (pun intended) these to a helper function so its easier to implement for async extract

mirascope/chat/models.py
willbakst: Good call. Done.

mirascope/chat/utils.py
willbakst: `model_` is a reserved namespace in Pydantic 2.0, so we have to alias anything that has `model_` as the beginning. This is actually a good catch because it's not going it quite right. Will update and add explanatory comment.

mirascope/chat/utils.py
willbakst: Done.

mirascope/chat/utils.py
willbakst: Good call. This works.

mirascope/chat/models.py
willbakst: I'll punt the refactor here to when implementing async extract

brenkao (approved):

PR #32, opened 2024-01-25
Rag markdowns

brianpark0126:


docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+
+In reality, we could have passed the entirety of `soccer_info` into the prompt due to its small size. Let's take a look at a practical, scaled up example using the vector database Pinecone to manage the larger data.
+
+Our dataset will be a collection of news articles from 2004 pulled from this [public Github database.](https://raw.githubusercontent.com/Dawit-1621/BBC-News-Classification/main/Data/BBC%20News%20Test.csv) We will pick a topic, then ask our chat model to give a summary of 4 relevant articles.
+
+### Dataset/Preprocessing
+
+Chat and embedding models have token limits on a single call - this wasn't an issue in the soccer example, but these constraints come into play with much larger texts and datasets. For the purposes of this recipe, we split the large texts into roughly equal sizes using helper function `split_text()`. Since we want to be able to fit multiple articles into our final prompt, let's set `max_tokens=1000` and split the text into how many thousands of tokens the original text comprises of.
```
willbakst, line 99: explain why 1000 is a good number if we want to fit multiple articles. also that last sentence reads a little weird. Maybe just "split the text into thousand token chunks" would suffice?

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+
+    for i in range(num_splits):
+        start = i * len(words) // num_splits
+        end = (i + 1) * len(words) // num_splits
+        split_texts.append(" ".join(words[start:end]))
+
+    return split_texts
+```
+
+```python
```
willbakst, line 130: instead of having two chunks of code back-to-back, you should split these up with text explaining each code chunk itself.

willbakst (commented):
a few minor comments, but generally looks great!

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
+# Retrieval Augmented Generation (RAG)
+
+When you need to provide additional information to your model that it hasn't been pretrained on, we retrieve the relevant information from an external source and provide it as context to our model.
```
brenkao, line 3: rather than summarize what rag is, i think this section should be a short one or two sentences about Mirascope RAG.

terrible e.g. For this recipe, we will demonstrate how easy it is to integrate RAG into Mirascope prompts.

thoughts? @willbakst @brianpark0126

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+embeddings_model = "text-embedding-ada-002"
+
+embeddings = [
+    chat.client.embeddings.create(model=embeddings_model, input=[text])
+    .data[0]
+    .embedding
+    for text in soccer_info
+]
+
+df = pd.DataFrame({"texts": soccer_info, "embeddings": embeddings})
```
brenkao, line 41: it would be useful for every step of the way to show the results of each block. this helps the user follow the flow

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+```python
+soccer_info = [
+    "Manchester City won the league, continuing their domestic dominance.",
+    "In La Liga, FC Barcelona emerged victorious with a sturdy defense.",
+    "Dortmund slipped up in the last matchday to hand Bayern the Bundesliga title.",
+    "For the first time since Maradona's days, Napoli lifted the Serie A trophy.",
+]
+```
+
+The end goal is to be able to ask our model: which team in {insert_country} won the league in the 2023 season? Then, we want to put this additional context in our prompt to obtain the correct answer.
```
brenkao, line 22: the {insert_country} is a bit awkward. maybe a `insert_country`

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+    prompt = SoccerPrompt(context=most_similar, question=query)
+    completion = chat.create(prompt)
+    print(completion)
+
+
+countries = ["English", "Spanish", "German", "Italian"]
+for country in countries:
+    ask_soccer(f"Who won the {country} top flight?")
+```
+
```
brenkao, line 86: add short blurb to showcase how the soccer_info above matches the print statement

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+### Batched Embeddings
+
+As before, we embed our texts. However, we can speed up the process and minimize token usage by embedding in batches. Since `max_tokens=1000` and the model `text-embedding-ada-002` has a token limit of 8096, we can pass in batches of 8.
+```python
+chat = OpenAIChat(model=MODEL)
+
+EMBEDDINGS_COLUMN = "embeddings"
+
+batch_size = 8
```
brenkao, line 163: inconsistent use of upper snake and lower snake case, applies to all other variables

docs/cookbook/rag.md
willbakst: How about this:

```markdown
Often you want your model to generate its response based on relevant
information that you provide. This recipe will demonstrate how to retrieve
the relevant information and how easy it is to format that information into
a Mirascope prompt.
```

The above has newlines that should be removed when added, but I added them here so it's easier to read.

docs/cookbook/rag.md
willbakst: how just `a given country`?

docs/cookbook/rag.md
willbakst: Strong agree. While not necessary in the code example, seeing the output here as though it were run in Jupyter can be extremely helpful. Some of the concept pages have examples of this I believe.

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+    ]
+    relevant_articles = df.loc[relevant_article_indices][TEXT_COLUMN].tolist()
+    prompt = NewsPrompt(
+        num_articles=num_articles, topic=query, articles=relevant_articles
+    )
+
+    return str(chat.create(prompt))
+
+
+print(summarize_news("teams and players going through trouble in soccer", 3))
```
brenkao, line 254: move NewsPrompt out of `summarize_news`
rename summarize_news to relevant_articles
print
```
NewsPrompt(num_articles=3, topic="teams and players going through trouble in soccer", articles=relevant_articles())
```

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+    articles: list[str]
+
+    @property
+    def listed_articles(self) -> str:
+        return "\n".join(
+            [f"{i+1}. {article}" for i, article in enumerate(self.articles)]
+        )
+```
+
+Another change from the soccer example is how we query our context, in this case `articles`. Instead of calculating the vector similarities locally, we call `Index.query()` on our Pinecone index to let them efficiently do the hard work for us. Once the articles are retrieved, we take their indices and search them up in our `pd.Dataframe` to match them to the articles, and let Mirascope take care of the rest.
```
brenkao, line 230: soccer -> basic/barebones

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
+# Retrieval Augmented Generation (RAG)
+
+When you need to provide additional information to your model that it hasn't been pretrained on, we retrieve the relevant information from an external source and provide it as context to our model.
+
+## A Barebones Retrieval Example
```
brenkao, line 5: this is the only place where we use the word barebones, consider changing to basic or changing all the basic words to barebones

docs/cookbook/rag.md
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    context: str
+    question: str
+```
+
+But to provide `context`, we need to implement the retrieval functionality. To do this, we embed our question (`query_embedding`), then rank the similarities between `query_embedding` and each of the embeddings from `soccer_info` using the dot product between the vectors. The text with the highest value (and thus, similarity) is passed in as the context. And just like that, our basic RAG example is complete.
+
+```python
+def ask_soccer(query):
```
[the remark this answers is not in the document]
willbakst: @brianpark0126 we should sync on this comment

PR #33, opened 2024-01-25
Initial prep work for streaming tools more conveniently

willbakst:
In preparation for adding convenience around streaming tools, I've added:
- `Partial` class from Pydantic issues page around dyanmically recreating a partial class (like `Partial` in TypeScript)
- Ability to stream the raw `tool_calls` more easily using our `OpenAITool` class.


tests/test_partial.py
```diff
@@ -0,0 +1,76 @@
...
+class PartialDeeperModel(BaseModel):
+    """A deeper model."""
+
+    shallow: Optional[PartialShallowModel] = {}  # type: ignore
+    param: Optional[str] = None
+
+
+def test_deeper_partial():
+    """Tests that `Partial` works to make all fields optional in a deeper model."""
+    print(Partial[DeeperModel].model_json_schema())
```
brenkao, line 47: remove prints

mirascope/partial.py
brenkao, line 1: does this make sense to be called typing, so people use from mirascope.typing import Partial. similar to from typing import Optional

tests/chat/models/test_async_openai_chat.py
```diff
@@ -166,6 +166,43 @@ async def test_async_openai_chat_stream_messages_kwarg(
...
+    "prompt,tools",
+    [
+        ("fixture_foobar_prompt", ["fixture_my_tool"]),
+        ("fixture_foobar_prompt", ["fixture_my_tool", "fixture_empty_tool"]),
+    ],
+)
+async def test_async_openai_chat_stream_tools(
+    mock_create, fixture_chat_completion_chunk_with_tools, prompt, tools, request
+):
+    """Tests that `OpenAIChat` returns the expected response when called with tools."""
```
brenkao, line 183: AsyncOpenAIChat

mirascope/chat/utils.py
```diff
@@ -12,26 +12,45 @@
...
-        return _get_openai_chat_messages(prompt)
+        return [
+            cast(ChatCompletionMessageParam, {"role": role, "content": content})
+            for role, content in prompt.messages
+        ]
     else:
         return [cast(ChatCompletionMessageParam, {"role": "user", "content": prompt})]
 
 
+def setup_openai_kwargs(
```
brenkao, line 28: not sure if there is a protocol for this but, perhaps patch_openai_kwargs or something to that effect informing us that you are mutating the dictionary.

mirascope/chat/utils.py
```diff
@@ -12,26 +12,45 @@
...
+    if tools:
+        openai_tools: list[type[OpenAITool]] = [
+            tool if isclass(tool) else convert_function_to_openai_tool(tool)
+            for tool in tools
+        ]
+        kwargs["tools"] = [tool.tool_schema() for tool in openai_tools]
+        if "tool_choice" not in kwargs:
+            kwargs["tool_choice"] = "auto"
+
+        return openai_tools
```
brenkao, line 49: we currently arent doing anything with this return so I would omit it completely. that way this function is very clear in what it is doing which is mutating kwargs, whereas currently it feels like its mutating kwargs but returning something slightly different

mirascope/chat/utils.py
willbakst: Done.

mirascope/partial.py
willbakst: Hmm, we can always add a typing module if we want, but I wanted to have a standalone reference source file to the original author here. And they can import `from mirascope import Partial` in the meantime

tests/chat/models/test_async_openai_chat.py
willbakst: Done.

tests/test_partial.py
willbakst: Done.

brenkao (approved):

PR #43, opened 2024-01-30
Rag refactor

brianpark0126:
prompt engineering is not yet complete, will finish tomorrow morning. no need to review as is, just leaving up in case there's anything noticable.


cookbook/rag_examples/rag_example.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    return str(completion)
+
+
+topics = [
+    "soccer teams/players going through trouble",
+    "environmental factors affecting economy",
+    "celebrity or politician scandals",
+]
+for topic in topics:
```
willbakst, line 78: I would update this to be a runnable script (like movie_bot). For example:

```python
def summarize_topics(summary_fn: Callable[[str, int], None]) -> None:
    """Runs the `summary_fn` on a list of topics."""
    topics = [...]
    for topic in topics:
        print(summary_fn(topic, 3))

def main():
    """Summarizes news articles. Uses pinecone if specified.
    if args.use_pinecone:
        summarize_topics(summarize_news_pinecone_embeddings)
    else:
        summarize_topics(summarize_news_local_embeddings)

if __name__ == "__main__":
    # do things like setup command line flags
    main()
```

Then we can run:

```sh
$ python rag_example.py --use-pinecone
```

cookbook/rag_examples/rag_example.py
```diff
@@ -0,0 +1,84 @@
...
+from setup_pinecone import setup_pinecone
+from utils import embed_df_with_openai, load_data
+
+from mirascope import OpenAIChat
+
+chat = OpenAIChat(api_key=os.getenv("OPENAI_API_KEY"))
+
+
+# Make reruns efficient by preprocessing data before saving to csv
+if not os.path.exists(FILENAME):
```
willbakst, line 23: I would refactor this into a separate setup script:

```sh
$ python setup_local.py
$ python setup_pinecone.py
```

Then in this file I would defer checking this stuff until you check the flags (see below comment):

```python
def main():
    ...
    if flags.use_pinecone:
        if not pinecone_is_setup:
            raise ValueError("`use_pinecone` specified but pinecone is not setup.")
        ...
    else:
        if not os.path.exists(FILENAME):
            raise ValueError("You must first run `setup_local.py`.")
        df = pd.read_pickle(FILENAME)
        ...
```

willbakst: Also see below comment. Is there some way to figure out `pinecone_is_setup`? If not easily through pinecone, could we do something like save a dummy file `pinecone_is_setup.txt` and then check `if not os.path.exists("pinecone_is_setup.txt")

cookbook/rag_examples/requirements.txt
```diff
@@ -3,3 +3,5 @@ numpy>=1
 pandas>=1.2.3
 tiktoken
 pinecone-client
+python-dotenv
+pydantic_settings
```
willbakst, line 7: I believe this should be `pydantic-settings` with a `-` and not `_`

cookbook/rag_examples/setup_pinecone.py
```diff
@@ -0,0 +1,43 @@
...
+        pc.create_index(
+            name=PINECONE_INDEX,
+            dimension=1536,
+            metric="cosine",
+            spec=ServerlessSpec(
+                cloud="aws",
+                region="us-west-2",
+            ),
+        )
+    index = pc.Index(PINECONE_INDEX)
```
willbakst, line 34: can we not just check here if the index is empty or not? Assume if empty this hasn't been run. If it isn't empty then it has been run?

cookbook/rag_examples/rag_prompts/news_rag_prompt.py
```diff
@@ -0,0 +1,98 @@
...
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    @property
+    @abc.abstractmethod
+    def context(self) -> str:
+        """Finds most similar articles to topic using embeddings."""
+        ...
+
+
+class LocalNewsRagPrompt(NewsRagPrompt):
```
willbakst, line 56: It's interesting, although it makes sense to group these prompts together given that they are very similar, it makes versioning one of the prompts much harder. I would recommend splitting these into three different prompts: `news_rag_base_prompt.py`, `local_news_rag_prompt.py`, and `pinecone_news_rag_prompt.py`

This solves a problem of clashing changes where you might make changes to local and pinecone but then want to revert just one of the changes and can't because you can't mix the different version easily.

willbakst (commented):
Looking great! Some refactor comments + nits

cookbook/rag_examples/rag_prompts/news_rag_prompt.py
```diff
@@ -0,0 +1,98 @@
...
+            num_results=self.num_statements,
+            chat=OpenAIChat(api_key=os.getenv("OPENAI_API_KEY")),
+        )
+        return "\n".join(
+            [f"{i+1}. {statement}" for i, statement in enumerate(statements)]
+        )
+
+
+class PineconeNewsRagPrompt(NewsRagPrompt):
+    __doc__ = inspect.getdoc(NewsRagPrompt)
```
willbakst, line 75: I just realized that inheriting the docstring like this makes it so that the docstring doesn't appear for this class in VSCode, which isn't ideal.

I'm wondering if there's a better way where we can automatically inherit the docstring from the parent prompt...this is potentially a good feature to think about

cookbook/rag_examples/setup_pinecone.py
willbakst:
```python
# Once we've successfully uploaded the news article dataset, we don't want to
# upload it again (to save on costs). However, in the event of a failure, we want to
# make sure that we upload again so the full dataset is there.
if num_examples = index.num_examples:
    return
...
```

cookbook/rag_examples/rag_prompts/news_rag_prompt.py
brenkao: I spent some time looking into this, but doesn't seem like an easy thing to add. I vote to copy and paste NewsRagPrompt's docstring here and call it a day.

cookbook/rag_examples/setup_pinecone.py
```diff
@@ -0,0 +1,62 @@
...
+            try:
+                index.upsert(vectors)
+            except PineconeApiException as e:
+                print(f"Failed to upsert batch starting at index {start}: {e}")
+                raise UpsertError(f"Failed to upsert vectors to Pinecone Index: {e}")
+
+
+class UpsertError(Exception):
+    """Exception raised when an upsert operation fails."""
+
```
willbakst, line 61: remove newline

cookbook/rag_examples/setup_pinecone.py
```diff
@@ -0,0 +1,62 @@
...
+        for start in range(0, len(df), batch_size):
+            end = min(start + batch_size, len(df))
+            batch = df[start:end]
+            vectors = [
+                {"id": str(i), "values": row[EMBEDDINGS_COLUMN]}
+                for i, row in batch.iterrows()
+            ]
+            try:
+                index.upsert(vectors)
+            except PineconeApiException as e:
```
willbakst, line 54: this should just catch the `UpsertError` and then try to upsert again some number of times until it succeeds

cookbook/rag_examples/setup_pinecone.py
willbakst: and pass

cookbook/rag_examples/mirascope/versions/local_news_rag_prompt/0001_local_news_rag_prompt.py
```diff
@@ -0,0 +1,39 @@
...
+
+    {context}
+
+    Pick only the snippets which are truly relevant to the topic, and summarize them.
+    """
+
+    num_statements: int
+    topic: str
+    df: pd.DataFrame
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
willbakst, line 26: add newline

cookbook/rag_examples/mirascope/versions/local_news_rag_prompt/0002_local_news_rag_prompt.py
```diff
@@ -0,0 +1,50 @@
...
+
+    {context}
+
+    Pick only the snippets which are truly relevant to the topic, and summarize them.
+    """
+
+    num_statements: int
+    topic: str
+    df: pd.DataFrame
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
willbakst, line 37: same here

cookbook/rag_examples/mirascope/versions/pinecone_news_rag_prompt/0001_pinecone_news_rag_prompt.py
```diff
@@ -0,0 +1,59 @@
...
+    {context}
+
+    Pick only the snippets which are truly relevant to the topic, and summarize them.
+    """
+
+    num_statements: int
+    topic: str
+    df: pd.DataFrame
+    _index: Pinecone.Index
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
willbakst, line 40: same here

willbakst (approved):

cookbook/basic_examples/prompt_cli/mirascope.ini
```diff
@@ -7,11 +7,11 @@ mirascope_location = mirascope
...
 # path to prompts directory
-prompts_location = prompts
+prompts_location = movie_prompts
 
 # name of versions text file
 version_file_name = version.txt
 
 # formats the version file
 # leave blank to not format 
-format_command = ruff format
```
willbakst, line 17: why did you remove the ending newline?

PR #44, opened 2024-01-30
Readme updates

willbakst:
Updating the README.md to focus specifically on Prompt Engineering given that's the focus on the library. Attempting to:
- make the value add far more clear, specifically wrt. prompts agnostic to any model.
- use techniques like italics/bolding to enable a user to see value ***at a glance***


docs/README.md
```diff
@@ -1,68 +1,89 @@
...
+Results in fewer bugs and iterations
+Helps to guarantee robust applications
+Enables faster and more reliable releases
+Reduces complexity and risk through maximal composability and customizability
 
-class BookRecommendationPrompt(Prompt):
-    """
-    Can you recommend some books on {topic}?
-    """
+Once you go mirascope, you'll never go back...ascope
```
brenkao, line 53: ...

docs/README.md
```diff
@@ -1,68 +1,89 @@
...
-$ pip install mirascope
-```
+Prompt Engineering is just engineering. Beyond simple toy examples (that aren't useful), prompting quickly becomes complex. Separating prompts from the engineering workflow will only put limitations on what you can build with LLMs. We firmly believe that prompts are far more than "just f-strings" and thus require developer tools that are purpose-built to make building these more complex prompts as easy as possible.
 
-This will install the `mirascope` package and CLI.
+Furthermore, the best developer experience requires the following, which we strive to uphold:
 
-## A Simple Mirascope Example
+* *Simplicity through idiomatic syntax* -> ***Faster and more reliable releases***
+* *Semi-opinionated methods* -> ***Reduced complexity that speeds up development***
```
brenkao, line 43: not really sure how this is different from the "Durable" section above. Maybe something like "Fully customizable ->  No adapters necessary to integrate"

docs/README.md
```diff
@@ -1,68 +1,89 @@
...
-## Installation
+* **Intuitive**: Editor support that you expect (e.g. **autocompletion**, **inline errors**)
+* **Peace of Mind**: Pydantic together with our Prompt CLI **eliminate prompt-related bugs**.
+* **Durable**: Seamlessly customize and extend functionality. **Never maintain a fork**.
+* **Integration**: Easily integrate with **JSON Schema** and other tools such as **FastAPI**
+* **Convenience**: Tooling that is **clean**, **elegant**, and **delightful** that **you don't need to maintain**.
+* **Open**: Dedication to building **open-source tools** you can use with **your choice of LLM**.
 
-Install Mirascope and start building with LLMs in minutes.
+## 🚨 Warning: Strong Opinion
```
brenkao, line 36: dont even think this is necessary

docs/README.md
```diff
@@ -1,68 +1,89 @@
...
 
-```sh
-$ pip install mirascope
-```
+Prompt Engineering is just engineering. Beyond simple toy examples (that aren't useful), prompting quickly becomes complex. Separating prompts from the engineering workflow will only put limitations on what you can build with LLMs. We firmly believe that prompts are far more than "just f-strings" and thus require developer tools that are purpose-built to make building these more complex prompts as easy as possible.
```
brenkao, line 38: move this section below the technical stuff, e.g. requirements. this section (after some refinement) should transition to examples to explain the "strong opinion"

docs/README.md
```diff
@@ -75,14 +96,12 @@ Any and all help is greatly appreciated! Check out our page on [how you can help
 
 ## Roadmap (What's on our mind)
 
+- [ ] Better DX for Mirascope CLI (e.g. autocomplete)
```
brenkao, line 99: omit, this will be in the release

willbakst: omit or mark as a check like I did for functions as OpenAI tools?

brenkao: i think check is fine until the list gets too long e.g. more than 5 items, then we can prune the checked ones

docs/README.md
willbakst: I feel like "Fully customizaable -> No adapter necessary to integrate" is "durable" whereas here I was trying to cover the idea that following an opinioni can reduce complexity that speeds up development.

I also wanted this to be more general than just the above Why use Mirascope section. General wrt. engineering principles for developer tools vs. what mirascope specifically helps with.

docs/README.md
willbakst: Hmm, I put this here because I thought the information was relevant before getting into usage. I felt like it would get lost at the end

docs/README.md
willbakst: what part of it isn't necessary?

docs/README.md
willbakst: so you like this right?

docs/README.md
```diff
@@ -1,68 +1,89 @@
...
 
-- **Easy**: Designed for ease of use, Mirascope features an intuitive interface and a minimal learning curve
-- **Intuitive**: Get editor support for prompts, eliminating the need to dig through documentation
-- **Durable**: Mirascope offers versatile and adaptable functionality, allowing seamless integration with custom solutions
-- **Integration**: Leveraging Pydantic, Mirascope offers easy integration with JSON Schema and other tools
-- **Maintainability**: Reduce prompt-related bugs and organize prompts with provided utilities
+**Mirascope** is a library purpose-built for Prompt Engineering on top of <a href="https://pydantic.dev" target="_blank">Pydantic</a>:
 
-## Installation
+* **Intuitive**: Editor support that you expect (e.g. **autocompletion**, **inline errors**)
```
brianpark0126, line 29: I like this line a lot because it tells you more directly than the previous iteration what are the problems mirascope is solving. That being said, it doesn't feel as in your face for the following lines - maybe I'm too inexperienced in the LLM field to take my word on this (or alternatively, I should be the gold standard for when something is obvious enough) but the following lines don't mean much to me since I haven't been building LLMs long, let alone with other libraries. Maybe they could be even more in your face?

brianpark0126: Like none of the following lines really make 100% to me for the following reasons: 
Peace of Mind: what kind of prompt-related bugs and what feature of Pydantic? My assumption is validation lets us not get silly things passed into chat models but could be clearer. 
Durable: extend what functionality? And why would I need to maintain a fork otherwise?
Integration makes sense
Convenience: why would I need to maintain other tools
Open makes sense

maybe this is just showing my greenness with LLMs and its not indicative of the average user but just sharing my two cents

brianpark0126 (commented):
my comments

docs/README.md
wcbudz: The phrase this is based on makes this problematic.

docs/README.md
wcbudz: Shots fired! I might make the second sentence less adversarial. "Beyond ~~simple toy examples (that aren't useful)~~ basic illustrative examples,..."

docs/README.md
brenkao: i think something cheesy is fine, but lets brainstorm better ones

docs/README.md
brenkao: the header

docs/README.md
brenkao: tone down the aggression and increase the persuasiveness, i think. if you think they are going to get lost because an installation guide is in front, we have bigger problems

docs/README.md
```diff
@@ -1,68 +1,89 @@
...
 
-## A Simple Mirascope Example
+* *Simplicity through idiomatic syntax* -> ***Faster and more reliable releases***
+* *Semi-opinionated methods* -> ***Reduced complexity that speeds up development***
+* *Reliability through validation* -> ***More robust applications with fewer bugs***
+* *High-quality up-to-date documentation* -> ***Trust and reliability***
 
-```python
-from mirascope import OpenAIChat, Prompt
+Speeds up development
```
wcbudz, line 47: These aren't rendering as bullets

wcbudz: Or separate lines, or whatever they're supposed to be

wcbudz: Screenshot 2024-01-30 at 5 36 46 PM

docs/README.md
willbakst: I'm happy to just remove, this was mostly a joke cause I couldn't think of something actually good

docs/README.md
willbakst: @wcbudz @brianpark0126 what are your thoughts on this?

I tried it without the header and it made the first section feel way too large, so I added this for some flare and also demarcation.

docs/README.md
willbakst: What I don't want is for them not to give the usage section a chance because it's based on strong opinions that we haven't made clear.

docs/README.md
willbakst: Good catch. These were notes I was using to write the above. I've removed.

docs/README.md
willbakst: @brenkao I was writing this geared more towards senior developers. Does the language here scratch the itch, or should it be more clear? Any suggestions for things that might cover what @brianpark0126 mentioned?

docs/README.md
```diff
@@ -31,39 +106,176 @@ Install Mirascope and start building with LLMs in minutes.
...
+    @property
+    def name_specific_question(self) -> str:
+        """Returns a question based on `name`."""
+        if self.name == self.name[::-1]:
+            return "a palindrome"
+        else:
+            return "not a palindrome"
+
+    @property
+    def name_specific_greeting(self) -> str:
```
brianpark0126, line 199: typo, method says greeting but property is called remark

docs/README.md
```diff
@@ -1,27 +1,102 @@
...
+    I've recently read the following books: {titles_in_quotes}.
+    What should I read next?
+    """
+
+    book_titles: list[str]
+
+    @property
+    def titles_in_quotes(self) -> str:
+        """Returns a comma separated list of book titles each in quotes."""
+        return ", ".join([f'"{title}"' for title in self.book_titles])
```
brianpark0126, line 71: since this is the very first thing someone sees when looking at mirascope, I think it might be worth leaving out the messages decorator/system message part of this prompt so the first example really hones in on the pydantic aspect

brianpark0126 (commented):
left 2 comments

docs/README.md
willbakst: Done.

docs/README.md
willbakst: Done.

brenkao (approved):

PR #48, opened 2024-01-31
feat: added basic tools parsing

brenkao:
* added the simplest form of streaming tools
* once a valid tool is detected, it is returned and ready to be handled by the user


mirascope/chat/parsers.py
```diff
@@ -0,0 +1,60 @@
...
+                if len(self.tool_calls) <= tcchunk.index:
+                    self.tool_calls.append(
+                        ChatCompletionMessageToolCall(
+                            id="",
+                            type="function",
+                            function=Function(name="", arguments=""),
+                        )
+                    )
+                    prompt = None
+                tc = self.tool_calls[tcchunk.index]
```
willbakst, line 43: avoid abbreviations like tc (use tool_call)

cookbook/tools_examples/stream_openai_tool_example.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+tools: list[Union[Callable, Type[OpenAITool]]] = [GetCurrentWeather]
+stream_completion = chat.stream(
+    CurrentWeatherPrompt(),
+    tools=tools,  # pass in the function itself for automatic conversion
+)
+parser = PartialToolParser(tools=tools)
+for partial_tool in parser.from_stream(stream_completion):
+    print("data: ", partial_tool.__dict__, "\n\n")
```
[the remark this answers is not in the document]
brenkao: we are only yielding each tool individually, do we expect to yield the previous tool?
current:
print tool A
-----------
print tool B
what I think you are asking for:
print tool A
-----------
print tool A
print tool B

mirascope/chat/parsers.py
```diff
@@ -0,0 +1,60 @@
...
+    ) -> Generator[Partial, None, None]:
+        """Parses a stream of `OpenAIChatCompletionChunk`s into `PartialTools`s."""
+        openai_tools = convert_tools_list_to_openai_tools(self.tools)
+        prompt: Optional[Type[OpenAITool]] = None
+        for chunk in stream:
+            tcchunklist = chunk.tool_calls
+            if not tcchunklist:
+                continue
+            for tcchunk in tcchunklist:
+                if len(self.tool_calls) <= tcchunk.index:
```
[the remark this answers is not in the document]
brenkao: i tried implementing this but it became harder to read due to the continue. I would have to return a tuple which contains (PartialTool, prompt) to keep track of prompt as a iterate since prompt gets reset each iteration since prompt doesnt get mutated.

mirascope/chat/parsers.py
```diff
@@ -0,0 +1,68 @@
...
+)
+from pydantic import BaseModel
+
+from ..partial import Partial
+from .tools import OpenAITool
+from .types import OpenAIChatCompletionChunk
+from .utils import convert_tools_list_to_openai_tools
+
+
+class PartialOpenAIToolParser(BaseModel):
```
willbakst, line 17: re: discussion, let's rename this to something like `OpenAIToolStreamParser` and clean up the below functionality such that it will always return a full tool (and not partial, so the consumer end is cleaner cause they can always expect either a full tool or a validation error)

mirascope/chat/parsers.py
```diff
@@ -0,0 +1,68 @@
...
+    def from_stream(
+        self, stream: Generator[OpenAIChatCompletionChunk, None, None]
+    ) -> Generator[Partial, None, None]:
+        """Parses a stream of `OpenAIChatCompletionChunk`s into `PartialTools`s."""
+        openai_tools = convert_tools_list_to_openai_tools(self.tools)
+        tool_type: Optional[Type[OpenAITool]] = None
+        for chunk in stream:
+            if not chunk.tool_calls:
+                continue
+            for tool_call_chunk in chunk.tool_calls:
```
willbakst, line 32:
```python
def _parse_tool_calls(tool_calls):
    ...
    yield ...

for chunk in stream:
    if not chunk.tool_calls:
        continue
    yield from self._parse_tool_calls(chunk.tool_calls)
```

cookbook/tools_examples/stream_openai_tool_example.py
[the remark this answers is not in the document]
willbakst: Other comment resolves this with update around it not being partial but full tools instead as a first step

mirascope/chat/parsers.py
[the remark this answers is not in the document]
brenkao: update: the inner for loop is unnecessary because chunk.tool_calls length is always 1.

mirascope/chat/parsers.py
```diff
@@ -0,0 +1,66 @@
...
+    ) -> Generator[OpenAITool, None, None]:
+        """Parses a stream of `OpenAIChatCompletionChunk`s into `OpenAITools`."""
+        openai_tools = convert_tools_list_to_openai_tools(self.tools)
+        tool_type: Optional[Type[OpenAITool]] = None
+        for chunk in stream:
+            if not chunk.tool_calls:  # is true for first and final chunk
+                continue
+            tool_call_chunk = chunk.tool_calls[
+                0
+            ]  # update if multiple tool calls are supported
```
willbakst, line 33: move this above the assignment for better formatting

willbakst: now I actually think we can remove this comment. see other comment about assumption

mirascope/chat/types.py
```diff
@@ -112,7 +112,14 @@ def content(self) -> Optional[str]:
 
     @property
     def tool_calls(self) -> Optional[list[ChoiceDeltaToolCall]]:
-        """Returns the tool calls for the 0th choice message."""
+        """Returns the partial tool calls for the 0th choice message.
+
+        The size of `list[ChoiceDeltaToolCall]` most likely never exceeds 1. The first
```
willbakst, line 117: Preface this with. `We're making what we think is a reasonable assumption here when we use this call ourselves: ...`

mirascope/chat/parsers.py
```diff
@@ -0,0 +1,66 @@
...
+            if len(self.tool_calls) <= tool_call_chunk.index:
```
willbakst, line 34: break up the below into sections with header comments explaining the steps in plain text (not too verbose)

willbakst (approved):

PR #51, opened 2024-02-01
TOOL message type for Prompt

willbakst:
- Added a basic parser for the TOOL message type.
- Does not handle additional {id} or {name} fields that are available when using OpenAI.


mirascope/chat/tools.py
```diff
@@ -10,7 +10,7 @@
 class OpenAITool(BaseModel):
     """A base class for more easily using tools with the OpenAI Chat client."""
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    model_config = ConfigDict(arbitrary_types_allowed=True, extra="allow")
```
brenkao, line 13: why is this necessary.

brenkao (approved):

mirascope/chat/tools.py
willbakst: I did this so that the tool can have a `tool_call` attribute that we set after creating the tool specified by the consumer. See `chat.types.py` lines 64 and 84.

Another option would be to add a private `tool_call` attribute. The consumer of the defined tool would be able to pass in a tool call, which might actually be better. WDYT?

brenkao: i would prefer the second option, having defined attributes and not allowing extras seems better practice

willbakst: Done.

brenkao (approved):

PR #52, opened 2024-02-02
rag cookbook example markdown

brianpark0126:


docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
-# RAG (Retrieval-Augmented Generation)
+# Retrieval Augmented Generation (RAG)
+
+When we need to provide additional information to the model that it hasn't yet been trained on, we can retrieve the relevant information from an external source and provide it as context to our model. For our example, we will use a dataset of [BBC news articles from 2004](https://raw.githubusercontent.com/Dawit-1621/BBC-News-Classification/main/Data/BBC%20News%20Test.csv). We will:
+
+- query the dataset with a topic of our choosing
+- retrieve a number of relevant articles
+- ask our chat model to summarize the relevant ones
+- show how to query the dataset both locally and using a vector database [(Pinecone)](https://docs.pinecone.io/)
```
willbakst, line 8: nit: this should be ([Pinecone](url))

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+PINECONE_NAMESPACE = "articles"
+
+class Settings(BaseSettings):
+    openai_api_key: Optional[str] = None
+    pinecone_api_key: Optional[str] = None
+
+    model_config = SettingsConfigDict(env_file=".env")
+```
+
+```python
```
willbakst, line 48: nit: switch the .env to be above the config. Realizing it makes more sense to show those first and then the Settings object in config.py

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+# .env
+
+OPENAI_API_KEY = "YOUR_OPENAI_API_KEY"
+PINECONE_API_KEY = "YOUR_PINECONE_API_KEY"
+```
+
+## Load and preprocess the data
+
+Before we load raw data into a pandas `Dataframe`, we have to handle token limits for both chat and embeddings models. In our case, where `gpt-3.5-turbo` has a lower token limit (4096) than the `text-embedding-ada-002` (8191), the chat becomes the limiting factor. A crude solution is to chunk each article into equal token lengths of at most `MAX_TOKENS=1000`, so we may fit 3 article snippets as well as any text in our hand-written section of the prompt. The function `split_text()` below contains the actual chunking logic.
```
willbakst, line 57: nit: remove "the" before "text-embedding-ada-002"

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+
+For counting the token lengths, we will use [tiktoken](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb) since we will be using OpenAI for both the chat and embedding models. [tiktoken](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb) is a useful library provided by OpenAI for encoding strings and decoding tokens for their models.
```
willbakst, line 59: "for counting the token lenghts" -> "For counting the number of tokens in an article"

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+    df = pd.concat(
+        [df, pd.DataFrame(split_articles, columns=[TEXT_COLUMN])], ignore_index=True
+    )
+
+    return df
+
+
+def split_text(text: str, tokens: list[int], max_tokens: int) -> list[str]:
+    """Roughly splits a text into chunks according to max_tokens.
+    Text is split into equal word counts, with number of splits determined by how many
```
willbakst, line 96: nit: add missing newline here between first sentence and additional information.

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+```python
+# utils.py
+
+import pandas as pd
+import tiktoken
+from config import MODEL, TEXT_COLUMN,
+
+def load_data(url: str, max_tokens: int) -> pd.DataFrame:
+    """Loads data from a url after splitting larger texts into smaller chunks.
+    Args:
```
willbakst, line 70: nit add missing newline here above Args

docs/cookbook/rag.md
willbakst: Also I'm a little confused here. Why is the chat the limiting factor? It seems like the token limit is the limiting factor? It's just different for different tasks.

Below it's clear that we chunk articles into chunks of 8191 for the embedding model, then we embed the query, compare, load the text, then chunk the text into MAX_TOKENS limit, right? Please make that more clear here.

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+    return [datum.embedding for datum in embeddings_response.data]
+
+
+def embed_df_with_openai(
+    df: pd.DataFrame,
+    client: OpenAI,
+) -> pd.DataFrame:
+    """Embeds a Pandas Series of texts in batches using minimal OpenAI calls.
+
+    Note that this functions assumes all texts are less than 8192 tokens long.
```
willbakst, line 182: "functions" -> "function"

docs/cookbook/rag.md
```diff
@@ -1,7 +1,424 @@
...
+        indices = [int(article["id"]) for article in query_response["matches"]]
+        statements = self.df.iloc[indices][TEXT_COLUMN].to_list()
+        return "\n".join(
+            [f"{i+1}. {statement}" for i, statement in enumerate(statements)]
+        )
+```
+
+## Retrieval Script For Topics
+
+Thanks to Pydantic, we handled the retrieval functionality internally through the `context` property, despite the different implementations of each class. With this power, we take everything we’ve done so far and execute it with these simple lines:
```
willbakst, line 374: Please update to:

"With the retrieval functionality implemented internally to our prompts, we can now build a script that uses the prompts without having to worry about how to format `context` or retrieve relevant articles:"

docs/cookbook/rag.md
```diff
@@ -1,7 +1,434 @@
...
+    pinecone_api_key: Optional[str] = None
+
+    model_config = SettingsConfigDict(env_file=".env")
+```
+
+
+
+## Load and preprocess the data
+
+Before we load raw data into a pandas `Dataframe`, we have to handle large texts which may exceed token limits in the chat or embeddings models. For the models we use in our case, `gpt-3.5-turbo` has a lower token limit (4096) than `text-embedding-ada-002` (8191), so it makes sense to determine a solution based on the chat's token limit. A crude solution is to split any article of token count greater than `MAX_TOKENS=1000` into equal chunks, with each resulting chunk cosisting of less tokens than `MAX_TOKENS`. This way, we may fit 3 article snippets as well as any text in our hand-written section of the prompt. The function `split_text()` below contains the implementation.
```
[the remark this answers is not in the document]
willbakst: Maybe something more like:

In our case, we are using `gpt-3.5-turbo`, which has a token limit of 4096. A crude solution is to split any article of token count greater than `MAX_TOKENS=1000` into equal chunks, with each resulting chunk consisting of fewer than `MAX_TOKENS` tokens. This way we can fit up to 3 article chunks along with the hand-written section of the prompt. Note that we have also made sure that `MAX_TOKENS` is less than the token limit for our embedding model `text-embedding-ada-002` that has a token limit of 8191.

The function `split_text()` below contains the implementation:

docs/cookbook/rag.md
```diff
@@ -1,7 +1,434 @@
...
+        The dataframe with the embeddings column added.
+    """
+    encoder = tiktoken.encoding_for_model(MODEL)
+    max_tokens = 8191
+
+    embeddings: list[list[float]] = []
+    batch: list[str] = []
+    batch_token_count = 0
+    for i, text in enumerate(df[TEXT_COLUMN]):
+        if batch_token_count + len(encoder.encode(text)) > max_tokens:
```
willbakst, line 203: add a comment here instead of the note below to explain what you're doing here.

willbakst (approved):

PR #53, opened 2024-02-02
feat: Added AsyncOpenAIToolParser

brenkao:
* Created utility function to share logic between the two tool parser classes
* Updated imports
* Updated docstrings


mirascope/chat/parsers/utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+from ..utils import convert_tools_list_to_openai_tools
+
+if TYPE_CHECKING:
+    from .openai_parsers import OpenAIToolStreamParser
+    from .openai_parsers_async import AsyncOpenAIToolStreamParser
+
+
+def handle_new_index(
+    self: Union["OpenAIToolStreamParser", "AsyncOpenAIToolStreamParser"],
```
willbakst, line 18: I don't like this passing of "self" and making updates externally. Each of these functions does a particular thing to self. I would have each of them take what they need and return changes that are handles in the caller. That gives the caller more flexibility without typing the function to particular class types.

For example:

```python
def handle_possible_new_tool_call(
    tool_calls: list[ChatCompletionMessageToolCall],
    tool_call_chunk: ChoiceDeltaToolCall,
) -> bool:
    """Handles creating a new tool call if `tool_call_chunk` starts a new tool call.

    Args:
        tool_calls: The current list of tool calls being built in the stream parser.
        tool_call_chunk: The current chunk in the stream to be parsed.

    Returns:
        `True` if a new tool call was created and added to `tool_calls`.
    """
    if len(tool_calls) <= tool_call_chunk.index:
        tool_calls.append(ChatCompletionMessageToolCall(...))
        return True
    return False

# elsewhere
if handle_possible_new_tool_call(...):
    current_tool_type = None
```

mirascope/chat/parsers/openai_parser.py
```diff
@@ -0,0 +1,60 @@
...
+    find_tool_class,
+)
+
+
+class OpenAIToolStreamParser(BaseModel):
+    """A utility class to parse `OpenAIChatCompletionChunk`s into `OpenAITools`."""
+
+    tool_calls: list[ChatCompletionMessageToolCall] = []
+    tools: list[Union[Callable, Type[OpenAITool]]] = []
+    _current_tool_type: Optional[Type[OpenAITool]] = None
```
willbakst, line 24: does this need to be a property now, or can it go back to being a local variable of the `from_stream` function?

mirascope/chat/parsers/openai_parser_async.py
```diff
@@ -0,0 +1,63 @@
...
+
+class AsyncOpenAIToolStreamParser(BaseModel):
+    """A utility class to parse `OpenAIChatCompletionChunk`s into `OpenAITools`.
+
+    This is an async version of `OpenAIToolStreamParser`.
+    """
+
+    tool_calls: list[ChatCompletionMessageToolCall] = []
+    tools: list[Union[Callable, Type[OpenAITool]]] = []
+    _current_tool_type: Optional[Type[OpenAITool]] = None
```
willbakst, line 27: ditto

willbakst (approved):

mirascope/chat/parsers/openai_parser.py
brenkao: good catch

PR #67, opened 2024-02-08
feat: prompt completion dump to json

brenkao:
* Added `prompt.dump`
* Added `completion.dump`
* Added `@tags`
* Added basic example for logging
* Added tests for dump


mirascope/prompts.py
```diff
@@ -151,3 +171,16 @@ def messages_fn(self) -> list[tuple[str, str]]:
 
     setattr(cls, "messages", property(messages_fn))
     return cls
+
+
+def tags(args: Union[list[str], str]) -> Callable[[Type[T]], Type[T]]:
+    def tags_fn(model_class: Type[T]) -> Type[T]:
+        # Step 2: Attach the tags to the model class
```
willbakst, line 178: what is step 1?

brenkao: Oh oops, I had other comment, I will update

cookbook/basic_examples/log.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def create_database():
+    """Create the database and table for the OpenAI chat completions."""
+    metadata = MetaData()
+    table_objects = [Base.metadata.tables[TABLE_NAME]]
+    metadata.create_all(engine, tables=table_objects)
+
+
+def log_to_database(prompt_completion: dict):
```
willbakst, line 60: can we get some better typing here than just `dict`?

cookbook/basic_examples/log.py
```diff
@@ -0,0 +1,89 @@
+"""A basic example on how to log the data from a prompt and a chat completion."""
```
willbakst, line 1: is it worth adding additional information here around setting up the database etc.?

mirascope/chat/types.py
```diff
@@ -82,6 +84,14 @@ def tool(self) -> Optional[OpenAITool]:
 
         return None
 
+    def dump(self) -> dict:
```
willbakst, line 87: can we get better typing here?

Often you'll see `dict[str, Any]` at the very least

willbakst, line 87: noticing you only have this for chat completion but not the chunks. Is it worth adding for the chunks?

mirascope/prompts.py
```diff
@@ -6,10 +6,12 @@
 import re
 from string import Formatter
 from textwrap import dedent
-from typing import Type, TypeVar
+from typing import Callable, Optional, Type, TypeVar, Union
 
 from pydantic import BaseModel
 
+from mirascope.chat.types import OpenAIChatCompletion
```
willbakst, line 13: I'm realizing that this adds `openai` as a required dependency of `mirascope`.

Seems better for the prompt to just accept the dict and the user passes in the completion dump?

mirascope/prompts.py
```diff
@@ -95,6 +99,22 @@ def messages(self) -> list[tuple[str, str]]:
         """Returns the docstring as a list of messages."""
         return [("user", str(self))]
 
+    def dump(
+        self, completion: Optional[Union[dict, OpenAIChatCompletion]] = None
+    ) -> dict:
```
willbakst, line 104: better return type here than `dict`

mirascope/prompts.py
```diff
@@ -95,6 +99,22 @@ def messages(self) -> list[tuple[str, str]]:
         """Returns the docstring as a list of messages."""
         return [("user", str(self))]
 
+    def dump(
+        self, completion: Optional[Union[dict, OpenAIChatCompletion]] = None
```
willbakst, line 103: see above, I think this should just be `Optional[dict[str, Any]]`

mirascope/prompts.py
```diff
@@ -151,3 +171,15 @@ def messages_fn(self) -> list[tuple[str, str]]:
 
     setattr(cls, "messages", property(messages_fn))
     return cls
+
+
+def tags(args: Union[list[str], str]) -> Callable[[Type[T]], Type[T]]:
```
willbakst, line 176: this function is missing a docstring

mirascope/prompts.py
```diff
@@ -151,3 +171,15 @@ def messages_fn(self) -> list[tuple[str, str]]:
...
+    def tags_fn(model_class: Type[T]) -> Type[T]:
```
willbakst, line 177: also missing a docstring

tests/test_prompts.py
willbakst, line 1: please also update `test_types.py` and `test_openai_chat.py` to account for changes.

tests/test_prompts.py
```diff
@@ -79,12 +91,36 @@ def test_save_and_load(fixture_foobar_prompt, tmpdir):
...
 
 
+def test_tags(fixture_tag_prompt: TagPrompt, fixture_tags_prompt: TagsPrompt):
+    """Tests that the tags decorator adds a `tags` attribute."""
+    assert hasattr(fixture_tag_prompt, "_tags")
+    assert fixture_tag_prompt._tags == ["test_tag"]
+    assert fixture_tags_prompt._tags == ["multiple", "tags"]
+
+
+def test_prompt_dump(
```
willbakst, line 109: you can/should parameterize this to run for multiple different prompts, as follows:

```python
@pytest.mark.parametrize("prompt", ["fixture_foobar_prompt", "fixture_messages_prompt"])
def teset_prompt_dump(
    prompt, fixture_chat_completion, request
):
    prompt = request.getfixturevalue(prompt)
    ...
```

You can do the same for the above `test_tags` function, just include an additional `expected_tags` parameter that you test against for each `prompt` parameter

cookbook/basic_examples/log.py
brenkao: what else would it be? do u want to create a pydantic model?

cookbook/basic_examples/log.py
brenkao: this is a super basic setup that i imagine is self explanatory

mirascope/chat/types.py
brenkao: ah i see, ignore comment above

cookbook/basic_examples/log.py
willbakst: Usually I'd expect `dict[str, Any]`

mirascope/chat/types.py
brenkao: i wasnt sure yet, so i opted to not add it. it doesnt make sense to add chunks to db or log since its unreadable

brenkao: but we might want to add something such that when the chunks are all gathered, then make it easy to dump

mirascope/prompts.py
brenkao: hmmm

tests/test_prompts.py
brenkao: openai_chat has 100% coverage already, what is missing?

tests/test_prompts.py
brenkao: since i moved completion dump outside of test_prompts, i dont need it for test_prompt_dump anymore but i implemented it for tags

tests/test_prompts.py
willbakst: we're not testing for the proper setting of `_start_time` and `_end_time` (which is covered because `create` is run but we don't test for it.

willbakst (approved):

PR #70, opened 2024-02-09
Update pydantic_prompts.md

brenkao:
* Added section for tags and dump

### DIFF
--- a/docs/concepts/mirascope_cli.md
+++ b/docs/concepts/mirascope_cli.md
@@ -136,12 +136,27 @@ Here you specify which prompt and also which version you want to use. This will 
 
 This will let you quickly swap prompts with no code change, the exception being when prompts have different properties.
 
+### Removing prompts
+
+Often times when experimenting with prompts, lots of unused versions will need to be cleaned up in your project.
+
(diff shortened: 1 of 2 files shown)


docs/concepts/mirascope_cli.md
```diff
@@ -136,12 +136,27 @@ Here you specify which prompt and also which version you want to use. This will
...
+
+```shell
+mirascope remove my_prompt 0001
+```
+
+Here you specify which prompt and version you want to remove. Removal will delete the file but also update any versions that have the deleted version in their `prev_revision_id` to `None`.
+
+!!!note
+
+Please make sure that the `current_revision` is not the same as the prompt you are trying to remove. You can use `mirascope add` if you have incoming changes or `mirascope use` to swap `current_revision`.
```
willbakst, line 153: does this mean it will cause an issue if you try to do this?

willbakst (approved):

docs/concepts/mirascope_cli.md
brenkao: it will prompt you to use and wont proceed

PR #73, opened 2024-02-12
Docstrings PR

brianpark0126:
It's possible that some of the docstrings are overkill, in the sense that I wanted to give a little bit of context of how some funcitons are used within the codebase (since something returning an obscure Pydantic schema doesn't really tell you much).


mirascope/chat/models/async_openai_chat.py
```diff
@@ -20,7 +20,21 @@
 
 
 class AsyncOpenAIChat:
-    """A convenience wrapper for the AsyncOpenAI Chat client."""
+    """A convenience wrapper for the AsyncOpenAI Chat client.
+
+    The Mirascope convenience wrapper for OpenAI provides a more user-friendly interface
+    for interacting with their API. For detailed usage instructions, check out the
```
willbakst, line 26: change `instructions` to `examples`

mirascope/chat/models/openai_chat.py
```diff
@@ -20,7 +20,21 @@
 
 
 class OpenAIChat:
-    """A convenience wrapper for the OpenAI Chat client."""
+    """A convenience wrapper for the OpenAI Chat client.
+
+    The Mirascope convenience wrapper for OpenAI provides a more user-friendly interface
+    for interacting with their API. For detailed usage instructions, check out the
+    following links.
```
willbakst, line 27: Include a python example here instead of links to the cookbook (even if it's a direct copy of the cookbook. This is so you don't have to leave the code to see the example in your editor).

mirascope/chat/models/async_openai_chat.py
```diff
@@ -20,7 +20,21 @@
...
+    following links.
+
+    `AsyncOpenAIChat.create`:
```
willbakst, line 29: Include a python example here instead of links to the cookbook (even if it's a direct copy of the cookbook. This is so you don't have to leave the code to see the example in your editor).

mirascope/chat/tools.py
```diff
@@ -10,7 +10,56 @@
...
+        """
+        Tell me my favorite animal if my favorite food is {food} and my
+        favorite color is {color}.
+        """
+
+        food: str
+        color: str
+
+    def animal_matcher(fav_food: str, fav_color: str) -> str:
+    """Tells you your most likely favorite animal from personality traits.
```
willbakst, line 34: indentation on this docstr/function below is one too few

mirascope/chat/tools.py
```diff
@@ -67,6 +120,9 @@ def tool_schema(cls) -> ChatCompletionToolParam:
     def from_tool_call(cls, tool_call: ChatCompletionMessageToolCall) -> OpenAITool:
         """Extracts an instance of the tool constructed from a tool call response.
 
+        Takes `ChatCompletionMessageToolCall` from an OpenAI chat completion response,
```
willbakst, line 123: This doesn't add anything compared to above line.

I think it would be better to instead describe what it actually does, as in takes the arguments from the function of the given tool call and creates the openaitool instance etc.

mirascope/chat/types.py
```diff
@@ -127,12 +180,11 @@ def content(self) -> Optional[str]:
     def tool_calls(self) -> Optional[list[ChoiceDeltaToolCall]]:
         """Returns the partial tool calls for the 0th choice message.
 
-        The first and last `list[ChoiceDeltaToolCall]` will be None indicating start
-        and end of stream respectively. The next `list[ChoiceDeltaToolCall]` will
-        contain the name of the tool and index and subsequent
-        `list[ChoiceDeltaToolCall]`s will contain the arguments which will be strings
-        that need to be concatenated with future `list[ChoiceDeltaToolCall]`s to form a
-        complete JSON tool calls.
+        The first `list[ChoiceDeltaToolCall]` will contain the name of the tool and
```
willbakst, line 183: @brenkao we need to confirm this different found here

mirascope/chat/utils.py
```diff
@@ -14,7 +14,15 @@
 def get_openai_messages_from_prompt(
     prompt: Union[Prompt, str],
 ) -> list[ChatCompletionMessageParam]:
-    """Returns a list of messages parsed from the prompt."""
+    """Returns a list of messages parsed from the prompt.
+
+    Args:
+        prompt: A `prompt` or `str` to parse into a list of messages.
```
willbakst, line 20: nit: should be "A `Prompt` ..."

mirascope/chat/utils.py
```diff
@@ -27,7 +35,14 @@ def get_openai_messages_from_prompt(
 def convert_tools_list_to_openai_tools(
     tools: Optional[list[Union[Callable, Type[OpenAITool]]]],
 ) -> Optional[list[Type[OpenAITool]]]:
-    """Converts a list of `Callable` or `OpenAITool` instances to an `OpenAITool` list."""
+    """Converts a list of `Callable` or `OpenAITool` instances to an `OpenAITool` list.
+
+    Args:
+        A list of functions or `OpenAITool`s.
```
willbakst, line 41: should read "tools: A list..."

mirascope/chat/utils.py
```diff
@@ -27,7 +35,14 @@ def get_openai_messages_from_prompt(
...
+
+    Returns:
+        A list of all items converted to `OpenAITool`s.
```
willbakst, line 44: nit: `OpenAITool` instances

mirascope/chat/utils.py
```diff
@@ -40,8 +55,20 @@ def patch_openai_kwargs(
     kwargs: dict[str, Any],
     prompt: Optional[Union[Prompt, str]],
     tools: Optional[list[Type[OpenAITool]]],
-):
-    """Sets up the kwargs for an OpenAI API call."""
+) -> None:
+    """Sets up the kwargs for an OpenAI API call.
+
+    Messages are parsed to have the required `role` and `content` items, and tools
+    (if any exist) are parsed into JSON schemas in order to fit the OpenAI API.
```
willbakst, line 62: make sure it's clear here that other kwargs remain untouched and that we only explicitly change messages and tools and that we won't overwrite tool_choice if provided etc.

mirascope/chat/utils.py
```diff
@@ -55,10 +82,10 @@ def patch_openai_kwargs(
 
 
 def convert_function_to_openai_tool(fn: Callable) -> Type[OpenAITool]:
-    """Constructs and `OpenAITool` type from the given function.
+    """Constructs an `OpenAITool` type from the given function.
 
-    If parameters are not defined in the Args section, then the description will simply
-    be the name of the parameter.
+    This method assumes all function parameters are properly documented in identical
```
willbakst, line 87: Don't say `assume` here. We don't assume, we expect. We throw errors if not so there shouldn't be silent failures if it doesn't conform.

mirascope/cli/utils.py
```diff
@@ -151,7 +174,15 @@ def get_prompt_versions(version_file_path: str) -> VersionTextFile:
 
 
 def check_prompt_changed(file1_path: Optional[str], file2_path: Optional[str]) -> bool:
-    """Checks if the given prompts have changed."""
+    """Compare two prompts to check if the given prompts have changed.
+
+    Args:
+        file1_path (str): The path to the first prompt.
```
willbakst, line 180: drop the `(str)` here. The type comes from the function definition and shouldn't also be here.

willbakst: Please make sure that any other instances of this are also cleaned up elsewhere.

mirascope/cli/utils.py
```diff
@@ -227,8 +258,22 @@ def write_prompt_to_template(
     file: str,
     command: Literal[MirascopeCommand.ADD, MirascopeCommand.USE],
     variables: Optional[dict] = None,
-):
-    """Writes the given prompt to the template."""
+) -> str:
+    """Writes the given prompt to the template.
+
+    Deconstructs a prompt with ast and reconstructs it using the jinja2 template, adding
```
willbakst, line 264: nit: jinja2 -> `Jinja2`

mirascope/cli/utils.py
```diff
@@ -227,8 +258,22 @@ def write_prompt_to_template(
...
+    revision history into the prompt when the command is `MirascopeCommand.ADD`.
+
+    Args:
+        file: The path to the prompt.
+        command: The CLI command to execute.
+        variables: `None` when the command is `MirascopeCommand.USE` since revision
```
willbakst, line 270: The way this is written makes it sound like the function sets the variable based on the command when it's actually that the user of the function provides it.

It would make more sense to explain what variables is, make it clear it's used for add, and if another command is provided it's ignored.

mirascope/cli/utils.py
```diff
@@ -255,8 +300,13 @@ def write_prompt_to_template(
...
 
 
-def update_version_text_file(version_file: str, updates: dict):
-    """Updates the version text file."""
+def update_version_text_file(version_file: str, updates: dict) -> None:
+    """Updates the version text file.
+
+    Args:
+        version_file: The path to the version text file.
+        updates: A dictionary of update to the file.
```
willbakst, line 308: this is not clear

mirascope/cli/utils.py
```diff
@@ -255,8 +300,13 @@ def write_prompt_to_template(
     return template.render(**data)
 
 
-def update_version_text_file(version_file: str, updates: dict):
-    """Updates the version text file."""
+def update_version_text_file(version_file: str, updates: dict) -> None:
```
willbakst, line 303: `dict` is not typed property, should have key and value types as well

mirascope/enums.py
```diff
@@ -18,7 +18,14 @@ class _Enum(str, Enum, metaclass=_Metaclass):
 
 
 class MirascopeCommand(_Enum):
-    """CLI commands to be executed."""
+    """CLI commands to be executed.
+
+    - ADD: save a modified prompt to the `/versions` folder.
```
willbakst, line 23: `/versions` -> `versions/`

mirascope/prompts.py
```diff
@@ -173,6 +183,20 @@ def tags(args: Union[list[str], str]) -> Callable[[Type[T]], Type[T]]:
...
 
+    Example:
+
+    ```python
+    from mirascope import Prompt, tags
+
+
+    @tags(["book_recommendation", "entertainment"])
+    class BookRecommendationPrompt(Prompt):
+    ...
```
willbakst, line 194: might as well just have the prompt here instead of ...

mirascope/prompts.py
```diff
@@ -173,6 +183,20 @@ def tags(args: Union[list[str], str]) -> Callable[[Type[T]], Type[T]]:
...
+
+    print(BookRecommendationPrompt.dump()["tags"])
+    #> ['book_recommendation', 'entertainment']
+    ```
+
     Returns:
         The decorated class.
```
willbakst, line 201: this should be updated to be the decorated class with `_tags` set etc.

willbakst (approved):

mirascope/chat/types.py
brenkao: is there supposed to be a link

PR #74, opened 2024-02-14
feat: auto tag version when using mirascope cli add command

brenkao:
* Added logic to append/update version tags on mirascope command add
* Updated test_utils to include auto tag changes, will add tests to commands in later PR


mirascope/cli/commands.py
```diff
@@ -23,7 +23,6 @@
         $ mirascope use my_prompt 0001
```
willbakst, line 23: it's starting to feel like time we refactor commands into a module with a separate file for each command. Future PR?

mirascope/cli/utils.py
```diff
@@ -33,7 +44,7 @@ def __init__(self):
         self.imports = []
         self.from_imports = []
         self.variables = {}
-        self.classes = []
+        self.classes: list[ClassInfo] = []  # type: ignore
```
willbakst, line 47: why type ignore here?

mirascope/cli/utils.py
```diff
@@ -223,10 +236,82 @@ def get_prompt_analyzer(file: str) -> PromptAnalyzer:
...
+
+def _update_tag_decorator_with_version(
+    decorators: list[str], variables: MirascopeCliVariables
+) -> Optional[str]:
+    """Updates the tag decorator and returns the import name."""
+    if variables.revision_id is None:
+        return None
+    import_name = "tags"
+    tag_exists = False
+    version_tag_prefix = "version:"  # mirascope tag prefix
```
willbakst, line 261: I feel like it makes more sense to put the colon below? so that tag prefix is `version` and the tag is f"{prefix}:{id}"

mirascope/cli/utils.py
brenkao: mypy throws https://mypy.readthedocs.io/en/latest/error_code_list.html#notify-about-an-annotation-in-an-unchecked-function-annotation-unchecked

i imagine i was supposed to add some arguments in the init but this class doesnt really do that so i opted to add type: ignore instead

mirascope/cli/utils.py
brenkao: hmmm, but if for whatever reason somebody had `version1` as their tag, it gets caught, whereas i'm reserving the `version:` keyword instead

mirascope/cli/commands.py
willbakst: Seems doable with a little additional thought. Will backlog this.

mirascope/cli/utils.py
willbakst: mypy won't check this because it's inside an untyped function (the init). Because there are no arguments, I think what you should do is add `-> None` to type the init and then also add types to all of the other self variables.

Another interesting possibility would be to double sub-class with `BaseModel` and include all of these attributes as defaulted values? Might specifically require calling the NodeVisitor super init and base model init, so not sure if it's worth it vs. above solution.

willbakst: Actually, another thought: what if you just instantiate each attribute as a class attribute and remove init completely? For example:

```python
class PromptAnalyzer(ast.NodeVisitor):
    """Utility class for analyzing a Mirascope `Prompt` file."""

    imports: list[...] = []
    from_imports: list[...] = []
    variables: dict[..., ...] = {}
    classes: list[ClassInfo] = []
    decorators: list[...] = []
    comments: str = ""

    def visit_import(...) -> None:  # <- this will enable mypy for this function and should be done for all others
        ...
```

brenkao: class attributes caused some issues that instance level did not so i opted to add types to my init instead

tests/cli/test_utils.py
```diff
@@ -197,10 +204,24 @@ def test_get_prompt_analyzer():
     assert analyzer.variables == {"foo": "bar"}
 
 
+@pytest.mark.parametrize(
```
willbakst, line 207: what about if the version tag is first? or in the middle?

willbakst (approved):

tests/cli/test_utils.py
brenkao: no difference, cuz i find index and replace existing as long as its first one

brenkao: added anyways

Issue #75, opened 2024-02-14
Variables outside of prompts become strings or crash program


brianpark0126:
---

## Describe the bug:
When using CLI, defining variables outside of the Prompt class creates various problems. Using function assignments (or anything that `ast.literal_eval` cannot parse) raises a ValueError and crashes the command entirely; variables which can be parsed become stringified in both the saved version and true prompt. 
---

**To Reproduce**

1. Create a mirascope prompt with a variables defined outside the Prompt class
Example 1:
```python
from mirascope import Prompt, messages

settings = Settings()


@messages
class LocalNewsRagPrompt(Prompt):
    """
```
or 
Example 2:
```python
from mirascope import Prompt, messages

a = 42
b = "frogs are goated"


@messages
class LocalNewsRagPrompt(Prompt):
```

3. Save the prompt using `mirascope add myprompt`. For example 1, you receive: `ValueError: malformed node or string: <ast.Call object at 0x1076c2d30>`. In example 2, you get this for both `0002_myprompt.py` within `/.mirascope/versions` and in the true prompt in the `/prompts` directory since the `ADD` command now also loads the saved version into the prompt:
```python
from mirascope import Prompt, messages, tags

prev_revision_id = "0001"
revision_id = "0002"
a = "1"
b = "frogs are goated"


@messages
@tags(["version:0002"])
class LocalNewsRagPrompt(Prompt):
```

PR #77, opened 2024-02-15
fix: github issue 75, CLI error for variables outside prompt

brianpark0126:
only tested unofficially by using CLI commands (ADD and USE) within cookbook examples since Brendan has said I should wait for the official golden suite of CLI tests.


tests/cli/test_utils.py
```diff
@@ -197,11 +197,19 @@ def test_get_prompt_analyzer():
...
     foo = "bar"
+    number = 1
+    settings = Settings()
+    a_list = [1, 2, 3]
     '''
     )
     analyzer = get_prompt_analyzer(sample_file_content)
     assert analyzer.comments == "This is a comment"
-    assert analyzer.variables == {"foo": "bar"}
+    assert analyzer.variables == {
```
brenkao, line 207: just to make sure, when running `template.render` do we get back
```
number = 1
settings = Settings()
a_list = [1, 2, 3]
 ```

File: mirascope/cli/utils.py [hunk 2]
```diff
@@ -292,20 +292,22 @@ def _update_tag_decorator_with_version(
...
+def _update_mirascope_imports(imports: list[tuple[str, Optional[str]]]):
     """Updates the mirascope import."""
-    if not any(import_name == "mirascope" for import_name in imports):
-        imports.append("mirascope")
+    if not any(import_name == "mirascope" for import_name, _ in imports):
+        imports.append(("mirascope", None))
 
 
-def _update_mirascope_from_imports(member: str, from_imports: list[tuple[str, str]]):
+def _update_mirascope_from_imports(
```
brenkao, line 301: There are two ` # TODO: Update `mirascope.tags` work with import alias` that need to be updated as well.

This use case needs to be supported: 
```
import mirascope as mira

@mira.tags()
class MyPrompt(Prompt):
    ...
```

brenkao: Right now the strings that we are checking are hardcoded to be `mirascope.tags`

mirascope/cli/generic/prompt_template.j2
brenkao, line 1: update prompt_cli examples and other examples that use prompt_template

mirascope/cli/generic/mirascope.ini.j2
brenkao, line 1: update prompt_cli examples and other examples that use mirascope.ini

tests/cli/test_utils.py
brianpark0126: yes, the items in the dictionary thats passed in are all wrapped in an extra layer of string so you have to put that in this specific test, but when you render it with `template.render` it gets peeled away hence everything being a string and actual strings being double quoted

mirascope/cli/utils.py
```diff
@@ -333,13 +340,20 @@ def write_prompt_to_template(
         }
 
     import_tag_name: Optional[str] = None
+    mirascope_alias = "mirascope"
+    for name, alias in analyzer.imports:
```
brenkao, line 344: also need to check from_imports as well. e.g.
```
  from mirascope import tags as foo
```

brenkao:
Fixes #75

PR #82, opened 2024-02-21
fix: prompt messages to return openai types

willbakst:
A few fixes here:
1. Fixed issue with prompt messages returning a tuple and not the openai message params
2. Fixed issue with template injection. Each parsed message is formatted after parsing so the user can't inject e.g. "\nSYSTEM:INJECT\n" into the message.
3. Fixing (1) fixes issue with users not being able to directly write the `messages` array themselves inside of the prompt. Now users can define a messages property, write the `messages` array like they would normally, and everything else should just work.
4. Fix issue with installing the package. Previously we moved certain packages to "extra" even though they are really part of the core package. This causes issues for new users trying to install the package when they try to use e.g. `OpenAIChat.create` and it fails because `openai` isn't installed. Furthermore, now that we're using the explicit OpenAI message types inside of `prompts.py` it is actually required even if just using the prompts.

### DIFF
--- a/mirascope/prompts.py
+++ b/mirascope/prompts.py
@@ -8,7 +8,22 @@ from string import Formatter
 from textwrap import dedent
 from typing import Any, Callable, Optional, Type, TypeVar
 
+from openai.types.chat import (
+    ChatCompletionAssistantMessageParam,
+    ChatCompletionMessageParam,
+    ChatCompletionSystemMessageParam,
+    ChatCompletionToolMessageParam,
+    ChatCompletionUserMessageParam,
+)
 from pydantic import BaseModel
+
+
+def _format_template(prompt: Prompt, template: str) -> str:
+    """Formats the template with the prompt's attributes."""
+    template_vars = [
+        var for _, var, _, _ in Formatter().parse(template) if var is not None
+    ]
+    return template.format(**{var: getattr(prompt, var) for var in template_vars})
 
 
 class Prompt(BaseModel):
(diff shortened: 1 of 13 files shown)


mirascope/prompts.py
```diff
@@ -78,16 +93,12 @@ def template(cls) -> str:
...
-        template = self.template()
-        template_vars = [
-            var for _, var, _, _ in Formatter().parse(template) if var is not None
-        ]
-        return template.format(**{var: getattr(self, var) for var in template_vars})
+        return _format_template(self, self.template())
 
     @property
-    def messages(self) -> list[tuple[str, str]]:
+    def messages(self) -> list[ChatCompletionMessageParam]:
```
brenkao, line 99: if user passes in a list dict [str, Any] does mypy throw an error?

mirascope/prompts.py
```diff
@@ -78,16 +93,12 @@ def template(cls) -> str:
 
     def __str__(self) -> str:
         """Returns the docstring prompt template formatted with template variables."""
-        template = self.template()
-        template_vars = [
-            var for _, var, _, _ in Formatter().parse(template) if var is not None
-        ]
-        return template.format(**{var: getattr(self, var) for var in template_vars})
+        return _format_template(self, self.template())
```
brenkao, line 96: add `self: Prompt` like below?

mirascope/prompts.py
willbakst: Do you mean if they write their own `messages` property and try to return a base dict[str, Any]?

brenkao: yeah

mirascope/prompts.py
willbakst: Unnecessary. Inside of the `Prompt` class it already knows that `self` is a `Prompt` (but for the `messages_fn` below it doesn't know what self is because it's a function external to the class and then set as an attribute).

mirascope/prompts.py
willbakst: No because it's a `TypedDict` and not a pydantic model.

brenkao (approved):

PR #83, opened 2024-02-21
added tests for status

brenkao:
* Added tests for status
* Added debugger for pytest and python files

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Pytest",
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -352,11 +352,11 @@ class EmptyTool(OpenAITool):
 def fixture_mirascope_user_settings() -> MirascopeSettings:
     """Returns a `MirascopeSettings` instance."""
     return MirascopeSettings(
-        format_command="ruff format",
-        mirascope_location=".test",
+        format_command="ruff check --select I --fix; ruff format",
```
willbakst, line 355: should we update this to be the default command in the generic mirascope.ini?

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
+{
```
willbakst, line 1: let's sync on this

willbakst (approved):

tests/conftest.py
brenkao: it already is

PR #90, opened 2024-02-23
added coverage for tests using async generators

brenkao:
* Fixed tests with async generators which fixes code coverage
* Fixed mock of openai create to set the return value in the correct spot
* Added fixtures
* Updated fixtures to handle more cases

### DIFF
--- a/tests/chat/models/test_openai_chat_async.py
+++ b/tests/chat/models/test_openai_chat_async.py
@@ -122,7 +122,9 @@ async def test_async_openai_chat_stream(
 ):
     """Tests that `AsyncOpenAIChat` returns the expected response when streaming."""
     prompt = request.getfixturevalue(prompt)
-    mock_create.__aiter__.return_value = [fixture_chat_completion_chunk] * 3
+    mock_create.return_value.__aiter__.return_value = [
+        fixture_chat_completion_chunk
+    ] * 3
 
     model = "gpt-3.5-turbo-16k"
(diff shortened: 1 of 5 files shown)


tests/chat/models/test_openai_chat_async.py
```diff
@@ -122,7 +122,9 @@ async def test_async_openai_chat_stream(
 ):
     """Tests that `AsyncOpenAIChat` returns the expected response when streaming."""
     prompt = request.getfixturevalue(prompt)
-    mock_create.__aiter__.return_value = [fixture_chat_completion_chunk] * 3
+    mock_create.return_value.__aiter__.return_value = [
```
willbakst, line 125: why not use `fixture_chat_completion_chunks` here as well?

tests/chat/models/test_openai_chat_async.py
```diff
@@ -149,7 +151,9 @@ async def test_async_openai_chat_stream_messages_kwarg(
     mock_create, fixture_chat_completion_chunk
 ):
     """Tests that `AsyncOpenAIChat.stream` works with a messages kwarg."""
-    mock_create.__aiter__.return_value = [fixture_chat_completion_chunk] * 3
+    mock_create.return_value.__aiter__.return_value = [
```
willbakst, line 154: ditto

willbakst (approved):

PR #98, opened 2024-02-27
Include `_call_params: CallParams` attribute in `Prompt`

willbakst:
- This inclusion enables co-locating all aspects of the call to the LLM that may impact the quality of the `Prompt`.

### DIFF
--- a/mirascope/chat/models/async_openai_chat.py
+++ b/mirascope/chat/models/async_openai_chat.py
@@ -66,6 +66,7 @@ class AsyncOpenAIChat:
         self,
         api_key: Optional[str] = None,
         base_url: Optional[str] = None,
+        client_wrapper: Optional[Callable] = None,
         **kwargs,
     ):
         """Initializes an instance of `AsyncOpenAIChat."""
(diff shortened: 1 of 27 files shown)


mirascope/chat/models/async_openai_chat.py
```diff
@@ -140,6 +162,27 @@ async def stream(
             OpenAIError: raises any OpenAI errors, see:
                 https://platform.openai.com/docs/guides/error-codes/api-errors
         """
+        if isinstance(prompt, Prompt):
+            if self.model != "gpt-3.5-turbo":
```
brenkao, line 166: ditto

mirascope/chat/models/openai_chat.py
brenkao, line 1: same as async

brenkao (approved):

mirascope/prompts.py
```diff
@@ -26,6 +32,36 @@ def _format_template(prompt: Prompt, template: str) -> str:
     return template.format(**{var: getattr(prompt, var) for var in template_vars})
 
 
+class CallParams(BaseModel):
```
brenkao, line 35: does this limit us when we implement other models. feels like prompt should take in a union of different call params and the first one we support is OpenAICallParams

willbakst: Good call. Updated

PR #99, opened 2024-02-27
added 3 examples for langchain using mirascope

brenkao:
* Added basic invoke example
* Added basic rag example
* Added parallism example


examples/langchain/langchain_mirascope_invoke.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+joke_prompt = TellMeAJokePrompt(topic="ice cream")
+prompt = ChatPromptTemplate.from_template(joke_prompt.template())
+model = ChatOpenAI(model="gpt-4")
+output_parser = StrOutputParser()
+
+chain = prompt | model | output_parser
+
+print(chain.invoke(joke_prompt.model_dump()))
```
willbakst, line 38: this only works if you don't use properties or computed fields.

We'll need to add a method `def template_variables(self) -> dict[str, str]` or something to the `Prompt` class

examples/langchain/langchain_mirascope_parallel.py
```diff
@@ -0,0 +1,43 @@
...
+
+model = ChatOpenAI()
+joke_prompt = JokePrompt(topic="bear")
+poem_prompt = PoemPrompt(topic="bear")
+
+joke_chain = ChatPromptTemplate.from_template(joke_prompt.template()) | model
+poem_chain = ChatPromptTemplate.from_template(poem_prompt.template()) | model
+
+map_chain = RunnableParallel(joke=joke_chain, poem=poem_chain)
+invoke = joke_prompt.model_dump() | poem_prompt.model_dump()
```
willbakst, line 41: same here

willbakst (approved):

PR #100, opened 2024-02-28
Cookbook wandb

brianpark0126:
initial gut check. 
slight differences from EDD -
 1) no need for configs file, 
2) set up the root_span in main() to give a clearer idea of execution detail so it's easier to check the logs online in WandB and compare fields, 
3) `TracePrompt.span()` ended up having to handle the fact that `extract()` doesn't output a OpenAIChatCompletion object

things to decide-
Variable names could be changed, prompts are very short and could be put into one file.

once extract() passes back more information, can clean up both `span()` and the need to manually state the start time of each span.


cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -0,0 +1,48 @@
...
+from utils import get_time_in_ms
+from wandb_prompts.coolnessprompt import Coolness, CoolnessPrompt
+from wandb_prompts.partyinviteprompt import PartyInvitePrompt
+from wandb_prompts.whoprompt import Person, WhoPrompt
+
+import wandb
+from mirascope import OpenAIChat
+from wandb.sdk.data_types.trace_tree import Trace
+
+load_dotenv()
```
brenkao, line 13: use settings, see rag_example or squad_example

cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -0,0 +1,48 @@
...
+
+if __name__ == "__main__":
+    wandb.login(key=os.getenv("WANDB_API_KEY"))
```
brenkao, line 16: update to use settings

cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -0,0 +1,48 @@
...
+    wandb.init(project="wandb_logged_chain")
+
+    root_span = Trace(
+        name="root",
+        kind="chain",
+        start_time_ms=get_time_in_ms(),
+        metadata={"user": "mirascope_user"},
+    )
+    chat = OpenAIChat(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-3.5-turbo-1106")
```
brenkao, line 25: update example to use new OpenAICallParam

cookbook/wandb_logged_chain/wandb_prompts/traceprompt.py
```diff
@@ -0,0 +1,49 @@
+from typing import Literal, Union
```
brenkao, line 1: docstring

cookbook/wandb_logged_chain/wandb_prompts/traceprompt.py
```diff
@@ -0,0 +1,49 @@
...
+from mirascope.chat import OpenAIChatCompletion
+from wandb.sdk.data_types.trace_tree import Trace
+
+
+class TracePrompt(Prompt):
+    """Parent class for inherited WandB functionality."""
+
+    span_type: Literal["tool", "llm"]
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
brenkao, line 16: what arbitrary type are we allowing and why?

brenkao, line 16: does this already get inherited from prompt?

cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -0,0 +1,48 @@
...
+
+    who_prompt = WhoPrompt(span_type="tool", person="Brian")
+    start_time = get_time_in_ms()
+    who_completion = chat.extract(Person, who_prompt)
+    who_span = who_prompt.span(who_completion, root_span, start_time)
```
brenkao, line 30: completion already has start time

cookbook/wandb_logged_chain/wandb_prompts/traceprompt.py
```diff
@@ -0,0 +1,49 @@
...
+        elif isinstance(completion, BaseModel):
+            output = {"assistant": str(completion.model_dump())}
+
+        span = Trace(
+            name=self.__class__.__name__,
+            kind=self.span_type,
+            status_code="success",
+            status_message=None,
+            start_time_ms=start_time,
+            end_time_ms=get_time_in_ms(),
```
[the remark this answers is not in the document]
brenkao: see dump function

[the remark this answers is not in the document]
brianpark0126: explained in comment above

cookbook/wandb_logged_chain/wandb_chain.py
willbakst, line 1: wondering if this should be `wandb_mirascope_chain.py` and we should have a separate `wandb_chain.py` that shows what it would look like without mirascope? Or is including it in the cookbook markdown enough?

brianpark0126: it would just be akin to the colab example they have in their docs but with some strings changed around, do we think we need that? feels like the markdown is enough when the whole point of it is that it would be that's more difficult to read/use

willbakst: Markdown only is fine then

cookbook/wandb_logged_chain/wandb_prompts/traceprompt.py
```diff
@@ -12,8 +13,7 @@ class TracePrompt(Prompt):
     """Parent class for inherited WandB functionality."""
 
     span_type: Literal["tool", "llm"]
-
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _call_params: OpenAICallParams = OpenAICallParams(model="gpt-3.5-turbo-1106")
```
brenkao, line 16: unfortunately you're going to have to change this again cuz its now call_params instead of_call_params

cookbook/wandb_logged_chain/wandb_prompts/coolnessprompt.py
brenkao, line 1: camel case the files, same as other files

cookbook/wandb_logged_chain/wandb_chain.py
brenkao: they do now, so please update the examples to use completion start time and end time

cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -0,0 +1,66 @@
...
+    coolness_completion = chat.extract(Coolness, coolness_prompt)
+    coolness_span = coolness_prompt.span(coolness_completion, who_span)
+
+    party_invite_prompt = PartyInvitePrompt(
+        span_type="llm", coolness=coolness_completion.coolness
+    )
+    party_completion = chat.create(party_invite_prompt)
+    party_invite_prompt.span(party_completion, coolness_span)
+
+    root_span._span.end_time_ms = round(datetime.datetime.now().timestamp() * 1000)
```
willbakst, line 53: could we do `root_span._span.end_time_ms = party_span._end_time`?

cookbook/wandb_logged_chain/wandb_prompts/trace_prompt.py
```diff
@@ -0,0 +1,47 @@
...
+from wandb.sdk.data_types.trace_tree import Trace
+
+from mirascope import OpenAICallParams, Prompt
+from mirascope.chat import OpenAIChatCompletion
+
+
+class TracePrompt(Prompt):
+    """Parent class for inherited WandB functionality."""
+
+    span_type: Literal["tool", "llm"]
```
willbakst, line 14: nit: put a newline here between `span_type` and `call_params`

cookbook/wandb_logged_chain/wandb_prompts/trace_prompt.py
```diff
@@ -0,0 +1,47 @@
...
+                    "tool_output": completion.tool.fn(
+                        **completion.tool.model_dump(exclude={"tool_call"})
+                    ),
+                }
+            else:
+                output = {"assistant": str(completion)}
+            dump = completion.dump()
+        elif isinstance(completion, BaseModel):
+            output = {"assistant": str(completion.model_dump())}
+            dump = completion._completion.dump()  # type: ignore
```
willbakst, line 34: actually instead of a type ignore this should do:

```
if not hasattr(completion, "_completion"):
    raise ValueError("Completion of type `BaseModel` was not created using the `extract` function and does not contain the necessary `_completion` private attribute.")
dump = completion._completion.dump()
```

cookbook/wandb_logged_chain/wandb_chain.py
brianpark0126: if we do this, we'd have to make the prompts store a span, do we want that?

willbakst (approved):

PR #101, opened 2024-02-28
fix: remove deprecation warning when using tools with prompt for extr…

willbakst:
…act since it needs to use it's own generated tool.


brenkao (approved):

brenkao:
approved after fixing tests

PR #106, opened 2024-02-29
fix: use ClassVar instead of PromptConfig so we don't need to ignore warnings

willbakst:
See title


brenkao (approved):

brenkao:
approved after checks passes

PR #109, opened 2024-03-01
chore: added basic example for validation

brenkao:
same as title

### DIFF
--- a/examples/validator.py
+++ b/examples/validator.py
@@ -71,5 +71,5 @@ try:
     content = chat.create(QuestionPrompt(question="cake recipe?"))
     guide = chat.extract(HowToGuide, str(content))
     print(guide)
-except ValidationError as e:
-    print(e)
+except ValidationError:
+    print("Either the text was not a question or it is not relevant to the topic.")
--- a/examples/validator.py
+++ b/examples/validator.py
@@ -1,0 +1,75 @@
+"""A basic example showing how to use field validators to validate the input data.
+
+Note: All validators are defined in the class because updates to the validators will
+affect the output of the model.
+"""
+import os
+from typing import Union
+
+from pydantic import (
+    BaseModel,
+    Field,
+    ValidationError,
+    field_validator,
+)
+
+from mirascope import OpenAICallParams, OpenAIChat, Prompt
+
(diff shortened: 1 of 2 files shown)


examples/validator.py
```diff
@@ -0,0 +1,75 @@
...
+        ..., description="The steps to perform the task or answer the question."
+    )
+
+
+try:
+    content = chat.create(QuestionPrompt(question="cake recipe?"))
+    guide = chat.extract(HowToGuide, str(content))
+    print(guide)
+except ValidationError as e:
+    print(e)
```
willbakst, line 75: I think the only thing missing is showing how to handle the validation errors when it isn't a question or is off topic (i.e. response with hard-coded "That's not a question" or something)

examples/validator.py
```diff
@@ -0,0 +1,75 @@
...
+        ..., description="The steps to perform the task or answer the question."
+    )
+
+
+try:
+    content = chat.create(QuestionPrompt(question="cake recipe?"))
+    guide = chat.extract(HowToGuide, str(content))
+    print(guide)
+except ValidationError:
+    print("Either the text was not a question or it is not relevant to the topic.")
```
willbakst, line 75: I was thinking more like:

```python
except ValidationError as e:
    print(e)
    #> The text is not a question: cake recipe
```

brenkao: ah

examples/validator.py
```diff
@@ -0,0 +1,75 @@
...
+
+    topic: str = topic
+    question: str = Field(..., description="The user's question.")
+
+    call_params = OpenAICallParams(model="gpt-3.5-turbo", temperature=0.1)
+
+    @field_validator("question", mode="after")
+    @classmethod
+    def validate_question(cls, question: str) -> str:
+        class Text(BaseModel):
```
willbakst, line 42: marge into `ValidationInfo` and do single extract

willbakst: Follow-up:

What if we showed hitting a "fine-tuned" local model for validation before sending the request to API to scrub it of PII

willbakst (approved):

PR #110, opened 2024-03-01
feat: WandB prompt (initial gutcheck)

brianpark0126:


mirascope/integrations/__init__.py
brenkao, line 1: add from imports so we can do `from mirascope.wandb import WandbPrompt`

willbakst: no don't do this it will cause dependency issues.

Instead add just a docstring describing the module

mirascope/integrations/wandb/wandb_prompt.py
```diff
@@ -0,0 +1,90 @@
...
+        Returns:
+            The created error trace, connected to the parent.
+        """
+        span = Trace(
+            name=self.__class__.__name__,
+            kind=self.span_type,
+            status_code="error",
+            status_message=str(error),
+            metadata={"model": self.call_params.model},
+            start_time_ms=None,
```
brenkao, line 84: do errors not care about start time end time?

mirascope/integrations/wandb/wandb_prompt.py
```diff
@@ -0,0 +1,90 @@
+"""Prompts with WandB and OpenAI integration to support logging functionality."""
```
willbakst, line 1: please update the cookbook example to use this feature.

mirascope/integrations/wandb/wandb_prompt.py
```diff
@@ -0,0 +1,90 @@
...
+
+        Returns:
+            The created error trace, connected to the parent.
+        """
+        span = Trace(
+            name=self.__class__.__name__,
+            kind=self.span_type,
+            status_code="error",
+            status_message=str(error),
+            metadata={"model": self.call_params.model},
```
brenkao, line 83: why not pass in the entire call params to metadata

mirascope/integrations/__init__.py
brenkao: ooh gotcha, cuz wandb is extra

mirascope/integrations/wandb/wandb_prompt.py
willbakst: Interesting, there's no completion to retrieve the start/end time conveniently, but we do likely want it

mirascope/integrations/__init__.py
brianpark0126: so just to make sure, its fine to leave init files blank? And then to use we can do from mirascope.integrations.wandb import WandbPrompt

mirascope/integrations/wandb/wandb_prompt.py
brianpark0126: if we want it for errors, we would have to call get_time of some form before every call to a chat completion, since calculating it on the fly only gets the end time. do we want to implement start time and end time as optional parameters in the trace_error so that they users can pass it in if they choose to?

mirascope/integrations/__init__.py
brianpark0126: added the import to just the lowest level of directory since I thought that made the most sense for imports:
`from mirascope.integrations.wandb import WandbPrompt`
instead of 
`from mirascope.integrations.wandb.wandb_prompt import WandbPrompt`

mirascope/integrations/wandb/wandb_prompt.py
willbakst: What if the error threw the start time @brenkao ?

mirascope/integrations/__init__.py
willbakst, line 1: This is missing a docstring describing the module

mirascope/__init__.py
```diff
@@ -1,6 +1,7 @@
 """mirascope package."""
 import importlib.metadata
 
+from . import integrations
```
willbakst, line 4: is this necessary?

brianpark0126: whoops, left in from trying fixes earlier. removed now

mirascope/integrations/wandb/wandb_prompt.py
willbakst: Decision: generate start time on prompt creation and end time on trace_error call.

willbakst (approved):

PR #113, opened 2024-03-05
feat: OpenAIPrompt class similar to GeminiPrompt

willbakst:
- new `OpenAIPrompt` class for co-located functionality when working with OpenAI.
- Implements `create`, `async_create`, `stream`, `async_stream`, `extract`, and `async_extract`.

### DIFF
--- a/tests/openai/test_prompt.py
+++ b/tests/openai/test_prompt.py
@@ -1,16 +1,12 @@
 """Tests for mirascope openai prompt."""
-from typing import Type
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


mirascope/openai/utils.py
brianpark0126, line 1: is there any reason to not support multiple basemodels in the list of OpenAITools

willbakst: I'm not sure what this is referencing

brianpark0126 (approved):
pulled and tested extract and tools and normal create, all worked

brenkao (approved):

PR #114, opened 2024-03-06
Extract additional types

willbakst:
- Added support for extracted `BaseTool` variants
- Added support for extracting base types (e.g. str, int, etc.) with `Prompt.extract(str)`

### DIFF
--- a/mirascope/base/types.py
+++ b/mirascope/base/types.py
@@ -1,4 +1,6 @@
 """Types for working with Mirascope prompts."""
+from enum import Enum
+from inspect import isclass
 from typing import (
     Annotated,
     Any,
@@ -31,9 +33,13 @@ BaseType = Union[
 
 def is_base_type(type_: Any) -> bool:
     """Check if a type is a base type."""
-    return type_ in {str, int, float, bool, list, dict, set, tuple} or get_origin(
-        type_
-    ) in {Literal, Union, Annotated}
+    if isclass(type_) and issubclass(type_, Enum):
+        return True
+    if type_ in {str, int, float, bool, list, dict, set, tuple}:
+        return True
+    if get_origin(type_) in {Literal, Union, Annotated}:
+        return True
+    return False
 
 
 class BaseCallParams(BaseModel):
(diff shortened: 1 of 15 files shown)


mirascope/base/types.py
```diff
@@ -1,11 +1,46 @@
...
 
 from .tools import BaseTool
 
+BaseType = Union[
+    str,
+    int,
+    float,
+    bool,
+    list,
+    dict,
```
brianpark0126, line 28: extracting a dict gives Pydantic validation error:
pydantic_core._pydantic_core.ValidationError: 1 validation error for DictTool
value
  Field required [type=missing, input_value={'tool_call': ChatComplet...ool'), type='function')}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.6/v/missing

mirascope/openai/prompt.py
```diff
@@ -183,7 +187,19 @@ async def async_stream(self) -> AsyncGenerator[OpenAIChatCompletionChunk, None]:
                 chunk=chunk, tool_types=tools if tools else None
             )
 
+    @overload
+    def extract(self, schema: Type[BaseTypeT], retries: int = 0) -> BaseTypeT:
```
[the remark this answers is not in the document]
willbakst: these are just overloads for type hints. The main docstring for the non-overloaded definition shows up for all overloaded definitions :)

mirascope/base/types.py
willbakst: Ok after further testing and research, it appears that it just doesn't work with dict, so I've removed support for dict as a base class (the tool just isn't populated in this case).

Will punt to further research.

mirascope/base/types.py
```diff
@@ -1,11 +1,46 @@
...
 from .tools import BaseTool
 
+BaseType = Union[
+    str,
+    int,
+    float,
+    bool,
+    list,
+    set,
+    tuple,
```
brenkao, line 29: what about dict? or bytes?

willbakst: not handled. dict just doesn't work, bytes is also hard and not necessarily that useful, so I've excluded.

We can look to add them later, but it wasn't a simple addition

brenkao (approved):

PR #116, opened 2024-03-06
chore: updated examples to use new creation from prompt

brianpark0126:


examples/chat_history.py
```diff
@@ -1,49 +1,106 @@
...
-    history: list[Message] = []
+    history: list[ChatCompletionMessageParam] = []
 
     call_params = OpenAICallParams(model="gpt-3.5-turbo")
 
     @property
-    def messages(self) -> list[Message]:
+    def messages(self) -> list[ChatCompletionMessageParam]:
         return [
-            {"role": "system", "content": "You are a helpful AI."},
```
willbakst, line 21: you should be able to leave these as dictionaries, which is what people using OpenAI messages will expect. No need to use their strict typing.

examples/chat_history.py
```diff
@@ -1,49 +1,106 @@
 """A basic example showcasing how to manually manage chat history."""
 import os
```
willbakst, line 3: this has become complicated for no value add. Please revert to previous example but using OpenAIPrompt.

examples/function_as_tool.py
```diff
@@ -22,20 +22,19 @@ def get_current_weather(
...
 
-chat = OpenAIChat()
-completion = chat.create(CurrentWeatherPrompt())
+completion = CurrentWeatherPrompt().create()
 
 if tools := completion.tools:
     for tool in tools:
         if tool.fn:
-            print(tool.fn(**tool.__dict__))
+            print(tool.fn(**tool.model_dump(exclude={"tool_call"})))
```
willbakst, line 38: this will have merge conflicts with my recent updates. Please resolve. This should use `tool.args`

examples/openai_tool.py
```diff
@@ -24,20 +25,19 @@ class GetCurrentWeather(OpenAITool):
...
 
-chat = OpenAIChat()
-completion = chat.create(CurrentWeatherPrompt())
+completion = CurrentWeatherPrompt().create()
 
 if tools := completion.tools:
     for tool in tools:
         if tool.fn:
-            print(tool.fn(**tool.__dict__))
+            print(tool.fn(**tool.model_dump(exclude={"tool_call"})))
```
willbakst, line 41: same here

examples/chat_history.py
willbakst: Resolved

examples/chat_history.py
willbakst: Resolved.

examples/function_as_tool.py
brianpark0126: resolved

examples/openai_tool.py
brianpark0126: resolved

willbakst (approved):

PR #118, opened 2024-03-06
fix: updated wanbd prompt and docs to match new format

brianpark0126:


mirascope/wandb/prompt.py
```diff
@@ -29,22 +29,31 @@ class WandbPrompt(BasePrompt):
...
     greeting: str
 
-    prompt = HiPrompt(span_type="llm", greeting="Hello")
-    completion = chat.create(prompt)
+    prompt = HiPrompt(
+        api_key="YOUR_OPENAI_API_KEY",
+        span_type="llm",
+        greeting="Hello",
+    )
+    completion = prompt.create()
```
willbakst, line 43: What if we updated the `WandbPrompt.create` method so it looked like this:
```python
completion, span = prompt.create(parent_span)
```

Note that both `trace` and `trace_error` would then happen automatically, right?

PR #121, opened 2024-03-07
Wandb regen

brianpark0126:


cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -34,21 +32,28 @@ class Settings(BaseSettings):
...
     )
-    chat = OpenAIChat(api_key=settings.openai_api_key)
 
-    who_prompt = WhoPrompt(span_type="tool", person="Brian")
-    who_completion = chat.extract(Person, who_prompt)
-    who_span = who_prompt.trace(who_completion, root_span)
+    who_prompt = WhoPrompt(
+        span_type="tool",
+        person="Brian",
+        api_key=settings.openai_api_key,
```
willbakst, line 39: you should just run `os.environ["OPENAI_API_KEY"] = settings.openai_api_key` at the top. Same for the remaining prompts.

cookbook/wandb_logged_chain/wandb_prompts/party_invite_prompt.py
```diff
@@ -5,12 +5,11 @@
 class PartyInvitePrompt(WandbPrompt):
```
willbakst, line 5: here and elsewhere, update to remove `Prompt` from the names i.e. `PartyInvite(WandbPrompt)`

mirascope/wandb/prompt.py
```diff
@@ -1,50 +1,46 @@
...
 
     greeting: str
 
-    prompt = HiPrompt(span_type="llm", greeting="Hello")
-    completion = chat.create(prompt)
-    span = prompt.trace(completion, parent=root_span)
+    prompt = HiPrompt(
+        span_type="llm",
+        greeting="Hello",
+    )
```
willbakst, line 40: why is this multiple lines? looks like it could fit on one

mirascope/wandb/prompt.py
```diff
@@ -53,10 +49,66 @@ class HiPrompt(WandbPrompt):
...
+
+    @overload
+    def extract_with_trace(
+        self, schema: Type[BaseModelT], parent: Optional[Trace], retries: int = 0
+    ) -> tuple[BaseModelT, Trace]:
+        ...  # pragma: no cover
+
+    @overload
+    def extract_with_trace(
+        self, schema: Callable, parent: Optional[Trace], retries: int = 0
```
willbakst, line 86: change all of the `parent: Optional[Trace] = None` to have none as written here as the default

cookbook/wandb_logged_chain/wandb_chain.py
```diff
@@ -34,21 +32,28 @@ class Settings(BaseSettings):
...
     party_invite_prompt = PartyInvitePrompt(
-        span_type="llm", coolness=coolness_completion.coolness
+        span_type="llm",
+        coolness=coolness_completion.coolness,
+        api_key=settings.openai_api_key,
     )
-    party_completion = chat.create(party_invite_prompt)
-    party_span = party_invite_prompt.trace(party_completion, coolness_span)
+    party_completion, party_span = party_invite_prompt.create_with_trace(coolness_span)
+    print("\nparty completion:\n", party_completion)
```
willbakst, line 56: remove?

mirascope/wandb/prompt.py
```diff
@@ -73,10 +125,9 @@ def trace(
             if completion.tool:
                 output = {
                     "assistant": completion.tool.model_dump(),
-                    "tool_output": completion.tool.fn(
-                        **completion.tool.model_dump(exclude={"tool_call"})
-                    ),
+                    "tool_output": completion.tool.fn(**completion.tool.args),
                 }
+                print(output)
```
willbakst, line 130: remove?

mirascope/wandb/prompt.py
```diff
@@ -105,10 +156,11 @@ def trace(
             inputs={message["role"]: message["content"] for message in self.messages},
```
willbakst, line 156: on lines 154/155 we're using the start/end times in dump. Instead we should use self._start_time and self._end_time. Same is true for `_trace_error`

cookbook/wandb_logged_chain/wandb_chain.py
brianpark0126: yeah whoops, left that in from testing during tool issue

mirascope/wandb/prompt.py
brianpark0126: since create catches exceptions too now, going to delete the `WandbPrompt._creation_time_ms` attr as well since no longer necessary, same with dump

willbakst (approved):

PR #122, opened 2024-03-07
fix: issue with missing start/end time and handling excepts for end t…

willbakst:
- Added better handling for setting start and end time for extract and errors
- Added back start and end time for completions (so you can see diff between construction, calling, return, etc.)

### DIFF
--- a/mirascope/openai/prompt.py
+++ b/mirascope/openai/prompt.py
@@ -73,9 +73,6 @@ class OpenAIPrompt(BasePrompt):
         AfterValidator(lambda key: _set_api_key(key) if key is not None else None),
     ] = None
 
-    _start_time: Optional[float] = None  # The start time of the completion in ms
-    _end_time: Optional[float] = None  # The end time of the completion in ms
-
     call_params: ClassVar[OpenAICallParams] = OpenAICallParams(
         model="gpt-3.5-turbo-0125",
     )
--- a/mirascope/openai/prompt.py
+++ b/mirascope/openai/prompt.py
@@ -108,11 +108,18 @@ class OpenAIPrompt(BasePrompt):
             tools.extend(self.call_params.tools)
         converted_tools = convert_tools_list_to_openai_tools(tools)
         patch_openai_kwargs(kwargs, self, converted_tools)
-        completion = client.chat.completions.create(
-            model=self.call_params.model,
-            stream=False,
-            **kwargs,
-        )
+        try:
+            completion_start_time = datetime.datetime.now().timestamp() * 1000
+            completion = client.chat.completions.create(
+                model=self.call_params.model,
+                stream=False,
+                **kwargs,
+            )
+            completion._end_time = datetime.datetime.now().timestamp() * 1000
+            completion._start_time = completion_start_time
+        except Exception:
+            self._end_time = datetime.datetime.now().timestamp() * 1000
+            raise
         self._end_time = datetime.datetime.now().timestamp() * 1000
         return OpenAIChatCompletion(completion=completion, tool_types=converted_tools)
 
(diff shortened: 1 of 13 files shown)


brianpark0126 (approved):
lgtm

mirascope/openai/prompt.py
```diff
@@ -308,22 +337,33 @@ async def async_extract(self, schema, retries=0):
             openai_tool = OpenAITool.from_model(schema)
             return_tool = False
 
-        completion = await self.async_create(tools=[openai_tool])
+        try:
```
brenkao, line 340: define self._start_time here? then you can remove 342 and 344 and 328

mirascope/openai/prompt.py
```diff
@@ -308,22 +337,33 @@ async def async_extract(self, schema, retries=0):
...
+            completion = await self.async_create(tools=[openai_tool])
+            self._start_time = extraction_start_time
+        except Exception:
+            self._end_time = datetime.datetime.now().timestamp() * 1000
+            self._start_time = extraction_start_time
+            raise
         try:
             tool = completion.tool
             if tool is None:
+                self._end_time = datetime.datetime.now().timestamp() * 1000
```
brenkao, line 350: why do all these end times need to be defined in each conditional. shouldnt end time be when completion is complete?

mirascope/openai/prompt.py
willbakst: create and async_create also set start time, but that one doesn't take into account all of the time spend converting tools, so I did it this way

mirascope/openai/prompt.py
willbakst: These are all above return statements, so it's right when it completes. The completion also has _start_time and _end_time for when it completes

brianpark0126 (commented):
encountered some errors when I just ran the code, essentially completion.dump() always gives None for start_time and end_time, and then theres an issue where when you do prompt.dump(completion.dump()) there's a doubling up of start_time and start_time_ms and same for end times, except start_time and end_time, are None. Snippet with outputs structured for reference: 
```
import os

# import openai api key from .env

from mirascope import tags
from mirascope.openai import OpenAIPrompt

load_dotenv()


@tags(["recommendation_project", "version:0001"])
class BookRecommendationPrompt(OpenAIPrompt):
    """
    Can you recommend some books on {topic}?
    """

    topic: str


prompt = BookRecommendationPrompt(topic="how to bake a cake")
comp = prompt.create()
print(comp.dump(), "\n\n")
print(prompt.dump(comp.dump()))
output1 = {
    "start_time": None,
    "end_time": None,
    "output": {
        "id": "chatcmpl-90FfY1MlXcuiwYITvNcy9RjASDB47",
        "choices": [
            {
                "finish_reason": "stop",
                "index": 0,
                "logprobs": None,
                "message": {
                    "content": '1. "The Cake Bible" by Rose Levy Beranbaum - ...
                    "role": "assistant",
                    "function_call": None,
                    "tool_calls": None,
                },
            }
        ],
        "created": 1709847260,
        "model": "gpt-3.5-turbo-0125",
        "object": "chat.completion",
        "system_fingerprint": "fp_b9d4cef803",
        "usage": {"completion_tokens": 176, "prompt_tokens": 19, "total_tokens": 195},
    },
}

output2 = {
    "template": "Can you recommend some books on {topic}?",
    "inputs": {"api_key": None, "topic": "how to bake a cake"},
    "tags": ["recommendation_project", "version:0001"],
    "call_params": {"model": "gpt-3.5-turbo-0125"},
    "start_time_ms": 1709847488599.576,
    "end_time_ms": 1709847491881.78,
    "start_time": None,
    "end_time": None,
    "output": {
        "id": "chatcmpl-90FjE0iXj8CY7i4eDbHkgHONRVuJi",
        "choices": [
            {
                "finish_reason": "stop",
                "index": 0,
                "logprobs": None,
                "message": {
                    "content": '1. "The Cake Bible" by Rose Levy Beranbaum\n2. ...
                    "role": "assistant",
                    "function_call": None,
                    "tool_calls": None,
                },
            }
        ],
        "created": 1709847488,
        "model": "gpt-3.5-turbo-0125",
        "object": "chat.completion",
        "system_fingerprint": "fp_2b778c6b35",
        "usage": {"completion_tokens": 140, "prompt_tokens": 19, "total_tokens": 159},
    },
}
```

mirascope/openai/prompt.py
brenkao: we are timing the actual completion, not the time it takes to do conversion.

mirascope/openai/prompt.py
brenkao: i understand that, but the end time should be set when the completion finishes, not when you hit the conditional

willbakst: Resolved, only completion has it now.

mirascope/openai/prompt.py
willbakst: Only completion has it now

willbakst:
The mentioned bug should no longer be an issue. Start/end time available in the completion only as required parameters

brianpark0126 (approved):

PR #133, opened 2024-03-08
feat: AnthropicPrompt and related convenience. NO TOOLS.

willbakst:
- Added `AnthropicPrompt` convenience wrapper with `create` and `stream` functionality.
- Due to their use of in-response XML function calling, I've decided to split up the work and get this in first and work on tools later.
- Will likely add `async_create` and `async_stream` methods to both Gemini and Anthropic, as well as `async_extract` to Gemini before working on tools.


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+  "workbench.colorTheme": "Default Dark Modern",
```
brenkao, line 2: remove?

docs/concepts/using_different_model_providers.md
```diff
@@ -7,7 +7,7 @@ We currently support the following providers:
 - OpenAI
 - Gemini
 - Mistral (coming soon...)
-- Claude (coming soon...)
+- Anthropic (coming soon...)
```
brenkao, line 10: maybe partially implemented? so people know they can use create/stream

brenkao (approved):

.vscode/settings.json
willbakst: This is the default for VSCode but not cursor and I wanted the same theme...

PR #138, opened 2024-03-14
BREAKING: Interface changes for v0.4 to incorporate recent feedback

willbakst:
This PR introduces breaking changes to the previous API. These changes come from some great feedback we received on the v0.3 release noting that our "prompts" were really "calls" that contain the prompt functionality as part of the call.

As part of this interface update, I took the liberty to also remove/change some functionality as I thought made sense:
- There are no longer model-specific prompt classes (e.g. `OpenAIPrompt`). Instead, there is a `BasePrompt` class that acts as the base class for prompting needs. Model specific "prompts" are now "calls" (e.g. `OpenAICall`), which extend the `BaseCall` interface.
- I have (temporarily) removed the parsers for streaming tools. If I need to wait for a full tool, then for a single tool call response streaming and calling are no different. For now, I've chosen to remove since updating to be truly streaming is not insignificant work but is a far more valuable feature worth implementing in the (near) future. Ultimately depends on Python support of something like `TypedDict.partial` or `Partial` along with Pydantic support for those updates.
- Additionally, I've removed the `extract` function from calls. It felt weird passing in a schema to the function that was not colocated with the call. Thus, we now instead have extractors that extend the `BaseExtractor` interface (e.g. `OpenAIExtractor`) and are dependent on / tied to the extraction schema through Generic and the `extract_schema` class variable.
- Prompt templates are no longer written as docstrings. Instead, they are written as a `prompt_template` class variable. This ensures that the originally functionality of the docstring remains untouched. The prompt template parsing functionality remains the same, except that we no longer do any "magic" around newlines for line width. Users expected adding a newline to include a newline, so the functionality was lost. Better will be to offer that parsing as an opt-in feature so it's more clear what's going on.
- Classes no longer override the `__str__` method. In the same vein as above, it seemed to make the most sense not to override the expected functionality. It was also clear that `str(response)` made less sense to users than `response.content`.
- I have removed all cookbook and examples. We will be re-adding relevant, high quality examples as follow-up work. For now, we will have the examples live in the README.
- From the Concepts section, I have removed the "Why we're building Mirascope section" as it needs a rewrite (as a Philosophy page or something).
- Possibly additional things that aren't coming to mind as I write this...

WIP:
- [X] MESSAGES injection
- [X] I need to update the API documentation to work with new structure
- [X] The Concepts section still needs to be updated (namely updating code examples).
- [X] The README still needs to be updated (namely with updated examples/usage etc.).
- [X] Port examples back over with new interfaces.


mirascope/base/prompts.py
```diff
@@ -0,0 +1,184 @@
...
+    prompt_template: ClassVar[str] = ""
+
+    def __str__(self) -> str:
+        """Returns the formatted template."""
+        return self._format_template(self.prompt_template)
+
+    def messages(self) -> Union[list[Message], Any]:
+        """Returns the template as a formatted list of messages."""
+        message_type_by_role = {
+            "system": SystemMessage,
```
brenkao, line 61: create enum for system, user assistant, model, tool. then use enum as keys here. also in gemini/calls.py, openai/calls.py, and anthropic/calls.py

mirascope/base/prompts.py
```diff
@@ -0,0 +1,184 @@
...
+        messages = []
+        re_roles = "|".join(
+            [role.upper() for role in roles] + ["MESSAGES"] + ["[A-Z]*"]
+        )
+        for match in re.finditer(
+            rf"({re_roles}):((.|\n)+?)(?=({re_roles}):|\Z)",
+            self.prompt_template,
+        ):
+            role = match.group(1).lower()
+            if role == "messages":
```
brenkao, line 113: update to use enum

mirascope/base/prompts.py
```diff
@@ -0,0 +1,184 @@
...
+                messages += getattr(self, template_var)
+            else:
+                if role not in roles:
+                    raise ValueError(f"Invalid role: {role}")
+                content = self._format_template(match.group(2))
+                messages.append({"role": role, "content": content})
+        if len(messages) == 0:
+            messages.append(
+                {
+                    "role": "user",
```
brenkao, line 135: update to use enum

mirascope/gemini/calls.py
```diff
@@ -0,0 +1,167 @@
...
+        Args:
+            **kwargs: Additional keyword arguments that will be used for generating the
+                response. These will override any existing argument settings in call
+                params.
+
+        Returns:
+            A `GeminiCallResponse` instance.
+        """
+        kwargs, tool_types = self._setup(kwargs, GeminiTool)
+        gemini_pro_model = GenerativeModel(kwargs.pop("model"))
```
brenkao, line 83: GenerativeModel takes in `model_name` instead of `model`. Our call params for GeminiCallParams uses `model`. Are we basically saying that we will always use `gemini-pro` and that `gemini-1.0-pro` which is our default isn't being used?

mirascope/gemini/calls.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+    response = BookRecommender(genre="fantasy").call()
+    print(response.call())
+    #> As the world's greatest librarian, I am delighted to recommend...
+    ```
+    '''
+
+    call_params: ClassVar[GeminiCallParams] = GeminiCallParams(
+        model="gemini-1.0-pro",
```
brenkao, line 55: nit: make models into constants default_gemini_model, default_openai_model, etc.

mirascope/anthropic/calls.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Args:
+            **kwargs: Additional keyword arguments parameters to pass to the call. These
+                will override any existing arguments in `call_params`.
+
+        Returns:
+            A `AnthropicCallResponse` instance.
+        """
+        messages, kwargs = self._setup_anthropic_kwargs(kwargs)
+        client = Anthropic(base_url=self.base_url)
```
brenkao, line 54: anthropic takes api_key. make consistent with openai?

mirascope/anthropic/types.py
```diff
@@ -11,66 +11,133 @@
...
 
 
-    print(BookRecommendation().create())
+    print(BookRecommender().call())
     ```
-    '''
+    """
+
+    # TOOLS NOT YET SUPPORTED
+    tool_types: Annotated[InstanceOf[None], BeforeValidator(lambda x: None)] = None
```
brenkao, line 90: tool_types is already defined in BaseCallResponse, is this necessary?

brenkao (commented):
Since we're cleaning up files, we can probably delete Partial and test_partial

mirascope/gemini/calls.py
willbakst: it's not keyworded, so it'll set model_name to `kwargs.pop("model")` which will come either from the kwargs the user provides OR the call params (kwargs take priority as overrides)

mirascope/anthropic/types.py
willbakst: This is forcing it to be None since we don't support it.

mirascope/anthropic/calls.py
willbakst: Done.

mirascope/gemini/calls.py
willbakst: I've refactored to single source in types as the default instead.

mirascope/gemini/calls.py
willbakst: I'm updating it to keyword `model_name=kwargs.pop("model_name")` for consistency and updating the gemini call params to use `model_name` instead of `model`

willbakst: Scratch that lol can't use `model_name` it's protected

mirascope/base/prompts.py
willbakst: Done.

mirascope/base/prompts.py
willbakst: Done.

mirascope/base/prompts.py
willbakst: Done.

mirascope/gemini/calls.py
willbakst: I'm still setting it as a keyword arg now tho for clarity

brenkao: gotcha

mirascope/anthropic/types.py
brenkao: gotcha

brenkao (approved):

PR #146, opened 2024-03-18
feat: mistral with extractors, still missing stream_async test

brianpark0126:


mirascope/mistral/types.py
```diff
@@ -0,0 +1,208 @@
...
+    def message(self) -> ChatMessage:
+        """Returns the message of the chat completion for the 0th choice."""
+        return self.choice.message
+
+    @property
+    def content(self) -> str:
+        """The content of the chat completion for the 0th choice."""
+        if isinstance(self.message.content, str):
+            return self.message.content
+        else:
```
willbakst, line 91: update to:

```python
if isinstance(self.message.content, str):
    return self.message.content
# We haven't seen the `list[str]` response type in practice,
# so for now we return the first item if it's a list
return self.message.content[0]
```

mirascope/mistral/types.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def dump(self) -> dict[str, Any]:
+        """Dumps the chat completion to a dictionary."""
+        return {
+            "start_time": self.start_time,
+            "end_time": self.end_time,
+            "output": self.response.model_dump(),
+        }
+
+    def __str__(self):
```
willbakst, line 143: remove -- new interface doesn't override str method except for `BasePrompt` specifically for printing the prompt

mirascope/mistral/types.py
```diff
@@ -0,0 +1,208 @@
...
+    def content(self) -> str:
+        """Returns the content of the delta."""
+        return self.delta.content if self.delta.content is not None else ""
+
+    @property
+    def tool_calls(self) -> Optional[list[ToolCall]]:
+        """Returns the partial tool calls for the 0th choice message."""
+        return self.delta.tool_calls
+
+    def __str__(self):
```
willbakst, line 206: ditto

willbakst (approved):

PR #148, opened 2024-03-19
Anthropic Function Calling Support Fixes

willbakst:
- Fixed issue with nested types like list, set, tuple, and dict
- Added support for def/ref so things like Enums and other references work
- Schema generation and parsing is all recursive, so it can theoretically handle whatever nesting structure you throw at it.

Still need to:
- [X] Write tests to get 100% coverage

### DIFF
--- a/mirascope/anthropic/tools.py
+++ b/mirascope/anthropic/tools.py
@@ -18,7 +18,11 @@ BaseTypeT = TypeVar("BaseTypeT", bound=BaseType)
 
 
 class AnthropicTool(BaseTool[ET.Element]):
-    '''
+    '''A base class for easy use of tools with the Anthropic Claude client.
+
+    `AnthropicTool` internally handles the logic that allows you to use tools with
+    simple calls such as `AnthropicCallResponse.tool` or `AnthropicTool.fn`, as seen in
+    the example below.
 
     Example:
 
@@ -40,7 +44,7 @@ class AnthropicTool(BaseTool[ET.Element]):
 
 
     class AnimalMatcher(AnthropicCall):
-        prompt_template = """\\
+        prompt_template = """
         Tell me my favorite animal if my favorite food is {food} and my
         favorite color is {color}.
         """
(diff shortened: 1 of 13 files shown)


mirascope/anthropic/tools.py
```diff
@@ -109,20 +103,47 @@ def from_tool_call(cls, tool_call: ET.Element) -> AnthropicTool:
...
-        parameters_node = tool_call.find("parameters")
-        if parameters_node is not None:
-            parameters.update(
-                {
-                    parameter_node.tag: parameter_node.text
-                    for parameter_node in parameters_node
-                }
-            )
+
+        def _parse_xml_element(element: ET.Element) -> Union[str, dict, list]:
```
brenkao, line 107: is this intentionally a nested function?

mirascope/anthropic/tools.py
```diff
@@ -139,3 +160,118 @@ def from_fn(cls, fn: Callable) -> Type[AnthropicTool]:
...
+            schema_xml += f"<name>{prop}</name>\n"
+            schema_xml += _process_property(definition, parent_type)
+            schema_xml += "</parameter>\n"
+        schema_xml += "</parameters>\n"
+    elif "items" in schema:
+        schema_xml += _process_property(schema, parent_type)
+    return schema_xml
+
+
+def _process_property(
```
brenkao, line 190: any reason we are using string concat vs using ET subelements?
feels like string concat can be very error prone. perhaps something to do for next iteration

brenkao (approved):

mirascope/anthropic/tools.py
willbakst: yeah, it's only used internally to this function and is private, and I thought it was easier to read. I pulled the other ones out because they got too long to be nested cleanly

mirascope/anthropic/tools.py
willbakst: this is ultimately to save on tokens to avoid using all of the indentation that the ET subelements functionality includes. Maybe I can do it with no indentation? For now going to leave, will add a TODO for later.

PR #158, opened 2024-03-23
feat: wandb support model agnostic mixin class

willbakst:
- Refactored WandbOpenAICall and WandbOpenAIExtractor into mixin classes instead.
- This makes them model agnostic :)

### DIFF
--- a/mirascope/wandb.py
+++ b/mirascope/wandb.py
@@ -2,35 +2,94 @@
 from __future__ import annotations
 
 import datetime
-from typing import Any, Generic, Literal, Optional, TypeVar, Union
+from abc import abstractmethod
+from typing import (
+    Any,
+    AsyncGenerator,
+    ClassVar,
+    Generator,
+    Generic,
+    Literal,
+    Optional,
+    Type,
+    TypeVar,
+    Union,
+)
 
 from pydantic import BaseModel
 from wandb.sdk.data_types.trace_tree import Trace
 
-from ..base import BasePrompt, BaseType, ExtractedType
-from ..openai import (
-    OpenAICall,
-    OpenAICallResponse,
-    OpenAIExtractor,
-    OpenAITool,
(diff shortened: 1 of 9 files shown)


mirascope/wandb.py
```diff
@@ -162,21 +219,24 @@ def call_with_trace(
...
-            span = trace(self, response, parent, **kwargs)
-            return response, span
+            response = self.call(**kwargs)
+            tool_type = None
+            if response.tool_types and len(response.tool_types) > 0:
+                tool_type = response.tool_types[0].__bases__[0]  # type: ignore
+            span = trace(self, response, tool_type, parent, **kwargs)
+            return response, span  # type: ignore
         except Exception as e:
             return None, trace_error(self, e, parent, start_time, **kwargs)
```
brenkao, line 229: I know this was like this b4 but, is there a reason we dont raise?

willbakst: Because we're catching and handling the error here. So to check for error you check if response is None (and the error is traced to wandb)

mirascope/wandb.py
```diff
@@ -200,7 +261,7 @@ class Book(BaseModel):
         author: str
 
 
-    class BookRecommender(WandbOpenAIExtractor[Book]):
+    class BookRecommender(OpenAIExtractor[Book], WandbExtractorMixin[Book]):
```
brenkao, line 264: Does `WandbExtractorMixin[OpenAIExtractor[Book]]` make more sense?

brenkao (approved):

mirascope/wandb.py
willbakst: Isn't possible. The mixin is adding a method to the `OpenAIExtractor[Book]` class. The reason we have `WandbExtractorMixin[Book]` is to type the `extract_with_trace` function with the correct return type.

PR #169, opened 2024-03-29
Feat: Integrate Weave with Mirascope

willbakst:
- Add `with_weave` decorator for easily incorporating weave monitoring into your Mirascope workflows.
- Works with `call`, `call_async`
- Semi-works with `stream` and `stream_async` (it logs our calls, but not the model calls. this is due to how streaming works, and I'm not sure how weave handles it yet).
- Extraction is blocked by a bug with weave, but the code is ready to go once the bug is resolved.

NOTE:
This PR has the library depend on version v0.44.0 of Weave, which has a critical bug failure with pydantic.BaseModel, but the newer version hasn't been released yet. Will wait to merge/release this feature until the version with the fix is released.


mirascope/openai/calls.py
```diff
@@ -68,6 +68,7 @@ def call(self, **kwargs: Any) -> OpenAICallResponse:
             OpenAIError: raises any OpenAI errors, see:
                 https://platform.openai.com/docs/guides/error-codes/api-errors
         """
+        # kwargs, tool_types = self._setup(kwargs, OpenAITool)
```
brenkao, line 71: remove?

mirascope/weave.py
brenkao, line 1: should this be in a wandb directory?

brenkao (approved):

mirascope/openai/calls.py
willbakst: good catch. removed.

mirascope/weave.py
willbakst: done.

PR #172, opened 2024-04-03
Fix errors and openai dict

### LINKED ISSUE #168: [BUG] Dictionary Type not working with OpenAI Extractor
Not sure if a bug or a feature request, as it might be related to OpenAI. Extracting to a Pydantic model that contains a dictionary throws validation errors when Pydantic tries to convert the returned string to a model.

**To Reproduce**
Copy and Paste these code blocks into a notebook.
Fill in your API key on this notebook, run the cells.

```
from typing import Dict, List, Type
from mirascope.openai import OpenAICall, OpenAICallParams, OpenAIExtractor
import json
import os

os.environ["OPENAI_API_KEY"] = "<Api-Key>"

class FoodExamples(BaseModel):
    foods_and_examples : Dict[str, List[str]]

class FoodExamplesExtractor(OpenAIExtractor[FoodExamples]):
    extract_schema: Type[FoodExamples] = FoodExamples
    prompt_template = """
    Extract the food types and their specific examples from the following json string:
    {json_string}
    """
    json_string : str 
    call_params = OpenAICallParams(model = "gpt-4")

simple_dict = {
    "cheese" : [
        "parmesean",
        "pecorino",
        "mozzarella"
    ],
    "bread" : [
        "sourdough",
        "whole wheat",
        "rye"
    ]
}

food_examples = FoodExamplesExtractor(json_string=json.dumps(simple_dict)).extract()
assert isinstance(food_examples, FoodExamples)
print(food_examples)
```

willbakst:
Closes #168 

- Raise RuntimeError when stop reason isn't function calls when using `tools` and `tool` convenience functions.
- Enable using `response_format={"type": "json_object"}` when using tools and extraction.
- Update Anthropic system prompt when using extract to force it to use a tool correctly.

### DIFF
--- a/mirascope/anthropic/types.py
+++ b/mirascope/anthropic/types.py
@@ -102,7 +102,9 @@ class AnthropicCallResponse(BaseCallResponse[Message, AnthropicTool]):
                 "Generation stopped before `</function_calls>` stop sequence. "
                 "This is likely due to a limit on output tokens that is too low. "
                 "Note that this could also indicate no tool is beind called, so we "
-                "recommend that you check the output of the call to confirm."
+                "recommend that you check the output of the call to confirm. "
+                f"Stop Reason: {self.response.stop_reason} "
+                f"Stop Sequence: {self.response.stop_sequence}"
             )
 
         try:
--- a/mirascope/anthropic/types.py
+++ b/mirascope/anthropic/types.py
@@ -1,4 +1,5 @@
 """Type classes for interacting with Anthropics's Claude API."""
+
 import xml.etree.ElementTree as ET
 from typing import Any, Callable, Literal, Optional, Type, Union, cast
 
@@ -90,12 +91,19 @@ class AnthropicCallResponse(BaseCallResponse[Message, AnthropicTool]):
     @property
     def tools(self) -> Optional[list[AnthropicTool]]:
         """Returns the tools for the 0th choice message."""
+        if not self.tool_types:
+            return None
+
         if (
-            not self.tool_types
-            or self.response.stop_reason != "stop_sequence"
+            self.response.stop_reason != "stop_sequence"
             or self.response.stop_sequence != "</function_calls>"
         ):
-            return None
+            raise RuntimeError(
+                "Generation stopped before `</function_calls>` stop sequence. "
+                "This is likely due to a limit on output tokens that is too low. "
+                "Note that this could also indicate no tool is beind called, so we "
+                "recommend that you check the output of the call to confirm."
+            )
 
         try:
             root_node = ET.fromstring(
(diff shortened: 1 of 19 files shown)


mirascope/anthropic/types.py
```diff
@@ -90,12 +91,19 @@ class BookRecommender(AnthropicCall):
...
-            not self.tool_types
-            or self.response.stop_reason != "stop_sequence"
+            self.response.stop_reason != "stop_sequence"
             or self.response.stop_sequence != "</function_calls>"
         ):
-            return None
+            raise RuntimeError(
+                "Generation stopped before `</function_calls>` stop sequence. "
+                "This is likely due to a limit on output tokens that is too low. "
+                "Note that this could also indicate no tool is beind called, so we "
```
brenkao, line 104: would be nice to have a copy and pastable command that people can use to check. something like call.response.stop_reason or something.

willbakst: I should just include it in the text. good call

mirascope/openai/calls.py
```diff
@@ -185,3 +203,31 @@ async def stream_async(
...
+        if self.call_params.response_format == {"type": "json_object"} and tool_types:
+            kwargs.pop("tools")
+        return kwargs, tool_types
+
+    def _maybe_update_messages(
+        self,
+        messages: list[ChatCompletionMessageParam],
+        tool_types: Optional[list[type[OpenAITool]]],
+    ) -> list[ChatCompletionMessageParam]:
+        if self.call_params.response_format == {"type": "json_object"} and tool_types:
```
brenkao, line 227: technically json is the format so we can omit the object part

willbakst: https://platform.openai.com/docs/guides/text-generation/json-mode

See line 6 of the code block in this section

mirascope/openai/calls.py
```diff
@@ -185,3 +203,31 @@ async def stream_async(
...
+        dict[str, Any],
+        Optional[list[Type[OpenAITool]]],
+    ]:
+        """Overrides the `BaseCall._setup` for Anthropic specific setup."""
+        kwargs, tool_types = self._setup(kwargs, OpenAITool)
+        if self.call_params.response_format == {"type": "json_object"} and tool_types:
+            kwargs.pop("tools")
+        return kwargs, tool_types
+
+    def _maybe_update_messages(
```
brenkao, line 222: change function name to be `_update_message_if_json` so its more clear than maybe

brenkao (approved):

mirascope/openai/calls.py
willbakst: Used OpenAI's ResponseFormat object

mirascope/openai/calls.py
willbakst: Done.

PR #174, opened 2024-04-05
Feat: Groq Support

willbakst:
- Added `GroqCall`, `GroqExtractor`, and other related classes
- Supports Groq like all other model providers.


mirascope/openai/calls.py
```diff
@@ -198,10 +195,9 @@ async def stream_async(
         )
         async for chunk in stream:
             openai_call_response_chunk = OpenAICallResponseChunk(
```
brenkao, line 197: yield directly?

brenkao (approved):

mirascope/openai/calls.py
willbakst: Done.

PR #175, opened 2024-04-05
bugfix: added missing groq call to cli

brenkao:


mirascope/cli/utils.py
```diff
@@ -30,6 +30,7 @@
     "GeminiCall",
     "AnthropicCall",
     "WandbOpenAICall",
+    "GroqCall",
```
willbakst, line 33: please also bump fix version to 0.9.1 in pyproject

willbakst (approved):

PR #176, opened 2024-04-08
feat: rag

brenkao:


examples/rag/state_of_the_union_rag.py
```diff
@@ -0,0 +1,51 @@
+import os
+
+from dotenv import load_dotenv
+
+from mirascope.base import BaseChunker
+from mirascope.base.types import BaseChunkerParams
+from mirascope.chroma.vectorstores import ChromaVectorStore
+from mirascope.openai import OpenAICall, OpenAIEmbedder
+
+load_dotenv()
```
willbakst, line 10: pydantic-settings?

examples/rag/state_of_the_union_rag.py
```diff
@@ -0,0 +1,51 @@
...
+    SYSTEM: Answer the question based on the context.
+    {context}
+    USER: {question}
+    """
+
+    question: str
+
+    @property
+    def context(self):
+        with open(f"{current_dir}/state_of_the_union.txt") as file:
```
willbakst, line 37: this should already be done separately outside of the property, right?

Instead user should define a vector store field and use that, right? e.g.

```python
@property
def context(self) -> list[str]:
    return [doc.text for doc in self.store.query(self.query)]
```

mirascope/base/embedders.py
```diff
@@ -0,0 +1,26 @@
...
+
+from .types import BaseEmbeddingParams
+
+BaseEmbeddingT = TypeVar("BaseEmbeddingT", bound=BaseModel)
+
+
+class BaseEmbedder(BaseModel, Generic[BaseEmbeddingT], ABC):
+    api_key: ClassVar[Optional[str]] = None
+    base_url: ClassVar[Optional[str]] = None
+    embedding_params: ClassVar[BaseEmbeddingParams] = BaseEmbeddingParams(
```
willbakst, line 14: this should either be `embedder_params` and `BaseEmbedderParams` to match `chunker_params` and `BaseChunkerParams` or we should switch to use `ChunkingParams` and `chunking_params`

I prefer the latter.

mirascope/base/vectorstores.py
```diff
@@ -0,0 +1,30 @@
...
+
+from mirascope.base.chunkers import BaseChunker
+from mirascope.base.embedders import BaseEmbedder
+from mirascope.openai.embedders import OpenAIEmbedder
+
+from .types import BaseVectorStoreParams, Document
+
+
+class BaseVectorStore(BaseModel, ABC):
+    vectorstore_api_key: Optional[str] = None
```
willbakst, line 14: why not just `api_key`?

mirascope/base/vectorstores.py
```diff
@@ -0,0 +1,30 @@
...
+    index_name: ClassVar[Optional[str]] = None
+    chunker: ClassVar[BaseChunker] = BaseChunker()
+    embedder: ClassVar[Optional[BaseEmbedder]] = OpenAIEmbedder()
```
willbakst, line 17: this shouldn't rely on openai

mirascope/base/vectorstores.py
```diff
@@ -0,0 +1,30 @@
...
+    vectorstore_params: ClassVar[BaseVectorStoreParams] = BaseVectorStoreParams()
+
+    @abstractmethod
+    def get_documents(
```
willbakst, line 21: do we prefer `get_documents` to `query` and `add_documents` to `upload`?

mirascope/chroma/vectorstores.py
```diff
@@ -0,0 +1,64 @@
...
+
+    vectorstore_params = ChromaParams(get_or_create=True)
+    client_settings: ClassVar[ClientSettings] = ClientSettings()
+
+    @property
+    def _client(self) -> ClientAPI:
+        return Client(self.client_settings)
+
+    @property
+    def _index(self) -> Collection:
```
willbakst, line 23: please organize private attributes/fields/properties/methods in a private section. See other implementations e.g. `BaseExtractor` for reference

mirascope/openai/embedders.py
```diff
@@ -0,0 +1,46 @@
...
+from openai.types import CreateEmbeddingResponse, Embedding
+
+from ..base import BaseEmbedder
+from .types import OpenAIEmbeddingParams
+
+
+class OpenAIEmbedder(BaseEmbedder[CreateEmbeddingResponse]):
+    embedding_params: ClassVar[OpenAIEmbeddingParams] = OpenAIEmbeddingParams()
+
+    def _embed(self, input: str) -> CreateEmbeddingResponse:
```
willbakst, line 13: ditto

mirascope/base/types.py
```diff
@@ -186,3 +186,79 @@ def content(self) -> str:
...
+
+    model: str
+    model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
+
+    def kwargs(
+        self,
+        exclude: Optional[set[str]] = None,
+    ) -> dict[str, Any]:
+        """Returns all parameters for the embedder as a keyword arguments dictionary."""
+        extra_exclude = {"tools", "weave"}
```
willbakst, line 229: unnecessary for these call params, right?

mirascope/base/types.py
```diff
@@ -186,3 +186,79 @@ def content(self) -> str:
...
+    """The parameters with which to make a vectorstore."""
+
+    model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
+
+    def kwargs(
+        self,
+        exclude: Optional[set[str]] = None,
+    ) -> dict[str, Any]:
+        """Returns all parameters for the index as a keyword arguments dictionary."""
+        extra_exclude = {"tools", "weave"}
```
willbakst, line 249: ditto

mirascope/openai/embedders.py
```diff
@@ -0,0 +1,46 @@
...
+        client = AsyncOpenAI(api_key=self.api_key, base_url=self.base_url)
+        embedding_params = self.embedding_params.model_copy(update={"input": input})
+        kwargs = embedding_params.kwargs()
+        return await client.embeddings.create(**kwargs)
+
+    async def embed_async(self, inputs: list[str]) -> list[CreateEmbeddingResponse]:
+        """Asynchronously call the embedder with multiple inputs"""
+        embedding_responses: list[CreateEmbeddingResponse] = []
+        for input in inputs:
+            embedding_responses.append(await self._embed_async(input))
```
willbakst, line 38: use `asyncio.gather` so that they all run in parallel (right now it will await each one in sequence, losing a lot of the value of running async on the embeddings in the first place)

mirascope/base/chunkers.py
```diff
@@ -0,0 +1,28 @@
...
+from typing import Callable, ClassVar
+
+from pydantic import BaseModel
+
+from mirascope.base.types import BaseChunkerParams, Document
+
+
+def base_chunker(text: str, params: BaseChunkerParams) -> list[Document]:
+    chunks: list[Document] = []
+    chunk_size = params.chunk_size or 0
```
willbakst, line 11: chunk size `0` doesn't make sense?

mirascope/base/types.py
```diff
@@ -186,3 +186,79 @@ def content(self) -> str:
...
+    chunk_size: Optional[int] = None
+    chunk_overlap: Optional[int] = None
+    length_function: Optional[Callable[[str], int]] = None
+    keep_separator: bool = False
+    add_start_index: bool = False
+    strip_whitespace: bool = True
+
+    model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
+
+    def kwargs(
```
willbakst, line 203: I don't believe we use this anywhere?

examples/rag/state_of_the_union_rag.py
brenkao: i can convert if we want to keep it, but i was thinking about using an example different from langchain

examples/rag/state_of_the_union_rag.py
brenkao: this is because this is a non-persistent chroma instance. so if you use `PersistentClient` or `HttpClient` then you are correct

mirascope/base/embedders.py
brenkao: okay I can switch to ChunkingParams

examples/rag/state_of_the_union_rag.py
willbakst: What do you mean? Do they use pydantic-settings?

I think regardless we should use pydantic-settings in our examples or directly set `os.environ["API_KEY"]`

mirascope/base/vectorstores.py
brenkao: I figured since we were using api_key for llm providers, it could be confusing but maybe the context of it being in a vectorstore is enough clarity?

mirascope/base/vectorstores.py
brenkao: what should our base embedder be

mirascope/base/types.py
brenkao: if we use openai embedder, we should have weave?

examples/rag/state_of_the_union_rag.py
willbakst: But even if it isn't persistent, it's still persistent in memory right? The user won't want to load each document every time the call is made. And using a computed field / cached_property here doesn't save us since it will make a new call if anything (e.g. a field) has changed.

Can a user not define a non-persistent index and pass that in as an attribute? That way they only have to create the index once?

mirascope/base/types.py
brenkao: i can remove

mirascope/base/vectorstores.py
willbakst: Yeah I think context is enough

mirascope/base/vectorstores.py
willbakst: Just BaseEmbedder? This class can't be instantiated anyway since it has abstractmethod methods

willbakst: See `BaseCall` for reference

mirascope/base/vectorstores.py
brenkao: its doing more than just `query`. upload is not really a database word so i'm against that specifically.

mirascope/base/types.py
willbakst: These are fields of `BaseCallParams` that don't exist on this object.

The `weave` field gets updated if you use the `with_weave` decorator, which currently supports `BaseCall` and `BaseExtractor` classes.

We should likely add support for `with_weave` with all of our rag classes. I'm happy to take care of that once we get the initial feature for RAG merged in.

mirascope/base/vectorstores.py
willbakst: Ok. I don't love `get_documents` because it loses the intent behind the word `query` and sounds more like getting all of the documents in the vector store.

What more beyond `query` is it doing?

As for not liking `upload` what about just `add`?

So it would be `store.query` and `store.add`

mirascope/openai/embedders.py
brenkao: done

mirascope/chroma/vectorstores.py
brenkao: do private properties get a PRIVATE PROPERTY section, or throw it all in PRIVATE METHODS

mirascope/base/chunkers.py
brenkao: updated to 1000

mirascope/base/vectorstores.py
```diff
@@ -0,0 +1,30 @@
...
+from mirascope.base.embedders import BaseEmbedder
+from mirascope.openai.embedders import OpenAIEmbedder
+
+from .types import BaseVectorStoreParams, Document
+
+
+class BaseVectorStore(BaseModel, ABC):
+    vectorstore_api_key: Optional[str] = None
+    index_name: ClassVar[Optional[str]] = None
+    chunker: ClassVar[BaseChunker] = BaseChunker()
```
willbakst, line 16: realizing this won't accept a chunker function instead of the class?

mirascope/base/chunkers.py
```diff
@@ -0,0 +1,28 @@
...
+        chunks.append(Document(text=text[start:end], id=str(uuid.uuid4())))
+        start += chunk_size - chunk_overlap
+    return chunks
+
+
+class BaseChunker(BaseModel):
+    chunker_params: ClassVar[BaseChunkerParams] = BaseChunkerParams(
+        chunk_size=1000, chunk_overlap=0
+    )
+    chunk_fn: Callable[[str, BaseChunkerParams], list[Document]] = base_chunker
```
willbakst, line 25: I feel like this should be a class variable? Also what do you think about `chunker` instead of `chunk_fn`? Or name it `base_chunk_fn` instead of `base_chunker`.

examples/rag/state_of_the_union_rag.py
brenkao: this example isnt the final one thats going to be pushed to dev. e.g. the state of the union, so I was going to scrap this once RAG is complete. the final one will use pydantic settings

examples/rag/state_of_the_union_rag.py
brenkao: this is scratch work, the final example will not use the normal chroma client

mirascope/base/types.py
brenkao: okay i will remove for now

mirascope/base/vectorstores.py
brenkao: hm they create a chunker class with params. and pass in the function to the chunker class, then pass the chunker class to the vectorstore. is this not what we want?

mirascope/base/chunkers.py
brenkao: oh yeah, it should be a classvar. is `chunker `clear enough?

willbakst: My only concern is overloading `chunker` and `Chunker`, so maybe `chunk_fn` is good and we should just use `base_chunk_fn`

mirascope/base/vectorstores.py
willbakst: I though we also wanted to allow for passing the function in directly and inferring the chunker by inspecting the function (i.e. set `chunking_params` to whatever the default value for `chunking_params` in the function is, and if no default is set use `BaseChunkingParams`?

mirascope/base/chunkers.py
```diff
@@ -0,0 +1,33 @@
+import uuid
+from typing import Any, Callable, Optional
+
+from pydantic import BaseModel
+
+
+class Document(BaseModel):
```
willbakst, line 7: why not put this in a `documents.py`?

mirascope/base/chunkers.py
willbakst, line 1: should we put all of this stuff under `mirascope.base.rag`?

mirascope/chroma/vectorstores.py
willbakst: Hmm, thinking through this more it doesn't make sense for `_client` and `_index` to be properties at all? You're going to create a client / collection every time?

My initial instinct is to used `cached_property` instead so that it get's cached on the first call. I was hoping maybe we could use `PrivateAttr` or something, but the `default_factory` can't access `cls` or `self`...

And yeah, I would probably have two separate sections, `PRIVATE PROPERTIES` above existing `PRIVATE METHODS`

mirascope/base/chunkers.py
brenkao: putting inside `rag/types.py`

examples/rag_on_wikipedia_article/retrieve_documents.py
```diff
@@ -0,0 +1,32 @@
...
+
+if settings.anthropic_api_key:
+    os.environ["ANTHROPIC_API_KEY"] = settings.anthropic_api_key
+if settings.openai_api_key:
+    os.environ["OPENAI_API_KEY"] = settings.openai_api_key
+
+
+class WikipediaCall(AnthropicCall):
+    prompt_template = """
+    SYSTEM: Answer the question based on the context.
```
willbakst, line 17: for multiline prompt strings, always start on the next line e.g.
```python
SYSTEM:
Answer the question based on the context:
{context}
```

This will ensure `dedent` works properly.

examples/rag_on_wikipedia_article/retrieve_documents.py
```diff
@@ -0,0 +1,32 @@
...
+    {context}
+    USER: {question}
+    """
+
+    question: str
+
+    @property
+    def context(self):
+        store = WikipediaStore()
```
willbakst, line 26: is there any concern with creating the store instance every call?

For example, in a call with history where I'll update an attribute and make another call this will create a new store instance each call instead of just creating the instance on class instantiation.

examples/rag_on_wikipedia_article/stores/wikipedia_chroma.py
```diff
@@ -0,0 +1,26 @@
+import uuid
+
+from mirascope.chroma import ChromaSettings, ChromaVectorStore
+from mirascope.openai import OpenAIEmbedder
+from mirascope.rag import BaseChunker, Document
+
+
+class MyChunker(BaseChunker):
```
willbakst, line 8: maybe a better name than `MyChunker` that's actually descriptive of what it's doing?

mirascope/openai/embedders.py
```diff
@@ -0,0 +1,50 @@
...
+    def __call__(self, input: list[str]) -> list[list[float]]:
+        embeddings: list[Embedding] = [
+            embedding for response in self.embed(input) for embedding in response.data
+        ]
+        sorted_embeddings = sorted(embeddings, key=lambda e: e.index)
+        return [result.embedding for result in sorted_embeddings]
+
+    ############################## PRIVATE METHODS ###################################
+
+    def _embed(self, input: str) -> CreateEmbeddingResponse:
```
willbakst, line 38: any benefit refactoring these as private methods? seem short enough to just have the main methods

mirascope/rag/chunkers.py
```diff
@@ -0,0 +1,31 @@
...
+    start: int = 0
+    while start < len(text):
+        end: int = min(start + chunk_size, len(text))
+        chunks.append(Document(text=text[start:end], id=str(uuid.uuid4())))
+        start += chunk_size - chunk_overlap
+    return chunks
+
+
+class BaseChunker(BaseModel):
+    """"""
```
willbakst, line 26: missing docstr?

mirascope/rag/chunkers.py
```diff
@@ -0,0 +1,31 @@
...
+
+from .types import Document
+
+
+def base_chunk_fn(
+    text: str,
+    chunk_size: int = 1000,
+    chunk_overlap: int = 0,
+) -> list[Document]:
+    """Chunk the text into chunks of size `chunk_size`."""
```
willbakst, line 15: also mention "and overlap `chunk_overlap`"

mirascope/rag/chunkers.py
```diff
@@ -0,0 +1,31 @@
...
+        start += chunk_size - chunk_overlap
+    return chunks
+
+
+class BaseChunker(BaseModel):
+    """"""
+
+    chunk_fn: Callable[Concatenate[str, ...], list[Document]] = base_chunk_fn
+
+    def chunk(self, text: str) -> list[Document]:
```
willbakst, line 30: I'm thinking we should make this an abstract method and drop `chunk_fn` entirely i.e. require a subclass implement the `chunk` function.

If we want, we can include something like `SimpleChunker` or `DumbChunker` that actually implements `base_chunk_fn` as an OOB class. Thoughts?

examples/rag_on_wikipedia_article/retrieve_documents.py
brenkao: it will just retrieve the chroma collection instance.

mirascope/rag/chunkers.py
brenkao: this is removed now

mirascope/rag/chunkers.py
brenkao: done

mirascope/openai/embedders.py
brenkao: mostly if you want to pass in one instead of a list

examples/rag_on_wikipedia_article/stores/wikipedia_chroma.py
brenkao: done

examples/rag_on_wikipedia_article/retrieve_documents.py
brenkao: done

mirascope/rag/chunkers.py
brenkao: `rag/chunkers/text_chunker` ?

mirascope/openai/embedders.py
willbakst: ok, fine to keep refactored then

mirascope/rag/chunkers.py
willbakst: Do we want to then import `from mirascope.rag.chunkers import TextChunker` or do we just want to example like you current have for the wikipedia store?

The former would mean importing the chunker from mirascope instead of defining it in the example

brenkao: they can import premade from chunks e.g. textchunker or import basechunker for extending purposes. the textchunker can then be recursive chunker or whatever chunker

willbakst: Ok, then update the example to import `TextChunker` once refactored

willbakst (approved):

mirascope/rag/chunkers/text.py
willbakst, line 1: rename this `text_chunker.py`?

mirascope/rag/chunkers/base.py
willbakst, line 1: rename to `base_chunker.py`?

willbakst (approved):

PR #177, opened 2024-04-08
feat: enable easier injection of list attributes and properties

willbakst:
- Instead of having to write your own custom property, we will automatically handle `list` attributes by injecting each string in the list into the prompt separated by "\n"
- If the attribute is a list of non-string elements (e.g. list of list), we will inject the string output of each item when calling `str(item)` on it.

### DIFF
--- a/tests/base/test_prompts.py
+++ b/tests/base/test_prompts.py
@@ -147,6 +147,6 @@ def test_base_prompt_list_attribute() -> None:
         my_list_of_lists: list[list[str]]
 
     my_list = ["my", "list"]
-    my_list_of_lists = [my_list]
+    my_list_of_lists = [my_list, my_list]
     prompt = MyPrompt(my_list=my_list, my_list_of_lists=my_list_of_lists)
-    assert str(prompt) == "my\nlist\n['my', 'list']"
+    assert str(prompt) == "my\nlist\nmy\nlist\n\nmy\nlist"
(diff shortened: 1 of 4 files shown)


tests/base/test_prompts.py
```diff
@@ -132,3 +132,21 @@ class MyPrompt(BasePrompt):
...
+        {my_list_of_lists}
+        """
+
+        my_list: list[str]
+        my_list_of_lists: list[list[str]]
+
+    my_list = ["my", "list"]
+    my_list_of_lists = [my_list]
+    prompt = MyPrompt(my_list=my_list, my_list_of_lists=my_list_of_lists)
+    assert str(prompt) == "my\nlist\n['my', 'list']"
```
brenkao, line 152: hmm so we don't flatten lists of lists? With RAG, I have list[list[str]], depending on how many queries a user makes for similarity. that entire context will be put in the prompt.

willbakst: I guess I could flatten further, and then separate with "\n\n" but when does it end? Do we support `list[list[list[str]]]`?

willbakst: I figured for something like a nested list, if you want more than just `str(inner_list)` that you would want more control than the default anyway? e.g.

```python
@property
def formatted_list_of_lists(self) -> list[str]:
    return [
        f"Index: {i}\n{"\n".join([item for item in l])}\n"
        for i, l in enumerate(self.my_list_of_lists)
    ]
```

but I guess it would still be convenient for us to at least have some default handling of `list[list[...]]`?

brenkao: i think we can stop at `list[list[...]]` because thats a real use case.

brenkao (approved):

tests/base/test_prompts.py
willbakst: Done.

PR #178, opened 2024-04-10
Feat: Streaming Tools

willbakst:
- You can now stream tools one by one when the LLM returns multiple tools.
- You can also stream partial tools. This will return partial tools with `None` returns in between each tool so you can handle switching to subsequent tool calls even while streaming partial.
- NOTE: it seems like OpenAI is the only provider that reliably implements streaming tool calls, so until other providers offer this support this functionality will be limited to OpenAI.

TODO:
- [X] Add support for streaming tools to OpenAIExtractor

Example Script For Fun Testing:
```python
import time

from rich.live import Live

from mirascope.openai import OpenAICall, OpenAICallParams, OpenAIToolStream


def print_book(title: str, author: str, description: str):
    """Prints the title and author of a book."""
    return f"Title: {title}\nAuthor: {author}\nDescription: {description}"


class BookRecommender(OpenAICall):
    prompt_template = "Please recommend some books to read."

    call_params = OpenAICallParams(tools=[print_book])


tool_stream = OpenAIToolStream.from_stream(
    BookRecommender().stream(), allow_partial=True
)

with Live("", refresh_per_second=15) as live:
    partial_tools, index = [None], 0
    previous_tool = None
    for partial_tool in tool_stream:
        if partial_tool is None:
            index += 1
            partial_tools.append(None)
            continue
        partial_tools[index] = partial_tool
        live.update(
            "\n----------\nBook Recommendation\n----------\n".join(
                [pt.fn(**pt.args) for pt in partial_tools]
            )
        )
        time.sleep(0.1)
```


mirascope/base/extractors.py
```diff
@@ -66,6 +73,18 @@ async def extract_async(self, retries: int = 0) -> ExtractedTypeT:
         """Asynchronously extracts the `extraction_schema` from an LLM call."""
         ...  # pragma: no cover
 
+    # @abstractmethod
```
brenkao, line 76: why commented?

willbakst: This would require adding it to every extractor, but few support streaming tools, so leaving commented out until we expect all providers to have this functionality (i.e. don't want to have `stream` implemented for `MistralExtractor` but just throw `NotImplementedError`.

mirascope/openai/tool_streams.py
```diff
@@ -0,0 +1,216 @@
...
+        allow_partial: Literal[False],
+    ) -> Generator[OpenAITool, None, None]:
+        yield ...  # type: ignore  # pragma: no cover
+
+    @classmethod
+    @overload
+    def from_stream(
+        cls,
+        stream: Generator[OpenAICallResponseChunk, None, None],
+        allow_partial: bool = False,
```
brenkao, line 133: why do you need this if you cover literal true and false above

willbakst: because those don't cover `bool`... thanks mypy

mirascope/base/extractors.py
```diff
@@ -194,6 +211,181 @@ class TempCall(call_type):  # type: ignore
...
+                    yield partial_tool
+            raise  # re-raise if we have no retries left
+
+    def _extract_schema(
+        self,
+        tool: Optional[BaseToolT],
+        schema: ExtractedType,
+        return_tool: bool,
+        response: Optional[Any],
+    ) -> Optional[ExtractedTypeT]:
```
brenkao, line 373: should the return type be `Optional[Union[ExtractedTypeT, BaseToolT]]`

mirascope/base/extractors.py
brenkao: put comment explaining abv

mirascope/base/extractors.py
willbakst: Technically BaseToolT is part of ExtractedTypeT but mypy is annoying

mirascope/base/extractors.py
willbakst: Done.

brenkao (approved):

PR #187, opened 2024-04-17
feat: added cost tracking to each provider except Gemini

brenkao:


mirascope/base/types.py
```diff
@@ -130,6 +130,7 @@ class BaseCallResponse(BaseModel, Generic[ResponseT, BaseToolT], ABC):
     tool_types: Optional[list[Type[BaseToolT]]] = None
     start_time: float  # The start time of the completion in ms
     end_time: float  # The end time of the completion in ms
+    cost: Optional[float] = None  # The cost of the completion in dollars
```
willbakst, line 133: is this optional because of Gemini? I would rather this be required and then just be `0.0` for gemini since we'll note it isn't supported in the docs

mirascope/base/types.py
willbakst, line 1: Docs are missing. please mention this in the `CallResponse` section in the generating content concept page at the very least

willbakst (approved):

mirascope/base/types.py
brenkao: no

willbakst: Note: sometimes usage is none.

mirascope/base/types.py
willbakst: Docs update coming in separate PR

PR #195, opened 2024-04-23
feat: added cohere embeddings and updated openai embeddings

brenkao:
* Added CohereEmbedder and various types
* Updated OpenAIEmbedder to use batch similar to CohereEmbedder


mirascope/cohere/types.py
```diff
@@ -288,3 +290,40 @@ def tool_calls(self) -> Optional[list[ToolCall]]:
...
+    """A convenience wrapper around the Cohere `EmbedResponse` response."""
+
+    @property
+    def embeddings(
+        self,
+    ) -> Optional[Union[list[list[float]], list[list[int]]]]:
+        """Returns the embeddings"""
+        if isinstance(self.response.embeddings, list):
+            return self.response.embeddings
+        else:
```
willbakst, line 305: can you explain what's happening here?

mirascope/openai/types.py
```diff
@@ -280,10 +268,25 @@ def tool_calls(self) -> Optional[list[ChoiceDeltaToolCall]]:
...
+    def embeddings(self) -> list[list[float]]:
+        """Returns the raw embeddings."""
+        embeddings_model: list[Embedding] = [
+            embedding for embedding in self.response.data
+        ]
+        return [embedding.embedding for embedding in embeddings_model]
+
+
+class OpenAIEmbeddingParams(BaseEmbeddingParams):
+    model: str = "text-embedding-ada-002"
```
willbakst, line 284: I feel like we should update the default to the most recent one?

mirascope/rag/types.py
```diff
@@ -22,7 +22,7 @@ class BaseEmbeddingResponse(BaseModel, Generic[ResponseT], ABC):
 
     @property
     @abstractmethod
-    def embedding(self) -> list[float]:
+    def embeddings(self) -> Optional[Union[list[list[float]], list[list[int]]]]:
```
willbakst, line 25: we're allowing int because cohere supports int?

willbakst (approved):
Minor things and questions

mirascope/rag/types.py
brenkao: yes

mirascope/cohere/types.py
willbakst: Use the `embedding_type`

e.g. (rough draft)

```
embedding_type = "float_" if self.embedding_params.embedding_type == "float" else self.embedding_params.embedding_type
return self.respopnse.embeddings.model_dump()[self.embedding_params.embedding_type]
```

mirascope/cohere/types.py
```diff
@@ -288,3 +291,41 @@ def tool_calls(self) -> Optional[list[ToolCall]]:
...
+            embedding_type = self.embedding_type
+            if embedding_type == "float":
+                embedding_type == "float_"
+
+            # TODO: Update to model_dump when Cohere updates to Pydantic v2
+            embeddings_by_type: EmbedByTypeResponseEmbeddings = self.response.embeddings
+            embedding_dict = embeddings_by_type.dict()
+            return embedding_dict.get(str(embedding_type), None)
```
willbakst, line 318: why not just `return self.response.embeddings.dict().get(embedding_type, None)`?

Do we have to convert to string if it's literal and not None? Also how do we handle None?

mirascope/cohere/types.py
```diff
@@ -288,3 +291,41 @@ def tool_calls(self) -> Optional[list[ToolCall]]:
         if isinstance(self.chunk, StreamedChatResponse_ToolCallsGeneration):
             return self.chunk.tool_calls
         return None
+
+
+class CohereEmbeddingResponse(BaseEmbeddingResponse[SkipValidation[EmbedResponse]]):
+    """A convenience wrapper around the Cohere `EmbedResponse` response."""
+
+    embedding_type: Optional[
```
willbakst, line 299: seems confusing to have this here as well as the sequence of `embedding_types` in the params?

willbakst (approved):

mirascope/cohere/types.py
brenkao: mostly mypy bc embedding_type is a Literal and get expects a str

mirascope/cohere/types.py
brenkao: dont have access to embedding_types in the response, i can pass the whole thing instead and grab the first one in the property instead.

willbakst: Oh I see. I think this is fine then since we're saying to access additional embeddings you can just access to original response and this is convenience for the first one.

PR #197, opened 2024-04-28
Add code for AstraDB

Tylersuard:
Allows use of AstraDB, similar to other databases already included.

Test plan:  Use the example code in mirascope/astra/vectorstore.py


mirascope/astra/vectorstores.py
```diff
@@ -0,0 +1,91 @@
...
+            return AstraDB(
+                api_endpoint=self.client_settings.api_endpoint,
+                token=self.client_settings.application_token,
+            )
+        except Exception as e:
+            logging.error(f"Failed to initialize AstraDB client: {e}")
+            raise
+
+    @cached_property
+    def _collection(self):
```
brenkao, line 85: For our interface we use `def _index` which would be equivalent to Astra's collections. I'm not sure which is the more widely used term yet but for now lets use `_index`.

mirascope/astra/vectorstores.py
```diff
@@ -0,0 +1,91 @@
...
+        my_store = MyStore()
+        with open("test_document.txt") as file:
+            data = file.read()
+            my_store.add(data)
+            documents = my_store.retrieve("Ask a question about the document you uploaded here.").documents
+        print(documents)
+            """
+
+    client_settings: ClassVar[AstraSettings] = AstraSettings()
+    collection_name: ClassVar[str] = "default_collection"  # Default value, can be overridden
```
brenkao, line 39: index_name is already defined in `BaseVectorStore` so this is unnecessary, but feel free to keep the default name.

mirascope/astra/vectorstores.py
```diff
@@ -0,0 +1,91 @@
...
+            )
+        except Exception as e:
+            logging.error(f"Failed to initialize AstraDB client: {e}")
+            raise
+
+    @cached_property
+    def _collection(self):
+        """Access or create the collection based on the name defined in settings."""
+        try:
+            return self._client.collection(self.collection_name)
```
brenkao, line 88: see comment regarding `collection_name`, also pass in vectorstore_params here. see `chroma/vectorstores.py` for a similar pattern.

mirascope/astra/vectorstores.py
```diff
@@ -0,0 +1,91 @@
...
+            }
+            self._collection.insert_one(document_to_insert)
+
+    ############################# PRIVATE PROPERTIES #################################
+
+    @cached_property
+    def _client(self) -> AstraDB:
+        """Instantiate and return an AstraDB client configured from settings."""
+        try:
+            return AstraDB(
```
brenkao, line 76: pass in client_settings via `**self.client_settings.kwargs()`

mirascope/astra/types.py
```diff
@@ -0,0 +1,23 @@
+from pydantic import BaseModel
+from typing import Any, Optional, Literal
+
+class AstraSettings(BaseModel):
```
brenkao, line 4: Add all astra settings
```
token: str
api_endpoint: str
api_path: Optional[str] = None
api_version: Optional[str] = None
namespace: Optional[str] = None
caller_name: Optional[str] = None
caller_version: Optional[str] = None
 ```

File: mirascope/astra/vectorstores.py [hunk 6]
```diff
@@ -0,0 +1,91 @@
...
+
+    def retrieve(self, text: Optional[Union[str, list[str]]] = None, **kwargs: Any) -> AstraQueryResult:
+        """Queries the vectorstore for closest match"""
+        embedded_query = self.embedder(text)[0] if text else None
+        results = self._collection.vector_find(
+            embedded_query,
+            limit=kwargs.get('limit', 10),  # Example of additional parameter
+            fields=["text", "source"]
+        )
+        return AstraQueryResult.convert(results)
```
brenkao, line 50: Remove convert and construct the pydantic model here.
```
return AstraQueryResult(documents=..., embeddings=...)
```

mirascope/astra/vectorstores.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def add(self, text: Union[str, list[Document]], **kwargs: Any) -> None:
+        """Takes unstructured data and upserts into vectorstore.
+        Each document is expected to have text, embeddings, and a source.
+        """
+        if not text:
+            logging.error("No text provided for addition.")
```
brenkao, line 57: you can raise ValueError here instead of logging, user can then try except as necessary and handle it themselves with logger.

brenkao: but also the arguments requires text, so maybe unnecessary?

pyproject.toml
```diff
@@ -28,6 +28,7 @@ mistralai = { version = ">=0.1.6,<1.0.0", optional = true }
 groq = { version = ">=0.4.2,<1.0.0", optional = true }
 cohere = { version = "^5.2.5", optional = true }
 pinecone-client = { version = "^3.2.2", optional = true }
+astrapy = "^1.1.0"
```
brenkao, line 31: add astra as an optional dependency
```
poetry add --optional astrapy
```
Then update `[tool.poetry.extras]` with
```
astrapy = ["astrapy"]
```
and update `all` to include `astrapy`

mirascope/astra/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""A module for interacting with Astra vectorstores."""
+from .types import AstraParams, AstraQueryResult, AstraSettings
+from .vectorstores import AstraVectorStore
```
brenkao, line 3: add `__all__ = [ "AstraParams", ...]`

brenkao (changes requested):
Let me know if you have any questions or if there is any clarification needed.

Thanks for contributing.

mirascope/astra/vectorstores.py
willbakst: Yeah this is not necessary since the typing requires text. If not provided, this will throw an error anyway that the user needs to resolve rather than log.

mirascope/astra/types.py
willbakst, line 1: Please add docstrings to the module and all public classes and functions. They should follow the [Google docstring style](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

mirascope/astra/vectorstores.py
```diff
@@ -0,0 +1,91 @@
...
+from astrapy.db import AstraDB
+from ..rag.types import Document
+from ..rag.vectorstores import BaseVectorStore
+from .types import AstraParams, AstraQueryResult, AstraSettings
+
+class AstraVectorStore(BaseVectorStore):
+    """A vector store for AstraDB.
+    
+    Example:
+        from mirascope.openai import OpenAIEmbedder
```
willbakst, line 15: please wrap the example in a python code block, e.g.

Example:

\```python
{code here}
\```

willbakst (changes requested):
This PR is missing tests. The tests should be located in a `tests/astra` directory.

Please note that all code should be tested with 100% coverage. Ideally we also test additional flows beyond just coverage, but this is not _required_ for a PR (and we will add additional tests should bugs be discovered in the future).

To check coverage, the easiest way is to run the following command from the root directory:
```bash
poetry run pytest tests --cov=./ --cov-report-html
```

This will produce a directory named `htmlcov` inside of which there will be an `index.html` file that is interactive and shows you all uncovered lines by file.

I recommend looking in the `tests/` directory for a reference of how to write tests. For this PR, I would specifically look at the [chroma tests](https://github.com/Mirascope/mirascope/tree/dev/tests/chroma) and the [pinecone tests](https://github.com/Mirascope/mirascope/tree/dev/tests/pinecone).

If you have any questions, let us know! We're here to help.

pyproject.toml
```diff
@@ -31,6 +31,7 @@ pinecone-client = { version = "^3.2.2", optional = true }
 logfire = { version = ">=0.26.0,<1.0.0", optional = true }
```
willbakst, line 31: please include `astrapy` with the correct version and `optional = true` like we do for other optional dependencies.

PR #202, opened 2024-04-30
Expose `with_logfire` from mirascope.logfire

barapa:
Add `with_logfire` to the `__all__` array in order to expose `with_logfire` from mirascope.logfire.

The static type checker pyright does not recognize the following import statement:

```
from mirascrope.logfire import with_logfire
```

And provides the following error:
```
error: "with_logfire" is not exported from module "mirascope.logfire"
    Import from "mirascope.logfire.logfire" instead (reportPrivateImportUsage)
```

In order to match the mirascope + logfire docs, adding the method to the __all__ field should make it importable. This may be something that you would want to take on across this repo.

### DIFF
--- a/mirascope/logfire/__init__.py
+++ b/mirascope/logfire/__init__.py
@@ -1,2 +1,4 @@
 """Module for integrations with Pydantic Logfire"""
 from .logfire import with_logfire
+
+__all__ = ["with_logfire"]


brenkao (approved):

willbakst:
Codecov is rate limiting us, I will rerun tests when we're out of the limit

willbakst:
@barapa I'm curious how any of the other from imports are working? e.g. `from mirascope.anthropic import AnthropicCall` or do those also break?

barapa:
Those don't work either. Instead I use:
```python
from mirascope.anthropic.extractors import AnthropicExtractor
from mirascope.anthropic.types import AnthropicCallParams
```

brenkao (approved):

barapa:
This may not be the approach we want actually. See this comment in the pyright repo:
https://github.com/microsoft/pyright/issues/5872#issuecomment-1703276682

What we may want instead is:

```python
from .logfire import with_logfire
```

willbakst:
Good find! I think we should stick to the `__all__` approach since that seems like best practice wrt. wildcard support (as well as what many other open-source libraries I respect seem to do).

Currently trying to figure out why Codecov keeps rate limiting us / failing here. Seems to be due to something with external PRs, so I will work to resolve this so we can merge this in and release it (and fix your errors when following the docs)

willbakst:
I cannot figure out the codecov issue for the life of me, but this doesn't affect coverage so I'm just going to force merge (since all the tests are passing, it's just the codecov issue).

willbakst:
@barapa this is released as well as the same changes for the other modules :)

PR #206, opened 2024-05-01
bugfix: fix logfire hierarchy & gemini

brenkao:
* fixed hierarchy for call_async, stream, stream_async
* made gemini pretty on logfire ui

### DIFF
--- a/mirascope/gemini/calls.py
+++ b/mirascope/gemini/calls.py
@@ -79,7 +79,8 @@ class GeminiCall(BaseCall[GeminiCallResponse, GeminiCallResponseChunk, GeminiToo
             A `GeminiCallResponse` instance.
         """
         kwargs, tool_types = self._setup(kwargs, GeminiTool)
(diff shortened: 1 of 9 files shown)


mirascope/gemini/calls.py
```diff
@@ -126,6 +128,8 @@ async def call_async(self, **kwargs: Any) -> GeminiCallResponse:
             generate_content_async = self.call_params.logfire(
                 generate_content_async, "gemini", response_type=GeminiCallResponse
             )  # pragma: no cover
+        else:
+            kwargs.pop("model")
```
willbakst, line 132: I wonder if a better way to do this would to be still to just pop it and then instead of if/else just re-add it to the kwargs in the if statement of logfire?

willbakst (approved):

mirascope/gemini/calls.py
brenkao: done

PR #208, opened 2024-05-02
feat: better retries

brenkao:


mirascope/base/extractors.py
```diff
@@ -125,22 +144,35 @@ def _extract(
         kwargs, return_tool = self._setup(tool_type, kwargs)
 
         class TempCall(call_type):  # type: ignore
-            prompt_template = self.prompt_template
+            prompt_template = """
+            {original_prompt_template}
+
+            {error_message}
```
willbakst, line 150: this will include unnecessary additional `\n` characters even if `error_message` is empty

mirascope/base/extractors.py
```diff
@@ -125,22 +144,35 @@ def _extract(
         kwargs, return_tool = self._setup(tool_type, kwargs)
 
         class TempCall(call_type):  # type: ignore
```
willbakst, line 146: this is out of sync with new temp call generator function, please update

mirascope/base/extractors.py
```diff
@@ -125,22 +144,35 @@ def _extract(
...
             model_config = ConfigDict(extra="allow")
 
         properties = getmembers(self)
         for name, value in properties:
             if not hasattr(TempCall, name):
                 setattr(TempCall, name, value)
-
-        response = TempCall(**self.model_dump(exclude={"extract_schema"})).call(
-            **kwargs
+        error_message = kwargs.pop("error_message", "")
```
willbakst, line 164: all of this error message stuff should be pushed into the temp call generator for reusability

mirascope/base/extractors.py
willbakst, line 1: we should add similar handling for call/stream as well (e.g. to handle rate limits better).

Ideally we could do this in a way where we don't have to repeat code, potentially by including it in BaseCall or creating a resuable context manager or something?

mirascope/base/extractors.py
```diff
@@ -149,11 +180,8 @@ class TempCall(call_type):  # type: ignore
                 raise AttributeError("No tool found in the completion.")
             return extracted_schema
         except (AttributeError, ValueError, ValidationError) as e:
-            if retries > 0:
-                logging.info(f"Retrying due to exception: {e}")
-                # TODO: include failure in retry prompt.
-                return self._extract(call_type, tool_type, retries - 1, **kwargs)
-            raise  # re-raise if we have no retries left
+            logging.info(f"Retrying due to exception: {e}")
```
willbakst, line 183: if we use logfire do these logs get captured, or should we call `logfire.info` in that case?

mirascope/base/extractors.py
```diff
@@ -89,12 +90,30 @@ async def extract_async(self, retries: int = 0) -> ExtractedTypeT:
     #     ...  # pragma: no cover
 
     ############################## PRIVATE METHODS ###################################
+    def _retry(
```
willbakst, line 93: I don't understand the point of this function being separate. Ultimately this will require updating it everywhere vs. just updating the `_extract` method to handle retries internally so every provider automatically works and we only have to make changes in the base extractor?

This is specific to `_extract` anyway, and `_extract_async`, `_stream`, and `_stream_async` will require their own specific handling anyway, so seems best to just embed this retry functionality into each of those functions as a replacement of our current recursive method?

mirascope/base/extractors.py
```diff
@@ -125,22 +144,34 @@ def _extract(
...
         properties = getmembers(self)
         for name, value in properties:
             if not hasattr(TempCall, name):
                 setattr(TempCall, name, value)
-
-        response = TempCall(**self.model_dump(exclude={"extract_schema"})).call(
-            **kwargs
-        )
+        error_message = kwargs.pop("error_message", "")
+        if error_message and error_message == "No tool found in the completion":
```
willbakst, line 165: this is unnecessary, just pass the error message in and tell it to fix it (it should identify that "no tool" and "please fix" means "use tool")

mirascope/base/extractors.py
```diff
@@ -125,22 +144,34 @@ def _extract(
...
 
         properties = getmembers(self)
         for name, value in properties:
             if not hasattr(TempCall, name):
                 setattr(TempCall, name, value)
-
-        response = TempCall(**self.model_dump(exclude={"extract_schema"})).call(
-            **kwargs
-        )
+        error_message = kwargs.pop("error_message", "")
```
willbakst, line 164: given other comment about just adding retry functionality here instead of refactoring it (since we can't share code anyway), we won't need to pop from kwargs, instead we can just forward it as an actual parameter of the function that defaults to the empty string

mirascope/base/extractors.py
brenkao: I think the user would try except the extract function and use log fire info and we can remove this line?

willbakst: This is logging each retry, which the user won't see, so they would only see their final log that there was an exception, not that there were X retries attempted in the logs.

If this shows up in logfire with no changes, great. Otherwise we should try to make it show up.

brenkao: This will not show up

mirascope/base/extractors.py
```diff
@@ -28,6 +31,10 @@
 from .tools import BaseTool, BaseType
 from .types import BaseCallParams
 
+with suppress(ImportError):
+    from logfire.integrations.logging import LogfireLoggingHandler
```
willbakst, line 35: does this mean that if they have logfire installed and there is a failure that it will always log to logfire even if they don't use the with_logfire decorator?

mirascope/base/extractors.py
```diff
@@ -303,42 +373,71 @@ async def _stream_async(
...
+                        raise
+        except RetryError as e:
+            raise e
+
+    def _generate_temp_call(
+        self, call_type: Type[BaseCallT], error_message: Optional[str] = None
+    ) -> Type[BaseCallT]:
         """Returns a `TempCall` generated using the extractors definition."""
+        _prompt_template = self.prompt_template
+        if error_message:
```
willbakst, line 435: I think we should consider using `error_messages` as an array (or dictionary keys for faster check/access) and continuously adding for each retry (and we can reduce the number of errors included by only including those of different types).

Thoughts?

willbakst (approved):
Great stuff!

mirascope/base/extractors.py
willbakst: looks like logfire handler covers this now.

mirascope/base/extractors.py
brenkao: okay so a set?

mirascope/base/extractors.py
brenkao: correct

mirascope/base/extractors.py
willbakst: well, sets aren't ordered, and we likely want the errors in order so the LLM can identify the order of failures, so best to actually just use a dictionary and then grab the set from the keys, right?

mirascope/base/extractors.py
willbakst: Hmm, I wonder if there is a way we could make this optional so it only happens if they use the decorator (or specify some other way they want this behavior)?

mirascope/base/extractors.py
willbakst: e.g. checking `if thing in dict` will have O(1) access like a set, but then when we want to inject them we can do `list(dict.keys())` and ensure they are in order

mirascope/base/extractors.py
brenkao: done

mirascope/base/extractors.py
```diff
@@ -121,32 +123,57 @@ def _extract(
...
+                with attempt:
+                    try:
+                        extraction = _extract_attempt(
+                            call_type, tool_type, error_messages, **kwargs
+                        )
+                    except (AttributeError, ValueError, ValidationError) as e:
+                        error_messages[str(e)] = None
+                        if (
+                            self.call_params.logfire or self.call_params.logfire_async
+                        ):  # pragma: no cover
```
willbakst, line 165: does this cover everything inside the conditional?

willbakst (approved):

mirascope/base/extractors.py
brenkao: yes

PR #211, opened 2024-05-04
Logfire updates for tool calls

willbakst:
- Added handling for all providers to support tool calling with the new Logfire UI updates.


mirascope/base/extractors.py
```diff
@@ -136,7 +136,6 @@ def _extract_attempt(
             obj = temp_call(
```
brenkao, line 136: Oops this also needs to be changed back to just one line

brenkao (approved):

Issue #221, opened 2024-05-08
Add support for nested structured in `GeminiTool`


willbakst:
**Is your feature request related to a problem? Please describe.**
https://github.com/Mirascope/mirascope/blob/269c3339b7e5113686a2435a8b85709abbdcf485/mirascope/gemini/tools.py#L75

Where does this restriction originate from? Is this just something not yet implemented, or is it inherent to gemini?

_Originally posted by @barapa in https://github.com/Mirascope/mirascope/discussions/219_

**Describe the solution you'd like**
Add support to `GeminiTool` to properly structured nested definitions to match the Open API 3.0.3 Parameter Object that Gemini supports.

Parameter Object: https://spec.openapis.org/oas/v3.0.3#parameter-object
Schema Object: https://spec.openapis.org/oas/v3.0.3#schema-object
Reference Object: https://spec.openapis.org/oas/v3.0.3#reference-object

willbakst:
@barapa do you have any interest in taking this on?

barapa:
It appears that it doesn't really follow the Open API 3.0.3 spec, but rather a "subset" supported by their [FunctionDefinition proto](https://ai.google.dev/api/python/google/ai/generativelanguage/FunctionDeclaration). The hard part is the [Schema proto](https://ai.google.dev/api/python/google/ai/generativelanguage/Schema), which I'm reproducing below:

A few challenges I have encountered so far, when trying to convert the model_json_schema() to conform to this Schema
- [x] Remove `$def` fields
- [x] Remove `$ref` fields
- [ ] Remove `AnyOf`
- [ ] Remove `AllOf`

WIP PR here: https://github.com/Mirascope/mirascope/pull/222/files

```python
class Schema(proto.Message):
    r"""The ``Schema`` object allows the definition of input and output data
    types. These types can be objects, but also primitives and arrays.
    Represents a select subset of an `OpenAPI 3.0 schema
    object <https://spec.openapis.org/oas/v3.0.3#schema>`__.


    .. _oneof: https://proto-plus-python.readthedocs.io/en/stable/fields.html#oneofs-mutually-exclusive-fields

    Attributes:
        type_ (google.ai.generativelanguage_v1beta.types.Type):
            Required. Data type.
        format_ (str):
            Optional. The format of the data. This is
            used only for primitive datatypes. Supported
            formats:

             for NUMBER type: float, double
             for INTEGER type: int32, int64
        description (str):
            Optional. A brief description of the
[...]

PR #222, opened 2024-05-09
Support nested pydantic models with Gemini tool calling [DRAFT]

barapa:
Attempt at addressing https://github.com/Mirascope/mirascope/issues/221

The google-supported schema is very far off from openapi spec. This may be worth waiting for them to solve it themselves.

### DIFF
--- a/mirascope/gemini/tools.py
+++ b/mirascope/gemini/tools.py
@@ -1,69 +1,27 @@
 """Classes for using tools with Google's Gemini API."""
 from __future__ import annotations
 
-import dataclasses
+import pprint
 from typing import Any, Callable, Type
 
+import jsonref
 from google.ai.generativelanguage import FunctionCall
 from google.generativeai.types import (  # type: ignore
+    FunctionDeclaration,
     Tool,
-    content_types,
-)
-from google.generativeai.types.content_types import (  # type: ignore
-    CallableFunctionDeclaration,
 )
 from pydantic import BaseModel, ConfigDict
-from pydantic_core import PydanticUndefined
+
+from mirascope.base.tools import DEFAULT_TOOL_DOCSTRING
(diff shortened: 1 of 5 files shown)


mirascope/gemini/tools.py
```diff
@@ -68,20 +73,42 @@ def tool_schema(cls) -> Tool:
...
+        # Replace all references with their values
+        without_refs: dict[str, Any] = jsonref.replace_refs(model_schema)  # type: ignore
+        pprint.pprint(without_refs)
+
+        # Remove all Defs
+        without_refs.pop("$defs")
+        pprint.pprint(without_refs)
+
+        # Get the name and description, and remove them from the schema
+        name: str = without_refs.pop("title")  # type: ignore
```
willbakst, line 89: quick note:

name and description are already in `super().tool_schema()`, so it's really just the `parameters` field that needs to be modified (and thus remove duplicate code).

mirascope/gemini/tools.py
```diff
@@ -68,20 +73,42 @@ def tool_schema(cls) -> Tool:
...
+        description: str = (  # type: ignore
+            without_refs.pop("description", None) or DEFAULT_TOOL_DOCSTRING
+        )
+        parameters: dict[str, Any] = without_refs
+
+        # Remove all instances of title key in each param definition
+        # This is careful not to delete keys that represent a field with the name title
+        remove_invalid_title_keys_from_parameters(parameters)
```
willbakst, line 97: what if all of the conversion of `parameters` was pushed inside this utility function so the main function remains clean and the utility function can be separately unit tested?

barapa:
This isn't quite ready for review - it's still messy as I am experimenting. If I get it over the finish line I will clean it up beforehand.

barapa:
@willbakst - Wth Gemini's current

PR #231, opened 2024-05-10
feat: added langfuse integration

brenkao:


mirascope/base/utils.py
```diff
@@ -222,7 +222,8 @@ def wrapper_generator(*args, **kwargs):
         try:
             for attempt in retries:
                 with attempt:
-                    for value in fn(*args, **kwargs):
+                    generator = fn(*args, **kwargs)
```
willbakst, line 225: is this necessary for some reason? if so please add a comment

mirascope/logfire/logfire.py
```diff
@@ -379,19 +379,57 @@ def with_logfire(cls):
         setattr(cls, "add", mirascope_logfire_span(cls.add))
     if get_parent_class_name(cls, "OpenAI"):
         if hasattr(cls, "call_params"):
-            cls.call_params.logfire = logfire.instrument_openai
+            setattr(
```
willbakst, line 382: we likely also want to do this with `weave`?

willbakst (approved):

mirascope/base/utils.py
brenkao: it was for debugging, switched back

mirascope/logfire/logfire.py
brenkao: done

PR #237, opened 2024-05-13
fix: use tenacity when retries > 1 instead of always 

### LINKED ISSUE #227: Retry error shows up even when retries aren't set
**To Reproduce**

See #226 where retries show up as the error even though they aren't set.

**Expected behavior**
If retries aren't explicitly set, I expect for calls to operate as though there are no retries at all ultimately surfacing the original error, not a retry error.

brenkao:
Fix #227

### DIFF
--- a/mirascope/base/extractors.py
+++ b/mirascope/base/extractors.py
@@ -147,7 +147,7 @@ class BaseExtractor(
                 raise
 
         if isinstance(retries, int):
-            if retries > 1:
+            if retries > 0:
                 retries = Retrying(stop=stop_after_attempt(retries))
             else:
                 return _extract_attempt(call_type, tool_type, {}, **kwargs)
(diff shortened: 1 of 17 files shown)


mirascope/base/extractors.py
```diff
@@ -224,7 +227,10 @@ async def _extract_attempt_async(
                 raise
 
         if isinstance(retries, int):
-            retries = AsyncRetrying(stop=stop_after_attempt(retries))
+            if retries > 1:
```
willbakst, line 230: should this be `> 0` everywhere?

If a user passes in `Retrying` with retries set to `1` it will now behave differently from passing in `retries=1` right? I think we want to match that behavior but default to `0` where `0` means no tenacity?

willbakst (approved):

mirascope/base/extractors.py
brenkao: thats fine we can do that too

PR #284, opened 2024-05-28
docs: added mirascope evaluations section and code example

brenkao:


mkdocs.yml
```diff
@@ -103,6 +103,10 @@ nav:
       - Chat History: "concepts/chat_history.md"
       - Tools (Function Calling): "concepts/tools_(function_calling).md"
       - Extracting structured information using LLMs: "concepts/extracting_structured_information_using_llms.md"
+      - Evaluations:
```
willbakst, line 106: hmm, I feel like this should just be it's own section in concepts? We already reference you should read it at the top, and I think nesting it will make it harder to find.

docs/concepts/evaluations/index.md
```diff
@@ -0,0 +1,3 @@
+# Evaluations
```
willbakst, line 1: see commend on structure, I think this should just be merged with the other pages into a single concepts page (at least for now). Each of these docs can be a high level section.

docs/concepts/extracting_structured_information_using_llms.md
```diff
@@ -249,7 +249,7 @@ class HappyStory(BaseModel):
 
 
 class StoryTeller(OpenAIExtractor[HappyStory]):
-    extract_template: Type[HappyStory] = HappyStory
+    extract_schema: Type[HappyStory] = HappyStory
```
willbakst, line 252: nit: also fix `Type` -> `type`

willbakst (approved):

PR #285, opened 2024-05-29
Support for AnthropicBedrock and AzureOpenAI

### LINKED ISSUE #224: Add support for `AzureOpenAI`
**Is your feature request related to a problem? Please describe.**
There are models hosted on Azure that don't work out of the box with existing SDKs and require special clients.

**Describe the solution you'd like**
Ideally support for models hosted on Azure while still using the same class.

**Describe alternatives you've considered**
Adding additional classes e.g. `AnthropicAzureCall`, but this isn't ideal.

**Additional context**
Many companies use models hosted on Azure, so this seems worth supporting?

### LINKED ISSUE #225: Add support for `AnthropicBedrock`
**Is your feature request related to a problem? Please describe.**
There are models hosted on Bedrock that don't work out of the box with existing SDKs and require special clients.

**Describe the solution you'd like**
Ideally support for models hosted on Bedrock while still using the same class.

**Describe alternatives you've considered**
Adding additional classes e.g. `AnthropicBedrockCall`, but this isn't ideal.

**Additional context**
Many companies use models hosted on Bedrock, so this seems worth supporting?

willbakst:
Closes #224 
Closes #225 

- Add `bedrock_client_wrapper` to `anthropic` module for wrapping `Anthropic` and `AsyncAnthropic` to use AWS Bedrock.
- Add `azure_client_wrapper` to `openai` module for wrapping `OpenAI` and `AsyncOpenAI` to use Microsoft Azure deployments.

### DIFF
--- a/docs/concepts/supported_llm_providers.md
+++ b/docs/concepts/supported_llm_providers.md
@@ -367,6 +367,40 @@ class RecipeRecommender(OpenAICall):
     call_params = OpenAICallParams(model="mistral")
 ```
 
+## Using Anthropic models on AWS Bedrock
+
+You can use Anthropic models through AWS Bedrock, which is supported in Mirascope through a client wrapper. Simply import the wrapper, provide the correct configuration, and add the wrapper to your call:
+
+```python
+from mirascope.anthropic import (
+    AnthropicCall,
+    AnthropicCallParams,
+    bedrock_client_wrapper,
+)
+from mirascope.base import BaseConfig
+
+
+class BookRecommender(AnthropicCall):
+    prompt_template = "Please recommend a fantasy book."
+
+    call_params = AnthropicCallParams(model="anthropic.claude-3-haiku-20240307-v1:0")
+    configuration = BaseConfig(
+        client_wrappers=[
+            bedrock_client_wrapper(
+                # both of these keys can be configured in your environment
+                aws_secret_key="YOUR_SECRET_KEY",
+                aws_access_key="YOUR_ACCESS_KEY",
+                aws_region="us-west-2",
+            )
+        ]
+    )
+
+
+recommender = BookRecommender()
+response = recommender.call()
+print(response.content)
+```
+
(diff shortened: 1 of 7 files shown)


docs/concepts/supported_llm_providers.md
```diff
@@ -367,6 +367,76 @@ class RecipeRecommender(OpenAICall):
     call_params = OpenAICallParams(model="mistral")
 ```
 
+## Using models hosted on major cloud providers
+
+Some model providers have their models hosted on major cloud providers such as AWS Bedricj and Microsoft Azure. You can use provided client wrappers to access these models by providing the proper configuration:
```
brenkao, line 372: `AWS Bedricj` lol

mirascope/anthropic/utils.py
```diff
@@ -1,10 +1,41 @@
...
+    base_url: Optional[Union[str, URL]] = None,
+) -> Callable[
+    [Union[Anthropic, AsyncAnthropic]], Union[AnthropicBedrock, AsyncAnthropicBedrock]
+]:
+    """Returns a client wrapper for using Anthropic models on AWS Bedrock."""
+
+    def inner_wrapper(client: Union[Anthropic, AsyncAnthropic]):
+        """Returns matching `AnthropicBedrock` or `AsyncAnthropicBedrock` client."""
+        kwargs = {
+            "aws_secret_key": aws_secret_key,
```
brenkao, line 24: if user sets in their `.env` file, do they also need to pass it in here? if so, does it make sense for us to get env if it exists?

brenkao (approved):
approve after fixing nit

docs/concepts/supported_llm_providers.md
willbakst: Good catch

mirascope/anthropic/utils.py
willbakst: I believe that if set to `None` the underlying provider class will grab it from the environment for us, so no need to duplicate that code.

PR #302, opened 2024-06-05
feat: updated mirascope logfire span data for custom panels

brenkao:


mirascope/logfire/logfire.py
```diff
@@ -233,10 +235,17 @@ def handle_before_call(self: BaseModel, fn: Callable, **kwargs):
     """Handles before call"""
     class_vars = get_class_vars(self)
     name = f"{self.__class__.__name__}.{fn.__name__}"
-    with logfire.span(
-        name,
-        class_vars=class_vars,
+    tags = class_vars.pop("tags", [])
+    span_data = {
+        "class_vars": class_vars,
+        "template_variables": self.model_dump(),
```
willbakst, line 241: once concern here is that a property won't show up unless it's a computed field. do we care about that?

willbakst (approved):

mirascope/logfire/logfire.py
brenkao: done

Issue #264, opened 2024-05-21
Add improved support for chat history message insertion + documentation


willbakst:
### Description

Right now users need to manually create the tools even though all of the information is generally already present in the tool. A method like `tool.message_param` would make this far more convenient (similar to how we can already to `response.message.model_dump()` for the assistant message.

We may also want to provide additional convenience, something like `response.user_message_param`, for inserting the user message back in. This one likely needs a little more thought than the tool call message feature.

willbakst:
Updating this feature request to be separate from tool call message param since the tool stuff is clear cut and the other stuff needs a little more thought.

Will be tracking the tool stuff in #294

willbakst:
One thought I just had for this would be to add a new Mirascope `Stream` class instead of using the raw `Generator`. The added bonus of creating our own class (and not just a type alias) is that we can do some nice things around return values.

We can have the generator return a tuple of values and then add those values to the generator object for easy access. For example, consider a generator that returns `assistant_message` accessible through e.g. `stream.assistant_message_param` after the stream so that the assistant message param can be easily inserted into chat history. This holds for all streaming across things like tools and cost tracking as well. If we return a tuple, we can insert each item as a named property of the stream (e.g. `stream.cost`).

This would provide some much needed convenience around reinserting messages, which is particularly annoying when streaming. We can provide simlar convenience for normal calls as well e.g. `response.assistant_message_param`.

Naming needs some thought, but I think the idea is there.

willbakst:
Worth noting that we should also ensure that mypy and other type checkers don't get update with these chat history and MESSAGES flows.

willbakst:
Plan of attack after further thinking discussion:
- [ ] Add a `response.message_param` property
This will return the proper type since `response.message` is a `ChatCompletionMessage` pydantic model and not a message param. Returning the correct type will make the type checker happy.
- [ ] Add `ProviderChatMessage` e.g. `OpenAIChatMessage` types and coerce to message params under the hood
If reading the OpenAI docs or coming from their SDK, users will expect to be able to insert `response.message` into chat history. This is because OpenAI makes that work even though their typing doesn't match that. We should just allow for this insertion and under the hood convert any output pydantic models into their input typed dict counterparts. This way users don't have to learn the difference and mirascope just makes it easy.
- [ ] Consider adding `response.user_message_param`.
[...]

PR #310, opened 2024-06-07
feat: add  to call response types

willbakst:
Progress on #264

### DIFF
--- a/mirascope/anthropic/calls.py
+++ b/mirascope/anthropic/calls.py
@@ -87,6 +87,7 @@ class AnthropicCall(
         )
         return AnthropicCallResponse(
             response=message,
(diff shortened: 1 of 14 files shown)


mirascope/anthropic/calls.py
```diff
@@ -87,6 +87,7 @@ def call(
         )
         return AnthropicCallResponse(
             response=message,
+            user_message_param=messages[-1] if messages[-1]["role"] == "user" else None,
```
brenkao, line 90: refactor to utility fn that all providers share

willbakst: Done.

brenkao (approved):

PR #315, opened 2024-06-10
docs: updated writing_prompts docs and added additional examples

brenkao:


examples/writing_prompts/defining_messages_without_prompt_template.py
willbakst, line 1: lol this will look much better once we add templating

willbakst (approved):

PR #316, opened 2024-06-10
docs: fix typos and bad examples, added new example

brenkao:


examples/generating_content/multiturn_call_gemini.py
willbakst, line 1: calling this "multiturn" feels weird because it kind of is with how you're prompting but not actually a multiturn interaction for the user.

This feels more like `gemini_system_call.py` or something?

willbakst (approved):

examples/generating_content/multiturn_call_gemini.py
brenkao: done

PR #317, opened 2024-06-11
feat: streaming classes for additional streaming convenience

willbakst:
Progress on #264


mirascope/base/types.py
```diff
@@ -295,3 +296,94 @@ def output_tokens(self) -> Optional[Union[int, float]]:
         If there is no output_tokens, this method must return None.
         """
         ...  # pragma: no cover
+
+
+BaseCallResponseT = TypeVar("BaseCallResponseT", bound=BaseCallResponse)
+BaseCallResponseChunkT = TypeVar("BaseCallResponseChunkT", bound=BaseCallResponseChunk)
+MessageParamT = TypeVar("MessageParamT", bound=Any)
```
brenkao, line 303: These are already defined in `mirascope/base/calls.py`. Does it make sense to consolidate all the typevars and prevent duplicates

mirascope/base/types.py
```diff
@@ -295,3 +296,94 @@ def output_tokens(self) -> Optional[Union[int, float]]:
...
+    def __init__(
+        self,
+        stream: AsyncGenerator[BaseCallResponseChunkT, None],
+        message_param_type: type[AssistantMessageParamT],
+    ):
+        """Initializes an instance of `BaseAsyncStream`."""
+        self.stream = stream
+        self.message_param_type = message_param_type
+
+    def __aiter__(
```
brenkao, line 369: why does `__aiter__` need to have an inner generator fn but the above `__iter__` does not?

brenkao (approved):

mirascope/base/types.py
willbakst: `__aiter__` is a synchronous method that returns the generator. It's `__anext__` that's `async`, but I need to iterate in the iter to gather.

We need to wrapper here so that we can call `async for chunk in self.stream`, and then we return an async generator by calling `generator()`.

mirascope/base/types.py
willbakst: Sure, done.

PR #323, opened 2024-06-12
docs: added multiple examples and docs updates to chat history

brenkao:


docs/concepts/chat_history.md
```diff
@@ -29,10 +29,8 @@ librarian = Librarian(question="", history=[])
 while True:
     librarian.question = input("(User): ")
     response = librarian.call()
-    librarian.history += [
-        {"role": "user", "content": librarian.question},
-        {"role": "assistant", "content": response.content},
-    ]
+    librarian.history.append({"role": "user", "content": librarian.question})
```
willbakst, line 32: why not use `response.user_message_param` here?

docs/concepts/chat_history.md
```diff
@@ -74,10 +72,8 @@ librarian = Librarian(question="", history=[])
...
```
willbakst, line 75: ditto

willbakst (approved):

PR #339, opened 2024-06-17
Adding weaviate integration

aowen14:
PR to add support for Weaviate Vector Store for Rag.

- Weaviate normally needs named columns, so accounted for that.
- Switched to `psycopg-binary` from `psycopg` per @willbakst's request.
- Testing with the existing examples should also uncomment anything with `store.close_connection()` as Weaviate generally wants to have a connection closed and this will impact the output printed to the terminal with these models.
- Ideally the usage of `Documents` would be refactored in the other Rag integrations to fit more of a standard interface (the types are all different right now). Can work on that in separate PR, considering that Out of Scope for this.


mirascope/weaviate/vectorstores.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+            data_objects.append(data_object)
+
+        return self._index.data.insert_many(data_objects)
+
+    def retrieve(self, text: str, **kwargs: Any) -> WeaviateQueryResult:
+        """Queries the vectorstore for closest match"""
+        query_result = self._index.query.near_text(query=text, **kwargs)
+        result = query_result.objects[0]
+        print(result)
```
brenkao, line 71: Remove print statement

brenkao (approved):

mirascope/weaviate/vectorstores.py
aowen14: Whoops good catch

Issue #278, opened 2024-05-23
Enable tool schemas to support more dynamic generation


willbakst:
### Description

Consider the following tool:

```python
from mirascope.openai import OpenAITool
from pydantic import Field

class MyTool(OpenAITool):
    """This is the overall tool description of my tool that gets injected."""

    foo: str = Field(..., description="This is the field specific description that gets injected.")
```

What happens when we want the descriptions to be more dynamic with templating like we have for `prompt_template`? For example:

```python
from mirascope.openai import OpenAITool
from pydantic import Field

template_var_to_insert = "inserted"

class MyTool(OpenAITool):
    """This is the overall tool description of my tool that gets {template_var_to_insert}."""

    foo: str = Field(
        ...,
        description=f"This is the field specific description that gets {template_var_to_insert}."
    )
```

While the above would work for `foo`, it doesn't work for the docstring. Furthermore, this only works if you have the variable available at the time of definition. What happens if you want to update the template variable dynamically at run time with multiple different options? Especially since you pass tools into the `call_params` classvar.

My initial instinct would be to enable dynamically updating the schema at generation time and then surfacing something like a config that can be passed in at run time to actually go and update the generation. Something like:

```python
from mirascope.openai import OpenAICall, OpenAICallParams, OpenAITool
from pydantic import Field

template_var_to_insert = "inserted"

class MyTool(OpenAITool):
    """This is the overall tool description of my tool that gets {template_var_to_insert}."""

    foo: str = Field(
        ...,
        description=f"This is the field specific description that gets {template_var_to_insert}."
    )

class MyCall(OpenAICall):
    prompt_template = "..."

    call_params = OpenAICallParams(tools=[MyTool])


my_call = MyCall()
response = my_call.call(update_tools={"MyTool": {"template_var_to_insert": "inserted"}})
```

I don't love this exact design, but I think it's headed in the direction of what would potentially work.

@off6atomic I would love your thoughts here.

willbakst:
@off6atomic I'm also wondering if I was off about wanting dynamic generation at runtime vs. just templating at definition time?

off6atomic:
If we want maximum flexibility like when calling OpenAI API, I think all the data sent to OpenAI should be modifiable in any way that users want at runtime.

This means that these needs to modifiable at runtime:
- docstring
- field type
- field description

One approach that could work is that Mirascope should provide convenience in defining the default base values at definition time but then let users modify them at runtime however they want.

---
[...]

PR #340, opened 2024-06-19
Add ToolKit for new interface

koxudaxi:
The PR adds ToolKit for a new style interface.

```python
class BookRecommendationTools(Toolkit):
    """A toolkit for recommending books."""

    reading_level: Literal["beginner", "advanced"]

    @toolkit_tool() # not sure about this naming...
    def format_book(self, title: str, author: str) -> str:
        """Returns the title and author of a book nicely formatted.

        Reading level: {self.reading_level}
        """
        return f"{title} by {author}"

@openai_call(model="gpt-4o")
def recommend_book(genre: str, reading_level: Literal["beginner", "advanced"]):
    """Recommend a {genre} book."""
    toolkit = BookRecommendationTools(reading_level=reading_level)
    return {"tools": [toolkit.create_tools()]}

```
The broad policy was discussed within #278, but any undecideds will be resolved within the PR. 
Closes https://github.com/Mirascope/mirascope/issues/278


koxudaxi:
@willbakst 
I have made a rough implementation based on what we have discussed. I have not implemented `create_tool` or unittest yet, but first, I want to ensure that my implementation policy is not different from your assumptions.
Also, I have not confirmed my implementation policy for the part of the connection with the new interface (`@openai_call`), which is not yet completed, so it is currently a `TODO`.
Of course, type hints and variable names may need improvement, but please confirm the general framework.

mirascope/core/openai/openai_call.py
```diff
@@ -42,6 +42,17 @@ def decorator(fn: Callable[P, R]) -> Callable[P, OpenAICallResponse]:
...
+
+            # Try to get the dictionary for tools from the function result
+            # tools = []
+            fn_result = fn(*args, **kwargs)
+            if isinstance(fn_result, dict):
+                if fn_result_tools := fn_result.get("tools"):
+                    for fn_result_tool in fn_result_tools:
+                        # TODO: Generate tools
+                        # tools.append(generate(fn_result_tool))
+                        pass
```
koxudaxi, line 54: @willbakst 
Do you have any thoughts on the design and implementation policy for this part of the to generate `tools`? On an abstract level, I have looked through this page of the documentation.
https://docs.mirascope.io/latest/concepts/tools_%28function_calling%29/

koxudaxi: OK, I confirmed that you implemneted it. Thank  you.

mirascope/core/_internal/utils.py
```diff
@@ -110,6 +121,7 @@ def convert_function_to_base_tool(
     Args:
         fn: The function to convert.
         base: The `BaseToolT` type to which the function is converted.
+        __doc__: The docstring to use for the constructed `BaseToolT` type.
```
willbakst, line 124: should this just be `doc` instead of `__doc__` since it's an input argument?

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+        formated_template = self._toolkit_tool_method.__doc__.format(self=self)
+        return convert_function_to_base_tool(self._toolkit_tool_method, BaseTool, formated_template)
+
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs):
+        toolkit_tool_method: Optional[Callable] = None
+        for name, value in cls.__dict__.items():
+            if getattr(value, _TOOLKIT_TOOL_METHOD_MARKER, False):
+                if toolkit_tool_method:
+                    raise ValueError("Only one toolkit_tool method is allowed")
```
willbakst, line 45: The toolkit should allow for multiple tools, not just once.

For example, consider a `FileSystem` toolkit that has `ls`, `mkdir`, etc. methods

mirascope/core/_internal/utils.py
```diff
@@ -50,6 +52,15 @@ def format_prompt_template(template: str, attrs: dict[str, Any]) -> str:
                 values[var] = "\n".join([str(item) for item in attr])
         else:
             values[var] = str(attr)
+    return values
+
+
+def format_prompt_template(template: str, attrs: dict[str, Any]) -> str:
+    """Formats the given prompt `template`"""
+    dedented_template = dedent(template).strip()
```
willbakst, line 60: wondering if we should make this `inspect.cleandoc` as well

mirascope/core/_internal/utils.py
```diff
@@ -50,6 +52,15 @@ def format_prompt_template(template: str, attrs: dict[str, Any]) -> str:
                 values[var] = "\n".join([str(item) for item in attr])
         else:
             values[var] = str(attr)
+    return values
+
+
+def format_prompt_template(template: str, attrs: dict[str, Any]) -> str:
```
willbakst, line 58: maybe we should change this name to just `format_template`

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+class BaseToolKit(BaseModel, ABC):
+    """A class for defining tools for LLM call tools."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _toolkit_tool_method: ClassVar[Callable[[BaseToolKit, ...], str]]
+    _toolkit_template_vars: ClassVar[list[str]]
+
+    def create_tool(self) -> type[BaseTool]:
+        """The method to create the tools."""
+        formated_template = self._toolkit_tool_method.__doc__.format(self=self)
```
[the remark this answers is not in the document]
willbakst: also `formated_template` has a typo (missing second `t`) -> `formatted_template`

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class BaseToolKit(BaseModel, ABC):
+    """A class for defining tools for LLM call tools."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _toolkit_tool_method: ClassVar[Callable[[BaseToolKit, ...], str]]
+    _toolkit_template_vars: ClassVar[list[str]]
+
+    def create_tool(self) -> type[BaseTool]:
```
willbakst, line 34: should be `create_tools` that returns `list[type[BaseTool]]` for all the tools in the toolkit

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+        if not toolkit_tool_method:
+            raise ValueError("No toolkit_tool method found")
+
+        # Validate the toolkit_tool_method
+        if (template := toolkit_tool_method.__doc__) is None:
+            raise ValueError("The toolkit_tool method must have a docstring")
+
+        dedented_template = dedent(template).strip()
+        if not (template_vars := get_template_variables(dedented_template)):
+            raise ValueError("The toolkit_tool method must have template variables")
```
willbakst, line 56: toolkit tools don't have to have a template variable. the toolkit could be used just to organize tools under a namespace without being dynamic

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if dedented_template != template:
```
willbakst, line 58: this is unnecessary since we'll be doing everything on tool creation anyway?

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+            toolkit_tool_method.__doc__ = dedented_template
+        for var in template_vars:
+            if not var.startswith("self."):
+                # Should be supported un-self variables?
+                raise ValueError(
+                    "The toolkit_tool method must use self. prefix in template variables"
```
willbakst, line 64: make this more clear that it's only if they are doing the dynamic tool creation

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,76 @@
...
+
+            self_var = var[5:]
+            # Expecting pydantic model fields or class attribute and property
+            if self_var in cls.model_fields or hasattr(cls, self_var):
+                continue
+            raise ValueError(
+                f"The toolkit_tool method template variable {var} is not found in the class"
+            )
+
+        cls._toolkit_tool_method = toolkit_tool_method
```
willbakst, line 75: before we set the methods, we should also make sure that if provided the user can create a namespace where the tools name becomes `{namespace}.{fn_name}`

mirascope/core/_internal/utils.py
koxudaxi: I thought the same thing at first, but then I thought that __doc__ would convey the intention better, so I decided on this one.

mirascope/core/base/toolkit.py
koxudaxi: > The toolkit should allow for multiple tools, not just once.

I understand. So, you wrote `create_tools()` in your example.
I will fix it.

mirascope/core/_internal/utils.py
koxudaxi: Nice catch. Completely forgot to change it later.

mirascope/core/base/toolkit.py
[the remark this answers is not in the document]
koxudaxi: > we should use the newly named format_template utility here setting self in the attrs dict

You are right. I refactored `format_template` to do that.

mirascope/core/base/toolkit.py
koxudaxi: I understand the intent and will fix it.

mirascope/core/base/toolkit.py
koxudaxi: Which of these cases are of concern?
My assumption is that the user has already defined a method and just takes it out and stores it in the tool's call method, but I didn't see the need to check the name.

mirascope/core/base/_utils.py
```diff
@@ -136,13 +147,16 @@ def convert_function_to_base_tool(
...
+    func_doc = __doc__ or fn.__doc__
+    if func_doc:
+        docstring = parse(func_doc)
 
     field_definitions = {}
     hints = get_type_hints(fn)
+    has_self_or_cls = False
     for i, parameter in enumerate(inspect.signature(fn).parameters.values()):
         if parameter.name == "self" or parameter.name == "cls":
+            has_self_or_cls = True
```
willbakst, line 159: what if instead we did something like this?

```python
if (has_self := (param.name == "self")) or parameter.name == "cls":
    continue

...
```

Then we can use just `has_self` separately since we aren't including the `cls` below?

mirascope/core/base/toolkit.py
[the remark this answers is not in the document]
koxudaxi: The methods decorated’s `__doc__` with @toolkit_tool always have a variable set that starts with `self.`, so I explicitly pass self.
Therefore, we do not use formatted_template, but we use the functions inside formatted_template that we extracted from the refactoring.

Please point out if there are any major errors in my specification assumptions.

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,81 @@
...
+class BaseToolKit(BaseModel, ABC):
+    """A class for defining tools for LLM call tools."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _toolkit_tool_methods: ClassVar[list[TookKitToolMethod]]
+
+    def create_tools(self) -> list[type[BaseTool]]:
+        """The method to create the tools."""
+        tools: list[type[BaseTool]] = []
+        for method, template_vars, template in self._toolkit_tool_methods:
```
willbakst, line 42: would it be cleaner to just return a list comprehension here? For example:

```python
def create_tools(self) -> list[type[BaseTool]]:
    """The method to create the tools."""
    return [
        convert_function_to_base_tool(method, BaseTool, template.format(self=self))
        for method, template_vars, template in self._toolkit_tool_methods
    ]
```

mirascope/core/base/_utils.py
```diff
@@ -57,6 +58,15 @@ def format_prompt_template(template: str, attrs: dict[str, Any]) -> str:
                 values[var] = "\n".join([str(item) for item in attr])
         else:
             values[var] = str(attr)
+    return values
+
+
+def format_template(template: str, attrs: dict[str, Any]) -> str:
```
willbakst, line 64: can you command search for the old name and find/replace? I believe there are a few places that use this method that haven't been updated (like the `__str__` method in `BasePrompt` for example

mirascope/core/base/toolkit.py
koxudaxi: I think it's very good.
I love list comprehension, but I refrain from using them if they are long because they are not generally accepted.

mirascope/core/base/toolkit.py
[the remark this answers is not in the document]
willbakst: You're right, and when we call `convert_function_to_base_model` it will do any necessary clearning to doc, so this is good.

mirascope/core/base/toolkit.py
willbakst: I agree. I think if it's too complex, best to avoid, but it's simpler to read in this case I believe

mirascope/core/base/_utils.py
koxudaxi: I fixed only the unittest name.
All other methods had been renamed to this name because of PyCharm's refactoring functionality.

mirascope/core/base/toolkit.py
willbakst: The user should be able to specify a namespace for a toolkit that updates the name of every function inside to contain the namespace.

For example:

```python
class FileSystem(Toolkit):
    namespace = "fs"

    @toolkit_tool
    def mkdir(self, ...):
        ...

print(FileSystem().create_tools()[0].__name__)
#> fs.mkdir
```

By default, `namespace` should be `None`, but if set, we should update the function names to be under the specified namespace. Does this make sense?

mirascope/core/base/_utils.py
koxudaxi: It is ok to separate them into `has_self`.

However, in that example, the value is assigned to `has_self` in every loop, so if there is more than one parameter, `has_self` will certainly be `False`.

willbakst: Oh good call you are right.

mirascope/core/base/toolkit.py
koxudaxi: I see.
As for the namespace declaration, is it OK to use namespace: ClassVar[str] as in the example?

>  we should also make sure that if provided the user can create a namespace where the tools name becomes {namespace}.{fn_name}

I'm a little confusing because the `__name__` of a function or method usually does not contain a namespace.

`assert inspect.signature.__name__ == 'signature'`

willbakst: Correct, we are doing this for the purposes of creating a "tool" for the LLM.

Consider two toolkits:

```python
class Library(Toolkit):
    books: list[Book]

    def find(self, title: str) -> Book:
        ...

class FileSystem(Toolkit):
    files: list[File]

    def find(self, filename: str) -> File:
        ...
```

We want to be able to insert both of the `find` methods, but we need to be able to identify which method is which, and we also should differentiate so the LLM can also better understand its tools. If we namespace, the LLM will see tools like `lib.find` and `fs.find`.

willbakst: Perhaps rather than updating the name of the function with the namespace we should instead pass the namespace into the `convert_function_to_base_tool` method like you did with `__doc__` and update the `_name` classmethod to use the namespaced name, that way there won't be a weird mismatch with the Python defaults / expectations.

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,80 @@
...
+def toolkit_tool(
+        method: Callable[Concatenate[BaseToolKit, P], str],
+) -> Callable[Concatenate[BaseToolKit, P], str]:
+    # Mark the method as a toolkit tool
+    setattr(method, _TOOLKIT_TOOL_METHOD_MARKER, True)
+
+    return method
+
+
+class TookKitToolMethod(NamedTuple):
```
willbakst, line 27: just noticed this should be `ToolKitToolMethod` (it says `Took`)

mirascope/core/base/toolkit.py
koxudaxi: I see. I understand the situation.

So what scope is affected by namespace?
In other words, where should I manage the name of this namespace?
Is it the whole module? Or the tool_kit decorator?

I don't see a complete use case, but it seems fine to keep it at module level.

```python
__namespace: set[str] = set()
...

print(__namespace)
#   {'lib.find', 'fs.find'}
```

willbakst: I will add new comments directly where they are applicable

mirascope/core/base/_utils.py
```diff
@@ -113,7 +123,7 @@ def parse_prompt_messages(
 
 
 def convert_function_to_base_tool(
-    fn: Callable, base: type[BaseToolT]
+    fn: Callable, base: type[BaseToolT], __doc__: str | None = None
```
willbakst, line 126: update to also take `namespace: str | None = None`

mirascope/core/base/_utils.py
```diff
@@ -181,20 +197,23 @@ def convert_function_to_base_tool(
     model = create_model(
         fn.__name__,
```
willbakst, line 198: here, we should use `f"{namespace}.{fn.__name__}" if namespace else fn.__name__`

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,80 @@
...
+    method: Callable[..., str]
+    template_vars: list[str]
+    template: str
+
+
+class BaseToolKit(BaseModel, ABC):
+    """A class for defining tools for LLM call tools."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _toolkit_tool_methods: ClassVar[list[ToolKitToolMethod]]
```
willbakst, line 38: add a `ClassVar` for namespace that the user can set

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def create_tools(self) -> list[type[BaseTool]]:
+        """The method to create the tools."""
+        return [
+            convert_function_to_base_tool(method, BaseTool, template)
```
willbakst, line 43: pass the classes namespace in here so created tools are namespaced

tests/core/base/test_toolkit.py
```diff
@@ -0,0 +1,35 @@
...
+
+            Reading level: {self.reading_level}
+            """
+            return f"{title} by {author}"
+
+    toolkit = BookRecommendationToolKit(reading_level="beginner")
+    tools = toolkit.create_tools()
+    assert len(tools) == 1
+    tool = tools[0]
+    assert tool._name() == "format_book"
```
willbakst, line 27: if we give the namespace `book_tools` then this should be `book_tools.format_book`

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,84 @@
...
+    template_vars: list[str]
+    template: str
+
+
+class BaseToolKit(BaseModel, ABC):
+    """A class for defining tools for LLM call tools."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _toolkit_tool_methods: ClassVar[list[ToolKitToolMethod]]
+    _namespace: ClassVar[str | None] = None
```
koxudaxi, line 40: I dared to use `_namespace` so that users can use `namespace`. What do you think?

willbakst: good call

mirascope/core/base/toolkit.py
```diff
@@ -0,0 +1,92 @@
...
```
willbakst, line 42: what do you think about using `__namespace__`?

koxudaxi: Looks good. I will change it soon.

koxudaxi: Done!

mirascope/core/base/_utils.py
```diff
@@ -115,7 +125,7 @@ def parse_prompt_messages(
 
 
 def convert_function_to_base_tool(
-    fn: Callable, base: type[BaseToolT]
+    fn: Callable, base: type[BaseToolT], __doc__: str | None = None, namespace: str | None = None
```
willbakst, line 128: we should also update this to `__namespace__` now

koxudaxi: OK, I just updated it.

PR #346, opened 2024-07-09
added initial work to fix/clean up tool streaming

brenkao:


mirascope/core/base/_stream.py
```diff
@@ -166,19 +159,19 @@ async def generator():
...
-                    kwargs["content"] = [kwargs["content"]] + tool_calls
-            self.message_param = self.message_param_type(**kwargs)
+            if not tool_calls:
+                tool_calls = None
+            self.message_param = self.construct_message_param(tool_calls, content)
 
         return generator()
 
+    @abstractmethod
+    def construct_message_param(
```
willbakst, line 169: we should make this private

willbakst (approved):

mirascope/core/base/_stream.py
brenkao: done

PR #348, opened 2024-07-12
Add `jiter = "^0.5.0"` to deps

jimkring:
this package must be installed for tool calling


willbakst:
This is a good catch, but I think this is only true for v1, can you confirm your version?

I just want to make sure we include this in the right branch.

jimkring:
I had the issue in v1 beta.

willbakst:
Got it. Can you close this PR and reopen a new on that pulls in the v1 branch please?

willbakst:
Adding this to v1 for the upcoming release.

jimkring:
Sounds good @willbakst. Thanks for taking care of v1

PR #351, opened 2024-07-16
fix: tools (sync vs. async) + streaming (cost tracking and linting)

willbakst:
I also fixed a bunch of other stuff I noticed along the way wrt. typing (e.g. Cohere and Mistral message types etc.)


mirascope/core/cohere/call_response.py
```diff
@@ -70,12 +71,12 @@ def content(self) -> str:
         return self.response.text
 
     @property
-    def model(self) -> str | None:
+    def model(self) -> str:
         """Returns the name of the response model.
 
         Cohere does not return model, so we return None
```
brenkao, line 77: Update docstring to match? We are returning the model in args, but not the returned call_response model

mirascope/core/gemini/call_response.py
```diff
@@ -143,12 +144,12 @@ def finish_reasons(self) -> list[str]:
         ]
 
     @property
-    def model(self) -> None:
+    def model(self) -> str:
         """Returns the model name.
 
         google.generativeai does not return model, so we return None
```
brenkao, line 150: ditto

mirascope/core/groq/stream.py
```diff
@@ -28,11 +30,20 @@ class GroqStream(
...
+        tool_calls: list[ChatCompletionMessageToolCallParam] | None = None,
+        content: str | None = None,
     ) -> ChatCompletionAssistantMessageParam:
-        return ChatCompletionAssistantMessageParam(
+        message_param = ChatCompletionAssistantMessageParam(
             role="assistant",
             content=content,
-            tool_calls=tool_calls,
         )
+        if tool_calls:
```
brenkao, line 47: we pass in None already if tool_calls is falsey. if you want to make this change, i would imagine `_stream.py` in our openai module also needs to change

mirascope/core/cohere/call_response.py
willbakst: We set `_model` in the specific decorators (like _create) to the model provided by the user, and we use that one if there is no response from the model (like here).

I've updated the docstring.

mirascope/core/gemini/call_response.py
willbakst: Done.

mirascope/core/groq/stream.py
willbakst: No, because there are some providers that accept `None` and I don't check there, but if we pass in `list[Unknown]` that could cause issues.

brenkao (approved):

PR #352, opened 2024-07-17
feat: beta rag

brenkao:


mirascope/beta/rag/__init__.py
willbakst, line 1: Lets restructure slightly here. I think everything should be under beta/rag

Things that are base (like chunkers) should live in beta/rag/base, and other modules like openai/cohere should live in beta/rag/openai etc.

willbakst (approved):

mirascope/beta/rag/__init__.py
brenkao: done

PR #353, opened 2024-07-17
Multimodal Support

### LINKED ISSUE #242: Add multimodal support in prompt_template for easier prompting.
### Description

First thing that comes to mind is to add something like this
```python
prompt_template = """
SYSTEM: 
...

USER:
...

IMAGE:
{image_url}
```

willbakst:
Closes #242 

Added multimodal support for Anthropic, Gemini, OpenAI (and LiteLLM for providers that support it).

Cohere, Groq, and Mistral don't support multimodal (at least to my knowledge), so I've thrown errors when attempting to use multimodal prompt templating with those providers.

The multimodal prompt template parser currently handles text, image, and images (which is just a list of images). It is implemented such that it should be able to be easily extended to additional modalities such as audio, video, etc.

### DIFF
--- a/mirascope/core/base/_utils/parse_content_template.py
+++ b/mirascope/core/base/_utils/parse_content_template.py
@@ -86,7 +86,8 @@ def _construct_parts(part: _Part, attrs: dict[str, Any]) -> List[TextPart | Imag
         return [{"type": "text", "text": format_template(part["template"], attrs)}]
     elif part["type"] == "image":
         path_key, *options = part["template"][7:-1].split(",")
-        return [_construct_image_part(attrs[path_key], options)]
+        source = attrs[path_key]
+        return [_construct_image_part(attrs[path_key], options)] if source else []
     elif part["type"] == "images":
         path_key, *options = part["template"][8:-1].split(",")
         sources = attrs[path_key]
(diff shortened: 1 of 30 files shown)


mirascope/core/base/_utils/parse_content_template.py
```diff
@@ -0,0 +1,115 @@
...
+    template: str
+    type: _PartType
+
+
+class _ImageOptions(TypedDict, total=False):
+    detail: _Detail
+
+
+def _parse_parts(template: str) -> List[_Part]:
+    pattern = r"\{([^:{}]+):([^{}]+)\}"
```
brenkao, line 24: short docstring to describe whats happening, especially in regards to the splitting

mirascope/core/base/_utils/parse_content_template.py
```diff
@@ -0,0 +1,115 @@
+"""This module provides a function to parse content parts from a prompt template."""
+
+import imghdr
```
brenkao, line 3: looks like this is getting deprecated in python 3.13. either implement the alternative or put a note https://docs.python.org/3/library/imghdr.html

brenkao (approved):

mirascope/core/base/_utils/parse_content_template.py
willbakst: Done.

mirascope/core/base/_utils/parse_content_template.py
willbakst: Done. I wrote our own detection function and also added handling for each provider to throw an error if the user provides an image type that provider doesn't support.

PR #356, opened 2024-07-19
fixed pyright errors and reorganized integrations, and added tests for integrations

brenkao:


mirascope/integrations/logfire/__init__.py
```diff
@@ -0,0 +1 @@
+from .with_logfire import with_logfire
```
willbakst, line 1: missing `__all__`

mirascope/integrations/logfire/_utils.py
```diff
@@ -0,0 +1,180 @@
...
+from ...core.base._stream import BaseStream
+from ...core.base._structured_stream import BaseStructuredStream
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R", bound=BaseCallResponse | BaseStream | BaseModel)
+
+
+@contextmanager
+def custom_context_manager(
+    fn: Callable[_P, _R] | Callable[_P, Awaitable[_R]],
```
willbakst, line 26: `_P` and `_R` seem unnecessary here? Should be sufficient to write `Callable` or `Callable[..., Any]` right?

mirascope/integrations/logfire/_utils.py
```diff
@@ -0,0 +1,180 @@
...
+    output: dict[str, Any] = {}
+    if isinstance(result, BaseModel) and result._response is not None:  # type: ignore
+        response: BaseCallResponse = result._response  # type: ignore
+        span_data |= get_call_response_span_data(response)
+        output |= get_output(response)
+        output["response_model"] = result
+        span_data["output"] = output
+    elif isinstance(result, BaseStructuredStream):
+        span_data = get_stream_span_data(result.stream)
+        output["response_model"] = result.response_model
```
willbakst, line 178: ditto (see sync comment above)

mirascope/integrations/logfire/_utils.py
```diff
@@ -0,0 +1,180 @@
...
+            "message", ""
+        )
+    return {
+        "messages": [stream.user_message_param],
+        "call_params": stream.call_params,
+        "model": stream.model,
+        "provider": stream._provider,
+        "prompt_template": stream.prompt_template,
+        "template_variables": stream.fn_args,
+        "output": {"cost": stream.cost, "content": content},
```
willbakst, line 62: We should double check that things like `cost` aren't `None` to ensure we're actually replacing the iter properly (if you haven't already)

mirascope/integrations/logfire/_utils.py
```diff
@@ -0,0 +1,180 @@
...
+    output: dict[str, Any] = {}
+    if isinstance(result, BaseModel) and result._response is not None:  # type: ignore
+        response: BaseCallResponse = result._response  # type: ignore
+        span_data |= get_call_response_span_data(response)
+        output |= get_output(response)
+        output["response_model"] = result
+        span_data["output"] = output
+    elif isinstance(result, BaseStructuredStream):
+        span_data = get_stream_span_data(result.stream)
+        output["response_model"] = result.response_model
```
willbakst, line 134: `result.response_model` in this case is the type of the response model, not the actual response model output, which you can get as the final output of the structured stream.

Easiest way around this without having to iterate through would likely be to add a new field (something like `final_response_model` that you then update at the end of the stream, but this will still require at least waiting until the stream is exhausted since it'll be `None` otherwise (which I think you do looking at your iter replacement function in the middleware).

mirascope/integrations/middleware_factory.py
```diff
@@ -1,154 +1,169 @@
...
-        elif isinstance(result, BaseStream) and handle_stream is not None:
+        elif isinstance(result, BaseStream):
             original_iter = result.__iter__
 
-            def new_iter(self):
+            def new_stream_iter(self):
                 # Create the context manager when user iterates over the stream
-                with custom_context_manager() as context:
+                with custom_context_manager(fn) as context:
                     for chunk, tool in original_iter():
```
willbakst, line 79: does this work? I'm wondering if you'll need to call the iter outside the function and then iterate through the result here instead of calling it here

Easy way to check would be to see if streaming has cost but doesn't when using logfire?

mirascope/integrations/logfire/__init__.py
brenkao: done

mirascope/integrations/logfire/_utils.py
brenkao: done

mirascope/integrations/logfire/_utils.py
brenkao: what about `response_object`?

willbakst: Hmm, maybe `constructed_response_model`?

brenkao: sg

brenkao: done

mirascope/integrations/middleware_factory.py
brenkao: streaming does have cost, per my tests so this should work

mirascope/integrations/logfire/_utils.py
brenkao: i have confirmed they exist

brenkao: with stream_options include_usage set to true, both logfire and no logfire have streaming so iter works

mirascope/integrations/langfuse/_utils.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any, Callable
+
+from ...core.base import BaseCallResponse
+from ...core.base._stream import BaseStream
+from ...core.base.metadata import Metadata
+
+
+def get_call_response_observation(result: BaseCallResponse, fn: Callable[..., Any]):
+    metadata: Metadata = fn.__annotations__.get("metadata", {})
```
willbakst, line 9: there is a `get_metadata` utility function is base, use that here instead

mirascope/integrations/langfuse/_utils.py
```diff
@@ -0,0 +1,31 @@
...
+        "input": result.prompt_template,
+        "metadata": result.response,
+        "tags": tags,
+        "model": result.model,
+        "output": result.message_param.get("content", None),
+    }
+
+
+def get_stream_observation(stream: BaseStream, fn: Callable[..., Any]):
+    metadata: Metadata = fn.__annotations__.get("metadata", {})
```
willbakst, line 22: ditto

mirascope/integrations/langfuse/with_langfuse.py
```diff
@@ -0,0 +1,115 @@
...
+            input=stream.input_tokens,
+            output=stream.output_tokens,
+            unit="TOKENS",
+        )
+        langfuse_context.update_current_observation(
+            **get_stream_observation(stream, fn),
+            usage=usage,
+        )
+
+    def handle_base_model(result: BaseModel | BaseStructuredStream, context: None):
```
willbakst, line 65: why not have these as two separate functions:

1. `handle_response_model`
2. `handle_structured_stream`

Also, is there a reason they should be internal functions rather than utility functions of the langfuse module that we unit test individually? That's what you've done for logfire, and I prefer that

mirascope/integrations/langfuse/with_langfuse.py
```diff
@@ -0,0 +1,115 @@
...
+                output=result.constructed_response_model,
+            )
+
+    async def handle_call_response_async(result: BaseCallResponse, context: None):
+        handle_call_response(result, None)
+
+    async def handle_stream_async(stream: BaseStream, context: None):
+        handle_stream(stream, None)
+
+    async def handle_base_model_async(
```
willbakst, line 96: see earlier comment

mirascope/integrations/logfire/_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    logfire_span: logfire.LogfireSpan | None,
+):
+    if logfire_span is None:
+        return
+    span_data = get_stream_span_data(stream)
+    span_data["async"] = False
+    logfire_span.set_attributes(span_data)
+
+
+def handle_base_model(
```
willbakst, line 117: see similar comment in langfuse module

mirascope/integrations/logfire/_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    logfire_span: logfire.LogfireSpan | None,
+):
+    if logfire_span is None:
+        return
+    span_data = get_stream_span_data(stream)
+    span_data["async"] = True
+    logfire_span.set_attributes(span_data)
+
+
+async def handle_base_model_async(
```
willbakst, line 164: ditto

mirascope/integrations/middleware_factory.py
```diff
@@ -0,0 +1,226 @@
...
+        [SyncFunc | AsyncFunc], AbstractContextManager[_T]
+    ] = default_context_manager,
+    custom_decorator: Callable | None = None,
+    handle_call_response: Callable[[BaseCallResponse, _T | None], None] | None = None,
+    handle_call_response_async: Callable[[BaseCallResponse, _T | None], Awaitable[None]]
+    | None = None,
+    handle_stream: Callable[[BaseStream, _T | None], None] | None = None,
+    handle_stream_async: Callable[[BaseStream, _T | None], Awaitable[None]]
+    | None = None,
+    handle_base_model: Callable[[BaseModel | BaseStructuredStream, _T | None], None]
```
willbakst, line 102: see earlier comment about splitting these

mirascope/integrations/otel/_utils.py
```diff
@@ -0,0 +1,266 @@
...
+def handle_stream(stream: BaseStream, span: Span | None):
+    if span is None:
+        return
+    attributes = get_stream_attributes(stream)
+    attributes["async"] = False
+    span.set_attributes(attributes)
+    set_stream_event_attributes(stream, span)
+
+
+def handle_base_model(result: BaseModel | BaseStructuredStream, span: Span | None):
```
willbakst, line 156: ditto

mirascope/integrations/otel/_utils.py
```diff
@@ -0,0 +1,266 @@
...
+async def handle_stream_async(stream: BaseStream, span: Span | None):
+    if span is None:
+        return
+    attributes = get_stream_attributes(stream)
+    attributes["async"] = True
+    span.set_attributes(attributes)
+    set_stream_event_attributes(stream, span)
+
+
+async def handle_base_model_async(
```
willbakst, line 221: ditto

tests/integrations/langfuse/test_utils.py
willbakst, line 1: this will need new tests wrt. other comment

tests/integrations/langfuse/test_with_langfuse.py
willbakst, line 1: my guess is that this will also be simplified and some tests will be moved over to utils as unit tests for the currently nested functions

willbakst (commented):
Left some comments, I'll do a deeper review of the tests once the current comments are resolved since they will likely impact the current tests/structure

mirascope/integrations/langfuse/with_langfuse.py
brenkao: originally it was supposed to be because they both return a base model, but they became too different, but they also share some commonalities.

for langfuse, i'm using fn, so not sure how I would have access to fn if the function is defined outside. I would need to pass in fn as a 3rd argument to every function

willbakst: I would rather you pass in `fn` and have them as separate functions that we can unit test.

I would also rather `response_model` and `structured_stream` remain separate even if there is some commonality since those are separate paths. Even though they may share some internal properties, externally they are separate paths, and these integrations should act without any knowledge of internals

mirascope/core/base/_structured_stream.py
```diff
@@ -73,7 +74,11 @@ def __iter__(self) -> Generator[_ResponseModelT, None, None]:
                 yield extract_tool_return(self.response_model, json_output, True)
         if json_output:
             json_output = json_output[: json_output.rfind("}") + 1]
-        yield extract_tool_return(self.response_model, json_output, False)
+        extracted_response_model = extract_tool_return(
+            self.response_model, json_output, False
+        )
+        self.constructed_response_model = extracted_response_model
+        yield extracted_response_model
```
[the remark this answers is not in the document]
brenkao: > can you explain why we need this? I'm assuming it's so that the user can easily grab the final constructed object after the stream without having to keep track of it in their own stream?
correct
> Regardless, we should just do self.constructed_response_model = extract_tool_return(...) and then yield self.constructed_response_model no need for the double assignment
done

mirascope/core/base/_utils/get_possible_user_message_param.py
```diff
@@ -7,6 +7,8 @@
 
 def get_possible_user_message_param(messages: list[_T]) -> _T | None:
     """Get the possible most recent user message."""
+    if len(messages) == 0:
```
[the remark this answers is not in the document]
brenkao: assume you mean `if not messages:`, done

mirascope/integrations/langfuse/_utils.py
brenkao: done

tests/integrations/logfire/test_utils.py
```diff
@@ -0,0 +1,506 @@
...
+                "content": [
+                    {
+                        "type": "text",
+                        "text": "Recommend a nonfiction book for me to read.",
+                    }
+                ],
+            }
+        ],
+        call_params={"tool_choice": "required"},
+        call_kwargs={
```
willbakst, line 81: why not just set `call_kwargs=MagicMock()` and `assert result["call_kwargs"] == call_response.call_kwargs`?

Seem unnecessary to have any real values here when we're just asserting that it's the same object, right?

Same is true for the other values here and in other tests.

tests/integrations/otel/test_utils.py
```diff
@@ -0,0 +1,585 @@
...
+                "content": [
+                    {
+                        "type": "text",
+                        "text": "Recommend a nonfiction book for me to read.",
+                    }
+                ],
+            }
+        ],
+        call_params={"tool_choice": "required"},
+        call_kwargs={
```
willbakst, line 130: see similar comment re: logfire tests

tests/integrations/test_middleware_factory.py
```diff
@@ -0,0 +1,414 @@
...
+        async_fn, handle_call_response_async=handle_call_response_async
+    )
+    result = await decorate()
+    assert result.content == call_response.content
+    decorate_with_context = middleware_decorator(
+        async_fn,
+        handle_call_response_async=handle_call_response_async,
+        custom_context_manager=my_context_manager,
+    )
+    await decorate_with_context()
```
willbakst, line 120: what are we testing by calling this?

tests/integrations/test_middleware_factory.py
```diff
@@ -0,0 +1,414 @@
...
+    result = await decorate()
+    async for chunk in result:
+        assert chunk.model_dump() == my_foo.model_dump()
+    decorate_with_context = middleware_decorator(
+        async_fn,
+        handle_structured_stream_async=handle_structured_stream_async,
+        custom_context_manager=my_context_manager,
+    )
+    async for chunk, _ in await decorate_with_context():
+        pass
```
willbakst, line 414: here and all other tests I'm a bit confused on what we're testing by running the function? Just that it runs?

willbakst (approved):

brianpark0126 (commented):
lgtm

PR #360, opened 2024-07-29
Anthropic tests

brenkao:


mirascope/core/anthropic/_utils/get_json_output.py
```diff
@@ -25,6 +29,7 @@ def get_json_output(
             response.chunk.type == "content_block_delta"
             and (delta := response.chunk.delta)
             and delta.type == "input_json_delta"
+            and delta.partial_json
```
willbakst, line 32: this check is unnecessary since the following line returns `""` and if this check is removed but `delta.partial_json` is the empty string the return will be the same.

mirascope/core/anthropic/call_response.py
```diff
@@ -78,17 +94,14 @@ def tools(self) -> list[AnthropicTool] | None:
...
         """
-        if not self.tool_types:
+        if not self.tool_types or not self.tool_calls:
             return None
 
         extracted_tools = []
-        for tool_call in self.response.content:
-            if tool_call.type != "tool_use":
-                continue
+        for tool_call in self.tool_calls:
```
willbakst, line 101: this doubles the iteration since we iterate first to get the tool calls and then again through the tool calls. I prefer having just a single iteration here

mirascope/core/anthropic/call_response.py
```diff
@@ -70,6 +76,16 @@ def finish_reasons(self) -> list[str]:
         """Returns the finish reasons of the response."""
         return [str(self.response.stop_reason)]
 
+    @property
+    def tool_calls(self) -> list[ToolUseBlock] | None:
```
willbakst, line 80: as discussed, I don't think we should add this (and instead we should remove the tool_calls property from all call responses and just have tools/tool properties)

mirascope/core/anthropic/call_response_chunk.py
```diff
@@ -66,6 +68,13 @@ def finish_reasons(self) -> list[str] | None:
             return [str(self.chunk.message.stop_reason)]
         return None
 
+    @property
+    def tool_call(self) -> ToolUseBlock | None:
```
willbakst, line 72: same thing here, we shouldn't add this, particularly because it doesn't match (e.g. OpenAI is `tool_calls` but this is `tool_call`.

If we find we need this convenience utility enough we should just add a utility function rather than on the call response chunk class.

tests/core/anthropic/_utils/test_setup_call.py
```diff
@@ -0,0 +1,139 @@
...
+    mock_base_setup_call.assert_called_once_with(
+        fn, {}, None, None, AnthropicTool, {"max_tokens": 1000}
+    )
+    mock_convert_message_params.assert_called_once_with(
+        mock_base_setup_call.return_value[1]
+    )
+    assert messages == mock_convert_message_params.return_value
+    assert inspect.signature(create) == inspect.signature(Anthropic().messages.create)
+
+    mock_base_setup_call.return_value[1] = [
```
willbakst, line 60: this seems like it should be a separate test for testing anthropic system message handling? something like `test_setup_call_system_message`

tests/core/anthropic/_utils/test_setup_call.py
```diff
@@ -0,0 +1,139 @@
...
+        {"role": "system", "content": [{"type": "text", "text": "test"}]}
+    ]
+    _, _, messages, _, call_kwargs = setup_call(
```
willbakst, line 63: if we're only testing call_kwargs here, why have `messages` as a named var?

tests/core/anthropic/_utils/test_setup_call.py
```diff
@@ -0,0 +1,139 @@
...
+        fn_args={},
+        dynamic_config=None,
+        tools=None,
+        json_mode=False,
+        call_params={"max_tokens": 1000},
+        extract=True,
+    )
+    assert "tool_choice" in call_kwargs and call_kwargs["tool_choice"] == {
+        "type": "tool",
+        "name": tool_types[0]._name(),  # type: ignore
```
willbakst, line 138: rather than type ignore this, we should first assert that `tool_types[0]` is an instance of `AnthropicTool`

tests/core/anthropic/test_stream.py
```diff
@@ -0,0 +1,137 @@
...
+        fn_args={},
+        dynamic_config=None,
+        messages=[{"role": "user", "content": "content"}],
+        call_params=AnthropicCallParams(max_tokens=1000),
+        call_kwargs={},
+    )
+    assert stream.cost is None
+    for _ in stream:
+        pass
+    # TODO: Verify if this is correct
```
willbakst, line 122: what do we need to verify?

tests/core/anthropic/test_tool.py
```diff
@@ -0,0 +1,68 @@
...
+    assert isinstance(tool, BaseTool)
+    assert tool.title == "The Name of the Wind"
+    assert tool.author == "Patrick Rothfuss"
+    assert tool.call() == "The Name of the Wind by Patrick Rothfuss"
+    assert FormatBook.tool_schema() == ToolParam(
+        name="FormatBook",
+        description="Returns the title and author nicely formatted.",
+        input_schema={
+            "$defs": {
+                "ToolUseBlock": {
```
willbakst, line 42: ah, this is wrong, we need to update `AnthropicTool` like I did for `OpenAITool` to exclude the tool call from the json schema. See https://github.com/Mirascope/mirascope/blob/f326e6beafa31914351e5de57383f14c08ee08c2/mirascope/core/openai/tool.py#L19

mirascope/core/anthropic/call_response.py
brenkao: done

mirascope/core/anthropic/call_response_chunk.py
brenkao: done

tests/core/anthropic/_utils/test_setup_call.py
brenkao: done

tests/core/anthropic/_utils/test_setup_call.py
brenkao: removed

tests/core/anthropic/test_stream.py
brenkao: oops, this was stale

tests/core/anthropic/test_tool.py
brenkao: done

tests/core/anthropic/_utils/test_setup_call.py
brenkao: done, just asserting its a list has the same effect

mirascope/core/anthropic/_utils/get_json_output.py
brenkao: done

mirascope/core/anthropic/call_response.py
brenkao: reverted

willbakst (approved):

PR #364, opened 2024-07-31
Cohere tests

brenkao:
Also fixed some bugs with cohere tools


mirascope/core/cohere/_utils/setup_call.py
```diff
@@ -50,9 +50,14 @@ def setup_call(
...
-        messages[-1].message += _utils.json_mode_content(
-            tool_types[0] if tool_types else None
+        # Cannot mutate ChatMessage in place
+        last_message = ChatMessage(
+            role=messages[-1].role,  # type: ignore
+            message=messages[-1].message
+            + _utils.json_mode_content(tool_types[0] if tool_types else None),
+            tool_calls=messages[-1].tool_calls,
         )
+        messages[-1] = last_message
```
willbakst, line 60: why not just assign the value directly rather than assign to `last_message` and then `messages[-1]`?

mirascope/core/cohere/call_response.py
```diff
@@ -1,5 +1,6 @@
 """This module contains the `CohereCallResponse` class."""
 
+from functools import cached_property
```
willbakst, line 3: what's this import for?

mirascope/core/cohere/call_response.py
```diff
@@ -144,10 +143,20 @@ def tool(self) -> CohereTool | None:
 
     @classmethod
     def tool_message_params(
-        cls, tools_and_outputs: list[tuple[CohereTool, list[dict[str, Any]]]]
+        cls,
+        tools_and_outputs: list[
+            tuple[CohereTool, list[dict[str, Any]] | dict[str, Any] | str]
```
willbakst, line 148: I think we should limit the output type to just `str`. All of the other providers have just `str` except for Gemini which has `object`, but I think we should probably update that to just `str` as well so the interface matches.

For now, the answer to how to use the other supported types would be to construct your own provider specific tool results

tests/core/cohere/_utils/test_convert_message_params.py
```diff
@@ -0,0 +1,68 @@
...
+def test_convert_message_params() -> None:
+    """Tests the `convert_message_params` function."""
+
+    message_params: list[BaseMessageParam | ChatMessage] = [
+        ChatMessage(message="Hello"),
+        BaseMessageParam(role="user", content="Hello"),
+        BaseMessageParam(
+            role="user",
+            content=[
+                TextPart(type="text", text="Hello"),
```
willbakst, line 19: you can just resolve this comment, but I'm realizing that this path should never be accessible since we only parse into parts if using multimodal. Still worth having this here I think for coverage, but worth noting I think.

tests/core/cohere/test_call_response.py
```diff
@@ -0,0 +1,148 @@
...
+    assert call_response.finish_reasons == ["COMPLETE"]
+    assert call_response.model == "command-r-plus"
+    assert call_response.id == "id"
+    assert call_response.usage == usage
+    assert call_response.input_tokens == 1
+    assert call_response.output_tokens == 1
+    assert call_response.cost == 1.8e-5
+    assert call_response.message_param == ChatMessage(
+        message="content",
+        role="assistant",  # type: ignore
```
willbakst, line 51: you can also resolve this but ugh not having role as a key is the weirdest most annoying thing like why

mirascope/core/cohere/_utils/setup_call.py
brenkao: done

mirascope/core/cohere/_utils/convert_message_params.py
```diff
@@ -20,12 +20,13 @@ def convert_message_params(
...
             )
         else:
-            if len(content) != 1 or content[0].type != "text":
+            if content[0].type != "text":
                 raise ValueError("Cohere does not currently support multimodalities.")
-            converted_message_params.append(
-                ChatMessage(
-                    role=message_param.role.upper(),  # type: ignore
-                    message=content[0].text,
+            if len(content) > 1:
```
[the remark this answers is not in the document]
brenkao: done

mirascope/core/cohere/call_response.py
brenkao: originally fixed a bug but later realized the bug was somewhere else, removed.

tests/core/cohere/_utils/test_convert_message_params.py
brenkao: yeah its pretty much never accessible, but for coverage

mirascope/core/cohere/call_response.py
brenkao: okay, done

tests/core/cohere/test_call_response.py
brenkao: story of all my tests

willbakst (approved):

PR #369, opened 2024-08-02
Readme updates

willbakst:


tests/core/base/_utils/test_format_template.py
```diff
@@ -22,6 +22,7 @@ def test_format_template(
     mock_get_template_values.return_value = attrs
     template = """
     Recommend a {genre} book.
+
```
brenkao, line 25: What is this line for?

brenkao (approved):

tests/core/base/_utils/test_format_template.py
willbakst: the ensure that we strip extra whitespace to reduce tokens.

PR #374, opened 2024-08-06
Construct call response from Stream

brenkao:


mirascope/core/openai/stream.py
```diff
@@ -39,13 +45,60 @@ def cost(self) -> float | None:
...
+                            name=function_dict["name"],
+                        ),
+                        type="function",
+                    )
+                )
+            message.tool_calls = tool_calls
+        completion = ChatCompletion(
+            id="id",
+            model=self.model,
+            choices=[Choice(finish_reason="stop", index=0, message=message)],
```
willbakst, line 87: shouldn't you just be able to set this to `ChatCompletionMessage(**self.message_param)` and call it a day?

brenkao: oh oops this was leftover code, updated

mirascope/core/openai/stream.py
```diff
@@ -61,30 +61,21 @@ def construct_call_response(self) -> OpenAICallResponse:
...
-                        id=stream_tool_call["id"],
-                        function=Function(
-                            arguments=function_dict["arguments"],
-                            name=function_dict["name"],
-                        ),
-                        type="function",
-                    )
-                )
-            message.tool_calls = tool_calls
+        message = {
```
willbakst, line 64: this is unnecessary, just unroll the message_param

brenkao: as per our discussion, since there is an extra 'name' field this is fine.

mirascope/core/anthropic/stream.py
```diff
@@ -34,9 +40,67 @@ def cost(self) -> float | None:
     def _construct_message_param(
         self, tool_calls: list[ToolUseBlock] | None = None, content: str | None = None
     ) -> MessageParam:
-        """Returns the tool message parameters for tool call results."""
+        """Constructs the message parameter for the assistant."""
+        message_param_tool_calls = []
+        if tool_calls:
+            message_param_tool_calls: list[ToolUseBlockParam] = [
+                tool_call.model_dump()
+                for tool_call in tool_calls  # type: ignore
```
willbakst, line 48: why does this need a type ignore?

mirascope/core/anthropic/stream.py
```diff
@@ -34,9 +40,67 @@ def cost(self) -> float | None:
...
+            raise ValueError(  # pragma: no cover
+                "No stream response, check if the stream has been consumed."
+            )
+        usage = Usage(input_tokens=0, output_tokens=0)
+        if self.input_tokens:
+            usage.input_tokens = int(self.input_tokens)
+        if self.output_tokens:
+            usage.output_tokens = int(self.output_tokens)
+
+        message_param: list[ContentBlock] = []
```
willbakst, line 67: I'm confused what the difference here is between this local `message_param` and the `self.message_param` below. The naming here makes it seem like the stuff below is unnecessary.

I'm assuming we have to do some conversion for some reason, but it's unclear why. Should this just be named `content`?

mirascope/core/anthropic/stream.py
```diff
@@ -34,9 +40,67 @@ def cost(self) -> float | None:
...
+                if content_type == "text":
+                    message_param.append(TextBlock.model_validate(content))
+                elif content_type == "tool_use":
+                    message_param.append(ToolUseBlock.model_validate(content))
+        completion = Message(
+            id=self.id if self.id else "",
+            content=message_param,
+            model=self.model,
+            role="assistant",
+            stop_reason=self.finish_reasons[0] if self.finish_reasons else None,  # type: ignore
```
willbakst, line 88: why is there a type ignore here? If it's because of something with the base type, then we should be using generics to properly type this.

mirascope/core/anthropic/stream.py
```diff
@@ -34,9 +40,67 @@ def cost(self) -> float | None:
...
+            ]
         return {
             "role": "assistant",
-            "content": ([TextBlock(text=content, type="text")] if content else [])
-            + (tool_calls or []),
+            "content": ([TextBlockParam(text=content, type="text")] if content else [])
+            + message_param_tool_calls,
         }
+
+    def construct_call_response(self) -> AnthropicCallResponse:
```
willbakst, line 56: missing docstring for this function. Make sure in the description to make it clear that this is a constructed response and might not be 1:1 equivalent with what the response would be if you made the same call without streaming

mirascope/core/base/_stream.py
```diff
@@ -77,9 +78,13 @@ class BaseStream(
     call_params: _BaseCallParamsT
     call_kwargs: dict[str, Any]
     user_message_param: _UserMessageParamT | None = None
-    message_param: _AssistantMessageParamT
+    message_param: _AssistantMessageParamT | None = None
     input_tokens: int | float | None = None
     output_tokens: int | float | None = None
+    id: str | None = None
+    finish_reasons: list[str] | None = None
```
willbakst, line 85: aha this is the culprit, we should probably make this a generic for good typing elsewhere

mirascope/core/cohere/stream.py
```diff
@@ -39,3 +45,33 @@ def _construct_message_param(
             message=content if content else "",
             tool_calls=tool_calls,
         )
+
+    def construct_call_response(self) -> CohereCallResponse:
```
willbakst, line 49: missing docstring

mirascope/core/cohere/stream.py
```diff
@@ -39,3 +45,33 @@ def _construct_message_param(
...
+        if self.message_param is None:
+            raise ValueError(  # pragma: no cover
+                "No stream response, check if the stream has been consumed."
+            )
+        usage = ApiMetaBilledUnits(
```
willbakst, line 54: can't usage be `None`? If `self.input_tokens` and `self.output_tokens` are both `None` we should set `usage` to `None`

mirascope/core/gemini/stream.py
```diff
@@ -35,7 +44,42 @@ def cost(self) -> float | None:
     def _construct_message_param(
         self, tool_calls: list[FunctionCall] | None = None, content: str | None = None
     ) -> ContentDict:
-        return {
-            "role": "model",
-            "parts": [{"type": "text", "text": content}] + (tool_calls or []),
-        }
+        """Constructs the message parameter for the assistant.
+        Gemini currently does not support tool calls for streaming.
```
willbakst, line 48: hmm, this feels bad because if they do support streaming we'll need to come back here and change this vs. what we're currently doing which is always settings tool_calls to `None` because it always will be.

This is an internal function, so why remove the functionality at all?

mirascope/core/gemini/stream.py
```diff
@@ -35,7 +44,42 @@ def cost(self) -> float | None:
...
+        """
+        return {"role": "model", "parts": [{"text": content}]}  # type: ignore
+
+    def construct_call_response(self) -> GeminiCallResponse:
+        if self.message_param is None:
+            raise ValueError(  # pragma: no cover
```
willbakst, line 54: put the no cover at the end of the `ValueError` on line 56

mirascope/core/gemini/stream.py
```diff
@@ -35,7 +44,42 @@ def cost(self) -> float | None:
...
+            )
+        )
+        return GeminiCallResponse(
+            metadata=self.metadata,
+            response=response,
+            tool_types=self.tool_types,
+            prompt_template=self.prompt_template,
+            fn_args=self.fn_args if self.fn_args else {},
+            dynamic_config=self.dynamic_config,
+            messages=self.messages,  # type: ignore
```
willbakst, line 79: again, don't love the type ignore if we can find a way around it, and `self.messages` should be the right type, and if it's not we should fix that

mirascope/core/groq/call_response_chunk.py
```diff
@@ -40,7 +40,12 @@ def content(self) -> str:
     @property
     def finish_reasons(self) -> list[str]:
         """Returns the finish reasons of the response."""
-        return [str(choice.finish_reason) for choice in self.chunk.choices]
+
```
willbakst, line 43: nit: remove extra newline

mirascope/core/groq/stream.py
```diff
@@ -49,3 +52,43 @@ def _construct_message_param(
...
+            "tool_calls": self.message_param.get("tool_calls", []),
+        }
+        completion = ChatCompletion(
+            id=self.id if self.id else "",
+            model=self.model,
+            choices=[
+                Choice(
+                    finish_reason=self.finish_reasons[0]
+                    if self.finish_reasons and self.finish_reasons[0]
+                    else "stop",  # type: ignore
```
willbakst, line 73: ditto here, work to remove type ignore

mirascope/core/groq/stream.py
```diff
@@ -49,3 +52,43 @@ def _construct_message_param(
...
+                    index=0,
+                    message=ChatCompletionMessage.model_validate(message),
+                )
+            ],
+            created=0,
+            object="chat.completion",
```
willbakst, line 79: where is usage?

mirascope/core/mistral/stream.py
```diff
@@ -38,3 +43,45 @@ def _construct_message_param(
             role="assistant", content=content if content else "", tool_calls=tool_calls
         )
         return message_param
+
+    def construct_call_response(self) -> MistralCallResponse:
+        if self.message_param is None:
+            raise ValueError(  # pragma: no cover
+                "No stream response, check if the stream has been consumed."
+            )
+        usage = UsageInfo(prompt_tokens=0, completion_tokens=0, total_tokens=0)
```
willbakst, line 52: why not just do something like:
```python
usage = UsageInfo(
    prompt_tokens=(self.input_tokens or 0),
    completion_tokens=(self.output_tokens),
    total_tokens=((self.input_tokens or 0) + (self.output_tokens or 0))
)
```
Maybe we can do this in the other providers as well

mirascope/core/mistral/stream.py
```diff
@@ -38,3 +43,45 @@ def _construct_message_param(
...
+        if self.output_tokens:
+            usage.completion_tokens = int(self.output_tokens)
+        usage.total_tokens = usage.prompt_tokens + (usage.completion_tokens or 0)
+        completion = ChatCompletionResponse(
+            id=self.id if self.id else "",
+            choices=[
+                ChatCompletionResponseChoice(
+                    finish_reason=self.finish_reasons[0]
+                    if self.finish_reasons
+                    else None,  # type: ignore
```
willbakst, line 64: ditto on the type ignore

mirascope/core/openai/stream.py
```diff
@@ -39,13 +44,54 @@ def cost(self) -> float | None:
...
+        """Constructs the message parameter for the assistant."""
+        message_param = ChatCompletionMessage(
             role="assistant",
             content=content,
         )
         if tool_calls:
-            message_param["tool_calls"] = tool_calls
-        return message_param
+            message_param.tool_calls = tool_calls
+        return message_param.model_dump(exclude={"function_call"})  # type: ignore
```
willbakst, line 57: remove type ignore and put type back to the assistant message for `message_param` I don't see why you changed it just to dump it?

mirascope/core/openai/stream.py
```diff
@@ -39,13 +44,54 @@ def cost(self) -> float | None:
...
+            "tool_calls": self.message_param.get("tool_calls", []),
+        }
+        completion = ChatCompletion(
+            id=self.id if self.id else "",
+            model=self.model,
+            choices=[
+                Choice(
+                    finish_reason=self.finish_reasons[0]
+                    if self.finish_reasons
+                    else "stop",  # type: ignore
```
willbakst, line 76: ditto on the type ignore

mirascope/core/openai/stream.py
```diff
@@ -39,13 +44,54 @@ def cost(self) -> float | None:
...
+                    index=0,
+                    message=ChatCompletionMessage.model_validate(message),
+                )
+            ],
+            created=0,
+            object="chat.completion",
```
willbakst, line 82: where is usage?

tests/core/anthropic/test_stream.py
```diff
@@ -122,17 +124,228 @@ def generator():
...
+        messages=[],
+        call_params=AnthropicCallParams(max_tokens=1000),
+        call_kwargs={},
+        user_message_param=None,
+        start_time=0,
+        end_time=0,
+    )
+    constructed_call_response = stream.construct_call_response()
+
+    assert constructed_call_response._provider == call_response._provider
```
willbakst, line 291: why not just assert the model dumps are equal?

tests/core/cohere/test_stream.py
```diff
@@ -62,3 +62,76 @@ def generator():
...
+        fn_args={},
+        dynamic_config=None,
+        messages=[],
+        call_params={},
+        call_kwargs={},
+        user_message_param=None,
+        start_time=0,
+        end_time=0,
+    )
+    assert constructed_call_response._provider == call_response._provider
```
willbakst, line 128: ditto here and across all providers

mirascope/core/anthropic/stream.py
brenkao: it was bc of model dump, updated to convert to ToolUseBlockParam

mirascope/core/anthropic/stream.py
brenkao: we are converting MessageParam to ContentBlock, from dict to pydantic model

mirascope/core/anthropic/stream.py
brenkao: should we pass in finish reason as a generic type? I feel like if thats the case I should do that for all inputs. Ideally I would be able to declare a property but I wasn't sure how to go about that

mirascope/core/base/_stream.py
brenkao: see above comment regarding mirascope/core/anthropic/stream.py

mirascope/core/anthropic/stream.py
brenkao: done

mirascope/core/openai/stream.py
brenkao: updated this to be more clear

mirascope/core/openai/stream.py
brenkao: oops, added

tests/core/anthropic/test_stream.py
brenkao: i dont copy over everything to stream when constructing the call response, we would need to implement provider specific properties to save all of them when we iterate chunks when we construct call params

brenkao: done

tests/core/cohere/test_stream.py
brenkao: done

mirascope/core/base/_stream.py
brenkao: done

mirascope/core/anthropic/stream.py
brenkao: done

mirascope/core/openai/stream.py
brenkao: done

mirascope/integrations/langfuse/_utils.py
```diff
@@ -37,7 +37,7 @@ def get_stream_observation(stream: BaseStream, fn: Callable):
         "input": stream.prompt_template,
         "tags": tags,
         "model": stream.model,
-        "output": stream.message_param.get("content", None),
+        "output": stream.message_param.get("content", None),  # type: ignore
```
brenkao, line 40: this will be refactored when i work on langfuse integration

mirascope/core/anthropic/stream.py
willbakst: right, `list[ContentBlock]` should not be named `message_param` it should be named `content`

mirascope/core/gemini/call_response_chunk.py
```diff
@@ -43,21 +48,10 @@ def content(self) -> str:
...
-            "SAFETY",
-            "RECITATION",
-            "OTHER",
-        ]
-
-        return [
-            finish_reasons[candidate.finish_reason]
-            for candidate in self.chunk.candidates
-        ]
+
```
willbakst, line 53: nit: remove extra newline

mirascope/core/groq/stream.py
```diff
@@ -49,3 +56,55 @@ def _construct_message_param(
...
+        if self.message_param is None:
+            raise ValueError(  # pragma: no cover
+                "No stream response, check if the stream has been consumed."
+            )
+        message = {
+            "role": self.message_param["role"],
+            "content": self.message_param.get("content", ""),
+            "tool_calls": self.message_param.get("tool_calls", []),
+        }
+        usage = CompletionUsage(
```
willbakst, line 71: usage can be `None` and we should handle that so we get `None` cost instead of `0` in that case

mirascope/core/openai/stream.py
```diff
@@ -39,13 +48,75 @@ def cost(self) -> float | None:
...
+        if self.message_param is None:
+            raise ValueError(  # pragma: no cover
+                "No stream response, check if the stream has been consumed."
+            )
+        message = {
+            "role": self.message_param["role"],
+            "content": self.message_param.get("content", ""),
+            "tool_calls": self.message_param.get("tool_calls", []),
+        }
+        usage = CompletionUsage(
```
willbakst, line 83: see other comment about this

tests/core/gemini/test_stream.py
```diff
@@ -64,5 +64,80 @@ def test_gemini_stream() -> None:
...
+        dynamic_config=None,
+        messages=[],
+        call_params={},
+        call_kwargs={},
+        user_message_param=None,
+        start_time=0,
+        end_time=0,
+    )
+    constructed_call_response = stream.construct_call_response()
+    assert constructed_call_response.response.__eq__(call_response.response)
```
willbakst, line 143: why not just use the `==` operator?

brenkao: Gemini things

mirascope/core/anthropic/stream.py
brenkao: done

mirascope/core/groq/stream.py
brenkao: done

mirascope/core/mistral/stream.py
brenkao: i have to check for float, so i think what i'm doing is cleaner

mirascope/core/gemini/stream.py
brenkao: done

mirascope/core/cohere/stream.py
brenkao: done

willbakst (approved):

willbakst (approved):

PR #375, opened 2024-08-06
Initial cookbook templates

willbakst:


docs/cookbook/llm_validation_with_retries.md
```diff
@@ -0,0 +1,155 @@
+# LLM Validation With Retries
+
+!!! info inline end "Concepts Covered"
```
brianpark0126, line 3: i also prefer the way brendan showed it, as a bullet. Its nice in its own box but to the right side of the text looks far too wonky

docs/cookbook/llm_validation_with_retries.md
```diff
@@ -0,0 +1,155 @@
...
+!!! note "Background"
+
+    While traditional validation tools like type checkers or Pydantic are limited to hardcoded rules (such as variable types or arithmetic), LLMs allow for much more nuanced and complex validation. This approach can be particularly useful for validating natural language inputs or complex data structures where traditional rule-based validation falls short.
+
+## Basic LLM Validation
+
+Let's start with a simple example of using an LLM to check for spelling and grammatical errors in a text snippet:
+
+```python
+from mirascope.core import anthropic, prompt_template
```
brianpark0126, line 20: for markdowns that refer to a cookbook with multiple .py files, i think adding a comment at the top of each file referencing which .py file the following snippet lives in would be a nice convenience

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+
+```python
+from openai.types.chat import (
+    ChatCompletionAssistantMessageParam,
+    ChatCompletionUserMessageParam,
+)
+from pydantic import BaseModel
+
+
+class FewShotExample(BaseModel):
```
brianpark0126, line 29: do you want this sort of defined class for all few shot examples? Or is there value in showing that if you're dealing with a rawer form of text as your few shot examples (because you got those prompts from somewhere else) its still easy to parse within a Mirascope call

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+print(response.content)
+# > Are follow up questions needed here: Yes.
+#   Follow up: What is the birth country of Jayantha Ketagoda?
+#   Intermediate answer: Jayantha Ketagoda is from Sri Lanka.
+#   Follow up: When did Sri Lanka leave the British Empire?
+#   Intermediate answer: Sri Lanka, formerly known as Ceylon, gained independence from the British Empire on February 4, 1948.
+#   So the final answer is: February 4, 1948.
+```
+
+## Enhancing Self-Ask Implementation
```
brianpark0126, line 156: per your slack comment, just making sure that it's fine for me to only do up to this part for the rest of text based prompt engineering techniques

docs/cookbook/llm_validation_with_retries.md
brenkao, line 3: Perhaps adding a Header 2 here would be nicer, so people could also quick nav up top. This also makes the introduction easier to read.

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
+# Self-Ask: Enhancing LLM Reasoning with Follow-Up Questions
+
+!!! info inline end "Concepts Covered"
```
brenkao, line 3: ditto

docs/cookbook/text_classification.md
```diff
@@ -1,39 +1,50 @@
+# Text Classification with LLMs
 
-# Text Classification
+!!! info inline end "Concepts Covered"
```
brenkao, line 3: ditto

docs/cookbook/llm_validation_with_retries.md
```diff
@@ -0,0 +1,155 @@
+# LLM Validation With Retries
+
+!!! info inline end "Concepts Covered"
+
+    - [Calls](ADD LINK)
+    - [Response Model](ADD LINK)
+    - [Tenacity Integration](ADD LINK)
+
+This recipe demonstrates how to leverage Large Language Models (LLMs) -- specifically Anthropic's Claude 3.5 Sonnet -- to perform automated validation on any value. We'll cover how to use **LLMs for complex validation tasks**, how to integrate this with Pydantic's validation system, and how to leverage [Tenacity](ADD LINK) to automatically **reinsert validation errors** back into an LLM call to **improve results**.
```
brenkao, line 9: We should probably mention we need to specifically `pip install mirascope[anthropic]`. People who only pip install mirascope may not have this installed. Not sure how important that is

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+Finally, we can use our Self-Ask implementation:
+
+```python
+response = self_ask_query(
+    query="The birth country of Jayantha Ketagoda left the British Empire when?",
+    examples=few_shot_examples,
+)
+print(response.content)
```
brianpark0126, line 147: also confirming that you want all outputs to be printed in this format, not in a table

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+
+
+@openai.call(model="gpt-4o-mini")
+@prompt_template(
+    """
+    MESSAGES: {example_prompts}
+    USER: {query}
+    """
+)
+def self_ask_query(
```
brianpark0126, line 56: why is it self_ask_query and not self_ask_call?

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+            """
+        ),
+    ),
+]
+```
+
+Finally, we can use our Self-Ask implementation:
+
+```python
+response = self_ask_query(
```
brianpark0126, line 143: are we not showing standard call (no prompt engineering) on the prompt? or was it just forgotten

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+    query="The birth country of Jayantha Ketagoda left the British Empire when?",
```
brianpark0126, line 144: should we call the prompt engineered call by directly defining the query arg? Especially if its 1) super long and 2) we reuse it in a non engineered call. In this case, I've been going by naming the variable `prompt`, please confirm we're defining the argument directly and if not, if the variable name `prompt` is fine or you want something different

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+            "example_prompts": [
+                message
+                for example in relevant_examples
+                for message in example.messages()
+            ]
+        }
+    }
+
+
+query = "What was the primary language spoken by the inventor of the phonograph?"
```
brianpark0126, line 209: this answers my above question, but just confirming that the question asked to the model is, going forward, named query

docs/cookbook/llm_validation_with_retries.md
willbakst: Ok this is a fair point. I've made "Concepts Covered" a hidden info section and renamed it to "Key Concepts" and put it right above "Background" so it's easier to read / navigate.

docs/cookbook/prompt_engineering/text_based/self_ask.md
willbakst: Yeah I think `query` is a good / common naming convention to use here going forward.

If you agree please also resolve the other comment you're mentioning.

docs/cookbook/llm_validation_with_retries.md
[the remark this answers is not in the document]
willbakst: Ok, I removed one of them. I'm leaving two because they are actually different links (one will point to tenacity whereas the other will point to our integration with tenacity)

docs/cookbook/prompt_engineering/text_based/self_ask.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Dynamic example selection: Choose relevant few-shot examples based on the input query.
+- Fallback mechanism: Implement a fallback to standard querying if Self-Ask doesn't improve the result.
+- Result validation: Add a step to validate the final answer against the initial query.
+
+Here's an example of how you might implement dynamic example selection:
+
+```python
+from mirascope.core import openai, prompt_template
+from sklearn.feature_extraction.text import TfidfVectorizer
```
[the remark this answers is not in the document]
willbakst: Yes this is a good catch. I'm going to add a setup section to all of the cookbooks.

I've resolved all other related comments, please resolve this if the new setup sections looks good and address concerns.

docs/cookbook/llm_validation_with_retries.md
willbakst: Hmm, I tried playing around with this but didn't like any outcomes for the following reasons:
1. There are some snippets that are broken up, so that becomes weirder to handle
2. The code is already matched to the headers, so the header acts as the file name essentially. I could link the headers, but that looked weird. I've opted instead to add a block at the top of each section, see image below

Screenshot 2024-08-06 at 11 40 41 AM

docs/cookbook/prompt_engineering/text_based/self_ask.md
willbakst: It seemed to make sense to me here particularly because it demonstrates how easy it is to create something you can re-use (e.g. if you want few shot examples in multiple places in your codebase).

I opted against creating a shared definition because I found it easier to grok things when all colocated ;)

For example, we may not include the `messages` function here if we decide to include few shot examples into prompts in other recipes in a different way (maybe we format them as text and inject them into a single message).

For now, we should just redefine them for each cookbook as makes sense.

docs/cookbook/prompt_engineering/text_based/self_ask.md
willbakst: I think I'll just take all of the text-based ones so I can do the entire thing. Ideally any recipe you take you'll do the whole thing.

docs/cookbook/prompt_engineering/text_based/self_ask.md
willbakst: I had the table originally, but when I was running the script this seemed to make more sense, so I just kept it this way. I think a table can be super useful if e.g. we're comparing things side-by-side, but for single or sequential outputs I think just printing things is totally fine

docs/cookbook/prompt_engineering/text_based/self_ask.md
willbakst: I don't like the variable name "prompt" compared to it being the user's query which gets injected into the final prompt (that we're engineering).

It's fine if you want to define it as a variable so you can re-use it, I just wasn't doing that here and do defined it directly.

willbakst: Acutally revisiting this, I updated to handle the no example comparison and ended up refactoring. Now I agree I think having it as a var should be the default regardless going forward it looks nicer.

brianpark0126 (approved):

PR #378, opened 2024-08-07
fixed some bugs and finalized logfire

brenkao:


docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
+# Logfire
+
+[Logfire](https://docs.pydantic.dev/logfire/), a new tool from Pydantic, is built on OpenTelemetry. As Mirascope is also built on Pydantic, it's appropriate for us to ensure seamless integration with them.
```
willbakst, line 3: I think we can drop the "built on Pydantic" bit at this point. Would rather just say we love logfire and pydantic instead

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+
+## How to use Logfire with Mirascope
+
+```python
+pip install logfire
+
+from mirascope.integrations.logfire import with_logfire
+```
+
+`with_logfire` is a decorator that can be used on all Mirascope classes to automatically log both Mirascope calls and also all our [supported LLM providers](https://docs.mirascope.io/latest/SUPPORTED_PROVIDERS/).
```
willbakst, line 13: update, not just "classes" but also "functions"?

Does it work on tools? tool calls? Might be worth being just a little more specific here

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+from mirascope.integrations.logfire import with_logfire
+
+logfire.configure()
+
+def format_book(title: str, author: str):
+    return f"{title} by {author}"
+
+@with_logfire
+@anthropic.call(model="claude-3-5-sonnet-20240620", tools=[format_book])
+@prompt_template("""Recommend a {genre} book.""")
```
willbakst, line 34: this can be a normal string, doesn't need to be a docstring

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+def recommend_book(genre: str): ...
+
+recommend_book("fantasy")
```
willbakst, line 37: maybe at least set `response = recommend_book("fantasy")`?

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+for chunk, _ in recommend_book("fantasy"):
+    print(chunk.content, end="", flush=True)
+```
+
+For some providers, certain `call_params` will need to be set in order for usage to be tracked.
+Also note that the span will not be logged until the stream has been exhausted.
+
+### Response Model
+
+Since Mirascope `response_model` is built on top of [Pydantic](https://docs.pydantic.dev/latest/), you can use the [Pydantic Plugin](https://docs.pydantic.dev/latest/concepts/plugins/) to track additional logs and metrics about model validation, which you can enable using the pydantic_plugin configuration.
```
willbakst, line 79: maybe instead of "built on top of" we can just say "uses" or maybe something like "Since Pydantic powers Mirascope `response_model` ..."

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+
+This can be particularly useful when [extracting structured information using LLMs](https://docs.mirascope.io/latest/learn/response_models/):
```
willbakst, line 81: for now for internal links to our own docs just write (ADD LINK) since these may change. External links are fine (like the ones you have for Pydantic).

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+    title: str
+    author: str
+
+
+@with_logfire
+@openai.call(model="gpt-4o-mini", response_model=Book)
+@prompt_template("""Recommend a {genre} book.""")
+def recommend_book(genre: str): ...
+
+recommend_book("fantasy")
```
willbakst, line 102: same thing here, at least capture the response

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+This will give you:
+
+* Tracking for validation of Pydantic models
+* All the information from call with a `response_model` output.
+
+![logfire-response-model](../assets/logfire-response-model.png)
+
+You can also set `stream=True` with the same rules as streaming a call. Note that the Pydantic Plugin will log many times if you are streaming a `response_model`, so it is recommended to wrap the Mirascope call in a function.
```
willbakst, line 112: hmm, will the response validations not already be included/wrapped in the call itself?

docs/integrations/logfire.md
```diff
@@ -0,0 +1,164 @@
...
+client = Client()
+logfire.configure()
+logfire.instrument_openai(client)
+
+@openai.call(model="gpt-4o-mini", client=client)
+@prompt_template("""Recommend a {genre} book.""")
+def recommend_book(genre: str): ...
+
+
+recommend_book("fantasy")
```
willbakst, line 163: again, capture response

willbakst (approved):

docs/integrations/logfire.md
brenkao: done

docs/integrations/logfire.md
brenkao: done

docs/integrations/logfire.md
brenkao: no, since everything is happening after the call? if that makes sense

docs/integrations/logfire.md
brenkao: done

docs/integrations/logfire.md
brenkao: updated below helper text to include tools

docs/integrations/logfire.md
brenkao: done

docs/integrations/logfire.md
brenkao: i updated to print the result with output

docs/integrations/logfire.md
brenkao: done

docs/integrations/logfire.md
brenkao: done

docs/integrations/logfire.md
willbakst: Oh because it's a stream, that makes sense

PR #380, opened 2024-08-08
updated tenacity integration

brenkao:


docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+from mirascope.core.base import BaseCallResponse
+
+
+def handle_call_response(
+    result: BaseCallResponse, fn: Callable, context_manager: ContextManagerType | None
+):
+    # handle after call here
+```
+
+The first argument will be a Mirascope `CallResponse` of the provider you are using.
```
willbakst, line 49: IIUC these won't actually show up on separate lines unless theres two newlines? Do you want these on a single line and are just writing it this way for readability in the markdown?

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+
+### `handle_response_model` and `handle_response_model_async`
+
+These functions will be called after making a mirascope call with `response_model` set with the following signature:
+
+```python
+from typing import Callable
+from pydantic import BaseModel
+
+from mirascope.core.base._utils._base_type import BaseType
```
willbakst, line 84: hmm, I don't like telling people to import something from a private module. Instead this should be:

```python
from mirascope.core.base import BaseType
```

This will require updating the internal imports in the init, and that's fine.

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+
+def handle_response_model(
+    response_model: BaseModel | BaseType, fn: Callable, context_manager: ContextManagerType | None
+):
+    # handle after call with response_model here
+```
+
+The first argument will be a Pydantic `BaseModel` or Python primative depending on what type `response_model` is.
```
willbakst, line 92: users will need to be able to distinguish between the two, so should we mention that they should first check if it's an instance of `BaseModel` and then otherwise handle a `BaseType` instance? Just to be extra clear / helpful?

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+```python
+from typing import Callable
+
+from mirascope.core.base import BaseCallResponse
+
+
+def handle_call_response(
+    result: BaseCallResponse, fn: Callable, context_manager: ContextManagerType | None
+):
+    # handle after call here
```
willbakst, line 46: I think we should show an implementation for at least one of the functions (if not all) even if it's something super simple.

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+```python
+from typing import Callable
+
+from mirascope.core.base._structured_stream import BaseStructuredStream
+
+def handle_structured_stream(
+    result: BaseStructuredStream,
+    fn: Callable,
+    context_manager: ContextManagerType | None,
+):
```
willbakst, line 111: either missing `...` or a comment?

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+from typing import Any, Callable, Generator
+
+from sqlmodel import Session
+
+from mirascope.core.base import BaseCallResponse
+
+
+@contextmanager
+def custom_context_manager(
+    fn: Callable,
```
willbakst, line 134: confusing why we pass `fn` here at all, so I feel like we should either show using it or explain what it's for?

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+    result: BaseCallResponse, fn: Callable, session: Session | None
+):
+    # write to db 
+```
+
+In this example, the 3rd argument of your handler will be typed with the yielded `Session` and metadata can be written to your database of choice.
+
+### `custom_decorator`
+
+There may be existing libraries that have a decorator implemented already. You can pass that decorator in to `custom_decorator` which will wrap the Mirascope call with that decorator.
```
willbakst, line 150: this can be more clear around what's happening i.e. in what order is the decorator called wrt. the middleware and other handlers (before or after?)

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+    if not session:
+        return
+
+    session.add(result)
+    session.commit()
+```
+
+Now, any Mirascope call that uses the `with_saving` decorator will write to your database.
+
+If there is a library that you would like for us to integrate out-of-the-box, let us know in our [Slack](https://join.slack.com/t/mirascope-community/shared_invite/zt-2ilqhvmki-FB6LWluInUCkkjYD3oSjNA) community.
```
willbakst, line 191: hmm, I feel like we should first mention posting a github issue as well e.g. "post an issue or let us know in our slack community" or something

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,191 @@
...
+
+```python
+from typing import Callable
+from sqlmodel import Session
+
+from mirascope.core.base import BaseCallResponse
+
+
+def handle_call_response(
+    result: BaseCallResponse, fn: Callable, session: Session | None
```
willbakst, line 179: this confuses me, how can it be `session` shouldn't it be a context manager?

docs/integrations/custom_middleware.md
willbakst, line 1: overall content is good but I think the order could be switched up a bit to flow better, and I think implementation details of at least one or a few (if not all) or the handlers would go a really, really long way to help others get started. It doesn't have to be anything crazy, but at least something more than just a comment saying to implement it.

willbakst: In terms of order, I think since all the handlers depend on the context manager we should show that first right after showing the middleware_decorator

docs/integrations/tenacity.md
```diff
@@ -0,0 +1,85 @@
...
+@anthropic.call(model="claude-3-5-sonnet-20240620")
+@prompt_template("What is your purpose?")
+def run(*, errors: list[InternalServerError] = []):
+    handle_internal_server_error(errors)
+
+
+print(run())
+```
+
+Let's break this down. The most common situation is that the call goes through and runs like normal. In the chance that Anthropic is down, we can collect specified `InternalServiceError` using the `collect_errors` function that Mirascope provides. We will make the request again, after waiting for some time and in the meantime call the `handle_internal_server_error` function to inform end users that there's a service outage, or better, make a call to another provider.
```
willbakst, line 34: hmm, I'm realizing that the handler here won't actually be called until after the next retry attempt (rather than being called right when it fails), so I'm wondering if (at least until we figure this out) we should just show no collection handling for InternalServiceError and then for validation errors we can show collecting them (and users can extrapolate from there for other errors they may want to include etc.)

docs/integrations/tenacity.md
```diff
@@ -0,0 +1,85 @@
...
+) -> openai.OpenAIDynamicConfig:
+    return {"computed_fields": {"previous_errors": errors}}
+
+
+book = "the name of the wind by patrick rothfuss"
+book_details = book_extractor(book=book)
+print(book_details)
+```
+
+For demonstrative purposes, we first prompt engineer the result to return a lower case title and author. We then feed the `ValidationError` "Title must be in all caps." back to the prompt, which is then used in the subsequent call, returning an all caps title and author.
```
willbakst, line 85: we don't actually have to prompt engineer this, it won't return it all caps on the first go (at least not in my experience). I think better here to just print the errors in the dynamic configuration section to see that it fails the first time and then succeeds the second time (which better shows the benefits of the reinsertion)

docs/integrations/tenacity.md
```diff
@@ -0,0 +1,85 @@
...
+)
+@prompt_template(
+    """
+    {previous_errors}
+    Extract the following Book details lowercased:
+    {book}
+    """
+)
+def book_extractor(
+    book: str, *, errors: list[ValidationError] = []
```
willbakst, line 75: never default to an object, this should default to `None`

docs/integrations/tenacity.md
brenkao: Is that a bad thing? Seems like its still useful, we could probably use before instead of after right?

docs/integrations/tenacity.md
brenkao: good point

docs/integrations/custom_middleware.md
brenkao: The Session contextmanager yields a Session object.

docs/integrations/custom_middleware.md
brenkao: Okay so just mention this decorator is called before?

docs/integrations/custom_middleware.md
brenkao: Should I not show it at all? What we use it for is to grab function name, but not sure how it would be used for SQLModel

docs/integrations/custom_middleware.md
brenkao: yes

docs/integrations/custom_middleware.md
brenkao: See my example with SQLModel, you think I should move that code snippet up here? Should I do the same with the other handlers, or extrapolate.

willbakst: same with other handlers, and yeah I think if we're going to demonstrate how to build your own by implementing `with_saving` we should essentially implement with saving and do the same for other handlers

docs/integrations/custom_middleware.md
willbakst: can `custom_context_manager` be a callable that takes no arguments?

docs/integrations/custom_middleware.md
willbakst: What confuses me is that the definition above has `context_manager: ContextManagerType | None` which makes it seem like `Session` is a `ContextManagerType`?

And then I don't see any `with ... as ...` which I generally expect from a context manager...

docs/integrations/tenacity.md
willbakst: Yeah it's a bad thing. Let's say we use exponential backoff, and we're waiting 10 seconds for the next call. The handler won't be called until after waiting the 10 seconds at which point we're already making the next call (rather than handling it immediately on failure and then waiting).

I think for now we should just drop this to reduce scope creep and instead just show a basic retry wrap (i.e. no integration) just to show how easy it is to use tenacity with our calls, and then we can show `collect_errors(ValidationError)` for the most common use-case (and currently the only actual known use-case) for collecting errors, which is to reinsert them and the original reason we implemented this.

docs/integrations/custom_middleware.md
brenkao: I always pass in fn

docs/integrations/custom_middleware.md
brenkao: its in `custom_context_manager`

brenkao: These functions are already wrapped in the context manager, so user doesnt do this for every handler.

docs/integrations/custom_middleware.md
willbakst: ok then we should at least better describe what `fn` is and where they might use it. If the answer is that we're only using it for the function name, then I wonder if we should instead just specifically provide the function name? What else would we need the function for here when we don't even really know what it is / how to call it and don't expect to (shouldn't?) call it.

docs/integrations/custom_middleware.md
willbakst: I'm honestly very confused here. Line 44 definition compared to this definition on line 179 just isn't making any sense to me. Perhaps it's just the way line 44 is defined / name? If line 179 is correct then 44 definitely needs fixing (as well as all of the other similar definitions across the various handlers). It's not clear that the context manager is run on their behalf.

docs/integrations/custom_middleware.md
brenkao: i dont think we should make any assumptions on what people might want to do. are you speaking from the perspective of a user? they know its a callable and I have already described what fn was above.

docs/integrations/custom_middleware.md
brenkao: right its probably the name that is confusing you, will update it to make it clear

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,234 @@
...
+    print(f"Saving call: {fn.__name__}")
+    with Session(engine) as session:
+        yield session
+```
+
+All the handlers will be wrapped by this context manager. The `fn` argument will be your Mirascope call function.
+
+### `handle_call_response` and `handle_call_response_async`
+
+These functions will be called after making a mirascope call with the following signature:
```
willbakst, line 77: wording here is a little weird (sounds like the mirascope call needs this signature).

Maybe something like the following is more clear?

"These functions must have the following signature (where async should be async) and will be called after making a standard Mirascope call. Here is a sample implementation of the sync version:"

docs/integrations/custom_middleware.md
```diff
@@ -0,0 +1,234 @@
...
+
+The first argument will be a Mirascope `CallResponse` of the provider you are using.
+The second argument is the same `fn` argument as the one used in `custom_context_manager`.
+The third argument is the yielded object of your `custom_context_manager`, in this case a SQLModel Session object. If no `custom_context_manager` is used, this value is `None`.
+
+`handle_call_response_async` is the same as `handle_call_response` but using an `async` function.
+
+### `handle_stream` and `handle_stream_async`
+
+These functions will be called after streaming a Mirascope call with the following signature:
```
willbakst, line 109: ditto here and for the others, match the updated language to be more clear

willbakst (approved):

PR #382, opened 2024-08-09
Documentation

willbakst:


docs/learn/old/prompts.md
```diff
@@ -0,0 +1,219 @@
+# Prompts
+
+When working with LLMs, a surprising amount of work can go into determining the final text which gets passed into the model. Let's take a look at some of the tools Mirascope provides to make formatting your prompts as seamless as possible.
+
+## BasePrompt
+
+`BasePrompt` is a base class which works across all model providers. To create a prompt using `BasePrompt`:
+
+- Define a class which inherits `BasePrompt`
+- Set the prompt via the docstring and denote any template variable with curly braces
```
brianpark0126, line 10: If all cookbooks and documentation uses `prompt_template` decorator, would it make more sense to mention using the decorator first?

docs/learn/old/prompts.md
```diff
@@ -0,0 +1,219 @@
...
+@prompt_template("Recommend a {genre} book.")
+class BookRecommendationPrompt(BasePrompt):
+    """Now this is a normal docstring."""
+    
+    genre: str
+```
+
+To call the model with a `BasePrompt`, create an instance of your class and invoke the `run()` method with the appropriate provider’s configuration decorator. The `run()` function will create a decorated function covered in more detail in the next section[link to calls].
+
+For the configuration decorator, you are required to set the `model` argument. You can also set  `stream` for streams[link] `response_model` for extractions[link], and `tools` for function calling/tools[link]. All other functionality is supported in the provider’s native API, and the original parameters can be passed directly as keyword arguments into the decorator.
```
brianpark0126, line 36: is it worth mentioning call params, json mode if we're going through all the functionalities? Or at least mention there's more

docs/learn/old/prompts.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Message Types
+
+By default, Mirascope treats the prompt template as a single user message. If you want to specify a list of messages, simply use the message keywords (`SYSTEM`, `USER`, `ASSISTANT`). Make sure to capitalize the keyword and follow it with a colon.
+
+```python
+class BookRecommendationPrompt(BasePrompt):
+  """
+  SYSTEM: You are a librarian.
+	USER: Recommend a {genre} book.
```
brianpark0126, line 64: indent

docs/learn/old/prompts.md
```diff
@@ -0,0 +1,219 @@
...
+When writing messages that span multiple lines, the message must start on a new line for the tabs to get properly dedented:
+
+```python
+class BookRecommendationPrompt(BasePrompt):
+    """
+    SYSTEM:
+    You are the world's greatest librarian.
+    When recommending books, also describe the book's writing style and content.
+  
+	  USER: Recommend a {genre} book.
```
brianpark0126, line 79: indent

docs/learn/old/prompts.md
```diff
@@ -0,0 +1,219 @@
...
+#   USER: Recommend a fantasy book.
+```
+
+## `dump()`
+
+The `dump()` method shows you some important information about your prompt: its tags, the formatted prompt, the prompt template, and inputs.
+
+```python
+class BookRecommendationPrompt(BasePrompt):
+	"""
```
brianpark0126, line 142: indent

docs/learn/old/prompts.md
```diff
@@ -0,0 +1,219 @@
...
+# > Sure! If you liked the Lord of the Rings, you might like ...
+```
+
+## `__str__` overload
+
+`BasePrompt`'s `__str__()` is overloaded so that it returns the formatted prompt with template variables.
+
+```python
+class BookRecommendationPrompt(BasePrompt):
+	"""
```
brianpark0126, line 124: indent

docs/learn/prompts.md
```diff
@@ -1,219 +1,425 @@
...
+#   Examples:
+#   Title: The Name of the Wind
+#   Author: Patrick Rothfuss
+#
+#   Title: Mistborn: The Final Empire
+#   Author: Brandon Sanderson
 ```
 
-`BasePrompt` supports every way to call a model that you would via a function decorator. For a full list of the functionalities, check out the `BasePrompt` section of the API reference[link].
+If there are any other such specifiers you would find useful, let us know!
```
brenkao, line 114: perhaps link to github?

docs/learn/prompts.md
```diff
@@ -1,219 +1,425 @@
...
+prompt = BookRecommendationPrompt(genre="fantasy", age_group="child")
 print(prompt)
-# > Recommend a science textbook.
+# > Recommend a fantasy picture book for a child reader.
 ```
 
-Using a `computed_field` will include the computed field in the dump, which is useful for seeing what affects your prompts.
+!!! tip "Retrieving External Data"
+
+    Computed fields can be particularly useful when you want to inject information you retrieve from an external source. For example, perhaps you want to retrieve context from a document store to augment the generation with relevant information.
```
brenkao, line 358: just say RAG here?

docs/learn/prompts.md
```diff
@@ -1,219 +1,425 @@
...
+
+prompt = BookRecommendationPrompt(genre="fantasy", age_group="adult")
+print(prompt.dump()['metadata'])
+# > {'tags': {'version:0001', 'category:books'}}
 ```
 
-You should note that this is just one example, and you can use all Pydantic functionalities to improve your prompts. Check out their docs here[link].
+!!! note "Adding Additional Fields"
+
+    Although `Metadata` is a `TypedDict` with only the `tags` key, there is nothing stopping you from adding additional keys. The only issue is that this will throw a type error, which you can ignore. If there are particular keys you find yourself using frequently, let us know so we can add it!
```
brenkao, line 384: maybe we should have a `custom` or `user_defined` key that can catch-all user defined metadata so there isnt a type error?

docs/learn/prompts.md
```diff
@@ -1,219 +1,425 @@
...
+
+!!! note "Agnostic Assuming Support"
+
+    While `BasePrompt` is provider-agnostic, some features (like multi-modal inputs) may not be supported by all providers. We try to maximize support across providers, but you should always check the provider's capabilities when using more advanced features.
+
+## Best Practices and Use Cases
+
+- Provider Comparison: Use `BasePrompt` to easily test the same prompt across different providers to compare performance and output quality.
+- Prompt Versioning: Utilize the metadata decorator to keep track of different versions of your prompts as you refine them.
+- Dynamic Content: Leverage `@computed_field` for injecting dynamic content or API calls into your prompts.
```
brenkao, line 422: should we mention cached_property as well?

wcbudz (approved):
It's clear and well written!

docs/learn/old/prompts.md
willbakst: This is the old doc

docs/learn/prompts.md
```diff
@@ -1,219 +1,425 @@
...
 
-- Define a class which inherits `BasePrompt`
-- Set the prompt via the docstring and denote any template variable with curly braces
-- Type your template variables (and type them correctly! `BasePrompt` is built on Pydantic's `BaseModel` so incorrect typing will result in a `ValidationError`).
+For the purposes of explaining how prompt templates work, we will use:
+
+- `@prompt_template`: a decorator for attaching a prompt template to a function or class
+- `BasePrompt`: a base class for writing provider-agnostic prompts.
+
+We will cover `BasePrompt` in more detail later. For now, all you need to know is that we inject fields with names that match template variables, and we've implemented the `__str__` method for easily viewing how template variables get injected:
```
brianpark0126, line 14: sentence sounds a bit awkward because it's not parallel

willbakst: What do you mean by not parallel?

docs/learn/prompts.md
willbakst: I was thinking about that, but there will already be a github link in the header of the docs and I didn't want to be linking and mentioning our issues in every single doc. I would almost rather just remove the ask than link everywhere. Thoughts?

docs/learn/prompts.md
willbakst: I added it in parens

docs/learn/prompts.md
willbakst: I thought about something like that, but at that point we should just make metadata `dict[str, Any]`

I opted against that because I like have the autocomplete on the `tags` key and figured a type ignore for additional keys until we add them is totally fine (and I couldn't really think of other non-tag metadata we should add at this moment so will leave that to users).

I can also make this a little more clear around adding something like `# pyright: ignore [reportArgumentType]` instead of just a catch all `# type: ignore`?

docs/learn/prompts.md
willbakst: Done.

docs/learn/prompts.md
willbakst: Oh I see it's the "that"

docs/learn/prompts.md
brenkao: yeah that seems like a good update

willbakst: Done.

PR #383, opened 2024-08-10
added pdf extraction cookbook

brenkao:


docs/cookbook/extract_from_pdf.md
```diff
@@ -0,0 +1,133 @@
...
+Afterwards, we update our previously defined call function to take in image bytes and run:
+
+```python
+from mirascope.core import openai
+
+@openai.call(model="gpt-4o-mini", response_model=ResumeInfo)
+async def extract_resume_info(pdf_imgs: list[bytes]):
+    """
+    Extract the resume information from the pdf file.
+    {images:pdf_imgs}
```
willbakst, line 103: this is backwards with new update around format specs

docs/cookbook/extract_from_pdf.md
```diff
@@ -0,0 +1,133 @@
+# Extracting from PDF: Guide to Extract Pages and Content from PDF Files
+
+This recipe demonstrates how to leverage Large Language Models (LLMs) -- specifically the OpenAI API -- to extract pages and content from PDF files. We'll cover single PDF document as well as multiple PDF files and also use OCR to extract text from scanned documents.
+
+??? info "Key Concepts"
+
+    - [Pydantic](ADD LINK)
```
willbakst, line 7: I think we can drop pydantic since we have response models, which will cover pydantic.

We should also include links in order of the learn section and include prompts here as well

willbakst (dismissed):

docs/cookbook/extract_from_pdf.md
brenkao: done

willbakst (approved):

PR #390, opened 2024-08-13
added knowledge graph cookbook

brenkao:


docs/cookbook/knowledge_graph.md
```diff
@@ -0,0 +1,211 @@
...
+        Node(id="Jane Doe", type="Person", properties={{"age": 25}})
+        Edge(source="John Doe", target="Jane Doe", relationship="Siblings")
+
+        USER:
+        {question}
+        """
+    )
+    def _generate_knowledge_graph(self, question: str) -> openai.OpenAIDynamicConfig:
+        text = ""
+        with open("wikipedia.txt") as f:
```
willbakst, line 90: as you mention, this is somewhat silly. Instead, have the argument be the url / local file path and then open it in the call (that way it's actually dynamic)

mirascope/core/base/_utils/_get_prompt_template.py
```diff
@@ -18,5 +18,7 @@ def get_prompt_template(fn: Callable | BaseModel) -> str:
     if not doc:
         raise ValueError("No prompt template set!")
     if docstring_prompt_enabled != "ENABLED":
-        raise ValueError("You must explicitly enable docstring prompt templates.")
+        raise ValueError(
+            "You must explicitly enable docstring prompt templates by setting MIRASCOPE_DOCSTRING_PROMPT_TEMPLATE=ENABLED."
```
willbakst, line 22: nit: specify "in your environment"?

docs/cookbook/knowledge_graph.md
willbakst, line 1: This doesn't really make much sense to me.

The way it's implemented there's a single fixed text that the tool for generating the knowledge graph can access, so in reality there's only a single knowledge graph created from the fixed text. In this case, it seems better to just make a call to generate the knowledge graph and then instantiate the agent with that knowledge graph (rather than giving it a tool).

We can then mention that you can extend this cookbook by letting it dynamically update it's knowledge graph as necessary based on the questions it receives but that this would require giving it access to the internet in order to do so.

Does this make sense?

docs/cookbook/knowledge_graph.md
brenkao: yeah that makes sense

docs/cookbook/knowledge_graph.md
brenkao: The tool does get called multiple times if on the first pass it doesnt have enough "knowledge" to answer the question. With that information does it still make sense for me to generate thee knowledge graph separately?

brenkao: made it simpler, and mentioned updating the example to agent in the final section

docs/cookbook/generating_captions.md
```diff
@@ -0,0 +1,49 @@
...
+
+!!! note "Background"
+
+    Caption generation evolved from manual human effort to machine learning techniques like Conditional Random Fields (CRFs) and Support Vector Machines (SVMs), which were time-consuming and resource-intensive. Large Language Models (LLMs) have revolutionized this field, enabling efficient multi-modal tasks through API calls and prompt engineering, dramatically improving caption generation speed and accuracy.
+
+!!! Warning
+
+    This recipe will only work for those which support images (OpenAI, Gemini, Anthropic) as of 8/13/2024. Be sure to check if your provider has multimodal support.
+
+With OpenAI’s multimodal capabilities, image inputs are treated just like text inputs, which means we can use it as context to ask questions or make requests. For the sake of reproducibility, we will get our image from a URL to save you the hassle of having to find and download an image. The image is [a public image from BBC Science of a wolf](https://www.sciencefocus.com/space/wolf-moon) in front of the moon (the link goes to the article, but use the URL of the actual image by right clicking the image and getting the image address).
```
willbakst, line 18: We should just include the direct URL instead of to the article, right? Why be indirect here?

docs/cookbook/generating_synthetic_data.md
```diff
@@ -0,0 +1,240 @@
+# Generate Synthetic Data
+
+In this cookbook recipe, we go over how to generate synthetic data for LLMs, in this case, OpenAI’s `gpt-4o-mini`. When using LLMs to synthetically generate data, it is most useful to generate non-numerical data which isn’t strictly dependent on a defined probability distribution - in those cases, it will be far easier to define a distribution and generate these points directly using functions.
```
willbakst, line 3: instead of "using functions" say "from the distribution"

docs/cookbook/generating_synthetic_data.md
```diff
@@ -0,0 +1,240 @@
...
+
+However, for:
+
+- data that needs general intelligence to be realistic
+- data that lists many items within a broad category
+- data which is language related
+
+and more, LLMs are far easier to use and yield better (or the only possible) results.
```
willbakst, line 11: instead of "possible" say "feasible"

docs/cookbook/generating_captions.md
```diff
@@ -0,0 +1,49 @@
+# Generate Captions for an Image
+
+In this recipe, we go over how to use LLMs to generate a descriptive caption set of tags for an image with OpenAI’s `gpt-4o-mini`.
+
+??? info "Key Concepts"
+
+    - [Calls](ADD LINK)
```
willbakst, line 7: Prompts is missing here, important for the vision stuff

docs/cookbook/generating_synthetic_data.md
```diff
@@ -0,0 +1,240 @@
...
+
+??? tip "Mirascope Concepts Used"
+
+    - [Prompts](../learn/prompts.md)
+    - [Calls](../learn/calls.md)
+    - [Response Model](ADD LINK)
+
+!!! note "Background"
```
willbakst, line 19: nit: shouldn't there be an extra space here between the note header and the content?

docs/cookbook/generating_synthetic_data.md
```diff
@@ -0,0 +1,240 @@
...
+#   name='Dishwasher' price=599.99 inventory=8
+```
+
+## Generate Data into a pandas `DataFrame`
+
+Since pandas is a common library for working with data, it’s also worth knowing how to directly create and append to a dataframe with LLMs.
+
+### Create a New `DataFrame`
+
+To create a new `DataFrame`, we define a `BaseModel` schema with a simple function to generate  `DataFrame` via a list of list of data and the column names. One might be tempted to use a tool call, which at first might appear to be simpler, but would be bad practice since a tool should pass back a string so it may be passed back into the chat. Note that there are many ways to initialize `DataFrame`s and this is just one of the many ways it can be implemented:
```
willbakst, line 97: I'm not tempted to use a tool call here lol, I would just remove that whole section since tools seem obviously separate from data generation at least to me.

docs/cookbook/generating_synthetic_data.md
```diff
@@ -0,0 +1,240 @@
...
+To append to a `DataFrame`, we can modify the prompt so that instead of describing the data we want to generate, we ask the LLM to match the type of data it already sees. Furthermore, we add a `append_dataframe()` function to append to an existing `DataFrame`. Finally, note that we use the generated `df` from above as the `DataFrame` to append to in the following example:
+
+```python
+class DataFrameGenerator(BaseModel):
+    ...
+
+    def append_dataframe(self, df: pd.DataFrame) -> pd.DataFrame:
+        df = df._append(
+            pd.DataFrame(self.data, columns=self.column_names), ignore_index=True
+        )  # type: ignore
```
willbakst, line 156: what's this type ignore?

docs/cookbook/knowledge_graph.md
```diff
@@ -0,0 +1,188 @@
+# Knowledge Graph
+
+Often times, data is messy and not always stored in a structured manner ready for use by an LLM. In this recipe, we show how to create a knowledge graph from an unstructured document using common python libraries and Mirascope using OpenAI GPT-4o-mini.
+
+??? info "Key Concepts"
+
+    - [Calls](ADD LINK)
+    - [Response Model](ADD LINK)
```
willbakst, line 8: missing Prompts and Dynamic Configuration

docs/cookbook/text_summarization.md
```diff
@@ -0,0 +1,318 @@
+# Text Summarization
+
+In this recipe, we show some techniques to improve an LLM’s ability to summarize a long text from simple (e.g. `"Summarize this text: {text}..."`) to more complex prompting and chaining. We will use OpenAI’s GPT-4o-mini model (128k input token limit), but you can use any model you’d like to implement these summarization techniques, as long as they have a large context window.
```
willbakst, line 3: add "ranging from simple...and chaining techniques"

docs/cookbook/text_summarization.md
willbakst, line 1: why is this in the prompt engineering section? Note that section is going to be specifically talking about the techniques from the learn prompting stuff

docs/cookbook/text_summarization.md
```diff
@@ -0,0 +1,318 @@
...
+
+??? info "Key Concepts"
+
+    - [Calls](ADD LINK)
+    - [Chaining](ADD LINK)
+    - [Response Model](ADD LINK)
+
+!!! note "Background"
+    Large Language Models (LLMs) have revolutionized text summarization by enabling more coherent and contextually aware abstractive summaries. Unlike earlier models that primarily extracted or rearranged existing sentences, LLMs can generate novel text that captures the essence of longer documents while maintaining readability and factual accuracy.
```
willbakst, line 12: nit: missing extra newline here?

docs/cookbook/text_summarization.md
```diff
@@ -0,0 +1,318 @@
...
+#   inspired by the comedy group Monty Python.
+```
+
+LLMs excel at summarizing shorter texts, but they often struggle with longer documents, failing to capture the overall structure while sometimes including minor, irrelevant details that detract from the summary's coherence and relevance.
+
+One simple update we can make is to improve our prompt by providing an initial outline of the text then adhere to this outline to create its summary.
+
+## Simple Call with Outline
+
+This prompt engineering technique is an example of Chain of Thought (CoT), forcing the model to write out its thinking process. It also involves little work and can be done by modifying the text of the single call. With an outline, the summary is less likely to lose the general structure of the text.
```
willbakst, line 74: put a link placeholder for CoT to link to the CoT cookbook

examples/cookbook/generating_synthetic_data.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Make sure to abide by the following constraints:
+    QLEDS should be roughly (not exactly) 2x the price of an OLED of the same size
+    for both OLEDs and QLEDS, price should increase roughly proportionately to size
+    """
+)
+def generate_tv_data(num_datapoints: int): ...
+
+
+for _ in generate_tv_data(10):
```
willbakst, line 129: use an actual name if you're going to use the variable

docs/cookbook/generating_synthetic_data.md
```diff
@@ -0,0 +1,240 @@
...
+    Output the datapoints as a list of instances of TV.
+
+    Make sure to abide by the following constraints:
+    QLEDS should be roughly (not exactly) 2x the price of an OLED of the same size
+    for both OLEDs and QLEDS, price should increase roughly proportionately to size
+    """
+)
+def generate_tv_data(num_datapoints: int): ...
+
+for _ in generate_tv_data(10):
```
willbakst, line 213: use a named variable if you're going to use it (e.g. print it)

docs/cookbook/generating_synthetic_data.md
brenkao: @brianpark0126

brenkao: updated to use concat

willbakst (approved):

Issue #432, opened 2024-08-19
Support for OpenAI Structured Outputs


willbakst:
### Description

There is a beta implementation of their new functionality, but it has no documentation and is an initial rough draft.

It would be great to provide improved support for structured outputs, especially if we could provide this option without requiring a separate decorator (i.e. enable it directly through the `openai.call` decorator instead of having to use the `openai.parse` beta decorator).

willbakst:
This is out in `v1.1`!

docs: https://mirascope.io/docs/latest/learn/provider_specific_features/#structured-outputs

PR #457, opened 2024-08-27
feat: openai structured outputs

willbakst:
Implements #432


mirascope/core/base/tool.py
```diff
@@ -23,9 +25,37 @@ class _CacheControl(TypedDict):
 
 
 class ToolConfig(TypedDict, total=False):
+    strict: bool
```
koxudaxi, line 28: @willbakst 
If this is a provider specific feature for OpenAI only, shouldn't `'strict` only be added to `OpenAIToolConfig` and not `ToolConfig`?

koxudaxi: However, this is fine if the design is to use a Tool for each provider.
I am concerned about the lack of clarity that this only applies to OpenAI.

willbakst: I can do that. I'll need to do the same for Anthropic since prompt caching is also provider-specific for Anthropic

koxudaxi: I see. looks good.

willbakst: yeah, I made `OpenAIToolConfig` and `AnthropicToolConfig` :)

tests/core/anthropic/test_tool.py
```diff
@@ -18,7 +18,7 @@ class FormatBook(AnthropicTool):
         title: str
         author: str
 
-        tool_config = ToolConfig(cache_control={"type": "ephemeral"})
+        tool_config = AnthropicToolConfig(cache_control={"type": "ephemeral"})
```
koxudaxi, line 21: It may be difficult to do this with a type checker, but it is better to check at runtime level whether the Tool is compatible with the tool_config passed.

koxudaxi (approved):
> It may be difficult to do this with a type checker, but it is better to check at runtime level whether the Tool is compatible with the tool_config passed.

This should be there to improve usability, but I don't think it is essential, so I approve it.

koxudaxi (dismissed):

tests/core/anthropic/test_tool.py
willbakst: This is difficult because the user will define a `BaseTool`, which will need to accept any `ToolConfig` subclass.

We could add runtime warnings or errors to all other providers, but I'm not sure how worthwhile that is since it won't break anything (it will just silently not use the feature unless the provider supports it).

Given the explicit naming of the ToolConfig to e.g. `AnthropicToolConfig`, I think it's clear enough for now, and we can revisit later if necessary.

PR #459, opened 2024-08-28
Fix: remove `type: ignore`

koxudaxi:


mirascope/core/litellm/_call.py
```diff
@@ -24,10 +24,10 @@
     TStream=OpenAIStream,
     TCallParams=OpenAICallParams,
     default_call_params=OpenAICallParams(),
-    setup_call=setup_call,  # type: ignore
-    get_json_output=get_json_output,  # type: ignore
-    handle_stream=handle_stream,
-    handle_stream_async=handle_stream_async,
+    setup_call=setup_call,  # pyright: ignore [reportArgumentType]
+    get_json_output=get_json_output,
+    handle_stream=handle_stream,  # pyright: ignore [reportArgumentType]
```
willbakst, line 29: these didn't need ignores before, why do they need them now?

mirascope/core/litellm/_utils/_setup_call.py
```diff
@@ -25,22 +69,23 @@ def setup_call(
     call_params: OpenAICallParams,
     extract: bool,
 ) -> tuple[
-    Callable[..., ChatCompletion] | Callable[..., Awaitable[ChatCompletion]],
-    str,
+    Callable[..., ModelResponse | CustomStreamWrapper]
```
willbakst, line 72: hmm, even though this is technically correct, we should type this as returning the OpenAI types because for all intents and purposes that's how we're treating them.

To remove the type ignore on the return type below, we should just cast the function to match the type?

mirascope/core/mistral/_utils/_setup_call.py
```diff
@@ -74,18 +120,9 @@ def setup_call(
...
-        ChatCompletionResponse
-        | Coroutine[Any, Any, ChatCompletionResponse]
-        | AsyncGenerator[ChatCompletionStreamResponse, None]
-        | Iterable[ChatCompletionStreamResponse]
-    ):
-        if stream:
-            return client.chat_stream(**kwargs)
-        return client.chat(**kwargs)
-
+    if isinstance(client, MistralAsyncClient):
```
willbakst, line 123: this whole thing with dataclasses and extra methods feels like overkill. I feel like we could just safely cast types here and call it a day? same for cohere

mirascope/core/openai/_utils/_setup_call.py
```diff
@@ -81,4 +127,4 @@ def setup_call(
         client = AsyncOpenAI() if inspect.iscoroutinefunction(fn) else OpenAI()
     create = client.chat.completions.create
 
-    return create, prompt_template, messages, tool_types, call_kwargs
+    return create, prompt_template, messages, tool_types, call_kwargs  # pyright: ignore [reportReturnType]
```
willbakst, line 130: same thing here, why can't we remove the type ignore?

mirascope/core/openai/call_response.py
```diff
@@ -103,7 +103,7 @@ def cost(self) -> float | None:
     @property
     def message_param(self) -> SerializeAsAny[ChatCompletionAssistantMessageParam]:
         """Returns the assistants's response as a message parameter."""
-        return self.response.choices[0].message.model_dump(exclude={"function_call"})  # type: ignore
+        return self.response.choices[0].message.model_dump(exclude={"function_call"})  # pyright: ignore [reportReturnType]
```
willbakst, line 106: I wonder if we could do something like the following instead of ignoring:

```python
message_param = self.response.choices[0].message.model_dump(exclude={"function_call"})
return ChatCompletionAssistantMessageParam(**message_param)
```

mirascope/core/openai/call_response.py
```diff
@@ -163,7 +163,7 @@ def tool_message_params(
                 role="tool",
                 content=output,
                 tool_call_id=tool.tool_call.id,
-                name=tool._name(),  # type: ignore
+                name=tool._name(),  # pyright: ignore [reportCallIssue]
```
willbakst, line 166: same thing here, not sure why we were ignoring this, wondering why it needs to be ignored at all

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
     "pydantic>=2.7.4,<3.0",
     "docstring-parser>=0.15,<1.0",
     "jiter>=0.5.0",
+    "typing-extensions>=4.10.0",
```
willbakst, line 46: wait is `typing-extensions` not a built-in package?

also, after looking into it briefly I'm wondering if we even need the extension since we only support 3.10+? are there issues with just importing things like `TypedDict` directly from typing? I don't think there are 3.11+ features we're using?

tests/core/base/_utils/test_convert_function_to_base_tool.py
```diff
@@ -79,7 +79,7 @@ def format_book(self):
 
     tool_type = convert_function_to_base_tool(format_book, BaseTool)
     tool_type.title = "The Name of the Wind"  # pyright: ignore [reportAttributeAccessIssue]
-    tool = tool_type()
+    tool = tool_type()  # pyright: ignore [reportCallIssue]
```
willbakst, line 82: why did we not need this before?

tests/core/base/_utils/test_convert_function_to_base_tool.py
```diff
@@ -90,7 +90,7 @@ def format_book(cls=None) -> str:
         return "The Name of the Wind"
 
     tool_type = convert_function_to_base_tool(format_book, BaseTool)
-    tool = tool_type()
+    tool = tool_type()  # pyright: ignore [reportCallIssue]
```
willbakst, line 93: same here

tests/core/base/test_tool.py
```diff
@@ -22,7 +22,7 @@ class FormatBook(BaseTool):
         def call(self) -> str:
             return f"{self.title} by {self.author}"
 
-    tool = FormatBook(title="The Name of the Wind", author="Patrick Rothfuss")
+    tool = FormatBook(title="The Name of the Wind", author="Patrick Rothfuss")  # pyright: ignore [reportCallIssue]
```
willbakst, line 25: aha, I think because you added `tool_call` to `BaseTool` it is now a required argument to create base tools when it wasn't before?

tests/core/gemini/_utils/test_setup_call.py
```diff
@@ -146,10 +145,8 @@ def test_setup_call_extract(
         call_params={},
         extract=True,
     )
-    assert "tool_config" in call_kwargs and isinstance(
-        call_kwargs["tool_config"], ToolConfigDict
-    )
-    assert call_kwargs["tool_config"].function_calling_config == {
+    assert "tool_config" in call_kwargs and isinstance(call_kwargs["tool_config"], dict)
```
willbakst, line 148: I'm confused, this previous test was correct? it should always be `ToolConfigDict` during gemini extraction

tests/core/mistral/_utils/test_setup_call.py
willbakst, line 1: test additions here seem overkill re: my comment about the whole thing here feeling overkill, there has to be a simpler way of doing this

tests/core/cohere/_utils/test_setup_call.py
willbakst, line 1: test additions here seem overkill re: my comment about the whole thing here feeling overkill, there has to be a simpler way of doing this

mirascope/core/base/tool.py
```diff
@@ -85,6 +85,9 @@ def call(self) -> str:
     tool_config: ClassVar[ToolConfig] = ToolConfig()
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+    if TYPE_CHECKING:
+        tool_call: Any
```
willbakst, line 89: what is the benefit of including this? if the answer is so that a user doesn't accidentally add a `tool_call` field to their `BaseTool` definition, then we should remove `if TYPE_CHECKING` and just always have `tool_call` as a field of `BaseTool`, right?

Separately, it seems adding this results in a bunch of pyright ignores (e.g. in our tests). I'm fine with that if we think this is worthwhile, but would like to see if there's a better solution (e.g. making `tool_call` optional for `BaseTool` but required for provider-specific tools e.g. `OpenAITool`?)

Last, if we are going to include this, we should ensure it is skipped for json schema generation like we do for tool calls for all provider-specific tool types

mirascope/core/base/_call_factory.py
```diff
@@ -318,6 +319,6 @@ def base_call(
             json_mode=json_mode,
             client=client,
             call_params=call_params,
-        )  # type: ignore
+        )  # pyright: ignore [reportReturnType, reportCallIssue]
```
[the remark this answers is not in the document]
koxudaxi: OK, I will think it

mirascope/core/base/_create.py
```diff
@@ -128,39 +130,43 @@ async def inner_async(
...
-                    messages=messages,
-                    call_params=call_params,
-                    call_kwargs=call_kwargs,
-                    user_message_param=get_possible_user_message_param(messages),
-                    start_time=start_time,
-                    end_time=end_time,
-                )
-                output._model = model
-                return output if not output_parser else output_parser(output)
+                if fn_is_sync(fn):
```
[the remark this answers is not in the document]
koxudaxi: I agree. I use `assert fn_is_sync(fn)`

mirascope/core/base/_extract.py
```diff
@@ -101,38 +103,40 @@ def decorator(
...
-                json_output = get_json_output(call_response, json_mode)
-                try:
-                    output = extract_tool_return(response_model, json_output, False)
-                except ValidationError as e:
-                    e._response = call_response  # type: ignore
-                    raise e
-                if isinstance(output, BaseModel):
-                    output._response = call_response  # type: ignore
-                return output if not output_parser else output_parser(output)  # type: ignore
+                if fn_is_sync(fn):
```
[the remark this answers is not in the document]
koxudaxi: Fixed

mirascope/core/base/stream.py
```diff
@@ -268,98 +264,109 @@ def decorator(
...
-                )
-
-                async def generator() -> (
-                    AsyncGenerator[
-                        tuple[_BaseCallResponseChunkT, _BaseToolT | None], None
-                    ]
-                ):
-                    async for chunk, tool in handle_stream_async(
-                        await create(stream=True, **call_kwargs), tool_types
+                if fn_is_async(fn):
```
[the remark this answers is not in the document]
koxudaxi: Fixed

mirascope/core/base/structured_stream.py
```diff
@@ -238,13 +240,15 @@ async def inner_async(
...
             @wraps(fn)
             def inner(*args: _P.args, **kwargs: _P.kwargs) -> Iterable[_ResponseModelT]:
-                assert SetupCall.fn_is_sync(fn)
-                return BaseStructuredStream[_ResponseModelT](
-                    stream=stream_decorator(fn=fn, **stream_decorator_kwargs)(
-                        *args, **kwargs
-                    ),
-                    response_model=response_model,
-                )
+                if fn_is_sync(fn):
```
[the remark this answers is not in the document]
koxudaxi: Fixed

mirascope/core/mistral/_utils/_setup_call.py
koxudaxi: Originally, I was going to use namedtuple, but there was a problem that namedtuple and generics could not be used together in 3.9.
I will remove dataclasses.

tests/core/gemini/_utils/test_setup_call.py
koxudaxi: Oh, I see.  `ToolConfigDict ` is not dict.

https://github.com/google-gemini/generative-ai-python/blob/e0928fc0f239242f0cf5d69b6821188a8dac4e66/google/generativeai/types/content_types.py#L881-L882

koxudaxi: I reverted

mirascope/core/base/stream.py
```diff
@@ -268,98 +264,109 @@ def decorator(
...
-                )
-
-                def generator() -> (
-                    Generator[
-                        tuple[_BaseCallResponseChunkT, _BaseToolT | None], None, None
-                    ]
-                ):
-                    yield from handle_stream(
-                        create(stream=True, **call_kwargs), tool_types
+                if fn_is_sync(fn):
```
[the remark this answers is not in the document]
koxudaxi: Fixed

mirascope/core/base/_utils/_setup_call.py
```diff
@@ -17,6 +34,129 @@
...
 
+_AsyncType = TypeVar("_AsyncType")
+_StreamedResponse = TypeVar("_StreamedResponse")
+_NonStreamedResponse = TypeVar("_NonStreamedResponse")
+
+_P = ParamSpec("_P")
+
+
+@dataclass
+class _AsyncFunctions(Generic[_NonStreamedResponse, _StreamedResponse]):
```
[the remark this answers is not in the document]
koxudaxi: I moved it

mirascope/core/base/tool.py
koxudaxi: I add the filed in BaseTool

tests/core/base/test_tool.py
koxudaxi: I removed it

mirascope/core/base/tool.py
koxudaxi: I tried adding a Field in `BaseTool`, but it does not seem possible to set a default value for a Field in `BaseTool` and not set the default in the inheriting class. (pyright error)
It seems to be referencing tool_call without checking it when calling it in the first place, so if all inherited classes don't have `tool_call`, it should be checked

mirascope/core/anthropic/tool.py
```diff
@@ -90,5 +90,5 @@ def from_tool_call(cls, tool_call: ToolUseBlock) -> AnthropicTool:
                 instance.
         """
         model_json = copy.deepcopy(tool_call.input)
-        model_json["tool_call"] = tool_call.model_dump()  # type: ignore
+        model_json["tool_call"] = tool_call.model_dump()  # pyright: ignore [reportIndexIssue]
```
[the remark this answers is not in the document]
koxudaxi: More errors will occur with the change.
```python
 tool.py:92:22 - error: No overloads for "__init__" match the provided arguments (reportCallIssue)
  tool.py:92:27 - error: Argument of type "object" cannot be assigned to parameter "iterable" of type "Iterable[list[bytes]]" in function "__init__"
    "object" is incompatible with protocol "Iterable[list[bytes]]"
      "__iter__" is not present (reportArgumentType)
  tool.py:93:9 - error: Argument of type "Literal['tool_call']" cannot be assigned to parameter "key" of type "bytes" in function "__setitem__"
    "Literal['tool_call']" is not assignable to "bytes" (reportArgumentType)
  tool.py:93:9 - error: Argument of type "dict[str, Any]" cannot be assigned to parameter "value" of type "bytes" in function "__setitem__"
    "dict[str, Any]" is not assignable to "bytes" (reportArgumentType)

```

mirascope/core/litellm/_utils/_setup_call.py
koxudaxi: I just reverted it

mirascope/core/openai/call_response.py
koxudaxi: Thank you. I applied it.

mirascope/core/openai/call_response.py
koxudaxi: `ChatCompletionToolMessageParam` doesn't have the `name`.
Am I misunderstanding something?
https://github.com/openai/openai-python/blob/9850c169c4126fd04dc6796e4685f1b9e4924aa4/src/openai/types/chat/chat_completion_tool_message_param.py#L13-L21

pyproject.toml
koxudaxi: `typing-extensions` is a non-build-in, optional typing backport.
Since `pydantic` already depends on `typing-extensions`, we do not expect to add any new dependencies.
Changed from `TypeGuard` to `TypeIs` for stricter type refinement.
This is a type feature from 3.13 and requires `typing-extensions`.
https://peps.python.org/pep-0742/

tests/core/base/_utils/test_convert_function_to_base_tool.py
koxudaxi: This is my mistake. reverted.

tests/core/base/_utils/test_convert_function_to_base_tool.py
koxudaxi: same

tests/core/mistral/_utils/test_setup_call.py
koxudaxi: OK, I will move the async generator test to utils' test

mirascope/core/openai/_utils/_setup_call.py
koxudaxi: There are two issues.

The stream argument type is `bool` plus `NotGiven`, which results in a type error.
Also, the return value is expected to be a `Generator`, but it returns an Iterator-like type. For this reason, it is necessary to ignore the type.
Since `Generator` is a subclass of Iterator, strictly speaking, it is necessary to add or wrap the missing methods as a generator.

mirascope/core/gemini/_utils/_setup_call.py
```diff
@@ -75,4 +119,4 @@ def setup_call(
         else client.generate_content
     )
 
-    return create, prompt_template, messages, tool_types, call_kwargs
+    return create, prompt_template, messages, tool_types, call_kwargs  # pyright: ignore [reportReturnType]
```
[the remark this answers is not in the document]
koxudaxi: The cause of the problem is the same as for OpenAI.

mirascope/core/groq/_utils/_setup_call.py
```diff
@@ -66,4 +114,4 @@ def setup_call(
         client = AsyncGroq() if inspect.iscoroutinefunction(fn) else Groq()
     create = client.chat.completions.create
 
-    return create, prompt_template, messages, tool_types, call_kwargs
+    return create, prompt_template, messages, tool_types, call_kwargs  # pyright: ignore [reportReturnType]
```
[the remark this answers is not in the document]
koxudaxi: The cause of the problem is the same as for OpenAI.
I'll see if I can handle this case as well by the function.

mirascope/core/gemini/_utils/_setup_call.py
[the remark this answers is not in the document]
koxudaxi: I Fixed it

mirascope/core/groq/_utils/_setup_call.py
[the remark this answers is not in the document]
koxudaxi: Fixed.

mirascope/core/openai/_utils/_setup_call.py
koxudaxi: Fixed

mirascope/core/groq/call_response.py
```diff
@@ -154,7 +154,7 @@ def tool_message_params(
                 role="tool",
                 content=output,
                 tool_call_id=tool.tool_call.id,
-                name=tool._name(),  # type: ignore
+                name=tool._name(),  # pyright: ignore [reportCallIssue]
```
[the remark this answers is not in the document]
koxudaxi: `ChatCompletionToolMessageParam` doesn't have `name` field.
https://github.com/groq/groq-python/blob/89b9c14a66af27a9b669f27702ef8433677d11ca/src/groq/types/chat/chat_completion_tool_message_param.py#L10-L18

mirascope/core/base/_create.py
```diff
@@ -128,39 +130,43 @@ async def inner_async(
...
+                            extract=False,
+                        )
+                    )
+                    start_time = datetime.datetime.now().timestamp() * 1000
+                    response = create(stream=False, **call_kwargs)
+                    end_time = datetime.datetime.now().timestamp() * 1000
+                    output = TCallResponse(
+                        metadata=get_metadata(fn, dynamic_config),
+                        response=response,
+                        tool_types=tool_types,  # pyright: ignore [reportArgumentType]
```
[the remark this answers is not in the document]
koxudaxi: When conditional branching by isType was applied, async side also needed it, so I added `ignore` to async side as well.

mirascope/core/base/_extract.py
```diff
@@ -101,38 +103,40 @@ def decorator(
...
+                    call_response = create_decorator(fn=fn, **create_decorator_kwargs)(
+                        *args, **kwargs
+                    )
+                    json_output = get_json_output(call_response, json_mode)
+                    try:
+                        output = extract_tool_return(response_model, json_output, False)
+                    except ValidationError as e:
+                        e._response = call_response  # pyright: ignore [reportAttributeAccessIssue]
```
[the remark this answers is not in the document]
koxudaxi: same above 
> When conditional branching by isType was applied, async side also needed it, so I added ignore to async side as well.

mirascope/core/base/stream.py
```diff
@@ -268,98 +264,109 @@ def decorator(
...
+                        ]
+                    ):
+                        yield from handle_stream(
+                            create(stream=True, **call_kwargs), tool_types
+                        )
+
+                    return TStream(
+                        stream=generator(),
+                        metadata=get_metadata(fn, dynamic_config),
+                        tool_types=tool_types,  # pyright: ignore [reportArgumentType]
```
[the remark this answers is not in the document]
koxudaxi: same above 
> When conditional branching by isType was applied, async side also needed it, so I added ignore to async side as well.

tests/core/cohere/_utils/test_setup_call.py
koxudaxi: reverted

tests/core/mistral/_utils/test_setup_call.py
koxudaxi: Done

mirascope/core/openai/call_response.py
willbakst: ooooh no you are correct, I was mistaken on what was being ignored.

mirascope/core/groq/call_response.py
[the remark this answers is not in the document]
willbakst: no you are correct, I was mistaken on what was being ignored. same thing with openai

mirascope/core/litellm/_utils/_setup_call.py
```diff
@@ -87,5 +85,8 @@ def setup_call(
         call_params=call_params,
         extract=extract,
     )
-    create = acompletion if inspect.iscoroutinefunction(fn) else completion
+    create = cast(
+        Callable[..., ChatCompletion] | Callable[..., Awaitable[ChatCompletion]],
+        acompletion if inspect.iscoroutinefunction(fn) else completion,
```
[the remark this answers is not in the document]
koxudaxi: use is_async_fn

mirascope/core/base/structured_stream.py
```diff
@@ -240,15 +240,13 @@ async def inner_async(
 
             @wraps(fn)
             def inner(*args: _P.args, **kwargs: _P.kwargs) -> Iterable[_ResponseModelT]:
-                if fn_is_sync(fn):
```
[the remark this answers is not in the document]
koxudaxi: Done

mirascope/core/gemini/_call.py
```diff
@@ -22,10 +22,10 @@
     TToolType=GeminiTool,
     TCallParams=GeminiCallParams,
     default_call_params=GeminiCallParams(),
-    setup_call=setup_call,  # type: ignore
+    setup_call=setup_call,
     get_json_output=get_json_output,
-    handle_stream=handle_stream,
-    handle_stream_async=handle_stream_async,
+    handle_stream=handle_stream,  # pyright: ignore [reportArgumentType]
```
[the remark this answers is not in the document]
koxudaxi: Removing the type ignore in setup_call caused type checking by Generics to be performed. it is difficult to fix because the Response is different for different providers.

mirascope/core/base/_call_factory.py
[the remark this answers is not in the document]
koxudaxi: Here, too, the types of other functions are now valid and the decision is made by Generics. It is difficult to fix because the response varies from provider to provider.

mirascope/core/litellm/_call.py
koxudaxi: same above

mirascope/core/gemini/_call.py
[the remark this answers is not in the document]
willbakst: but shouldn't the generics here specifically match? i.e. generics from gemini setup_call match handle_stream etc.?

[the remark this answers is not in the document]
koxudaxi: The issue here is the varying "combinations" of responses from providers.
Some return the same response for both `stream=False` and `stream=True`, while others don't. Additionally, there are various patterns regarding whether it's an `AsyncGenerator` or an `Awaitable[Stream]`. With the current function definition, it's too challenging to express all these patterns in type annotations. If possible, it would be ideal to create a new function that wraps everything, but I don't think it's worth investing that much time and effort into it.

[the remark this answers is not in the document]
willbakst: ok, we can revisit this later if we think it's worth the time.

willbakst (approved):

willbakst (approved):

PR #462, opened 2024-08-29
docs: documentation agent and eval documentation agent 

brenkao:


examples/cookbook/agents/documentation_agent/agent.py
```diff
@@ -0,0 +1,134 @@
...
+                "text": node.node.get_text(),  # pyright: ignore[reportAttributeAccessIssue]
+                "semantic_score": node.score,
+            }
+            for id, node in enumerate(retrieved_nodes[idx : idx + choice_batch_size])
+        ]
+        print(nodes_batch)
+        results += llm_query_rerank(nodes_batch, query)
+    results = sorted(results, key=lambda x: x.score or 0, reverse=True)[:top_n]
+
+    return "\n".join(result.document for result in results)
```
willbakst, line 97: nit: why not just have this function return `list[str]` and then use the `:list` format specifier in the call below?

examples/cookbook/agents/documentation_agent/agent.py
```diff
@@ -0,0 +1,134 @@
...
+        {context}
+
+        USER:
+        {question}
+        """
+    )
+    def _call(self, context: str, question: str): ...
+
+    def _step(self, question: str):
+        context = get_documents(question)
```
willbakst, line 120: nit: why get the documents here rather than inside of `_call` as a computed field?

for example:

```python
@openai.call(...)
@prompt_template(...)
def _call(self, question: str) -> openai.OpenAIDynamicConfig:
    return {"computed_fields": {"context": get_documents(question)}}
```

examples/cookbook/agents/documentation_agent/agent.py
```diff
@@ -0,0 +1,170 @@
...
+            @openai.call("gpt-4o-mini")
+            @prompt_template("Recommend a {{genre}} book")
+            def recommend_book(genre: str):
+                ...
+
+            response = recommend_book("fantasy")
+            print(response.content)
+
+            Explanation: This is a Code Examples question, 
+            so only a code snippet is provided.
```
willbakst, line 145: any reason this is on a newline?

willbakst (commented):
Now that I'm seeing this and how it's different from the local chat with codebase, I think we should include a cookbook markdown for the agent and the evals. WDYT?

examples/cookbook/agents/documentation_agent/agent.py
willbakst: this was resolved but I'm still seeing it on a newline with no explanation, so either the change wasn't pushed or there's a reason there's a newline?

brenkao: oh oops, there were two so i only changed one

docs/cookbook/evals/evaluating_documentation_agent.md
```diff
@@ -0,0 +1,347 @@
...
+    - [Chaining](../../learn/chaining.md)
+    - [Evals](../../learn/evals.md)
+
+## Documentation Agent
+
+??? note "Check out the Full Code Here"
+
+    We will be using our `DocumentationAgent` for our evaluations. For the full code, refer to the [example](https://github.com/Mirascope/mirascope/tree/dev/examples/cookbook/agents/documentation_agent).
+
+```python
```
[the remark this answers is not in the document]
brenkao: i think this is a github viewer issue. its collapsable with no issues on mkdocs.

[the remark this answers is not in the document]
willbakst: ah ok that's fine then

willbakst (approved):

Issue #388, opened 2024-08-12
Support for models deployed on Azure AI Studio


vintrocode:
### Description

Azure has an offering separate from the OpenAI Studio that allows you to deploy basically every other model outside of OpenAI's. I deployed a Llama-3.1-405B-Instruct endpoint (following [this guide](https://learn.microsoft.com/en-us/azure/ai-studio/how-to/deploy-models-llama?tabs=meta-llama-3-1&pivots=programming-language-python#meta-llama-chat-models)) and I'd love to use Mirascope for interacting with it. I imagine it'd be similar to the azure client wrapper, but not entirely sure.

willbakst:
Just took a look at the guide. It looks like it's not an OpenAI compatible endpoint but a separate Azure endpoint, which means the client wrapper won't work.

Instead, we would likely implement support specifically for the Azure SDK (like we do for all other providers).

I've put this into the v1.1 milestone :)

willbakst:
This is released in `v1.2`!

PR #469, opened 2024-09-04
feat: Support AzureAI

koxudaxi:
Implements #388


mirascope/core/azureai/call_response.py
```diff
@@ -0,0 +1,176 @@
...
+            ValidationError: if the tool call doesn't match the tool's schema.
+            ValueError: if the model refused to response, in which case the error
+                message will be the refusal.
+        """
+        if tools := self.tools:
+            return tools[0]
+        return None
+
+    @classmethod
+    def _get_tool_message(cls, tool: AzureAITool, output: str) -> ToolMessage:
```
willbakst, line 152: nit: why refactor this into a separate private method? seems unnecessary for just a few lines of code?

mirascope/core/azureai/call_response_chunk.py
```diff
@@ -0,0 +1,88 @@
...
+        """Returns the name of the response model."""
+        return self.chunk.model
+
+    @property
+    def id(self) -> str:
+        """Returns the id of the response."""
+        return self.chunk.id
+
+    @property
+    def usage(self) -> CompletionsUsage:
```
willbakst, line 76: is this never `None`? weird that the call response can be `None` but this isn't...

mirascope/core/azureai/__init__.py
willbakst, line 1: can you also include `Azure AI` and `Vertex` in the `Supported Providers` section of the calls documentation? Just realized we don't mention we support these new providers anywhere

mirascope/core/azureai/call_response.py
willbakst: note: this improves performance slightly

willbakst (approved):

PR #473, opened 2024-09-05
docs: add web_search.ipynb

koxudaxi:


examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "# Web Search Agent Implementation\n",
+        "\n",
+        "This notebook demonstrates the implementation of a Web Search Agent using the Mirascope library. The agent is capable of performing web searches and parsing webpage content to answer user queries.\n",
```
willbakst, line 9: nit: this feels duplicated below. I think we can remove this and include it only in the introduction section?

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+        "\n",
+        "## Table of Contents\n",
```
willbakst, line 11: we can probably remove "Table of Contents" here and just have the table of contents right under the header?

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## 1. Introduction <a name=\"introduction\"></a>\n",
+        "\n",
+        "The Web Search Agent is an AI-powered tool that combines the capabilities of Large Language Models (LLMs) with web search functionality. It can understand user queries, perform web searches, parse webpage content, and generate informative responses.\n",
+        "\n",
+        "Key components of our implementation include:\n",
+        "- Mirascope library for LLM interactions\n",
```
willbakst, line 29: add links for things like `Mirascope` and`OpenAI` here?

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+        "First, we need to install the required packages and import the necessary modules."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "!pip install \"mirascope[openai]\" requests beautifulsoup4 pydantic"
```
willbakst, line 49: where is duckduckgo-search?

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+        "\n",
+        "import requests\n",
+        "from bs4 import BeautifulSoup\n",
+        "from openai.types.chat import ChatCompletionMessageParam\n",
+        "from pydantic import BaseModel\n",
+        "\n",
+        "from mirascope.core import openai, prompt_template\n",
+        "\n",
+        "# Set your OpenAI API key\n",
+        "os.environ[\"OPENAI_API_KEY\"] = \"your-api-key-here\""
```
willbakst, line 68: nit: maybe have setting the api key as a separate cell?

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "def web_search(query: str) -> str:\n",
+        "    \"\"\"Perform a web search and return the results.\n",
+        "\n",
+        "    This is a simplified implementation. In a real-world scenario,\n",
+        "    you would integrate with a proper search engine API.\n",
+        "    \"\"\"\n",
+        "    # Placeholder implementation\n",
```
willbakst, line 92: i don't like this. there's no reason not to show the implementation using duckduckgo with an explanation of the implementation, otherwise the final cells for running the agent won't actually work, which detracts from the value

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "MAX_RESULTS = 10\n",
+        "\n",
+        "\n",
+        "class WebSearchAgent(BaseModel):\n",
+        "    history: list[ChatCompletionMessageParam] = []\n",
```
willbakst, line 125: nit: see recent PR for updating this to use `BaseMessageParam | openai.OpenAIMessageParam`

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+        "    max_results: int = MAX_RESULTS\n",
+        "\n",
+        "    @openai.call(\"gpt-4o-mini\", stream=True, tools=[parse_webpage, web_search])\n",
```
willbakst, line 128: nit: we should either be passing these in as dynamically configured tools so they have access to self OR we should remove state like `max_results` and keep the tools separate. For simplicity, I'm leaning toward the latter.

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "agent = WebSearchAgent()\n",
+        "agent.run()"
+      ]
+    },
+    {
```
willbakst, line 211: we should actually run the agent and save the output and include the output in the guide so people can see example runs (which is particularly useful when viewing the guide in github)

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+        "            print(\"(Assistant): \", end=\"\", flush=True)\n",
+        "            self._step(query)\n",
+        "            print(\"\")"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "Let's break down the key components of our WebSearchAgent class:\n",
```
willbakst, line 179: ok, I'm not sure I love the structure here. I like that you're breaking down the key components, but I think the structure here still follows the cookbook structure too closely.

Remember, the goal of these guides is to provide a more comprehensive introduction / getting started walkthrough. This gives us an opportunity to slowly build things in smaller chunks and explain each part, building on the last part until we have our final agent.

For example, we can first show how to implement a basic call, then we can show streaming the call, then we can show putting the streaming call inside the `BaseModel` agent and adding history, then we can show adding tools (which we could further break down if we wanted to into showing just the web search tool and then showing we need to add the parse webpage tool to get the actual content for better responses).

All along the way we should be printing and showing intermediate outputs so that the notebook flows more like a guide.

If you have any questions on this wrt. next steps let me know, I think this is the most important change we'll need to make for this guide to be really useful.

examples/getting_started/web_search.ipynb
```diff
@@ -0,0 +1,255 @@
...
+        "        - `web_search`: Search the web for information (limited to {self.max_results} results).\n",
+        "        - `parse_webpage`: Parse the content of a webpage.\n",
+        "\n",
+        "        When calling the `web_search` tool, the `body` is simply the body of the search\n",
+        "        result. You MUST then call the `parse_webpage` tool to get the actual content\n",
+        "        of the webpage. It is up to you to determine which search results to parse.\n",
+        "\n",
+        "        Once you have gathered all of the information you need, generate your response,\n",
+        "        which should strike the right balance between brevity and completeness. You answer\n",
+        "        to the user's query should be concise yet comprehensive.\n",
```
willbakst, line 143: nit: it's useful to include the current date as this will often provide better results (particularly for requests that ask about "recent" stuff where the LLM might use the date of its training cutoff rather than the actual current date because it doesn't know the current date unless we tell it).

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+    "- Implementing custom logic that the LLM can leverage (e.g., determining the continent based on the country)\n",
+    "\n",
+    "By providing tools, you give the LLM additional capabilities that it can use to generate more accurate, formatted, or context-aware responses. In this case, the `FormatCapital` tool ensures that the capital information is always presented in a consistent format."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Asynchronous Processing\n",
```
willbakst, line 272: I think we should also show async streaming at least in this section?

another option for all of these sections if we want to keep them brief would be to add a link to their respective learn docs that contains all of the information?

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import re\n",
+    "\n",
+    "\n",
+    "def format_capital_info(chunk: str) -> str:\n",
```
willbakst, line 375: this won't actually work because the chunks may be just a single character, so you'll have to manage the content as you stream it and format more than just a single chunk to get the effect you're talking about here.

For the quickstart, this may be a bit overkill, and as I mentioned in my previous comment we may want to just keep this simple and then link to the Learn section?

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+    "This modification formats the streamed content in real-time, bold-ing headers and converting numbered lists to bullet points. This demonstrates how streaming can be combined with real-time processing to enhance the presentation of information as it's generated.\n",
+    "\n",
+    "Streaming is particularly useful when dealing with longer, more detailed responses, such as comprehensive information about capital cities. It allows for a more interactive and responsive user experience, especially in applications where immediate feedback is valuable."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Response Models\n",
```
willbakst, line 401: as I mentioned in my comment re: tools, I think the `Response Models` section should start with an extremely simple example where the goal is just to get structured information.

We can include additional information about validators if we think we should, but we can also just link to the learn section if we think that would be better?

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+    "4. We can then access the structured data directly, with type checking and validation already performed.\n",
+    "\n",
+    "Response models are particularly useful when you need consistent, structured output from your LLM calls, especially for data-driven applications."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Dynamic Configuration & Chaining\n",
```
willbakst, line 461: now that I'm reading this, I think we should split these sections in two

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+    "\n",
+    "@openai.call(\"gpt-4o-mini\")\n",
+    "@prompt_template(\"Translate the following to French: {text}\")\n",
+    "def translate_to_french(text: str):\n",
+    "    pass\n",
+    "\n",
+    "\n",
+    "original_text = \"Recently, it's been really hot, so I’m thinking of gathering with friends to enjoy some champagne outdoors. I’m looking for something refreshing and light, especially a champagne that’s enjoyable even on a hot day. If possible, I’d love to know about any recommended champagne that’s perfect for relaxing with everyone.\"\n",
+    "summary = summarize(original_text)\n",
+    "french_summary = translate_to_french(summary.content)\n",
```
willbakst, line 513: while you certainly can chain like this, we recommend chaining using dynamic configuration because then the final `response.model_dump` will include the entire dump from the whole chain and not just this call (which we should cover in the learn docs).

useful here I think to then show that output.

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+    "\n",
+    "response = city_info(\"Tokyo\")\n",
+    "print(response.content)  # This will be a JSON-formatted string"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "By setting `json_mode=True`, we instruct the LLM to format its response as JSON, which can then be easily parsed and used in your application."
```
willbakst, line 555: I think it's important to note here that not all providers have an explicit JSON mode, and for those providers we simply attempt to instruct the model to provide JSON; however, there is no guarantee that it will actually output only JSON (as it may start with some text like "Here is the JSON: ..."). This is where Output Parsers can be useful (and works as a great next section following such a note).

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+    "    capital: str\n",
+    "\n",
+    "\n",
+    "def parse_country_and_capital(response: openai.OpenAICallResponse) -> CountryInfo:\n",
+    "    country, capital = response.content.split(\" -> \")\n",
+    "    return CountryInfo(country=country, capital=capital)\n",
+    "\n",
+    "\n",
+    "@openai.call(\"gpt-4o-mini\", output_parser=parse_country_and_capital)\n",
+    "@prompt_template(\"What is the capital of {country}? the format 'country -> capital'.\")\n",
```
willbakst, line 586: nit: update "the format ..." to "Your output MUST follow the format 'country -> capital'."

examples/getting_started/quickstart.ipynb
willbakst, line 1: I think this is a good start. There are a few high level things I think we should update:

1. I think for each example we should show just a single example that is as simple as possible -- the purpose here is just to demonstrate a most basic (quickstart) version of that primitive building block.
2. Either at the beginning or end of each section we should link to the Learn documentation and mention that we have more comprehensive and detailed information there. This way if there's a particular feature that a new user is interested in they can directly link to the detailed information for that feature.
3. Now that I'm reading through the quickstart, I don't think we actually need all of these sections. I think Prompts, Calls, Streams, Response Models, JSON Mode, and Async should suffice. We can then have a separate follow-up getting started guide that covers e.g. Dynamic Configuration and Chaining in more detail with more examples of different types of chains (that use the basic building blocks from the quickstart).

I think point (3) is the most important update, particularly because it will make the other notebooks make more sense wrt. what information we should explain at what level of detail (e.g. we can link quickstart and say check that out first and then discuss dynamic configuration assuming the user already knows about calls etc.). This will result in the following notebooks:

1. Quickstart
2. Dynamic Configuration & Chaining
3. Tools and Agents
4. Evaluations

Mostly I'm realizing that we should have a core set of "Getting Started" guides that are really about getting the user started. The rest of the material and super in-depth details can be reserved for the Learn section (where it makes more sense to cover absolutely everything and anything in extreme detail)

examples/getting_started/web_search.ipynb
willbakst, line 1: as I mentioned in my comment on the Quickstart notebook, we should update this to be `Tools and Agents` as a notebook that provides a getting started guide around using tools and implementing agents (where web search is just the underlying functionality we've chosen to showcase tools and agents).

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,626 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "@prompt_template(\"What is the capital of {country}?\")\n",
+    "def get_capital_prompt(country: str):\n",
+    "    pass\n",
```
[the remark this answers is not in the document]
koxudaxi: Sorry, It's just my habit.

examples/getting_started/quickstart.ipynb
koxudaxi: Your explanation makes a lot of sense and makes sense to me, so I will quote it.

examples/getting_started/web_search.ipynb
koxudaxi: I have created the new branch and the PR. https://github.com/Mirascope/mirascope/pull/473
Because I want to add only Quick start notebook.

koxudaxi:
@willbakst 
I applied your suggestion to `quickstart.ipynb`
But, I keep the branch until merge the new PR.

PR #475, opened 2024-09-06
docs: Add quickstart notebook

koxudaxi:
The PR is created to add just quickstart notebook instead of https://github.com/Mirascope/mirascope/pull/473


examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this example:\n",
+    "1. The `@prompt_template` decorator defines the structure of the prompt.\n",
+    "2. `{country}` is a placeholder that gets replaced with the matching argument passed to the function.\n",
+    "\n",
+    "Here's an example of a multi-line prompt:"
```
willbakst, line 102: this is not multi-line, this is using messages with different roles using the role keywords

multiline prompt is when the content for a single message spans multiple lines (and needs to be started on a newline in order to be parsed properly).

For example:

## BAD

These will be parsed incorrectly and include erroneous tabs since cleandoc/dedent will have no indentation from the first line of the multi-line prompt

```python
# BAD, will be parsed incorrectly and include erroneous tabs

@prompt_template(
    """First line
    Second line
    """
)

@prompt_template(
    """
    USER: First line
    Second line
    """
)

@prompt_template(
    """
    SYSTEM: First line
    Second line

    USER: single line prompt
)
```

## GOOD

```python
@prompt_template(
    """
    First line
    Second line
    """
)

@prompt_template(
    """
    USER:
    First line
    Second line
    """
)

@prompt_template(
    """
    SYSTEM:
    First line
    Second line

    USER: single line prompt
)
```

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This example:\n",
+    "1. Includes \"SYSTEM\" and \"USER\" keywords to define messages with different roles in the prompt.\n",
+    "2. Uses multiple variables (`{topic}` and `{subtopic}`) to create a more specific and dynamic prompt.\n",
+    "\n",
+    "Prompt templates improve code readability, maintain consistency in prompts, and make it easier to experiment with and modify prompts for effective prompt engineering.\n",
+    "\n",
+    "For more detailed information on prompt templates, check out our [Learn documentation on Prompts](https://docs.mirascope.io/learn/prompts)."
```
willbakst, line 146: nit: I think we drop "Learn" and just have the link be "documentation on Prompts"

I would do the same for all of the other links as well

examples/getting_started/quickstart.ipynb
willbakst, line 1: I like the simpler examples, but now I'm feeling like the descriptions / explanations are lacking. I think we should still provide a little more detail around exactly what's happening like you had before.

Of course, the truly detailed information should live in the learn documentation, but now i'm feeling like this is too sparse. Thoughts?

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+    "\n",
+    "response = city_info(\"Tokyo\")\n",
+    "print(response.content)  # This will be a JSON-formatted string"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that not all providers have an explicit JSON mode. For those providers, we attempt to instruct the model to provide JSON; however, there is no guarantee that it will output only JSON (it may start with some text like \"Here is the JSON: ...\"). This is where Output Parsers can be useful.\n",
```
willbakst, line 389: nit: link to the Output Parsers section here?

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+     "end_time": "2024-09-06T14:00:17.154267Z",
+     "start_time": "2024-09-06T14:00:16.512749Z"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The capital of France is The capital of France is Paris.\n",
```
willbakst, line 269: the output here is a little weird because the LLM is including the text you're also including.

I think we could improve this by either only printing the content or instead using `response_model` here.

I'm thinking we should put Response Models before this section and then use `response_model` for the async example (and focus on extraction rather than generation). Thoughts?

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+    "Note that not all providers have an explicit JSON mode. For those providers, we attempt to instruct the model to provide JSON; however, there is no guarantee that it will output only JSON (it may start with some text like \"Here is the JSON: ...\"). This is where Output Parsers can be useful.\n",
+    "\n",
+    "For more information on JSON mode, refer to our [Learn documentation on JSON Mode](https://docs.mirascope.io/learn/json-mode)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Response Models\n",
```
willbakst, line 398: let's move this before the async section as i mentioned in my other comment

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+     "text": [
+      "The capital of Japan is Tokyo\n"
+     ]
+    }
+   ],
+   "source": [
+    "from pydantic import BaseModel\n",
+    "\n",
+    "\n",
+    "class Capital(BaseModel):\n",
```
willbakst, line 425: for this one, rather than showing how to generate a response model, we should show how to extract unstructured information into a structured format. For example:

```python

class Capital(BaseModel):
    city: str
    country: str

@openai.call("gpt-4o-mini", response_model=Capital)
@prompt_template("{query}")
def extract_capital(query: str): ...

capital = extract_capital("The capital of France is Paris")
print(capital)
```

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+   "metadata": {},
+   "source": [
+    "For more information on output parsers, see our [Learn documentation on Output Parsers](https://docs.mirascope.io/learn/output-parsers)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This concludes our Quickstart Guide to Mirascope. We've covered the main features of the library, including prompt templates, basic calls, streaming, asynchronous processing, JSON mode, response models, and output parsers. Each of these features can be combined and customized to create powerful, flexible AI applications.\n",
```
willbakst, line 508: nit: make sure to update the order of topics covered here to match the updated order

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,535 @@
...
+   "metadata": {},
+   "source": [
+    "For more details on response models, including advanced validation techniques, check our [Learn documentation on Response Models](https://docs.mirascope.io/learn/response-models)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Output Parsers\n",
```
willbakst, line 450: i think we should update this section so that it is a better follow-up section to JSON mode.

For example:

```python
import json
from mirascope.core import openai, prompt_template


def only_json(response: anthropic.AnthropicCallResponse) -> str:
    json_start = response.content.index("{")
    json_end = response.content.rfind("}")
    return response.content[json_start : json_end + 1]


@openai.call("gpt-4o-mini", json_mode=True, output_parser=only_json)
@prompt_template("Extract {fields} from the following text: {text}")
def json_extraction(text: str, fields: list[str]): ...


json_response = json_extraction(
    text="The capital of France is Paris",
    fields=["capital", "country"],
)
print(json.loads(json_response))
# > {'capital': 'Paris', 'country': 'France'}
```

We can then note that this is particularly important for providers like Anthropic that do not have a strict JSON mode.

examples/getting_started/quickstart.ipynb
willbakst, line 1: nit: I would like to see the outputs for the cells for each section (except setup) cause I think that provides helpful additional details when reading through the notebook

koxudaxi: I tried to save the output of cell. But, I can't do it. 
I pasted all output as comments in the code sections

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,455 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this example:\n",
+    "1. We import the `prompt_template` decorator from Mirascope.\n",
+    "2. We define a function `get_capital_prompt` and decorate it with `@prompt_template`.\n",
+    "3. The `{country}` placeholder in the template will be replaced with the actual argument passed to the function.\n",
+    "4. When we call `get_capital_prompt(\"Japan\")`, it returns the string \"What is the capital of Japan?\"\n",
```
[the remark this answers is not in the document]
koxudaxi: I think we don't need the example. because, the output is `None`.
But, we should explain about `@prompt_template`. 🤔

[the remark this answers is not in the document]
koxudaxi: Should we change the order of the prompt and call sections?
I think we can delay the explanation in the call section (e.g., "We can use @prompt_template to define prompt templates. We will explain the details of @prompt_template in the next section.")

examples/getting_started/quickstart.ipynb
willbakst: oh weird I saw the outputs in a previous iteration.

maybe this will help? https://stackoverflow.com/questions/62288996/how-to-save-a-jupyter-notebook-with-outputs

examples/getting_started/quickstart.ipynb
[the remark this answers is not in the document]
willbakst: hmmm, i like the idea of `prompt_template` actually turning the function into something useful, this way we can follow a more natural flow of explanation (and it's easier to explain `prompt_template` because we can run the functions and show what happens so it's more clear)

What if we had a `prompt_template` decorated function always return the list of `BaseMessageParams` including any computed fields from the dynamic config but ignore things like tools, metadata, and messages (particularly messages because if a user uses messages with a call decorator then the prompt template is ignored anyway and shouldn't be used).

Then, inside of the decorator for each of the paths we can check if the function being decorated has something like `_original_fn` attached to it, grab the dynamic config from there, if it has messages, use those, and if not, call the decorated function to get the list of messages (which could all happen inside of the base setup call).

WDYT?

examples/getting_started/quickstart.ipynb
koxudaxi: the solution is for only `Colab `

koxudaxi: I understand we can see the output when we change  `source_hidden` to `false`

```
     "jupyter": {
-     "source_hidden": true
+     "source_hidden": false
     },

```

koxudaxi: I clean up output cells

examples/getting_started/quickstart.ipynb
```diff
@@ -0,0 +1,650 @@
...
+    "print(\n",
+    "    f\"Name: {response.name}, Population: {response.population}, Country: {response.country}\"\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This approach combines the benefits of JSON mode with the type safety and validation of response models.\n",
```
willbakst, line 537: nit: this is essentially duplicated from the text before the code snippet, either remove it there or remove it here?

willbakst (commented):
please also sync once the prompt template PR is merged so that the only change in this PR is the notebook

willbakst (approved):

PR #476, opened 2024-09-09
Improve prompt decorator return type

koxudaxi:


tests/core/base/test_prompt.py
```diff
@@ -55,6 +56,35 @@ def genre(self) -> str:
     assert str(prompt) == "Recommend a fantasy book."
 
 
+def test_base_prompt_with_prompt_template() -> None:
+    """Tests the `BasePrompt` class with prompt_template."""
+
+    class BookRecommendationPrompt(BasePrompt):
+        prompt_template: ClassVar[str] = "Recommend a {genre} book."
```
willbakst, line 63: can we also test setting it without having to write `ClassVar[str]`?

For example:

```python
class BookRecommendationPrompt(BasePrompt):
    prompt_template = "recommend a {genre} book"
    genre: str
```

I want to make sure this works properly so that users can more easily define prompts with fewer imports

mirascope/core/base/_create.py
willbakst, line 1: make sure all other decorators are similarly updated (e.g. stream)

I think this indicates we should likely implement a new utility function (e.g. `get_dynamic_configuration`)

mirascope/core/base/_utils/_setup_call.py
```diff
@@ -51,12 +51,11 @@ def setup_call(
     if not messages:
         prompt_template = get_prompt_template(fn)
         assert prompt_template is not None, "The function must have a docstring."
```
willbakst, line 53: nit: i just noticed this should say "The function must have a prompt template" not "docstring"

please make sure this is updated anywhere else we have this issue

mirascope/core/base/prompt.py
```diff
@@ -298,19 +305,67 @@ async def run():
...
+    prompt: type[_BasePromptT]
+    | Callable[_P, BaseDynamicConfig]
+    | Callable[_P, Awaitable[BaseDynamicConfig]],
+) -> TypeIs[
+    Callable[_P, BaseDynamicConfig] | Callable[_P, Awaitable[BaseDynamicConfig]]
+]:
+    return isinstance(prompt, types.FunctionType)
+
+
+def _get_prompt_messages_from_prompt_function(
```
willbakst, line 348: this function is not necessary since you can just use `get_fn_args` and `parse_prompt_messages` directly inside the decorator below

mirascope/core/base/prompt.py
```diff
@@ -341,16 +396,64 @@ def recommend_book(genre: str):
...
+        if not _is_function(prompt):
+            return prompt
+
+        if fn_is_async(prompt):
+
+            @wraps(prompt)
+            async def get_base_message_params_async(
+                *args: _P.args, **kwargs: _P.kwargs
+            ) -> list[BaseMessageParam]:
+                return _get_prompt_messages_from_prompt_function(
```
willbakst, line 432: instead (wrt. previous comment) this should just look like this:

```python
return parse_prompt_messages(
    roles=["system", "user", "assistant"],
    template=template,
    attrs=get_fn_args(prompt),
    dynamic_config=await prompt(*args, **kwargs),
)
```

mirascope/core/base/prompt.py
```diff
@@ -341,16 +396,64 @@ def recommend_book(genre: str):
...
+            get_base_message_params_async._original_fn = prompt  # pyright: ignore [reportAttributeAccessIssue,reportFunctionMemberAccess]
+            return get_base_message_params_async
+
+        else:
+
+            @wraps(prompt)
+            def get_base_message_params(
+                *args: _P.args, **kwargs: _P.kwargs
+            ) -> list[BaseMessageParam]:
+                return _get_prompt_messages_from_prompt_function(
```
willbakst, line 448: same thing here except sync:

```python
return parse_prompt_messages(
    roles=["system", "user", "assistant"],
    template=template,
    attrs=get_fn_args(prompt),
    dynamic_config=prompt(*args, **kwargs),
)
```

willbakst (approved):

PR #478, opened 2024-09-10
docs: Add dynamic_config_and_chaining.ipynb

koxudaxi:


examples/getting_started/dynamic_config_and_chaining.ipynb
```diff
@@ -0,0 +1,734 @@
...
+     "text": [
+      "The Girl with the Dragon Tattoo by Stieg Larsson (mystery)\n"
+     ]
+    }
+   ],
+   "source": [
+    "from mirascope.core import BaseToolKit, toolkit_tool\n",
+    "\n",
+    "\n",
+    "class BookToolkit(BaseToolKit):\n",
```
willbakst, line 184: nit: `BaseToolKit` might be a little overkill for this getting started guide. Maybe just show return a basic tool function here? and then add a link to the toolkit docs and/or the tools + agents getting started guide saying that that's where this is most useful?

examples/getting_started/dynamic_config_and_chaining.ipynb
```diff
@@ -0,0 +1,734 @@
...
+    "# asyncio.run(main())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Advanced Techniques\n",
+    "\n",
+    "### Combining Dynamic Config and Chaining\n",
```
willbakst, line 530: you already showed this above with chaining using computed fields?

I would maybe make a more clear distinction here between chaining using functions vs. chaining with computed fields, specifically focusing on the fact that when using computed fields you get the full chain as part of the response of a single function rather than having to call and track them separately in sequence

examples/getting_started/dynamic_config_and_chaining.ipynb
```diff
@@ -0,0 +1,506 @@
...
+    "\n",
+    "\n",
+    "@openai.call(\"gpt-4o-mini\")\n",
+    "@prompt_template(\"Rewrite this {genre} story in a {tone} tone: {story}\")\n",
+    "def rewrite_story(story: str, genre: str, tone: str) -> openai.OpenAIDynamicConfig:\n",
+    "    return {\n",
+    "        \"computed_fields\": {\n",
+    "            \"story\": story,\n",
+    "            \"genre\": genre,\n",
+    "            \"tone\": analyze_tone(story).content,\n",
```
willbakst, line 341: nit: this should just be `analyze_tone(story)` for the improved traceability as we mentioned in the other section on chaining using computed fields

willbakst (approved):

PR #480, opened 2024-09-10
docs: Add tools_and_agents.ipynb

koxudaxi:


examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,519 @@
...
+   "metadata": {},
+   "source": [
+    "## 1. Introduction\n",
+    "\n",
+    "Tools and Agents are two key concepts in building advanced AI systems, particularly those involving Large Language Models (LLMs):\n",
+    "\n",
+    "- **Tools**: Functions that extend the capabilities of LLMs, allowing them to perform specific tasks or access external data.\n",
+    "- **Agents**: Autonomous or semi-autonomous entities that use LLMs and Tools to perform complex tasks or interact with users.\n",
+    "\n",
+    "In this notebook, we'll explore how to create and use Tools, and how to implement an Agent using the WebSearchAgent as our example. We'll be using Mirascope, a framework that simplifies the process of building LLM-powered applications with tools and agents.\n",
```
willbakst, line 33: nit: change "framework" -> "toolkit"

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,519 @@
...
+    "7. **Logging**: Implement logging to track the behavior and performance of your Agents.\n",
+    "\n",
+    "For more advanced usage, you can explore concepts like:\n",
+    "\n",
+    "- Multi-agent systems\n",
+    "- Tool chaining and composition\n",
+    "- Dynamic tool selection\n",
+    "- Memory and state management for long-running agents\n",
+    "\n",
+    "For more advanced techniques and best practices in using Mirascope, refer to the following documentation:\n",
```
willbakst, line 479: nit: drop "advanced" here so it's just "more techniques and..."

my reasoning here is that tools and agents are the most advanced techniques

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,519 @@
...
+    "\n",
+    "from mirascope.core import BaseMessageParam, openai\n",
+    "\n",
+    "\n",
+    "class WebAssistantBase(BaseModel):\n",
+    "    messages: list[BaseMessageParam | openai.OpenAIMessageParam] = []\n",
+    "    search_history: list[str] = []\n",
+    "    max_results_per_query: int = 2\n",
+    "\n",
+    "    def _web_search(self, queries: list[str]) -> str:\n",
```
willbakst, line 250: this will result in worse tool usage because the tool is generated from the function definition, so the docstring for `_web_search` is used here instead of the one for `web_search` (which is far more verbose).

Instead, we should implement the `web_search` tool to accept the `self` argument in preparation for including it in out agent.

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,519 @@
...
+    "        return \"\\n\".join(line for line in lines if line)\n",
+    "    except Exception as e:\n",
+    "        return f\"{type(e)}: Failed to extract content from URL {url}\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This `extract_content` function is a Tool that takes a URL as input and returns the main content of the webpage as a string. It uses BeautifulSoup to parse the HTML and extract the relevant text content.\n",
```
willbakst, line 148: I think we need to provide more explanation here around what exactly is happening. For example, we should mention that when passed to a call it will be automatically converted into a `BaseTool` class, and when passed to the LLM API we use this schema to auto-generate the actual schema the API expects. We should likely show an example of what this schema looks like (use the `tool_schema` method and something like `OpenAITool.type_from_fn(...)` to get the correct type to output the correct schema rather than writing it by hand)

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,519 @@
...
+    }
+   },
+   "outputs": [],
+   "source": [
+    "from pydantic import BaseModel\n",
+    "\n",
+    "from mirascope.core import BaseMessageParam, openai\n",
+    "\n",
+    "\n",
+    "class WebAssistantBase(BaseModel):\n",
```
willbakst, line 245: I think we should first show implementing a basic chat bot without any tools before we implement any tools just to show the basic flow of adding history and state management. We should make sure we explain all of that stuff first to set a good baseline for adding tools to the chatbot.

This will also enable writing the guide such that when you implement a tool you already have access to the agent and can write the tool inside of the agent (with access to self as mentioned in other comments)

koxudaxi:
@willbakst 
I introduced an example of basic chatbot.
Could you please check if this is the deployment you are expecting.

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,583 @@
...
+    "## 4. Understanding Tools in Mirascope\n",
+    "\n",
+    "Tools in Mirascope are functions that extend the capabilities of LLMs. They allow the LLM to perform specific tasks or access external data. Tools are typically used to:\n",
+    "\n",
+    "1. Retrieve information from external sources\n",
+    "2. Perform calculations or data processing\n",
+    "3. Interact with APIs or databases\n",
+    "4. Execute specific actions based on the LLM's decisions\n",
+    "\n",
+    "Let's start by creating a simple Tool that extracts content from a webpage:"
```
willbakst, line 171: nit: remove "simple" here since some users might not consider this simple

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,583 @@
...
+    "        return \"\\n\".join(line for line in lines if line)\n",
+    "    except Exception as e:\n",
+    "        return f\"{type(e)}: Failed to extract content from URL {url}\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This `extract_content` function is a Tool that takes a URL as input and returns the main content of the webpage as a string. It uses BeautifulSoup to parse the HTML and extract the relevant text content.\n",
```
willbakst, line 226: nit: add a link to BeautifulSoup?

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,583 @@
...
+    "            return \"\\n\\n\".join(urls)\n",
+    "        except Exception as e:\n",
+    "            return f\"{type(e)}: Failed to search the web for text\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This `web_search` method is a more complex Tool that takes a list of search queries and returns a string of newline-separated URLs from the search results. It uses the DuckDuckGo search API to perform the searches and updates the agent's `search_history`.\n",
```
willbakst, line 331: nit: add something about "Luckily the DuckDuckGO search API does not require an API key!"

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,583 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This implementation includes:\n",
+    "\n",
+    "1. The `_stream` method, which sets up the LLM call with the necessary tools and computed fields.\n",
+    "2. The `_step` method, which processes the LLM response, handles tool calls, and updates the conversation history.\n",
+    "3. The `run` method, which implements the main interaction loop for the agent.\n",
+    "\n",
+    "The `@openai.call` and `@prompt_template` decorators are used to set up the LLM interaction and define the prompt for the agent.\n",
```
willbakst, line 430: nit: add a blurb about how we've seen that specifying the specific tools in the system prompt and how to use them improves performance

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,583 @@
...
+     "end_time": "2024-09-11T08:23:30.442635Z",
+     "start_time": "2024-09-11T08:22:56.224422Z"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "(Assistant): using web_search tool with args: {'queries': ['large language models overview', 'applications of large language models', 'future of large language models', 'impact of large language models on AI']}\n",
```
willbakst, line 458: nit: something weird happening here where the input isn't being saved as output, so maybe include the input in the cell as a comment so it's more clear what you asked?

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,583 @@
...
+    "- Multi-agent systems\n",
+    "- Tool chaining and composition\n",
+    "- Dynamic tool selection\n",
+    "- Memory and state management for long-running agents\n",
+    "\n",
+    "For more techniques and best practices in using Mirascope, refer to the following documentation:\n",
+    "\n",
+    "- [Dynamic Configuration](https://docs.mirascope.io/latest/learn/dynamic_configuration/)\n",
+    "- [Chaining](https://docs.mirascope.io/latest/learn/chaining/)\n",
+    "- [Streams](https://docs.mirascope.io/latest/learn/streams/)"
```
willbakst, line 547: nit: should we also link the Tools and Agents documentation here?

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,586 @@
...
+    "\n",
+    "\n",
+    "class BasicChatbot(BaseModel):\n",
+    "    messages: list[BaseMessageParam | openai.OpenAIMessageParam] = []\n",
+    "\n",
+    "    @openai.call(model=\"gpt-4o-mini\")\n",
+    "    @prompt_template(\n",
+    "        \"\"\"\n",
+    "        SYSTEM: You are a friendly chatbot assistant. Engage in a conversation with the user.\n",
+    "        \n",
```
willbakst, line 118: nit: remove remove newline here between the system message and the MESSAGES keyword

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,586 @@
...
+    "This schema provides the LLM with information about the tool's name, description, and expected input parameters. The LLM uses this information to determine when and how to use the tool during its reasoning process.\n",
+    "\n",
+    "For more details on implementing and using Tools in Mirascope, see the [Tools documentation](https://docs.mirascope.io/latest/learn/tools/)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 5. Creating Custom Tools\n",
```
willbakst, line 271: nit: after reading this again, I'm wondering if this is the right. Maybe it would be better to write something like "Tools With Access To Agent State"?

examples/getting_started/tools_and_agents.ipynb
```diff
@@ -0,0 +1,586 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This implementation includes:\n",
+    "\n",
+    "1. The `_stream` method, which sets up the LLM call with the necessary tools and computed fields.\n",
+    "2. The `_step` method, which processes the LLM response, handles tool calls, and updates the conversation history.\n",
+    "3. The `run` method, which implements the main interaction loop for the agent.\n",
+    "\n",
+    "The `@openai.call` and `@prompt_template` decorators are used to set up the LLM interaction and define the prompt for the agent. Note that we've explicitly specified the available tools and their usage instructions in the system prompt. This approach has been shown to improve the performance and reliability of tool usage by the LLM.\n",
```
willbakst, line 431: nit: you write "This approach has been shown" which leads me to believe there is something like research to share on this point vs. it being something we have found to work in practice, so I would update this instead to say "We have found that this improves the performance..."

willbakst (approved):

PR #484, opened 2024-09-11
docs: Improve WHY document

koxudaxi:


docs/WHY.md
```diff
@@ -26,6 +26,78 @@ Mirascope addresses these pain points with a philosophy centered on transparency
...
+def hello_world():
+   ...
+
+response = hello_world()
+print(response.content)
+```
+
+   Mirascope's approach streamlines the process using decorators, making it more concise.
+
+2. **Provider Agnosticism**: Mirascope allows you to easily switch between providers without changing your core logic. This is not possible with official SDKs, which are tied to specific providers.
```
willbakst, line 64: nit: just make this "Provider Agnostic"

willbakst: also should we add code examples here like you did for the first point showcasing Mirascope easily using `prompt_template` with two providers compared to manually making it work with two providers official sdks?

docs/WHY.md
```diff
@@ -26,6 +26,78 @@ Mirascope addresses these pain points with a philosophy centered on transparency
...
+
+3. **Prompt Management**: Mirascope's `@prompt_template` decorator provides a way to manage prompts separately from logic. Official SDKs often mix prompts with API calls.
```
willbakst, line 66: same thing here, do you think a code example would help make this point more clear?

willbakst: i also wonder if this should be point (2) so it leads better into the point on being provider agnostic?

docs/WHY.md
```diff
@@ -26,6 +26,78 @@ Mirascope addresses these pain points with a philosophy centered on transparency
...
+model='gpt-3.5-turbo',
+messages=[
+ {'role': 'user', 'content': 'Count to 5.'}
+],
+temperature=0,
+stream=True
+)
+
+for chunk in response:
+ if chunk.choices[0].delta.content is not None:
```
willbakst, line 82: nit: weird indentation

docs/WHY.md
```diff
@@ -26,6 +26,78 @@ Mirascope addresses these pain points with a philosophy centered on transparency
     - **Type Safety**: Catch errors before runtime with our comprehensive type annotations.
     - **Intuitive API**: Our API is designed to be predictable and easy to learn, following Python best practices.
 
+## Comparison with Official SDKs
```
willbakst, line 29: I think the biggest thing here to cover is tools and agents at least in some capacity. that's really where the simplified syntax of mirascope + convenience functions really shine (and reduce code without reducing transparency).

I think maybe drop the third point about "Streaming Support" and include streaming, tools, and agents as examples of the first point on simplified syntax?

docs/WHY.md
```diff
@@ -44,3 +116,6 @@ Mirascope is ideal for:
 Ready to experience the Mirascope difference? Check out our [Learn](./learn/index.md) documentation to begin learning how to build cleaner, more maintainable LLM applications today.
 
 By choosing Mirascope, you're opting for a tool that respects your expertise as a developer while providing the conveniences you need to work efficiently with LLMs. We believe that the best tools get out of your way and let you focus on building great applications.
+```
+
+This updated version removes unsubstantiated claims, includes a comparison with OpenAI's official streaming implementation, and maintains a more balanced tone throughout the document.
```
willbakst, line 121: drop this (ai generated? be more careful not to include this stuff next time)

PR #486, opened 2024-09-12
docs: Improve provider_specific_features.md

koxudaxi:


docs/learn/provider_specific_features.md
```diff
@@ -18,7 +18,7 @@ This feature can be extremely useful when extracting structured information or u
 
 To use structured outputs with tools, use the `OpenAIToolConfig` and set `strict=True`. You can then use the tool just like you normally would:
```
willbakst, line 19: should we add a link to the `Tools` docs somewhere here? we say "like you normally would" but then provide no reference

docs/learn/provider_specific_features.md
```diff
@@ -47,11 +47,52 @@ if tool := response.tool:
...
+```python
+from openai import OpenAI
+import json
+
+client = OpenAI()
+
+def format_book(title: str, author: str) -> str:
+    return f"{title} by {author}"
+
+function_json = {
```
willbakst, line 63: this example is not correct (primarily "functions" are deprecated). here are their docs on structured outputs: https://platform.openai.com/docs/guides/structured-outputs

note that the endpoint they show is beta, so you can still use the official endpoint manually. i think we should show using the official endpoint first (i.e. using `json_schema` with `strict=True`) and then showing their beta endpoint that enables using Pydantic more directly

willbakst (approved):

PR #487, opened 2024-09-12
docs: Improve prompts.md

koxudaxi:


docs/learn/prompts.md
```diff
@@ -31,9 +53,13 @@ print(prompt)
 # > Recommend a fantasy book
 ```
 
-For your convenience, we will also automatically dedent and strip the prompt template. This makes writing multi-line prompts a breeze:
+In this example, the `@prompt_template` decorator defines the template string `"Recommend a {genre} book"`. The genre attribute of the `BookRecommendationPrompt` class corresponds to the `{genre}` placeholder in the template. When you create an instance of the class with `genre="fantasy"`, this value is automatically injected into the template.
```
willbakst, line 56: hmm, this feels entirely duplicated from the previous section? we should merge this into the explanation above instead of duplicating the code?

docs/learn/prompts.md
```diff
@@ -31,9 +53,13 @@ print(prompt)
...
 
-```python
+We've implemented the `__str__` method in `BasePrompt` for easy verification of how variables are templated.
```
willbakst, line 58: same here, move this into the earlier section when we show `BasePrompt`

docs/learn/prompts.md
```diff
@@ -31,9 +53,13 @@ print(prompt)
...
+
+For your convenience, we automatically dedent and strip the prompt template. This makes writing multi-line prompts a breeze:
+
+```python hl_lines="5-8 11"
```
willbakst, line 62: so in line with my other comment, i think after the very first example showing `BasePrompt` we should use the `prompt_template` functional approach going forward until the end where we'll cover `BasePrompt` in more detail. Originally the reason I used `BasePrompt` for all of these examples was that adding `prompt_template` didn't do anything, but since we've implemented that feature and now the decorated function is useful, we should instead just use the functional approach for these explanations?

docs/learn/prompts.md
```diff
@@ -71,12 +97,16 @@ print(prompt)
...
+
 We also provide additional specifiers we've found useful in our own prompt engineering:
 
 - `list`: formats an input list as a newline-separated string
 - `lists`: formats an input list of lists as newline-separated strings separated by double newlines
 
-```python
+Here's an example demonstrating these custom specifiers:
+
+```python hl_lines="7 10 19-23"
```
willbakst, line 109: when using the function based approach, since the output is `list[BaseMessageParam]` we can always run `print(prompt(...)[0].content)` so it's pretty

docs/learn/prompts.md
```diff
@@ -172,11 +206,14 @@ Make sure when writing multi-line prompts with message roles that you start the
 )
 ```
 
+This ensures that the content for each role is properly formatted and aligned.
```
willbakst, line 209: nit: i think we should make it more clear that since we are dedenting on the user's behalf we need to make sure we don't include extra tabs and such

docs/learn/prompts.md
```diff
@@ -345,7 +388,7 @@ It leverages Pydantic's [`BaseModel`](https://docs.pydantic.dev/latest/concepts/
 
 To recap, you can create a prompt using `BasePrompt` by defining a class that inherits from it:
```
willbakst, line 389: i'm realizing that this entire section on `BasePrompt` is out of place. For example, `computed_field` should really be discussed in dynamic configuration alongside the section where we discuss computed fields for call functions.

the `run` method should be discussed in the "Calls" documentation alongside where we should making a call.

Essentially this documentation should really just focus on the core features of prompt templates and prompts and push everything else into the respective section where it makes sense

docs/learn/prompts.md
```diff
@@ -10,14 +10,36 @@ Prompts are the foundation of effective communication with Large Language Models
...
+This approach allows for flexible and reusable prompt creation, making it easy to generate variations of prompts by changing the input parameters.
+
+
+For more complex use cases, we'll use:
 
 - `@prompt_template`: a decorator for attaching a prompt template to a function or class
 - `BasePrompt`: a base class for writing provider-agnostic prompts.
 
-We will cover `BasePrompt` in more detail later. For now, all you need to know is that we inject fields with names that match template variables and that we've implemented the `__str__` method for easily viewing how template variables get injected:
+We'll cover `BasePrompt` in more detail later. For now, let's focus on how template variables are injected and how to view the resulting prompt:
```
willbakst, line 40: in line with a comment i make later, I think essentially we should show `BasePrompt` first briefly and then use the functional approach to discuss all of the various prompt templating features excluding the more complex ones (like computed fields).

we should mention here that we will cover `BasePrompt` in more detail in each of these respective sections.

key examples include talking about `prompt.run` in the Calls documentation or `computed_field` in the dynamic configuration section. i'm realizing it's weird to have these things separated and discussed multiple times.

docs/learn/prompts.md
```diff
@@ -529,16 +580,83 @@ print(response.content)
...
 ## Best Practices
 
-- Provider Comparison: Use `BasePrompt` to easily test the same prompt across different providers to compare performance and output quality.
-- Prompt Versioning: Utilize the metadata decorator to keep track of different versions of your prompts as you refine them.
-- Dynamic Content: Leverage `@computed_field` for injecting dynamic content or API calls into your prompts.
-- Cached Properties: Use `@functools.cached_property` to cache frequently used properties that you only want to compute once.
-- Prompt Libraries: Build libraries of commonly used prompts that can be easily shared across projects or teams.
-
-Mastering `BasePrompt` is the first step towards building robust LLM applications with Mirascope that are flexible, reusable, and provider-agnostic.
+To make the most of Mirascope's prompt features, consider the following best practices:
```
willbakst, line 591: in line with other comments, many of these points should be dropped and moved to the respective section where that content will live (e.g. run in "Calls")

if we want to discuss best practices here, it seems like the key things to focus on (to me) are:
1. Prompt Libraries
2. Testing Your Prompts

every other point is already covered in the respective section (e.g. multi-line formatting, etc.)

Then we should mention that more advanced features for prompting can be found in their respective sections (e.g. dynamic configuration)

docs/learn/prompts.md
```diff
@@ -529,16 +580,83 @@ print(response.content)
...
+
+- **Provider Comparison**: Use `BasePrompt` to easily test the same prompt across different providers to compare performance and output quality.
+   ```python
+   prompt = MyPrompt(...)
+   openai_response = prompt.run(openai.call(...))
+   anthropic_response = prompt.run(anthropic.call(...))
+   ```
+
+- **Prompt Versioning**: Utilize the metadata decorator to keep track of different versions of your prompts as you refine them.
```
willbakst, line 600: I think we should move metadata like other things into a section where it makes more sense.

right now, metadata is generally contained either on `BasePrompt` or as part of a response to a call, so it seems like the best place to discuss metadata would be in the "Calls" documentation?

docs/learn/prompts.md
```diff
@@ -529,16 +580,83 @@ print(response.content)
...
+   ```
+
+- **Testing**: Create unit tests for your prompts, especially for complex templates or those with computed fields.
+   ```python
+   def test_book_recommendation_prompt():
+       prompt = BookRecommendationPrompt(genre="mystery", age_group="adult")
+       assert "mystery book for an adult reader" in str(prompt)
+   ```
+
+Mastering `BasePrompt` is the first step towards building robust LLM applications with Mirascope that are flexible, reusable, and provider-agnostic. By following these best practices, you can create more maintainable, efficient, and powerful prompts for your LLM-powered applications.
```
willbakst, line 662: given other comments, I would update this to focus on "Mastering prompts..."

docs/learn/prompts.md
willbakst, line 1: another thing i'm realizing here is that we need to show using the `messages` field in dynamic configuration here in this doc so users know they can write messages without a template as soon as possible.

for example:
```python
from mirascope.core import BaseMessageParam, BaseDynamicConfiguration

def recommend_book_prompt(genre: str) -> BaseDynamicConfiguration:
    return {"messages": [BaseMessageParam(role="user", content=f"recommend a {genre} book")]}
```

we can either include this here or we can include it as part of the dynamic configuration docs and link it here super early.

i'm not sure what's best, but we definitely need to make this clear as soon as possible so users who don't want a prompt template don't get turned away right off the bat

willbakst: perhaps we can show this as an example of what the shorthand `@prompt_template` decorator is doing?

willbakst, line 1: after further thinking, I'm actually thinking we should just completely deprecate `BasePrompt` and instead only support the more functional approach.

this will make general maintenance easier since there will only be one path for prompting vs. two

also, i've submitted a new feature request in line with this thinking to make the functional approach better: https://github.com/Mirascope/mirascope/issues/490

docs/learn/prompts.md
```diff
@@ -10,14 +10,36 @@ Prompts are the foundation of effective communication with Large Language Models
 
 The primary means of writing prompts in Mirascope is through prompt templates, which are just formatted strings (as they should be) with a few additional conveniences. This allows you to define prompts such that they are dynamic and reusable.
 
-For the purposes of explaining how prompt templates work, we will use:
+Prompt templates in Mirascope work similarly to f-strings in Python, but with added functionality specifically designed for LLM interactions. The `@prompt_template` decorator automatically injects variables from your class or function into the template.
+
+Let's look at a basic example:
+
+```python hl_lines="1 5"
+@prompt_template("Recommend a {genre} book")
```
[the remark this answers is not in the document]
koxudaxi: I prefer 'function' style prompt. I will fix it.

docs/learn/prompts.md
```diff
@@ -333,212 +235,48 @@ print(prompt.message_params())
...
-- Prompt Versioning: Utilize the metadata decorator to keep track of different versions of your prompts as you refine them.
-- Dynamic Content: Leverage `@computed_field` for injecting dynamic content or API calls into your prompts.
-- Cached Properties: Use `@functools.cached_property` to cache frequently used properties that you only want to compute once.
-- Prompt Libraries: Build libraries of commonly used prompts that can be easily shared across projects or teams.
+By mastering prompts in Mirascope, you'll be well-equipped to build robust, flexible, and reusable LLM applications.
 
-Mastering `BasePrompt` is the first step towards building robust LLM applications with Mirascope that are flexible, reusable, and provider-agnostic.
+!!! note "Deprecation of BasePrompt"
+    We are in the process of deprecating the `BasePrompt` class in favor of the more functional approach shown in this documentation. This change will simplify maintenance and provide a more consistent API. For more information on upcoming changes, see our [issue tracker](https://github.com/Mirascope/mirascope/issues/490).
```
koxudaxi, line 282: @willbakst 
I don't know you prefer the section. But, I added it to avoid confusion of user when user see `BasePrompt` example.
What do you think about it?

docs/learn/prompts.md
```diff
@@ -125,33 +97,29 @@ If there are any other such specifiers you would find useful, let us know!
...
-print(prompt.message_params())
-# > [BaseMessageParam(role='system', content="You are the world's greatest librarian"), BaseMessageParam(role='user', content='Recommend a fantasy book')]
+print(book_recommendation_prompt("fantasy"))
+# > [
+#     BaseMessageParam(role='system', content="You are the world's greatest librarian"),
+#     BaseMessageParam(role='user', content='Recommend a fantasy book')
+#   ]
 ```
 
+In this example, we use the `SYSTEM` and `USER` keywords to specify different roles for parts of the prompt. The `message_params()` method returns a list of BaseMessageParam instances, each representing a different role and its content.
```
willbakst, line 116: this is stale as we're no longer showing `message_params()` and are instead just calling the prompt template function

docs/learn/prompts.md
```diff
@@ -333,212 +235,48 @@ print(prompt.message_params())
...
-    """Recommend a {genre} book"""
-    ...
+- **Prompt Libraries**: Build libraries of commonly used prompts that can be easily shared across projects or teams.
+   ```python
+   # prompts/book_recommendations.py
+   @prompt_template("Recommend a {genre} book by {author}")
+   def book_recommendation_prompt(genre: str, author: str): ...
 
+   # In your main code
+   from prompts.book_recommendations import author_based_recommendation
```
willbakst, line 263: nit: this function name doesn't match the defined function?

docs/learn/prompts.md
willbakst: which `BasePrompt` example?

I think the better thing to do would be to add a deprecation warning on the class and then just remove all examples and documentation that uses it?

docs/learn/prompts.md
willbakst, line 1: in this review, given that we've added the snippet below for `BaseDynamicConfig`, I'm thinking that the "Prompt Templates" section feels incomplete without a discussion around computed fields.

I'm thinking that instead of having a "Dynamic Configuration" section we should instead just discuss dynamic configuration in each section and directly show dynamically configuring that specific thing? WDYT?

For example, we would show "messages" and "computed_fields" here, we would show "tools" in the section on Tools, and we would chose "call_params" in the section on Calls (that way it's all colocated).

docs/learn/prompts.md
```diff
@@ -10,76 +10,49 @@ Prompts are the foundation of effective communication with Large Language Models
 
 The primary means of writing prompts in Mirascope is through prompt templates, which are just formatted strings (as they should be) with a few additional conveniences. This allows you to define prompts such that they are dynamic and reusable.
```
willbakst, line 11: note: we should add a second section `Messages` where we discuss writing the messages style of prompting as described in #490 

In this section, we should link to the `Messages` section with a blurb saying something like "If you are not a fan of prompt templates and want to write prompts using formatted strings directly, see the Messages section below" and add a quick link

willbakst: i would then also mention early on in this new section that you can write prompts as prompt templates with additional templating conveniences if you'd like, see the section below or something along those lines

willbakst (approved):

PR #488, opened 2024-09-12
docs: Improve calls.md

koxudaxi:


docs/learn/calls.md
```diff
@@ -70,7 +78,7 @@ Mirascope's call decorator supports multiple LLM providers, allowing you to easi
 
 To use a specific provider, simply use the `call` decorator from the corresponding provider's module. Here's an example of how to use multiple different providers with the same function:
```
willbakst, line 79: nit: we should update the last bit "the same function" to "the same prompt template"

docs/learn/calls.md
```diff
@@ -174,6 +184,8 @@ response = recommend_book("fantasy")
 print(response.content)
 ```
 
+In this example, we're creating a custom OpenAI client with a specific API key, organization ID, and base URL. This custom client is then passed to the call decorator, allowing us to use a custom endpoint or configuration for our LLM calls.
+
 Any custom client is supported so long as it has the same API as the original base client.
 
 ## Handling Responses
```
willbakst, line 191: nit: as mentioned in the review on the "Prompts" docs, things like "metadata" and the link will need to be updated to point to the new section since it'll no longer live in the prompts docs.

docs/learn/calls.md
```diff
@@ -283,13 +295,133 @@ In this example, your IDE will provide proper autocompletion and type checking f
...
+   For more advanced retry logic, check out our documentation on [using Tenacity](../integrations/tenacity.md).
+
+- **Streaming for Long Tasks**: Utilize streaming for long-running tasks or when providing real-time updates to users, improving the user experience.
+   ```python
+   @openai.call("gpt-4o-mini", stream=True)
+   @prompt_template("Write a short story about {topic}")
+   def write_story(topic: str):
+       ...
+
+   for chunk in write_story("a space adventure"):
```
willbakst, line 355: nit: this is missing the ignore for the second tool tuple item, should be `for chunk, _ in write_story(...)`

willbakst: also, i would keep this in line with the book recommendation theme, so maybe instead do something like "give me a summary of this book"?

docs/learn/calls.md
```diff
@@ -283,13 +295,133 @@ In this example, your IDE will provide proper autocompletion and type checking f
...
+   This approach allows you to use open-source models by serving OpenAI compatible endpoints (e.g., with `vLLM`, `Ollama`, etc).
+
+- **Structured Outputs**: Use `response_models` to ensure consistency and type safety when you need structured data from your LLM calls.
+   ```python
+   from pydantic import BaseModel
+
+   class BookRecommendation(BaseModel):
+       title: str
+       author: str
+       reason: str
```
willbakst, line 404: nit: drop reason, not sure it adds anything here except more text

docs/learn/calls.md
willbakst, line 1: in reference to my review of the "Prompts" documentation, we should update this to include things from those docs that should live here (e.g. the `run` method as an alternative to the functional prompt template approach etc.)

docs/learn/calls.md
```diff
@@ -48,15 +50,20 @@ response = recommend_book("fantasy")
...
+In this example:
+
+- The [`@prompt_template`](./prompts.md#prompt-templates) decorator turns the function into a prompt template.
+- The `@openai.call` decorator transforms the prompt template into an LLM API call.
+- We're using OpenAI's `gpt-4o-mini` model to make the call.
+- When we call the function with `genre="fantasy"`, the argument is automatically injected into the prompt template and used in the call.
 
 !!! info "Function Body"
 
     In the above example, we've used an ellipsis (`...`) for the function body, which returns `None`. If you'd like, you can always explicitly `return None` to be extra clear. For now, you can safely ignore how we use the function body, which we cover in more detail in the documentation for [dynamic configuration](./dynamic_configuration.md).
```
willbakst, line 62: nit: as i mentioned in my recent review of the "Prompts" documentation, we should drop the "Dynamic Configuration" page and just discuss it in each respective section.

this means here we should discuss dynamic call parameters in the section on call parameters, and then we should mention that section here instead of the original dynamic configuration docs.

thoughts?

docs/learn/calls.md
```diff
@@ -174,8 +183,36 @@ response = recommend_book("fantasy")
 print(response.content)
 ```
 
+In this example, we're creating a custom OpenAI client with a specific API key, organization ID, and base URL. This custom client is then passed to the call decorator, allowing us to use a custom endpoint or configuration for our LLM calls.
+
 Any custom client is supported so long as it has the same API as the original base client.
 
+## Using the `run` Method
```
willbakst, line 190: this is stale, there is no longer a `run` method since we're deprecating `BasePrompt`

docs/learn/calls.md
```diff
@@ -209,7 +246,7 @@ All `BaseCallResponse` objects share these common properties:
 - `message_param`: The assistant's response formatted as a message parameter.
 - `tools`: A list of provider-specific tools used in the response, if any. Otherwise this will be `None`. Check out the [`Tools`](./tools.md) documentation for more details.
 - `tool`: The first tool used in the response, if any. Otherwise this will be `None`. Check out the [`Tools`](./tools.md) documentation for more details.
-- `metadata`: Any metadata provided using the [`@metadata`](./prompts.md#metadata) decorator.
+- `metadata`: Any metadata provided using the [`@metadata`](./metadata.md) decorator.
```
willbakst, line 249: this link is broken? should point to the section in this doc?

docs/learn/calls.md
```diff
@@ -174,8 +183,36 @@ response = recommend_book("fantasy")
...
+# Using Anthropic
+anthropic_response = recommend_book.run(anthropic.call(model="claude-3-5-sonnet-20240620"))("fantasy")
+print("Anthropic recommendation:", anthropic_response.content)
+```
+
+This approach allows you to easily switch between providers or compare outputs without changing your core prompt logic.
+
+## Metadata
+
+You can add metadata to your calls using the `@metadata` decorator. This can be useful for tracking, categorizing, or adding any additional information to your LLM calls. For more information on using metadata, see the [Metadata documentation](./metadata.md).
```
willbakst, line 214: what are your thoughts on deprecating the `@metadata` decorator and instead only supporting metadata through dynamic configuration?

i feel like forcing this single path will be a little more clear

koxudaxi: Looks good.
I will remove the section and I will change the link to dynamic configuration markdown.

docs/learn/calls.md
```diff
@@ -146,13 +153,43 @@ response = recommend_book("fantasy")
...
+    }
+
+
+response = recommend_book("fantasy")
+print(response.content)
+print(f"Temperature used: {response.call_kwargs['temperature']}")
+print(f"Metadata: {response.metadata}")
+```
+
+In this example, we're dynamically setting the `temperature` based on the genre, and also adding metadata to the call. The `OpenAIDynamicConfig` (or the equivalent for other providers) allows you to specify various dynamic configurations, including `call_params`, `metadata`, `computed_fields`, and more.
```
willbakst, line 182: should we link the computed fields section in the prompts documentation here?

docs/learn/calls.md
```diff
@@ -22,20 +22,22 @@
 
 The `call` decorator is a core feature of the Mirascope library, designed to simplify and streamline interactions with various Large Language Model (LLM) providers. This powerful tool allows you to transform Python functions into LLM API calls with minimal boilerplate code while providing type safety and consistency across different providers.
 
-The primary purpose of the call decorator is to:
+## Purpose and Benefits
```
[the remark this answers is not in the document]
willbakst: also please let me know if you'd like me to write this, as i'm more than happy to do that :)

docs/learn/calls.md
```diff
@@ -146,13 +153,43 @@ response = recommend_book("fantasy")
 print(response.content)
 ```
 
+In this example, we're setting the `temperature` parameter to 0.7, which affects the randomness of the model's output. A higher temperature (closer to 1.0) will result in more diverse and creative outputs, while a lower temperature (closer to 0.0) will make the outputs more focused and deterministic.
+
+### Dynamic Configuration
```
willbakst, line 158: should we add a blurb here mentioning the difference between using `BaseDynamicConfig` and provider specific versions like `openai.OpenAIDynamicConfig`?

seems worth mentioning that you can always use provider-agnostic configuration options (such as computed fields and metadata) but that you can also tie the prompt to a specific provider by using the provider specific config (and also then get provider specific type hints for things like call params but also messages etc.).

thoughts?

docs/learn/calls.md
```diff
@@ -132,7 +139,7 @@ For all providers, we have only included additional call parameters that are not
```
willbakst, line 139: nit: the api documentation above is missing links to the new providers we added (which also likely means that the providers call params are missing the learn docs link in their docstrings)

docs/learn/calls.md
```diff
@@ -174,6 +211,8 @@ response = recommend_book("fantasy")
 print(response.content)
 ```
 
+In this example, we're creating a custom OpenAI client with a specific API key, organization ID, and base URL. This custom client is then passed to the call decorator, allowing us to use a custom endpoint or configuration for our LLM calls.
+
 Any custom client is supported so long as it has the same API as the original base client.
 
 ## Handling Responses
```
willbakst, line 218: nit: should we link `BaseCallResponse` here to the API reference as the first mention?

docs/learn/calls.md
```diff
@@ -283,13 +320,131 @@ In this example, your IDE will provide proper autocompletion and type checking f
...
+To make the most of Mirascope's `call` decorator and ensure you're building robust, efficient, and maintainable LLM applications, consider the following best practices:
+
+- **Provider Flexibility**: Design functions to be provider-agnostic, allowing easy switching between different LLM providers for comparison or fallback strategies.
+   ```python
+   @prompt_template("Recommend a {genre} book")
+   def recommend_book(genre: str):
+       ...
+
+   # Can easily switch between providers
+   openai_response = recommend_book.run(openai.call("gpt-4o-mini"))("fantasy")
```
willbakst, line 332: the method doesn't have a `run` method, it should instead be `openai.call("gpt-4o-mini")(recommend_book)("fantasy")`

same for the anthropic line below

docs/learn/calls.md
```diff
@@ -283,13 +320,131 @@ In this example, your IDE will provide proper autocompletion and type checking f
...
+       base_url="https://your-custom-endpoint.com/v1"
+   )
+
+   @openai.call("gpt-4o-mini", client=custom_client)
+   @prompt_template("Recommend a {genre} book")
+   def recommend_book(genre: str):
+       ...
+   ```
+
+   This approach allows you to use open-source models by serving OpenAI compatible endpoints (e.g., with `vLLM`, `Ollama`, etc).
```
willbakst, line 420: nit: we should add links to vLLM and Ollama

docs/learn/calls.md
```diff
@@ -116,23 +224,29 @@ These common parameters provide a consistent way to control the behavior of LLM
...
 
+    [`mirascope.core.azure.call_params`](../api/core/anthropic/azure.md)
+
     [`mirascope.core.cohere.call_params`](../api/core/cohere/call_params.md)
 
     [`mirascope.core.gemini.call_params`](../api/core/gemini/call_params.md)
 
     [`mirascope.core.groq.call_params`](../api/core/groq/call_params.md)
 
+    [`mirascope.core.litellm.call_params`](../api/core/litellm/call_params.md)
```
willbakst, line 235: nit: litellm doesn't have call params, it uses openai call params, so link the openai ones

willbakst (approved):

Issue #490, opened 2024-09-12
More Flexible (Better) Prompt Engineering Support


willbakst:
### Description

Right now if you don't want to use `@prompt_template` you have to use the provider-specific message parameters or `BaseMessageParam`.

We need to make writing messages in this way (rather than as templates) more of a first-class citizen. Here are a few ideas:

## `Messages` Return Type

I'm imagining the ability to return a more direct type rather than `{"messages": ...}` so that the new user does not necessarily have to learn about "dynamic configuration" in order to access custom messages.

For all of the below form factors, you could add `@prompt_template` so that the function returns `list[BaseMessageParam]`

**`str` return type**

```python
from mirascope.core import Messages

def recommend_book(genre: str) -> Messages:
    return f"recommend a {genre} book"

print(recommend_book("fantasy"))
# > [BaseMessageParam(role='user', content='recommend a fantasy book')]
```

**`list[str]` return type**

```python
from mirascope.core import Messages

def recommend_book(genre: str) -> Messages:
    return ["hello!", f"recommend a {genre} book"]

print(recommend_book("fantasy")
# > [BaseMessageParam(role='user', content=[TextPart(type='text', text='hello!'), TextPart(type='text', text='recommend a fantasy book')]
```

**`list[Message.{ROLE}]` return type**

```python
from mirascope.core import Messages

def recommend_book(genre: str) -> Messages:
    return [Messages.System("You are a librarian"), Messages.User(f"recommend a {genre} book")]
```

**type coersion in `BaseMessageParam`**

```python
from mirascope.core import Messages

def recommend_book(genre: str) -> Messages:
   # could wrap in list or not? both valid?
    return BaseMessageParam(role='user', content=["hello!", f"recommend a {genre} book"])
```

**multi-modal type coercion**

```python
from mirascope.core import Messages
from PIL import Image

def recommend_book(previous_book: Image.Image) -> Messages:
    return ["I just read this book:", previous_book, "What should I read next?"]
    # return BaseMessageParam(role='user', content=["I just read this book:", previous_book, "What should I read next?"])
```

We should allow as much mixing/matching across all the various described conveniences as possible.

willbakst:
Some additional clarification.

**Typing**

In order to put everything under `Messages`, we'll need to make a minor change to the design and return `Messages.Type` instead:

```python
from mirascope.core import Messages

def recommend_book(genre: str) -> Messages.Type:
    return f"recommend a {genre} book"
```

[...]

PR #492, opened 2024-09-13
feat: Implement flexible prompt engineering support

koxudaxi:
Closes: https://github.com/Mirascope/mirascope/issues/490


mirascope/core/base/_utils/_parse_content_template.py
```diff
@@ -147,17 +147,25 @@ def _construct_parts(
...
+            for template_part in template
+            for part in _parse_parts(template_part)
+            for item in _construct_parts(part, attrs)
+        ]
+    else:
+        parts = [
+            item
+            for part in _parse_parts(template)
+            for item in _construct_parts(part, attrs)
+        ]
```
koxudaxi, line 168: I would like to reuse this logic, is it reasonable?

mirascope/core/base/message_param.py
```diff
@@ -74,3 +74,29 @@ class BaseMessageParam(BaseModel):
 
     role: str
     content: str | list[TextPart | ImagePart | AudioPart | CacheControlPart]
+
+
+Messages: TypeAlias = list[BaseMessageParam]
```
koxudaxi, line 79: Do you expect more complex type? `union` or `pydantic.RootModel`?

mirascope/core/base/message_param.py
```diff
@@ -74,3 +74,29 @@ class BaseMessageParam(BaseModel):
...
+
+
+class Message:
```
koxudaxi, line 82: `SystemMessage` would be more appropriate, but is the idea of `Messages.System` from accessibility?

mirascope/core/base/message_param.py
```diff
@@ -74,3 +74,29 @@ class BaseMessageParam(BaseModel):
...
+    class System(BaseMessageParam):
+        def __init__(
+            self,
+            content: str | list[TextPart | ImagePart | AudioPart | CacheControlPart],
+        ) -> None:
+            super().__init__(content=content, role="system")
```
koxudaxi, line 88: I did this to avoid specifying a role, but validator might be better.

mirascope/core/base/prompt.py
```diff
@@ -307,18 +309,24 @@ async def run():
 _BasePromptT = TypeVar("_BasePromptT", bound=BasePrompt)
 _MessageParamT = TypeVar("_MessageParamT", bound=Any)
 _BaseDynamicConfigT = TypeVar("_BaseDynamicConfigT", bound=BaseDynamicConfig)
+_MessageFuncReturnTypes: TypeAlias = (
+    str | BaseMessageParam | list[BaseMessageParam] | list[Any]
+)
```
koxudaxi, line 314: If there are other return values that should be supported, add them here.

mirascope/core/base/prompt.py
```diff
@@ -307,18 +309,24 @@ async def run():
...
+    "_MessageFuncReturnT", bound=_MessageFuncReturnTypes, contravariant=True
+)
 
 
 class PromptDecorator(Protocol):
     @overload
     def __call__(
         self, prompt: Callable[_P, BaseDynamicConfig]
-    ) -> Callable[_P, list[BaseMessageParam]]: ...
+    ) -> Callable[_P, Messages]: ...
```
koxudaxi, line 325: Checked if `list[BaseMessageParam]` and `Messages ` can be used to mean the same thing.

mirascope/core/base/prompt.py
```diff
@@ -345,7 +351,138 @@ def _is_function(
...
+    if isinstance(value, list):
+        if all(isinstance(item, BaseMessageParam) for item in value):
+            return value
+        if all(isinstance(item, str) for item in value):
+            if parsed_messages := parse_content_template("user", value, attrs):
+                return [parsed_messages]
+            return []
+        # How to detect non-text types?
+
+    raise ValueError(f"Invalid message type: {type(value)} for value: {value}")
```
koxudaxi, line 409: My understanding is that it is difficult to determine the type of content (`previous_book `) from the examples you raised.
How can I make a decision other than `str`?
```python
def recommend_book(previous_book: Image.Image) -> Messages:
    return ["I just read this book:", previous_book, "What should I read next?"]
    # return BaseMessageParam(role='user', content=["I just read this book:", previous_book, "What should I read next?"])
```

mirascope/core/base/prompt.py
```diff
@@ -421,18 +564,18 @@ async def get_base_message_params_async(
...
+                    template=template_or_message_function,
                     attrs=get_fn_args(prompt, args, kwargs),
                     dynamic_config=prompt(*args, **kwargs),
                 )
 
             get_base_message_params._original_fn = prompt  # pyright: ignore [reportAttributeAccessIssue,reportFunctionMemberAccess]
             return get_base_message_params
 
-    return inner
+    return inner  # pyright: ignore [reportReturnType]
```
koxudaxi, line 578: pyright does not consider `if` branches

tests/core/base/test_prompt.py
```diff
@@ -229,3 +234,115 @@ class BookRecommendationPrompt(BasePrompt): ...
     def fn() -> None: ...
 
     assert hasattr(fn, "_metadata") and fn._metadata == {"tags": {"version:0001"}}  # pyright: ignore [reportFunctionMemberAccess]
+
+
+def test_prompt_template_str_return():
+    @prompt_template
+    def recommend_book(genre: str) -> str:
+        return f"recommend a {genre} book"
```
koxudaxi, line 242: I don't know if this is your preference, but here it would be correct to define ` -> str`, create a function and have the `@prompt_template` decorator change the return value of the function.
This will allow pyright and the IDE to do type checking and completion correctly.

tests/core/base/test_prompt.py
```diff
@@ -229,3 +234,115 @@ class BookRecommendationPrompt(BasePrompt): ...
...
+#
+#     result = recommend_book(dummy_image)
+#     assert isinstance(result, list)
+#     assert len(result) == 1
+#     assert isinstance(result[0], BaseMessageParam)
+#     assert result[0].role == "user"
+#     assert len(result[0].content) == 3
+#     assert result[0].content[0] == "I just read this book:"
+#     assert isinstance(result[0].content[1], Image.Image)
+#     assert result[0].content[2] == "What should I read next?"
```
koxudaxi, line 348: As I wrote in the implementation, I think it is difficult to specify the type of content from `previous_book`. It might be possible to read the magic number at the beginning of the file, but to what extent do you assume this is possible?

mirascope/core/base/prompt.py
```diff
@@ -372,15 +509,23 @@ def recommend_book(genre: str):
...
     """
 
+    if template_or_message_function is None:
+        # For @prompt_template() case
+        return _messages_decorator()
+
+    if _is_message_param_function(template_or_message_function):
+        # For @prompt_template case
+        return _messages_decorator()(template_or_message_function)
+
```
koxudaxi, line 519: Two cases can be supported: `@prompt_template()` and `@prompt_template`.

tests/core/base/test_prompt.py
willbakst: we already have helper utilities for determining this in base that you can use to determine the file type from the raw bytes

mirascope/core/base/prompt.py
willbakst: instead of refactoring and supporting both cases, should we just write the single case here where they write `@prompt_template()` and then type it to expect the `Messages` return type?

mirascope/core/base/prompt.py
willbakst: curious how this wasn't throwing an error before?

tests/core/base/test_prompt.py
willbakst: hmm, ideally i want `def recommend_book(genre: str) -> Messages` in all cases unless we think this is more clear?

mirascope/core/base/prompt.py
koxudaxi: It is possible that you and I have different ideas about the return type of a function.
I am assuming that list[Any] is the following case in the example you provided.
Obviously, `Image.Image` does not have a `BaseMessageParam`, so it allows Any.
In other words, the type you are defining here is the type that the function returns in `actual`, not the type after the decorator has made the change.

```python
@prompt_template
def recommend_book(previous_book: Image.Image) -> list[Any]:
    return ["I just read this book:", previous_book, "What should I read next?"]
```

The reason this type definition is necessary is so that the decorator can type-check all target functions that it is allowed to accept

tests/core/base/test_prompt.py
koxudaxi: I understand your intention, but wouldn't it be confusing from a type checking and usability standpoint for an undecorated function to return the decorated form (`Messages`)?

willbakst: what do you mean by decorated form? I view `Messages` as the undecorated form, and `list[BaseMessageParam]` as the decorated return type

mirascope/core/base/prompt.py
willbakst: but the `prompt_template` decorator should require that the function it decorates returns the `Messages` type, which can be either `str`, or `Sequence[str, Image.Image]`, or `list[BaseMessageParam]`, or some form of `BaseDynamicConfig`.

the decorated function will then always have return type `list[BaseMessageParam]` or `BaseDynamicConfig` right?

mirascope/core/base/_utils/_parse_content_template.py
koxudaxi: reverted

mirascope/core/base/prompt.py
koxudaxi: removed

tests/core/base/test_prompt.py
koxudaxi: fixed

mirascope/core/base/_utils/_convert_messages_to_message_params.py
```diff
@@ -0,0 +1,92 @@
...
+    return [
+        _convert_message_sequence_part_to_content_part(message)
+        for message in message_sequence
+    ]
+
+
+def _is_base_message_params(
+    value: object,
+) -> TypeIs[list[BaseMessageParam]]:
+    return isinstance(value, list) and isinstance(value[0], BaseMessageParam)
```
willbakst, line 69: do we not need to check all, why just the first object?

mirascope/core/base/_utils/_get_dynamic_configuration.py
```diff
@@ -1,39 +1,54 @@
...
+from typing import Any, TypeVar, cast
+
+from typing_extensions import TypeIs
 
 from ..dynamic_config import BaseDynamicConfig
+from ..message_param import BaseMessageParam
 
 _BaseDynamicConfigT = TypeVar(
     "_BaseDynamicConfigT", bound=BaseDynamicConfig, covariant=True
 )
```
willbakst, line 13: why do we need both the sync and async methods here when we didn't before?

mirascope/core/base/_utils/_convert_messages_to_message_params.py
```diff
@@ -0,0 +1,92 @@
...
+            type="image",
+            media_type=f"image/{get_image_type(image)}",
+            image=image,
+            detail=None,
+        )
+    else:
+        raise ValueError(f"Invalid message sequence type: {message_sequence_part}")
+
+
+def get_content_from_message(
```
willbakst, line 49: i don't understand why this is a separate method from `_convert_message_sequence_to_content`

shouldn't we just have a single function named something like `convert_message_content_to_message_param_content`?

mirascope/core/base/_utils/_convert_messages_to_message_params.py
```diff
@@ -0,0 +1,92 @@
...
+    if isinstance(messages, str):
+        return [BaseMessageParam(content=messages, role=role)]
+    elif isinstance(messages, BaseMessageParam):
+        return [messages]
+    elif _is_base_message_params(messages):
+        return messages
+    elif isinstance(messages, Sequence):
+        return [
+            BaseMessageParam(
+                content=_convert_message_sequence_to_content(messages), role=role
```
willbakst, line 88: wrt. comment before, content here can be a `str` type so no need for a separate method, we can just use the merged `convert_message_content_to_message_param_content`

mirascope/core/base/messages.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from . import AudioPart, BaseMessageParam, ImagePart, TextPart
+from ._utils._convert_messages_to_message_params import (
+    Image,
+    get_content_from_message,
```
willbakst, line 8: see earlier comment about merging functions, which would then be used here

mirascope/core/base/prompt.py
```diff
@@ -323,13 +337,13 @@ def __call__(
         | Callable[_P, BaseDynamicConfig]
         | Callable[_P, Awaitable[BaseDynamicConfig]],
     ) -> (
-        Callable[_P, list[BaseMessageParam]]
-        | Callable[_P, Awaitable[list[BaseMessageParam]]]
+        Callable[_P, str | list[BaseMessageParam]]
```
willbakst, line 340: i'm confused by this typing, wouldn't the decorator return only `list[BaseMessageParam]` or `Awaitable[list[BaseMessageParam]]`? In what case does it return `str`?

mirascope/core/base/stream.py
```diff
@@ -285,7 +288,7 @@ def decorator(
             @wraps(fn)
             async def inner_async(*args: _P.args, **kwargs: _P.kwargs) -> TStream:
                 fn_args = get_fn_args(fn, args, kwargs)
-                dynamic_config = await get_dynamic_configuration(fn, args, kwargs)
+                dynamic_config = await get_dynamic_configuration_async(fn, args, kwargs)
```
willbakst, line 291: same thing here re: sync vs. async

mirascope/core/base/_create.py
```diff
@@ -93,7 +94,7 @@ async def inner_async(
                 *args: _P.args, **kwargs: _P.kwargs
             ) -> TCallResponse | _ParsedOutputT:
                 fn_args = get_fn_args(fn, args, kwargs)
-                dynamic_config = await get_dynamic_configuration(fn, args, kwargs)
+                dynamic_config = await get_dynamic_configuration_async(fn, args, kwargs)
```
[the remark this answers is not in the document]
koxudaxi: If I understand correctly, await can only be called in the async function, so I split the function into two

mirascope/core/base/_utils/_get_dynamic_configuration.py
koxudaxi: In the previous get_dynamic_configuration
The result value of `fn` was just returned by calling `fn`, so there was no need to await inside get_dynamic_configuration. However, since it is necessary to check the return type of `fn` with `isinstance` in this PR, it is necessary to do await.
Therefore, async was prepared to enable await.

mirascope/core/base/prompt.py
```diff
@@ -339,7 +353,90 @@ def _is_function(
...
+        self,
+        messages_fn: MessagesSyncFunction[_P, _MessageFuncReturnT]
+        | MessagesAsyncFunction[_P, _MessageFuncReturnT],
+    ) -> (
+        Callable[_P, list[BaseMessageParam] | BaseDynamicConfig]
+        | Callable[_P, Awaitable[list[BaseMessageParam] | BaseDynamicConfig]]
+    ): ...
+
+
+def _messages_decorator() -> MessagesDecorator:
```
willbakst, line 383: should we move this to it's own utility file like other utilities?

mirascope/core/base/stream.py
koxudaxi: I commented before that.

mirascope/core/base/_create.py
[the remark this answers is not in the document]
willbakst: but `get_dynamic_configuration` can return an awaitable (i.e. by not calling await on the function) and defer calling `await` to the caller (which would be here where we're inside an async function). Does this make sense?

mirascope/core/base/_utils/_messages_decorator.py
willbakst, line 1: should we move the tests out from `test_prompt` into a specific utils test for this refactored code?

tests/core/base/_utils/test_convert_messages_to_message_params.py
willbakst, line 1: we should move the tests for the messages decorator into a matching file for that utility and only test here that we properly call that method for the `Messages.Type` return type, right?

mirascope/core/base/_create.py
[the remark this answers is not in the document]
koxudaxi: Since the return value is expected to be `_BaseDynamicConfigT`, it is awaiting inside `get_dynamic_configuration_async` and returns list[BaseMessageParam] wrapped in dict like this.
If you don't await, the function returns `_BaseDynamicConfigT | list[BaseMessageParam` type.
```python
) -> _BaseDynamicConfigT:
...
    if _is_message_param_list(result):
        return cast(_BaseDynamicConfigT, {"messages": result})
```

mirascope/core/base/_utils/_get_dynamic_configuration.py
willbakst: oooh i see ok that makes sense.

do you think it's better to have these separate functions, or should we instead just define an async function internally and return calling that? for example:

```python
def get_dynamic_configuration(
    fn: Callable[..., _BaseDynamicConfigT | Awaitable[_BaseDynamicConfigT]],
    args: tuple[Any, ...],
    kwargs: dict[str, Any],
) -> _BaseDynamicConfigT | Awaitable[_BaseDynamicConfigT]:
    result = (
        fn._original_fn(*args, **kwargs)  # pyright: ignore [reportFunctionMemberAccess]
        if hasattr(fn, "_original_fn")
        else fn(*args, **kwargs)
    )
    if inspect.isawaitable(result):

        async def inner_async() -> Awaitable[_BaseDynamicConfigT]:
            result = await result
            if _is_message_param_list(result):
                return cast(_BaseDynamicConfigT, {"messages": result})
            return result

        return inner_async()

    if _is_message_param_list(result):
        return cast(_BaseDynamicConfigT, {"messages": result})
    return result
```

this way consuming the utility is easy and only ever requires a single import (rather than having to import both the sync and async versions), and then the overloads provide proper type hints as before

mirascope/core/base/_create.py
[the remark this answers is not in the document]
willbakst: see my other comment on this

mirascope/core/base/_utils/_get_dynamic_configuration.py
koxudaxi: I tried the your code. But, the `result` is not resolved in the scope 🤔 

```python

    async def inner_async() -> Awaitable[_BaseDynamicConfigT]:
>       result = await result
E       UnboundLocalError: cannot access local variable 'result' where it is not associated with a value

../mirascope/core/base/_utils/_get_dynamic_configuration.py:30: UnboundLocalError
```

koxudaxi: OK, It works fine
```python
            async_result = await result
            if _is_message_param_list(result):
                return cast(_BaseDynamicConfigT, {"messages": result})
            return async_result
```

tests/core/base/_utils/test_convert_messages_to_message_params.py
koxudaxi: sorry, I don't know why you mentioned `messages decorator` here?
in the file, we don't import `message decorator`

PR #499, opened 2024-09-17
docs: Improve streams.md

koxudaxi:


docs/learn/streams.md
```diff
@@ -16,15 +16,51 @@
```
willbakst, line 16: nit: above reference doesn't have all providers, which means the providers also likely don't have backlinks, please add them

docs/learn/streams.md
```diff
@@ -36,9 +72,24 @@ def recommend_book(genre: str):
...
+
+!!! info "Supported Providers"
+
+    Mirascope supports standard streaming (without tools) for all supported providers. While we use OpenAI in this example, the interface is the same across all supported providers.
+
+!!! tip "Streaming with Other Providers"
+
+    To use streaming with other providers, simply replace `openai.call` with the appropriate provider's call decorator (e.g., `anthropic.call`, `gemini.call`, etc.).
 
 ## Handling Streamed Responses
```
willbakst, line 95: same thing here re: api and backlinks not having all providers

docs/learn/streams.md
```diff
@@ -120,21 +178,106 @@ except OpenAIError as e:
...
+
+```python
+@openai.call(model="gpt-4o-mini", stream=True)
+@prompt_template("Recommend a {genre} book")
+def recommend_book(genre: str) -> openai.OpenAIStream:
+    ...
 
-Mirascope's `call` decorator provides proper type hints and safety when working with streams. When you enable streaming, the return type of your function will accurately reflect the stream type.
+stream = recommend_book("fantasy")
+# Your IDE will provide autocompletion for stream methods and properties
```
willbakst, line 198: should we include a screenshot here like we do in the calls docs?

docs/learn/streams.md
```diff
@@ -120,21 +178,106 @@ except OpenAIError as e:
...
+   @openai.call(model="gpt-4o-mini", stream=True)
+   @prompt_template("Write a short story about {topic}")
+   def write_story(topic: str):
+       ...
+
+   print("Generating story...")
+   for chunk, _ in write_story("a magical forest"):
+       print(chunk.content, end="", flush=True)
+   # > Generating story...
+   # Once upon a time, in a magical forest...
```
willbakst, line 219: nit: match the indentation here with the generating story line above

docs/learn/streams.md
```diff
@@ -120,21 +178,106 @@ except OpenAIError as e:
...
+
+   @openai.call(model="gpt-4o-mini", stream=True)
+   @prompt_template("Summarize the following text: {text}")
+   def summarize_text(text: str):
+       ...
+
+   text = "..." # Long text to summarize
+   with tqdm(total=100, desc="Summarizing") as pbar:
+       for i, (chunk, _) in enumerate(summarize_text(text)):
+           print(chunk.content, end="", flush=True)
```
willbakst, line 235: it seems weird to me that we're printing and updating the progress bar?

wouldn't it be better to stream, capture the text, update the progress bar, and then print the final amount when done?

also, how do we know the total is `100` here when we have no idea how many chunks there might be? honestly not sure how to implement a progress bar like this, might need a little more thought here

docs/learn/streams.md
```diff
@@ -120,21 +178,106 @@ except OpenAIError as e:
...
+
+   ```python
+   @openai.call(model="gpt-4o-mini", stream=True)
+   @prompt_template("Generate a list of {n} random words")
+   def generate_words(n: int):
+       ...
+
+   word_count = 0
+   for chunk, _ in generate_words(1000):
+       words = chunk.content.split()
```
willbakst, line 249: what happens when there is a long word and it's never split? seems like we would increate the word count by `1` each time but we would still be on the same word as it streams?

docs/learn/streams.md
```diff
@@ -120,21 +178,106 @@ except OpenAIError as e:
...
+               print(chunk.content, end="", flush=True)
+       except TimeoutError:
+           print("\nGeneration timed out")
+
+   asyncio.run(stream_with_timeout("artificial intelligence", 30.0))
+   ```
+
+By leveraging streaming effectively, you can create more responsive and efficient LLM-powered applications with Mirascope's streaming capabilities.
+
+## Conclusion
```
willbakst, line 279: nit: we can drop this header as it doesn't really add any value

willbakst (approved):

PR #502, opened 2024-09-18
docs: Improve async.md

koxudaxi:


docs/learn/async.md
```diff
@@ -1,68 +1,74 @@
...
-
+from mirascope.core import openai, Messages, prompt_template
 
 @openai.call(model="gpt-4o-mini")
-@prompt_template("Recommend a {genre} book")
-async def recommend_book(genre: str):
-    ...
-
+@prompt_template()
+async def recommend_book_prompt(genre: str) -> Messages.Type:
```
willbakst, line 21: nit: rename method to just `recommend_book` since it's making a call?

docs/learn/async.md
```diff
@@ -1,68 +1,74 @@
...
     print(response.content)
 
-
 asyncio.run(main())
 ```
 
-In this example, the `recommend_book` function is defined as async, and we use `await` when calling it within another async function. You can handle asynchronous calls when using `response_model` or `output_parser` the exact same way.
+This code does the following:
+
+1. Defines an asynchronous function `recommend_book_prompt` that uses the OpenAI model to recommend a book of a given genre.
```
willbakst, line 33: make sure to update the method naming here as well

docs/learn/async.md
```diff
@@ -72,112 +78,115 @@ class FormatBook(BaseTool):
...
 asyncio.run(main())
 ```
 
-!!! tip "Type Hints"
+This example:
 
-    When using and calling asynchronous tools, you can check the type of the tool to get proper type hints.
+1. Defines an asynchronous tool `FormatBook` with an async `call` method.
+2. Uses this tool in an asynchronous OpenAI call.
+3. In the main function, it checks if the response includes a tool call, and if so, calls the tool asynchronously.
```
willbakst, line 101: we should mention that we check the instance type of the tool to ensure awaiting the tool call is correct

docs/learn/async.md
```diff
@@ -72,112 +78,115 @@ class FormatBook(BaseTool):
...
 
 ## Async with BasePrompt
```
willbakst, line 103: stale? we should drop this?

willbakst (approved):

PR #503, opened 2024-09-18
docs: Improve response_models.md

koxudaxi:


docs/learn/response_models.md
```diff
@@ -24,9 +24,9 @@ class Book(BaseModel):
 
 
 @openai.call(model="gpt-4o-mini", response_model=Book)
-@prompt_template("Recommend a {genre} book")
-def recommend_book(genre: str):
-    ...
+@prompt_template()
+def recommend_book(genre: str) -> str:
```
willbakst, line 28: nit: should return `Messages.Type`?

docs/learn/response_models.md
```diff
@@ -35,33 +35,38 @@ print(f"Title: {book.title}")
 print(f"Author: {book.author}")
 ```
 
+In this example:
+
+1. We define a `Book` class that inherits from Pydantic's `BaseModel`.
+2. We use the `@openai.call` decorator with the `response_model` parameter set to our `Book` class.
+3. The `recommend_book` function is decorated with both `@openai.call` and `@prompt_template`.
```
willbakst, line 42: this explanation is a little confusing.

seems like we should switch the order? i.e. (1) we write a prompt template, (2) we wrap it in the call decorator, and (3) we set the response model to `Book`?

docs/learn/response_models.md
```diff
@@ -35,33 +35,38 @@ print(f"Title: {book.title}")
...
-    ```python
+    ```python hl_lines="2 4"
     # original `OpenAICallResponse`
     response = cast(openai.OpenAICallResponse, book._response)  # pyright: ignore [reportAttributeAccessIssue]
     # original `ChatCompletion`
     completion = response.response
     ```
 
-### Extracting Built-in Types
+    Note that this is a private attribute included for access to the original response and is not an official property of the response model type.
```
willbakst, line 59: nit: this is duplicated from the above explanation before the code snippet in the note?

docs/learn/response_models.md
```diff
@@ -35,33 +35,38 @@ print(f"Title: {book.title}")
...
 
+For simpler cases where you want to extract just a single built-in type, Mirascope provides a shorthand:
 
+```python hl_lines="1 7"
 @openai.call("gpt-4o-mini", response_model=list[str])
-@prompt_template("Recommend a list of {num} {genre} books")
-def recommend_books(genre: str, num: int):
-    ...
+@prompt_template()
+def recommend_books(genre: str, num: int) -> str:
```
willbakst, line 68: same thing here, should return `Messages.Type`?

docs/learn/response_models.md
```diff
@@ -95,15 +102,22 @@ class Book(BaseModel):
 
 
 @openai.call("gpt-4o-mini", response_model=Book, json_mode=True)
-@prompt_template("Recommend a {genre} book. Match example format.")
-def recommend_book(genre: str): ...
+@prompt_template()
+def recommend_book(genre: str) -> str:
```
willbakst, line 106: nit: ditto on type return

docs/learn/response_models.md
```diff
@@ -123,9 +137,9 @@ class Book(BaseModel):
 
 
 @openai.call(model="gpt-4o-mini", response_model=Book, stream=True)
-@prompt_template("Recommend a {genre} book")
-def recommend_book(genre: str):
-    ...
+@prompt_template()
+def recommend_book(genre: str) -> str:
```
willbakst, line 141: nit: ditto

docs/learn/response_models.md
```diff
@@ -156,6 +170,33 @@ In the above example, your IDE will provide proper autocompletion and type check
 
 When using Response Models, the default setting is `json_mode=False`, which will use [`Tools`](./tools.md) under the hood to perform the structured output extraction. However, for some providers setting `json_mode=True` enables support for field types that may not be supported by tools.
 
+```python hl_lines="8"
+from mirascope.core import openai, prompt_template
+from pydantic import BaseModel
+
+class ComplexBook(BaseModel):
```
willbakst, line 177: what makes this "complex" vs. something more descriptive like `BookOutline` or `BookWithChapterTitles`?

docs/learn/response_models.md
```diff
@@ -156,6 +170,33 @@ In the above example, your IDE will provide proper autocompletion and type check
...
+    return f"Outline a {genre} book with chapter titles"
+
+book_outline = outline_book("mystery")
+print(f"Title: {book_outline.title}")
+print("Chapters:")
+for chapter in book_outline.chapters:
+    print(f"- {chapter}")
+```
+
+!!! note "Provider Support"
```
willbakst, line 193: nit: remove this, it's duplicated content, see below

docs/learn/response_models.md
```diff
@@ -222,9 +263,9 @@ class Book(BaseModel):
 
 
 @openai.call("gpt-4o-mini", response_model=Book)
-@prompt_template("Recommend a {genre} book")
-def recommend_book(genre: str): ...
-
+@prompt_template()
+def recommend_book(genre: str) -> str:
```
willbakst, line 267: nit: ditto

docs/learn/response_models.md
```diff
@@ -240,13 +281,80 @@ except ValidationError as e:
...
+   @openai.call("gpt-4o-mini", response_model=Recipe)
+   @prompt_template()
+   def get_recipe(dish: str) -> str:
+       return f"Provide a recipe for {dish}"
+
+   try:
+       recipe = get_recipe("spaghetti carbonara")
+       print(recipe)
+   except Exception as e:
+       print(f"Failed to get recipe after retries: {e}")
```
willbakst, line 337: add another link after this to our tenacity integration for more advanced error handling?

docs/learn/response_models.md
```diff
@@ -240,13 +281,80 @@ except ValidationError as e:
...
+   ```
+ 
+5. **Leverage JSON Mode**: When possible, use `json_mode=True` for better type support and consistency.
+   ```python
+   @openai.call("gpt-4o-mini", response_model=ComplexDataStructure, json_mode=True)
+   @prompt_template()
+   def get_complex_data(query: str) -> str:
```
willbakst, line 344: nit: ditto

docs/learn/response_models.md
```diff
@@ -240,13 +281,80 @@ except ValidationError as e:
...
+       return f"Generate complex data for {query}"
+   ```
+
+6. **Test Thoroughly**: Validate your Response Models across different inputs and edge cases.
+   ```python
+   import pytest
+
+   @pytest.mark.parametrize("genre", ["sci-fi", "romance", "mystery", "non-fiction"])
```
willbakst, line 352: hmm, this test feels more like it's testing mirascope than the desired behavior (since we should expect if it doesn't fail that we will in fact get a book back)

docs/learn/response_models.md
koxudaxi: There is nothing wrong with setting this to `Messages.Type`, but I have one question.
If I were a user, I would want to define it as str because it would be a hassle to type `Message.Type`.
Typing-wise, Message.Type is fine because it encompasses `str`.
Are you saying that you are writing `Message.Type` as a reference here? Or are you strongly suggesting that we deprecate writing str and force users to use `Message.Type`?

willbakst: hmm, i guess it's fine to type it as str, i was mostly suggesting wrt. consistency of showing returning `Messages.Type` for easy extension down the line, but i guess typing it as str does remove the need for the extra import unless you're using it

willbakst (approved):

Issue #501, opened 2024-09-17
Anthropic cost coverage for AWS bedrock and Vertex


ramon-prieto:
### Description

Currently, the calculate cost function only accepts the naming convention from the official Anthropic API. Can we extend the cost mapping to allow for cost estimation on workflows using either the Vertex or Bedrock clients?

https://github.com/Mirascope/mirascope/blob/dev/mirascope/core/anthropic/_utils/_calculate_cost.py#L20

willbakst:
A bit swamped this week, but will try to look into this next week.

PR welcome if you have the time and want to contribute!

Would be much appreciated :)

ramon-prieto:
Created this [PR](https://github.com/Mirascope/mirascope/pull/504) to extend the mapping

PR #504, opened 2024-09-18
feat: add bedrock and vertex models to Anthropic cost mapping

ramon-prieto:
Closes issue [#501](https://github.com/Mirascope/mirascope/issues/501)

Extends the anthropic cost mapping to include models from bedrock and vertex


willbakst (approved):
Great stuff, thank you!!

PR #505, opened 2024-09-18
docs: Improve output_parsers.md

koxudaxi:


docs/learn/output_parsers.md
```diff
@@ -36,17 +53,53 @@ print(f"Title: {book.title}")
...
+4. The `recommend_book` function now returns a `Book` object instead of a raw response.
 
 ### Type Safety and Proper Hints
 
-When using output parsers with Mirascope's call decorator, you benefit from accurate type hints that reflect the parser's output type.
+When using output parsers with Mirascope's call decorator, you benefit from accurate type hints that reflect the parser's output type. This enhances code reliability and helps catch potential type-related errors early in the development process.
+
+For example, in the code above, your IDE will recognize `book` as a `Book` instance, providing appropriate autocompletion and type checking for `book.title` and `book.author`.
+
+## Comparison with Official SDK
```
willbakst, line 69: this isn't a 1:1 comparison because it doesn't use the pydantic `Book` model even though it could

docs/learn/output_parsers.md
```diff
@@ -36,17 +53,53 @@ print(f"Title: {book.title}")
...
+
+```python hl_lines="13-14"
+from anthropic import Anthropic
+
+client = Anthropic()
+
+def recommend_book(genre: str):
+    response = client.completions.create(
+        model="claude-3-5-sonnet-20240620",
+        prompt=f"Recommend a {genre} book in the format Title by Author",
```
willbakst, line 81: this is not correct, this should be a messages array

docs/learn/output_parsers.md
```diff
@@ -36,17 +53,53 @@ print(f"Title: {book.title}")
...
+
+Let's compare the Mirascope approach with using the official Anthropic SDK directly:
+
+```python hl_lines="13-14"
+from anthropic import Anthropic
+
+client = Anthropic()
+
+def recommend_book(genre: str):
+    response = client.completions.create(
```
willbakst, line 79: this is not the right endpoint, seems stale

docs/learn/output_parsers.md
```diff
@@ -107,16 +160,270 @@ else:
...
-- **Handle Parsing Errors**: Always implement error handling in your parsers. LLMs may occasionally produce outputs that don't conform to the expected structure.
-- **Gradual Refinement**: Start with simple parsers and gradually increase complexity as you refine your prompts and understand the model's output patterns.
-- **Provide Clear Instructions**: In your prompts, be explicit about the structure you expect the LLM to produce. This helps ensure the output is parseable.
-- **Validate Parsed Output**: After parsing, validate the extracted information to ensure it meets your application's requirements.
-- **Consider Fallback Strategies**: Implement fallback strategies for cases where parsing fails, such as requesting a reformatted response from the LLM. Check out our [tenacity integration](../integrations/tenacity.md) for more details on how to easily reinsert caught errors into subsequent retries.
+When working with Output Parsers in Mirascope, consider the following best practices:
+
+1. **Align with Prompt Engineering**: Design your prompts to generate outputs that match your parser's expectations. This improves consistency and reliability.
```
willbakst, line 174: given there is so much duplication, it seems best to merge the earlier example about using XML parsing and just include that here instead so we don't duplicate content. makes sense to me that it would be best to mention this here alongside the prompt engineering bit anyway, and we can mention here that Anthropic is particularly good with XML as an example of such prompt engineering

docs/learn/output_parsers.md
```diff
@@ -107,16 +160,270 @@ else:
...
+        root = ET.fromstring(response.content)
+        return {
+            "title": root.find("title").text,
+            "author": root.find("author").text,
+            "year": int(root.find("year").text),
+            "summary": root.find("summary").text,
+            "genres": [genre.text for genre in root.findall("genre")]
+        }
+    
+    @anthropic.call("claude-3-5-sonnet-20240620", output_parser=parse_title_author)
```
willbakst, line 271: should we colocate this code with the corresponding parser rather than colocating the parsers and methods separately?

willbakst (approved):

PR #515, opened 2024-09-23
Add batching from OpenAI

vanities:
Hey there! This is very much WIP, but I thought it would be easier to talk about implementation here. I had to re-remember how to do write tests with mocks, but I can fix those. 

Would it be better if these functions were all in a `Batch` class with methods like `create`, `run`, and `retrieve`? It would be useful to track the different states the process could be in. Anyway, any and all feedback on this would be very helpful :)


willbakst:
This is super old and has no progress, so I'm going to close it for now until further notice.

PR #518, opened 2024-09-24
docs: fix typo and added extra section

brenkao:


docs/learn/prompts.md
```diff
@@ -38,7 +38,16 @@ Let's take a look at how we can write these same prompts in a provider-agnostic
 
 The core concept to understand here is [`BaseMessageParam`](../api/core/base/message_param.md#basemessageparam). This class operates as the base class for message parameters that Mirascope can handle and use across all supported providers.
 
-In Mirascope, we use the `@prompt_template` decorator to write prompt templates as reusable methods. Let's look at a basic example:
+In Mirascope, we use the `@prompt_template` decorator to write prompt templates as reusable methods. There are four main ways of writing prompts using Mirascope:
```
willbakst, line 41: In Mirascope, we use the `@prompt_template` decorator to write prompt templates as reusable methods. There are four methods of writing prompts:

1. _(Shorthand)_ Returning the `str` or `list` content for a single user message.
2. _(Messages)_ Using `Messages.{Role}` methods, which accept the full or shorthand content and output a `BaseMessageParam` instance.
3. _(String Template)_ Passing a string template to `@prompt_template` that gets parsed and then formatted like a normal Python formatted string.
4. _(BaseMessageParam)_ Directly writing `BaseMessageParam` instances.

Which method you use is mostly up to your preference, so feel free to select which one you prefer in the following sections.

Let's look at a basic example:

willbakst (approved):

Issue #521, opened 2024-09-25
Mistral Vision Support


willbakst:
### Description

Mistral recently released a new vision model Pixtral. The API looks extremely similar to other providers such as OpenAI, so adding support should be fairly straightforward.

Just need to update [this conversion code](https://github.com/Mirascope/mirascope/blob/912d8d3a5b1fd1845c38148117b1158b15f09623/mirascope/core/mistral/_utils/_convert_message_params.py#L18) to support image parts [like we do for OpenAI](https://github.com/Mirascope/mirascope/blob/912d8d3a5b1fd1845c38148117b1158b15f09623/mirascope/core/openai/_utils/_convert_message_params.py#L24)

skylarbpayne:
I'll take a crack at this.

willbakst:
This is released in `v1.10`!

PR #531, opened 2024-09-29
Bump mistralai to > 1.0.0 in preparation for latest models such as Pixtral

skylarbpayne:
Related to #521

# Why?

Mistral has released a vision model, Pixtral. Supporting vision for Mistral in mirascope requires bumping up the version to the latest major version to take advantage of image capabilities.

# What?

Migrates Mistral usage to v1.0.0, [according to the migration guide](https://github.com/mistralai/client-python/blob/9f487dc998e948a52e17bcaa9aace5ccfa92571b/MIGRATION.md).

The major changes are that the separate sync/async clients were merged into a single one with 4 separate functions under a `chat` attribute: `complete`, `stream`, `complete_async`, `stream_async` -- with some corresponding changes to the response formats.

Additionally, several enums were replaced with `Literal` types.

# Additional Context

I have filed a bug related to support of the TypedDict versions of the message types: https://github.com/mistralai/client-python/issues/149

# Testing

- All mistral tests pass
- Examples using mistral run; on spot check they look correct. But let me know if there's better validation I can do here.


mirascope/core/mistral/_utils/_setup_call.py
```diff
@@ -95,28 +94,28 @@ def setup_call(
...
         assert tool_types, "At least one tool must be provided for extraction."
-        call_kwargs["tool_choice"] = cast(ToolChoice, ToolChoice.any)
+        call_kwargs["tool_choice"] = cast(ToolChoiceEnum, "any")
     call_kwargs |= {"model": model, "messages": messages}
 
     if client is None:
-        client = (
-            MistralAsyncClient() if inspect.iscoroutinefunction(fn) else MistralClient()
+        client = Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
+    if inspect.iscoroutinefunction(fn):
```
willbakst, line 115: you should use our internal helper utility [`fn_is_async`](https://github.com/Mirascope/mirascope/blob/7c1f6e3eae4aa086f5ee9685884516b75820af08/mirascope/core/base/_utils/_protocols.py#L110) for better typing here.

If it's not clear because of where the function is located, feel free to pull `fn_is_async` out into it's own file like other utilities.

mirascope/core/mistral/dynamic_config.py
```diff
@@ -1,9 +1,8 @@
 """This module defines the function return type for functions as LLM calls."""
 
-from mistralai.models.chat_completion import ChatMessage
-
 from ..base import BaseDynamicConfig, BaseMessageParam
 from .call_params import MistralCallParams
+from .message import ChatMessage
 
 MistralDynamicConfig = BaseDynamicConfig[
     ChatMessage | BaseMessageParam, MistralCallParams
```
willbakst, line 8: this would then just be a union of all the message types + `BaseMessageParam` since they don't have an internal union type?

mirascope/core/mistral/_utils/_convert_message_params.py
```diff
@@ -1,23 +1,24 @@
...
+from ..message import ChatMessage, is_chat_message, make_chat_message
 
 
 def convert_message_params(
     message_params: list[BaseMessageParam | ChatMessage],
 ) -> list[ChatMessage]:
     converted_message_params = []
     for message_param in message_params:
-        if isinstance(message_param, ChatMessage):
+        if is_chat_message(message_param):
```
willbakst, line 12: here we should just check `if not isinstance(BaseMessageParam)`

examples/learn/response_models/basic_usage/mistral/official_sdk.py
```diff
@@ -1,8 +1,9 @@
-from mistralai.client import MistralClient
-from mistralai.models.chat_completion import ToolChoice
+from mistralai.client import Mistral
+from mistralai.models import ToolChoice
 from pydantic import BaseModel
+import os
 
-client = MistralClient()
+client = Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 6: ditto

examples/learn/calls/custom_client/mistral/string_template.py
```diff
@@ -1,13 +1,17 @@
...
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
+)
 @prompt_template("Recommend a {genre} book")
 def recommend_book(genre: str): ...
 
 
-@mistral.call("mistral-large-latest", client=MistralAsyncClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 14: ditto

examples/learn/calls/custom_client/mistral/string_template.py
```diff
@@ -1,13 +1,17 @@
 from mirascope.core import mistral, prompt_template
-from mistralai.async_client import MistralAsyncClient
-from mistralai.client import MistralClient
+from mistralai import Mistral
+import os
 
 
-@mistral.call("mistral-large-latest", client=MistralClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 7: ditto

examples/learn/calls/basic_call/mistral/official_sdk_call.py
```diff
@@ -1,6 +1,7 @@
-from mistralai.client import MistralClient
+from mistralai import Mistral
+import os
 
-client = MistralClient()
+client = Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 4: we should let the client load the api key like we do for other providers, so just client = Mistral() should suffice here, no?

examples/learn/calls/custom_client/mistral/base_message_param.py
```diff
@@ -1,13 +1,17 @@
 from mirascope.core import BaseMessageParam, mistral
-from mistralai.async_client import MistralAsyncClient
-from mistralai.client import MistralClient
+from mistralai import Mistral
+import os
 
 
-@mistral.call("mistral-large-latest", client=MistralClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 7: ditto

examples/learn/calls/custom_client/mistral/base_message_param.py
```diff
@@ -1,13 +1,17 @@
...
+)
 def recommend_book(genre: str) -> list[BaseMessageParam]:
     return [BaseMessageParam(role="user", content=f"Recommend a {genre} book")]
 
 
-@mistral.call("mistral-large-latest", client=MistralAsyncClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 14: ditto

examples/learn/calls/custom_client/mistral/messages.py
```diff
@@ -1,13 +1,17 @@
 from mirascope.core import Messages, mistral
-from mistralai.async_client import MistralAsyncClient
-from mistralai.client import MistralClient
+from mistralai import Mistral
+import os
 
 
-@mistral.call("mistral-large-latest", client=MistralClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 7: ditto

examples/learn/calls/custom_client/mistral/messages.py
```diff
@@ -1,13 +1,17 @@
...
+)
 def recommend_book(genre: str) -> Messages.Type:
     return Messages.User(f"Recommend a {genre} book")
 
 
-@mistral.call("mistral-large-latest", client=MistralAsyncClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 14: ditto

examples/learn/calls/custom_client/mistral/shorthand.py
```diff
@@ -1,13 +1,17 @@
 from mirascope.core import mistral
-from mistralai.async_client import MistralAsyncClient
-from mistralai.client import MistralClient
+from mistralai import Mistral
+import os
 
 
-@mistral.call("mistral-large-latest", client=MistralClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 7: ditto

examples/learn/calls/custom_client/mistral/shorthand.py
```diff
@@ -1,13 +1,17 @@
...
+)
 def recommend_book(genre: str) -> str:
     return f"Recommend a {genre} book"
 
 
-@mistral.call("mistral-large-latest", client=MistralAsyncClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
willbakst, line 14: ditto

tests/core/mistral/_utils/test_setup_call.py
```diff
@@ -88,10 +83,10 @@ async def test_async_setup_call(
     mock_base_setup_call: MagicMock,
 ) -> None:
     """Tests the `setup_call` function."""
-    mock_mistral_chat = AsyncMock(spec=ChatCompletionResponse)
+    mock_mistral_chat = MagicMock(spec=ChatCompletionResponse)
```
willbakst, line 86: hmm, we'll need to make sure we're using sync vs. async properly here and testing properly.

I doubt this should be changed from AsyncMock to MagicMock, seems like instead we need to update the code to ensure we're returning the correct sync vs. async methods?

mirascope/core/mistral/message.py
```diff
@@ -0,0 +1,40 @@
+from typing import Annotated, Any, Literal, TypeAlias
+
+from mistralai.models import AssistantMessage, SystemMessage, ToolMessage, UserMessage
+from pydantic import Discriminator, Tag, TypeAdapter
+
+
+def is_chat_message(v: Any) -> bool:  # noqa: ANN401
```
willbakst, line 7: given other comment on the check here, we can probably remove this?

mirascope/core/mistral/message.py
```diff
@@ -0,0 +1,40 @@
...
+ChatMessage: TypeAlias = Annotated[
+    Annotated[SystemMessage, Tag("system")]
+    | Annotated[UserMessage, Tag("user")]
+    | Annotated[AssistantMessage, Tag("assistant")]
+    | Annotated[ToolMessage, Tag("tool")],
+    Discriminator(chat_message_discriminator),
+]
+
+
+def make_chat_message(**kwargs: Any) -> ChatMessage:  # noqa: ANN401
```
willbakst, line 35: this feels like an internal method? I feel like we should move this into a `_make_chat_message.py` file in `_utils` to retain more control over changes to it as necessary (rather than making it public).

My other question is how necessary this function is at all? Can we not just internally use their types and check the role and in the message conversion just create the correct role?

We could then put this as a private utility of the conversion utility and likely drop the whole discriminator thing?

mirascope/core/mistral/_utils/_setup_call.py
```diff
@@ -95,28 +94,28 @@ def setup_call(
...
     elif extract:
         assert tool_types, "At least one tool must be provided for extraction."
-        call_kwargs["tool_choice"] = cast(ToolChoice, ToolChoice.any)
+        call_kwargs["tool_choice"] = cast(ToolChoiceEnum, "any")
     call_kwargs |= {"model": model, "messages": messages}
 
     if client is None:
-        client = (
-            MistralAsyncClient() if inspect.iscoroutinefunction(fn) else MistralClient()
+        client = Mistral(api_key=os.environ.get("MISTRAL_API_KEY"))
```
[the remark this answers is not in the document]
skylarbpayne: `Mistral` does not do this unfortunately. I too was surprised. Is there a different way you want to handle this?

mirascope/core/mistral/dynamic_config.py
skylarbpayne: Yeah. Do you want to explicitly break it out like:

```
MistralDynamicConfig = BaseDynamicConfig[
    SystemMessage | AssistantMessage | UserMessage | ToolMessage | BaseMessageParam, MistralCallParams
```

or is this just an ask for clarification?

mirascope/core/mistral/message.py
skylarbpayne: Moving it to internal makes sense.

Hmm I'm not sure what the broader suggestion here is. Are you suggesting we could get rid of this function and then anywhere we need to construct a "ChatMessage" we would do it inline? Or are you maybe suggestioning a different implementation of this function?

mirascope/core/mistral/_utils/_setup_call.py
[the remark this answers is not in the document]
willbakst: Oh weird. In that case just make sure to do `os.environ.get("MISTRAL_API_KEY", "")` instead here and elsewhere

mirascope/core/mistral/dynamic_config.py
willbakst: yeah breaking it out like you showed is what I meant

mirascope/core/mistral/message.py
willbakst: Yeah I'm suggesting we could get rid of this function and just construct the appropriate messages inline.

However, since they offer two methods for writing messages (e.g. `UserMessage` and `UserMessageTypedDict` we'll need to handle both of these cases, in which case we could have `_convert_message_to_base_message_param` as an internal helper function in the `_convert_message_params.py` utility file.

Does this make sense?

tests/core/mistral/_utils/test_setup_call.py
skylarbpayne: I think we are returning the right methods; I've never used `AsyncMock` on the return value; always just on the method (i.e. the actual thing that's awaited). So the thing that matters to be AsyncMock is the `complete_async` and `stream_async` methods.

I think it doesn't matter whether these mocked return values are `MagicMock` or `AsyncMock`. I changed it here because as far as I can tell the return values themselves are not awaited -- just the function that returns them.
See this simplified example:

```
class Dummy:
  pass

# Below we will create 2 async mocks to be something that is awaited.

# Version a: we make the return value an AsyncMock
a = AsyncMock()
a.return_value = AsyncMock(spec=Dummy)

# Version b: we make the return value a MagicMock
b = AsyncMock()
b.return_value = MagicMock(spec=Dummy)


isinstance(await a(), Dummy) # True
isinstance(await b(), Dummy) # True
```

mirascope/core/mistral/stream.py
```diff
@@ -26,10 +29,10 @@ class MistralStream(
     BaseStream[
         MistralCallResponse,
         MistralCallResponseChunk,
-        ChatMessage,
-        ChatMessage,
-        ChatMessage,
-        ChatMessage,
+        AssistantMessage | SystemMessage | ToolMessage | UserMessage,
```
willbakst, line 32: so these were originally all the same because there was only one message type, but now we should use the correct message types for the generics (see [here](https://github.com/Mirascope/mirascope/blob/9202d7f32fffa3798c95c3051d9e98a953eab27c/mirascope/core/base/stream.py#L55)):

1. _UserMessageParamT = UserMessage
2. _AssistantMessageParamT = AssistantMessage
3. _ToolMessageParamT = ToolMessage
4. _MessageParamT = SystemMessage | UserMessage | AssistantMessage | ToolMessage

mirascope/core/mistral/call_response.py
```diff
@@ -138,7 +146,7 @@ def tool(self) -> MistralTool | None:
     @classmethod
     def tool_message_params(
         cls, tools_and_outputs: list[tuple[MistralTool, str]]
-    ) -> list[ChatMessage]:
+    ) -> list[AssistantMessage | SystemMessage | ToolMessage | UserMessage]:
```
willbakst, line 147: this should always be `list[ToolMessage]`

mirascope/core/mistral/call_response.py
```diff
@@ -98,7 +104,9 @@ def cost(self) -> float | None:
 
     @computed_field
     @property
-    def message_param(self) -> ChatMessage:
+    def message_param(
+        self,
+    ) -> AssistantMessage | SystemMessage | ToolMessage | UserMessage:
```
willbakst, line 107: this should always be `AssistantMessage`

not sure if `self.response.choices[0].message` is the correct type, so may need a conversion

mirascope/core/mistral/call_response.py
```diff
@@ -22,9 +28,9 @@ class MistralCallResponse(
         MistralTool,
         dict[str, Any],
         MistralDynamicConfig,
-        ChatMessage,
+        AssistantMessage | SystemMessage | ToolMessage | UserMessage,
         MistralCallParams,
-        ChatMessage,
+        AssistantMessage | SystemMessage | ToolMessage | UserMessage,
```
willbakst, line 33: this is _UserMessageParamT and should just be `UserMessage`

mirascope/core/mistral/_utils/_setup_call.py
```diff
@@ -1,23 +1,26 @@
...
     ResponseFormat,
-    ResponseFormats,
-    ToolChoice,
+    SystemMessage,
+    ToolChoiceEnum,
+    ToolMessage,
+    UserMessage,
 )
 
+from mirascope.core.base._utils._protocols import fn_is_async
```
willbakst, line 22: this should be imported through the `...base._utils` import below

mirascope/core/mistral/_utils/_convert_message_params.py
```diff
@@ -1,23 +1,35 @@
...
+    message_params: list[
+        BaseMessageParam | AssistantMessage | SystemMessage | ToolMessage | UserMessage
+    ],
+) -> list[BaseMessageParam]:
     converted_message_params = []
     for message_param in message_params:
-        if isinstance(message_param, ChatMessage):
+        if not isinstance(
+            message_param,
+            BaseMessageParam,
```
willbakst, line 22: nit: unnecessary comma causing unnecessary formatting? should be a single line

mirascope/core/mistral/_utils/_convert_message_params.py
```diff
@@ -1,23 +1,35 @@
...
+        ):
             converted_message_params.append(message_param)
         elif isinstance(content := message_param.content, str):
-            converted_message_params.append(ChatMessage(**message_param.model_dump()))
+            converted_message_params.append(
+                BaseMessageParam(**message_param.model_dump())
```
willbakst, line 27: this is not converted since by this point `message_param` is already confirmed to be a `BaseMessageParam` instance, so we'll need to check the role and convert this into the correct mistral-specific message param

mirascope/core/mistral/_utils/_convert_message_params.py
```diff
@@ -1,23 +1,35 @@
...
+            )
         else:
             if len(content) != 1 or content[0].type != "text":
                 raise ValueError("Mistral currently only supports text parts.")
             converted_message_params.append(
-                ChatMessage(role=message_param.role, content=content[0].text)
+                BaseMessageParam(role=message_param.role, content=content[0].text)
```
willbakst, line 33: same thing here

mirascope/core/mistral/_utils/_convert_message_params.py
```diff
@@ -1,23 +1,35 @@
...
 from ...base import BaseMessageParam
 
 
 def convert_message_params(
-    message_params: list[BaseMessageParam | ChatMessage],
-) -> list[ChatMessage]:
+    message_params: list[
+        BaseMessageParam | AssistantMessage | SystemMessage | ToolMessage | UserMessage
+    ],
+) -> list[BaseMessageParam]:
```
willbakst, line 17: this should be `list[SystemMessage | UserMessage | AssistantMessage | ToolMessage]`

examples/learn/calls/custom_client/mistral/base_message_param.py
```diff
@@ -1,7 +1,20 @@
...
+
 from mirascope.core import BaseMessageParam, mistral
-from mistralai.client import MistralClient
+from mistralai import Mistral
 
 
-@mistral.call("mistral-large-latest", client=MistralClient())
+@mistral.call(
+    "mistral-large-latest",
+    client=Mistral(api_key=os.environ.get("MISTRAL_API_KEY", "")),
```
willbakst, line 9: ok, so after further thought I think that when using Mistral with Mirascope we should have a helper utility for getting the api key.

I noticed that their client accepts a function, so we could do something like `client=Mistral(mistral.load_api_key)`?

seems more convenient both internally + for users using a custom client, WDYT?

mirascope/core/mistral/_utils/_convert_message_params.py
skylarbpayne: On the output?

examples/learn/calls/custom_client/mistral/base_message_param.py
willbakst: hmm, actually just ignore this lol

we only load it internally once in setup call, otherwise I think it actually makes more sense for the user to just use `os` since that's what the mistral docs show and it's standard anyway

examples/learn/async/custom_client/decorator/mistral/messages.py
```diff
@@ -1,7 +1,7 @@
 from mirascope.core import Messages, mistral
-from mistralai.async_client import MistralAsyncClient
+from mistralai import Mistral
 
 
-@mistral.call("mistral-large-latest", client=MistralAsyncClient())
+@mistral.call("mistral-large-latest", client=Mistral())
```
willbakst, line 5: i believe in all of these example instances we need to manually load and pass in the api key from the environment? at least that was my understanding from their docs

mirascope/core/mistral/_call_kwargs.py
```diff
@@ -2,12 +2,17 @@
...
+)
 
 from ..base import BaseCallKwargs
 from .call_params import MistralCallParams
 
 
 class MistralCallKwargs(MistralCallParams, BaseCallKwargs[dict[str, Any]]):
     model: str
-    messages: list[ChatMessage]
+    messages: list[AssistantMessage | SystemMessage | ToolMessage | UserMessage]
```
willbakst, line 18: nit: this should be typed as `Sequence`, and we should probably update any other providers as well

mirascope/core/mistral/_utils/_convert_message_params.py
```diff
@@ -1,23 +1,45 @@
...
+    AssistantMessage,
+    SystemMessage,
+    ToolMessage,
+    UserMessage,
+)
 
 from ...base import BaseMessageParam
 
 
+def make_message(
```
willbakst, line 13: nit: make this private `_make_message`?

mirascope/core/mistral/_utils/_load_api_key.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+
+def load_api_key() -> str:
+    """Load the API key from the standard environment variable."""
+    return os.environ.get("MISTRAL_API_KEY", "")
```
willbakst, line 6: nit: I think we should actually just remove this method entirely and just use os directly, no reason to wrap this

mirascope/core/mistral/_utils/_setup_call.py
```diff
@@ -33,19 +41,19 @@
...
     json_mode: bool,
     call_params: MistralCallParams | CommonCallParams,
     extract: bool,
     stream: bool,
 ) -> tuple[
-    AsyncCreateFn[ChatCompletionResponse, ChatCompletionStreamResponse],
+    AsyncCreateFn[ChatCompletionResponse, CompletionEvent],
     str | None,
-    list[ChatMessage],
+    list[AssistantMessage | SystemMessage | ToolMessage | UserMessage],
```
willbakst, line 56: nit: same thing here, this will have to be updated to sequence

koxudaxi: Since many setup_calls return a list in `convert_message_params`, shouldn't we be able to guarantee that a list is returned? If we change from list to Sequence here, I think we'll have compatibility issues with subsequent processing that expects list operations like list.append. What do you think?

willbakst: hmm, that's a good point. really the only important thing here is that the user can provide the list of messages in the dynamic config as a sequence (so that not all types must be present for no type errors, so most recent v1.9.5 bug fix release for details)

I'm fine guaranteeing a list internally here, which means we should likely revert the base setup call return type to a list back from sequence (but still expect a sequence input, right?)

examples/learn/async/custom_client/decorator/mistral/base_message_param.py
```diff
@@ -1,7 +1,11 @@
+import os
+
 from mirascope.core import BaseMessageParam, mistral
-from mistralai.async_client import MistralAsyncClient
+from mistralai import Mistral
 
 
-@mistral.call("mistral-large-latest", client=MistralAsyncClient())
+@mistral.call(
+    "mistral-large-latest", client=Mistral(api_key=os.environ["MISTRAL_API_KEY"])
```
willbakst, line 8: nit: doesn't this mean we'll have to change the docs highlighting for this? same for the sync custom client, right?

mirascope/core/mistral/_utils/_setup_call.py
koxudaxi: I agree with you. I keep the sequence input and revert base setup call retrun type to a list.

examples/learn/async/custom_client/decorator/mistral/base_message_param.py
koxudaxi: Nice catch. I fixed docs.

willbakst (approved):

PR #534, opened 2024-09-30
docs: Convert text_summarization markdown to notebook

koxudaxi:


koxudaxi:
@willbakst 
Two corrections were made.
- Because there was no text file to be summarized, we added a cell to be installed with curl.
- The wikipedia page on Python is too long and token-overloaded, so I changed it to guido's page.

Also, normally, it is necessary to extract only the text from the downloaded html page, but in this case, I summarized the html page. If you don't mind more steps, you may pip install beautifulsoup and parse the html.

willbakst (approved):

PR #604, opened 2024-10-06
fix: broken link, typos, updated some sentences

brenkao:


docs/learn/agents.md
```diff
@@ -14,7 +14,7 @@ In this section we will implement a toy `Librarian` agent to demonstrate key con
         If you haven't already, we recommend first reading the section on [Calls](./calls.md)
     </div>
 
-Since an agent needs to operate across multiple LLM API calls, the first concept to cover is state. You can introduce state in many ways ranging from local variables to a database.
+Since an agent needs to operate across multiple LLM API calls, the first concept to cover is state. You can introduce state in many ways ranging from local variables to a database, or short-term memory and long-term memory respectively.
```
willbakst, line 17: Hmm, it's not necessarily memory. Could just be state for the agent to work with, and I think "memory" is a bit overloaded as a term?

Curious the value add here of this additional description rather than keeping it more general to storage?

docs/learn/agents.md
```diff
@@ -102,6 +102,8 @@ In this example we:
     - If yes, we call the tools, collect the outputs, and insert the tool calls into the chat history. We then recursively call `_step` again with an empty user query until the LLM is done calling tools and is ready to response
     - If no, the LLM is ready to respond and we return the response content.
 
+Now that our chatbot is capable of using tools, we now have a basic agent.
```
willbakst, line 105: nit: you use "now" here twice

docs/learn/agents.md
brenkao: its a common phrase used in agents, so i figured state or stateless is not quite the same as what people think when it comes to llm agents. what situation would you add state and consider it not memory?

willbakst: Hmm, I guess it's less about "memory" and more about short term vs. long term.

You could use a database for persistence that you load into local state for "working memory" for example

willbakst: What about something like:

The goal of providing state to the agent is to give it memory. For example, we can think of local variables as "working memory" and a database as "long-term memory"

willbakst (approved):

Issue #319, opened 2024-06-11
More Bedrock Model Support


ionflow:
### Description

Assuming AWS has better uptime percentage, it would be good to support more of their models for enterprise customers. Specifically I'm interested in the following models:

Claude (latest)
```
anthropic.claude-3-opus-20240229-v1:0
anthropic.claude-3-sonnet-20240229-v1:0
anthropic.claude-3-haiku-20240307-v1:0
```
Meta (latest)
```
meta.llama3-70b-instruct-v1:0
meta.llama3-8b-instruct-v1:0
```
Mistral (latest)
```
mistral.mistral-large-2402-v1:0
mistral.mixtral-8x7b-instruct-v0:1
mistral.mistral-7b-instruct-v0:2
```
Cohere (latest)
```
cohere.command-r-plus-v1:0
cohere.command-r-v1:0
```

willbakst:
In v1, we've added support for LiteLLM, which should enable all of the models on Bedrock. Is this sufficient? Or are you looking more for direct support @ionflow?

agentgill:
Ability to invoke models on AWS would be useful. I guess it some sort of wrapper around boto3 bedrock-runtime

willbakst:
Yeah I think LiteLLM has comparable support, which we've added support for in v1.

I'm not 100% sure what our own version for direct support would look like, but I'm happy to look into this if the LiteLLM support is insufficient for whatever reason.

willbakst:
Looks like we should be able to use the BedrockRuntime `converse` and `converse_stream` methods to implement support for `bedrock.call` so we can access Bedrock more directly (rather than using something like LiteLLM).

This is a great first issue as there are already 9 other providers implemented that can act as a good reference for how one should implement this using our internal helper methods and factories.

We also have documentation for adding support for [Custom LLM Providers](https://docs.mirascope.io/latest/integrations/custom_provider/), which is the same process for adding support as is required for adding Bedrock support.

Converse: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html
Converse Stream: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse_stream.html

willbakst:
Released in `v1.4`!

PR #606, opened 2024-10-08
feat: Support AWS Bedrock

koxudaxi:
Related Issues: https://github.com/Mirascope/mirascope/issues/319


mirascope/core/bedrock/__init__.py
```diff
@@ -0,0 +1,24 @@
+from ._call import bedrock_call
+from ._call import bedrock_call as call
+from ._utils import AssistantMessageTypeDef, BedrockMessageParam, UserMessageTypeDef
+from .call_params import BedrockCallParams
+from .call_response import BedrockCallResponse
+from .call_response_chunk import BedrockCallResponseChunk
+from .dynamic_config import BedrockDynamicConfig
+from .stream import BedrockStream
+from .tool import BedrockTool, BedrockToolConfig
+
```
willbakst, line 10: missing the message param type

for reference, see [this](https://github.com/Mirascope/mirascope/blob/e357897c40d2e4923a8f022bac79c878d65b8811/mirascope/core/openai/__init__.py#L17)

mirascope/core/bedrock/_utils/_calculate_cost.py
```diff
@@ -0,0 +1,8 @@
+"""Calculate the cost of a completion using the Bedrock API."""
+
+
+def calculate_cost(
+    input_tokens: int | float | None, output_tokens: int | float | None, model: str
+) -> float | None:
+    """Calculate the cost of a completion using the Bedrock API."""
+    return None
```
willbakst, line 8: can you please add a comment here mentioning that we're looking into a way to determine the cost dynamically since there are too many to support for this initial PR?

mirascope/core/bedrock/_utils/_extract_stream.py
```diff
@@ -0,0 +1,50 @@
...
+
+from mirascope.core.bedrock._utils._types import (
+    AsyncStreamOutputChunk,
+    StreamOutputChunk,
+)
+
+_P = ParamSpec("_P")
+
+
+def _extract_sync_stream_fn(
```
willbakst, line 20: nit: these functions should not be private, it's the module that is private, and these are public in the private module (see other utilities for reference)

mirascope/core/bedrock/_utils/_setup_call.py
```diff
@@ -0,0 +1,156 @@
...
+            if "toolConfig" in call_kwargs:
+                call_kwargs["toolConfig"]["toolChoice"] = cast(
+                    ToolChoiceTypeDef, {"type": "tool", "name": tool_types[0]._name()}
+                )
+
+    call_kwargs |= cast(BedrockCallKwargs, {"modelId": model, "messages": messages})
+
+    if client is None:
+        client = (
+            asyncio.run(_get_async_client())
```
willbakst, line 143: hmm, i think these refactored utilities are a little overkill. we only use them here, correct? we should just do this here in an if/else and call it a day, refactoring this without a need to reuse it is overkill

mirascope/core/bedrock/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+__all__ = [
+    "AssistantMessageTypeDef",
+    "bedrock_call",
+    "BedrockCallParams",
+    "BedrockCallResponse",
+    "BedrockCallResponseChunk",
+    "BedrockDynamicConfig",
+    "BedrockMessageParam",
```
willbakst, line 18: aha, just saw this. looking more deeply, it looks like this message param is not quite right since it won't include `BaseMessageParam` like the others.

Perhaps we can rename this one to `InternalBedrockMessageParam` and then create a new alias for `BedrockMessageParam` that matches the other providers?

mirascope/core/bedrock/_utils/_get_client.py
willbakst, line 1: this whole utility is unnecessary. we only use it once and have no plans to reuse it, so we should just do this inline in setup call, i left a comment there as well

mirascope/core/bedrock/_utils/_setup_call.py
willbakst: something like this should work:

```python
# defined above so it's not defined on every call
def _get_async_client(session: ...) -> AsyncBedrockRuntimeClient:
    async with session.create_client("bedrock_runtime") as client:
        return client

# inside function
if client is None:
    if inspect.iscoroutinefunction(fn):
        session = get_session()
        # we need to get the async client through an async context manager
        # so we define a function to do this and run it here with asyncio
        client = asyncio.run(_get_async_client(session))
    else:
        session = Session()
        client = session.client("bedrock-runtime")
```

mirascope/core/bedrock/_utils/_extract_stream.py
willbakst, line 1: this could also be merged into the file for setup_call

mirascope/core/bedrock/_utils/_types.py
willbakst, line 1: this should just be under `.../bedrock/_types.py`

willbakst (dismissed):

willbakst (approved):

Issue #519, opened 2024-09-25
Improve error message when using custom client for sync vs. async cases


willbakst:
### Description

The error message when using the incorrect client with sync vs. async calls is currently unhelpful. See #513

We should add better error messages that throw a `ValueError` in the event that the function is sync and using an async client and vice versa.

For example, for OpenAI this would happen as an else block check [here](https://github.com/Mirascope/mirascope/blob/ad45151c443d6852f87cc392d097c0c9fbafc17a/mirascope/core/openai/_utils/_setup_call.py#L127)

We should add this for all providers.

willbakst:
Changes from #607 make this better through type/lint errors, but I still think a value error would be useful here.

willbakst:
Upon further thought, I think the type/lint errors are sufficient here, and adding additional value errors would be unnecessary overhead for no additional value, so I'm closing this issue.

Issue #513, opened 2024-09-23
Async not working


anthony2261:
### Description

Hello,

I'm hitting this error when trying to await async calls:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 await dummy()

File .../lib/python3.11/site-packages/mirascope/core/base/_extract.py:105, in extract_factory.<locals>.decorator.<locals>.inner_async(*args, **kwargs)
    101 @wraps(fn)
    102 async def inner_async(
    103     *args: _P.args, **kwargs: _P.kwargs
    104 ) -> _ResponseModelT:
--> 105     call_response = await create_decorator(
    106         fn=fn, **create_decorator_kwargs
    107     )(*args, **kwargs)
    108     json_output = get_json_output(call_response, json_mode)
    109     try:

File .../lib/python3.11/site-packages/mirascope/core/base/_create.py:105, in create_factory.<locals>.decorator.<locals>.inner_async(*args, **kwargs)
     93 create, prompt_template, messages, tool_types, call_kwargs = setup_call(
     94     model=model,
     95     client=client,
   (...)
    102     extract=False,
    103 )
    104 start_time = datetime.datetime.now().timestamp() * 1000
--> 105 response = await create(stream=False, **call_kwargs)
    106 end_time = datetime.datetime.now().timestamp() * 1000
    107 output = TCallResponse(
    108     metadata=get_metadata(fn, dynamic_config),
    109     response=response,
   (...)
    119     end_time=end_time,
    120 )

TypeError: object ChatCompletion can't be used in 'await' expression"
```

Running this code in a jupyter notebook (in VSCode):
```python
from groq import Groq
from mirascope.core import groq, prompt_template

@groq.call(
    model="llama3-8b-8192",
    response_model=str,
    client=Groq(api_key=groq_api_key),
)
@prompt_template("Hello!")
async def dummy():...

await dummy()
```

### Python, Mirascope & OS Versions, related packages (not required)

```TOML
Python 3.11
Mirascope 1.2.0
WSL
```

willbakst:
You're using `Groq` as your custom client, but you need to use `AsyncGroq` instead for async calls.

I am working on some documentation updates and will make sure to include this in the custom client section so it's more clear.

willbakst:
I'm reopening this issue for others to see until we release the updated docs with `v1.3` that make this far more clear.

willbakst:
New docs make this clear!

PR #607, opened 2024-10-08
Improve type safety for client arguments in setup_call functions

koxudaxi:
Related Issues: https://github.com/Mirascope/mirascope/issues/513, #519


mirascope/core/base/_call_factory.py
```diff
@@ -310,7 +490,7 @@ def base_call(
                 json_mode=json_mode,
                 client=client,
                 call_params=call_params,
-            )  # pyright: ignore [reportReturnType]
+            )  # pyright: ignore [reportReturnType, reportCallIssue]
```
koxudaxi, line 493: I think I resolved all types. But, I can't resolve the error.
```bahs
  /Users/koudai/PycharmProjects/mirascope/mirascope/core/base/_call_factory.py:480:20 - error: No overloads for "decorator" match the provided arguments (reportCallIssue)
1 error, 0 warnings, 0 informations 
```

mirascope/core/base/_utils/_protocols.py
```diff
@@ -33,6 +35,46 @@
...
+
+    @overload
+    def __call__(self, fn: Callable[_P, Messages.Type]) -> Callable[_P, _ResponseT]: ...
+
+    def __call__(
+        self, fn: Callable[_P, _BaseDynamicConfigT] | Callable[_P, Messages.Type]
+    ) -> Callable[_P, _ResponseT]: ...  # pragma: no cover
+
+
 class LLMFunctionDecorator(Protocol[_BaseDynamicConfigT, _ResponseT, _AsyncResponseT]):
```
willbakst, line 78: can we get rid of this if we have the sync vs. async and just use those now?

mirascope/core/base/_utils/_protocols.py
```diff
@@ -108,12 +150,87 @@ def __call__(
...
+        CreateFn[_ResponseT, _ResponseChunkT]
+        | AsyncCreateFn[_ResponseT, _ResponseChunkT],
+        str | None,
+        list[Any],
+        list[type[_BaseToolT]] | None,
+        BaseCallKwargs,
+    ]: ...  # pragma: no cover
+
+
+class BaseClientSetupCall(
```
willbakst, line 233: why do we need both the base client setup call and the new protocol? can you explain what this solves so i know why we include both?

mirascope/core/base/_call_factory.py
```diff
@@ -42,7 +47,9 @@
 _BaseCallParamsT = TypeVar("_BaseCallParamsT", bound=BaseCallParams)
 _BaseDynamicConfigT = TypeVar("_BaseDynamicConfigT", bound=BaseDynamicConfig)
 _BaseStreamT = TypeVar("_BaseStreamT", bound=BaseStream)
+_SyncBaseClientT = TypeVar("_SyncBaseClientT", bound=object)
```
willbakst, line 50: nit: move this below the `_BaseClientT` line right below so this is right above it's async counterpart

mirascope/core/base/_utils/_protocols.py
willbakst, line 233: we should name this so it's reason for existing is more clear

`SameSyncAndAsyncClientSetupCall` to it's more clear this is for providers who have a client that is shared across sync + async?

mirascope/core/base/_extract.py
```diff
@@ -21,7 +22,9 @@
 from .tool import BaseTool
 
 _BaseCallResponseT = TypeVar("_BaseCallResponseT", bound=BaseCallResponse)
+_SyncBaseClientT = TypeVar("_SyncBaseClientT", bound=object)
 _BaseClientT = TypeVar("_BaseClientT", bound=object)
```
willbakst, line 26: we could probably then also rename this to `_SameSyncAndAsyncClientT`

mirascope/core/base/_call_factory.py
willbakst: ok, that's fine for now, we can revisit this later if necessary

mirascope/core/base/_utils/_protocols.py
willbakst, line 1: you'll need to update bedrock with this stuff as well, so merge that PR first, then sync here and i'll approve

willbakst: ok i merged that PR in, so this should be ready for the updates

willbakst (approved):

PR #624, opened 2024-10-17
feat: Support OpenAI Realtime tool

koxudaxi:


mirascope/beta/openai/realtime/tool.py
willbakst, line 1: why aren't we just using `OpenAITool` from Mirascope core? their docs say they are 1:1?

mirascope/beta/openai/realtime/realtime.py
```diff
@@ -246,33 +359,61 @@ async def _process_receiver(self, conn: ClientConnection) -> None:
...
+                        if not function_call_arguments_chunks[response_id]:
+                            del function_call_arguments_chunks[response_id]
+                        tool = tool_schema.from_tool_call(
+                            FunctionCallArguments(
+                                call_id=message["call_id"],
+                                arguments=function_call_arguments,
+                            )
+                        )
+                        await self._process_specific_receiver(tool, "tool")
+                        # TODO: type: "function_call_output".
```
willbakst, line 416: as mentioned in my previous comment, the user is required to respond with the `function_call_output` event or it won't work, so we have to support this for tools to work at all

see my other comment on an idea for how to handle this in the interface

examples/beta/openai/realtime/audio_stream_input_output_tools.py
```diff
@@ -0,0 +1,44 @@
...
+
+from typing import AsyncGenerator, Any
+
+from pydub import AudioSegment
+from pydub.playback import play
+
+from mirascope.beta.openai import Realtime, record_as_stream, OpenAIRealtimeTool
+
+app = Realtime(
+    "gpt-4o-realtime-preview-2024-10-01",
```
willbakst, line 12: the user should also be able to set tools at the top level here that will be set in the initial `session.update`

examples/beta/openai/realtime/audio_stream_input_output_tools.py
```diff
@@ -0,0 +1,44 @@
...
+def format_book(title: str, author: str) -> str:
+    return f"{title} by {author}"
+
+
+@app.receiver("audio")
+async def receive_audio(response: AudioSegment, context: dict[str, Any]) -> None:
+    play(response)
+
+
+@app.receiver("audio_transcript")
```
willbakst, line 25: we need to have a receiver that can receive the tool call

from openai docs: "The client must respond to the function call before by sending a conversation.item.create message with type: "function_call_output"."

this means we have to have a means of receiving the tool call and sending the tool call update in one fell swoop (i.e. it's both a receiver and a sender)

what if we did something like:

```python
@app.function_call(format_book)
async def handle_format_book(...) -> str: ...
```

the idea is that this function will handle when the API wants to call the `format_book` tool, and then we'll internally use the `str` response to then send the `function_call_output` event

mirascope/beta/openai/realtime/realtime.py
```diff
@@ -95,7 +136,7 @@ def __init__(
             "tool_choice": tool_choice,
             "temperature": temperature,
             "max_response_output_tokens": max_response_output_tokens,
-            "tools": tools,
+            "tools": self._tool_schemas.values() if self._tool_schemas else NotSet,
```
willbakst, line 139: again why aren't we just using our existing tool schema utilities from `OpenAITool`??

mirascope/beta/openai/realtime/realtime.py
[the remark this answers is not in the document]
willbakst: same thing on `"model"` which needs to be `'model'`

mirascope/beta/openai/realtime/realtime.py
```diff
@@ -14,6 +14,7 @@
     TypeAlias,
     TypedDict,
     TypeVar,
+    overload,
 )
```
willbakst, line 18: also, `NotRequired` needs to be imported from `typing_extensions` for 3.10 support

mirascope/beta/openai/realtime/realtime.py
```diff
@@ -26,41 +27,75 @@
...
 
 class Sender(TypedDict):
-    func: SenderFunc[_ResponseT]
+    func: SenderFunc[_ResponseT, OpenAIRealtimeTool]
     wait_for_text_response: bool
     wait_for_audio_transcript_response: bool
+    tools: list[type[OpenAIRealtimeTool] | Callable]
 
 
 NotSet = object()
```
willbakst, line 84: i don't really get why we need this? can't the user just not set it? what is this doing?

mirascope/beta/openai/realtime/realtime.py
willbakst, line 1: also you have `async for conn in connect(...)` but isn't it supposed to be `async with connect(...) as conn`??

willbakst: oh nevermind i was just having import issues

mirascope/beta/openai/realtime/tool.py
koxudaxi: I initially thought the same, but they are written as the same, yet they are not the same. For example, OpenAITool uses ChatCompletionToolParam, but RealTimeAPI uses different Params from these.

mirascope/beta/openai/realtime/realtime.py
koxudaxi: This is where you save what is passed as a sender decorator option.
Should the tool be set per top level `session.update` rather than per sender (`response.create`)?
```
    def sender(
        self,
        wait_for_text_response: bool = False,
        wait_for_audio_transcript_response: bool = False,
        tools: list[type[OpenAIRealtimeTool] | Callable] | None = None,
```

mirascope/beta/openai/realtime/realtime.py
koxudaxi: https://platform.openai.com/docs/api-reference/realtime-client-events/session/update
image

↑we need the schema for Realtime API. But, `OpenAITool.tool_schema()` retruns `ChatCompletionToolParam`
which is diffrent type.
image

examples/beta/openai/realtime/audio_stream_input_output_tools.py
koxudaxi: It is already supported.
```
class Realtime:
    def __init__(
        self,
        model: str,
        *,
        context: dict[str, Any] | None = None,
        modalities: list[str] | NotSet = NotSet,
        instructions: str | NotSet = NotSet,
        voice: str | NotSet = NotSet,
        turn_detection: dict[str, Any] | None | NotSet = NotSet,
        input_audio_format: str | NotSet = NotSet,
        output_audio_format: str | NotSet = NotSet,
        input_audio_transcription: str | NotSet = NotSet,
        tool_choice: str | NotSet = NotSet,
        temperature: float | NotSet = NotSet,
        max_response_output_tokens: int | float | Literal["inf"] = NotSet,
        tools: list[type[OpenAIRealtimeTool] | Callable] | NotSet = NotSet,
    ) -> None:
 ```

koxudaxi: it's supported

mirascope/beta/openai/realtime/realtime.py
willbakst: as discussed, we should use the `OpenAITool` class and just massage the data as necessary inline to convert

willbakst: actually for now let's just ignore this, we can fix this later when we work to move this out of beta into production

willbakst (approved):

Issue #625, opened 2024-10-17
✨ MultiAgent & SubAgents support


jd-solanki:
### Description

Hi 👋🏻 

I love the mirascope and its simplicity ❤️ 

I choose it over langchain and other frameworks and now it also has suport for structured outputs I also replaced it from [instructor](https://github.com/instructor-ai/instructor).

---

I'm building highly dynamically API where I receive everything as request body like model name, model provider, prompt, tools etc hopefully everything is possible with mirascope. However, I'm stuck at multiple agent system like AutoGen.

I want to know how to implement multi agent system & sub agent system. This is related to [agency swarm](https://github.com/VRSEN/agency-swarm), agent orchestration, task delegation, etc. I seen the docs on agents but it's limited to single agent only.

Thanks 😇

willbakst:
I'm so glad you're liking Mirascope!

Agents are top of mind for us right now, and we're actively working on additional features/functionality around how to build agents (which is why the Learn section does not yet cover the more complex cases, but I promise docs and examples are coming soon!)

In the meantime, one method of writing multi/sub-agent architectures is to use agents as tools. You can see an example of this in our [Blog Writing Agent](https://mirascope.com/docs/tutorials/agents/blog_writing_agent/) tutorial. Let me know if you have any comments or questions!

Of course, this example gives the agent executor and agents a lot of control. The features we're aiming to build will hopefully make it easier to build multi/sub-agent flows where you have more fine-grained control over the flow without as much boilerplate

willbakst:
After a lot of thinking, I've come to the following conclusion:

We should close this issue in favor of individual issues outlining specific examples you'd like to see.

Here's my reasoning:
1. Mirascope is supposed to be a lightweight toolkit for interacting with LLM APIs. Mirascope sits at the single-layer abstraction level so that it can remain simple and clean, and I intend to keep it this way.
2. This means that things like agents (which I consider higher order abstractions) don't really belong as "features" of Mirascope. You can use Mirascope to build agents the same way you could use the raw OpenAI SDK to build agents. Rather than new "features" or additional "support" Mirascope should have more extensive examples of how to build such agents using Mirascope as-is.
3. We should outline specific architectures / examples and work to implement those as Tutorials so that users can see how to build them with Mirascope. To me, this is what "multiagent and subagent support" really looks like in Mirascope since Mirascope already technically has support for this.

[...]

PR #626, opened 2024-10-18
docs: improve types in blog writing agent notebook

jd-solanki:
Hi 👋🏻 

I was playing with writing agent notebook after my issue #625 and realized there are some type issues:
image
```
Type of "append" is partially unknown
Argument type is "list[Unknown]"
```

This is because we don't explicitly provides type for list item of `tools_and_outputs`.

I've made fix that explicitly adds the type for `tools_and_outputs`.

---

I follow strict typing in all of my projects to get type hints everywhere, I recommend we add following in vscode `settings.json`
```py
"python.analysis.typeCheckingMode": "strict",
```

This is because even with 30 sec tutorial we're not 100% type safe:
image

Thanks ❤️


willbakst:
Hmm, we've put a lot of effort into type safety, and we ensure every PR that gets merged passes Pyright, so I'm not sure what is going on here, particularly for the `call` decorator.

What type checker are you using here?

willbakst:
Looks like there is an issue from a previous PR. We are going to work to resolve those type errors that should never have been the case.

jd-solanki:
> _What type checker are you using here?_

I'm using VSCode + Pylance + mypy

If you create new project using uv and add mypy & use vscode + pylance you'll get the same error.

Most of the time these errors are from missed types of list item, forgot to pass down generic, etc.

Moreover, as you can see in my first screenshot, implicit type of `tools_and_outputs` is list[Any]` which is not ideal and technically we should write type for list item as well.

This is my first PR and if you want I'll submit more PRs regarding types as I get aware of it. I really love the simplicity of mirascope and will be using from now on.

Awaiting your response.

willbakst:
So for maintainability we've created a `call_factory` method for creating the `call` decorator for each provider. Originally we were relying heavily on nested typing overloads to dynamically generate the type based on variable types, but apparently the new version of Pyright (which Pylance uses) no longer supports this (or I guess considered support a bug and removed support 🤦 )

We've implemented a fix to use protocols instead so that the correct typing returns in this newer version. Once we merge in that fix and release it, I would love to see if you're still having the same typing issues here. I'm hopeful that this fix will return the proper type hints we expect in the examples.

As for PRs / contributions, they are always welcome with open arms! This particular PR may no longer be necessary with the fix I mentioned, but definitely keep the PRs coming!

willbakst:
We've also labeled a bunch of issues with the "good first issue" and "hacktoberfest" labels, so if you're looking for a well-defined PR to work on, those are a great place to start!

We're available to answer any questions you may have along the way :)

jd-solanki:
Thanks for your time and efforts. Closing this one.

Finally, Can you let me know where should I track the fix you mentioned so I can check it once landed in official release.

Thanks.

willbakst:
[This PR](https://github.com/Mirascope/mirascope/pull/627) contains the fix, which I just released in `v1.5`

As you mentioned, while this fixes the issue with the `call` decorator, it still doesn't change anything on the `tools_and_outputs` list (since the type there is inferred and not explicit).

If we're going to add this additional typing, we should add it for every single example, not just this tutorial. Namely, all of the files in `examples/learn/agents/` would need updating as well as the files in `examples/learn/tools/tool_message_params`.

jd-solanki:
> _we should add it for every single example_

I left this decision up to you. If you want I can update this PR because it's simple but lengthy.

I tried v1.5.0 and I'm no longer getting error on `.call()` however while exploring more I found missing type arg for generic error on tool. I guess we should open seperate issue for dicussing this type issues instead of in this PR. WDYS?
image

Awaiting your response 😇

willbakst:
Glad the `call` decorator type errors are fixed! PR for updating these examples with improved types on `tools_and_outputs` are welcome if you believe they will add value. One reason we've left some type hints out is for brevity/readability of examples, but you're welcome to add more clarity here!

As for the `BaseTool` typing, this is an interesting point. As a user, you'll never actually use or instantiate a `BaseTool` instance because it's an abstract class. We've included the generic for the tool schema type for each specific provider (e.g. OpenAITool).

You could use `Any` as the type in your definition (e.g. `class GetBookAuthor(BaseTool[Any])`) but using `Any` is generally bad practice.

One reason we haven't marked the `tool_schema` method as abstract is that the user should not need to define it. Curious for your thoughts here.

jd-solanki:
Hi,

Thanks for your patience. I reflected on your thoughts and realized I should first understand the codebase and then make large PRs.

For the rest of the comments I'll open new issue if anything found.

Thanks, Closing this for now.

PR #629, opened 2024-10-21
feat: Support gpt-4o-audio-preview

koxudaxi:


mirascope/core/base/_utils/_convert_messages_to_message_params.py
```diff
@@ -49,6 +63,24 @@ def _convert_message_sequence_part_to_content_part(
...
+                .set_channels(CHANNELS)
+                .set_sample_width(SAMPLE_WIDTH)
+                .export(format="wav"),
+            ).read(),
+        )
+    elif isinstance(message_sequence_part, Wave_read):
+        return AudioPart(
+            type="audio",
+            media_type="audio/wav",
+            audio=message_sequence_part.readframes(-1),
```
koxudaxi, line 82: `Wave_read` dones't support resampling. we use input without any conversion.

mirascope/core/base/types.py
```diff
@@ -0,0 +1,22 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pydub import AudioSegment
+
+    has_pydub_module: bool
+else:
+    try:
+        from pydub import AudioSegment  # pyright: ignore [reportAssignmentType]
```
koxudaxi, line 9: How do you feel this file name?

mirascope/core/openai/call_params.py
```diff
@@ -11,17 +13,36 @@
...
+    try:
+        from openai.types.chat.chat_completion_audio_param import (  # pyright: ignore [reportMissingImports]
+            ChatCompletionAudioParam,
+            ChatCompletionModality,
+        )
+    except ImportError:
+
+        class ChatCompletionAudioParam: ...
+
+        class ChatCompletionModality: ...
```
koxudaxi, line 31: The audio model is supported only the latest OpenAI SDK.
Therefore, we import it here to support older sdk versions.

uv.lock
```diff
@@ -3973,7 +3986,7 @@ wheels = [
 
 [[package]]
 name = "openai"
-version = "1.42.0"
+version = "1.52.0"
```
koxudaxi, line 3989: I updated the lock file for unittest

koxudaxi:
@willbakst 
I felt that raising the minimum required version of the OpenAI SDK to support audio was excessive, so I decided to support older versions as well.

mirascope/core/openai/call_response_chunk.py
```diff
@@ -87,3 +92,23 @@ def output_tokens(self) -> int | None:
...
+        if (audio := self._audio) and (audio_data := audio.get("data")):
+            return base64.b64decode(audio_data)
+        return None
+
+    @property
+    def audio_transcript(self) -> str | None:
+        """Returns the transcript of the audio content."""
+        if audio := self._audio:
+            return audio.get("transcript")
+        return None
```
koxudaxi, line 114: OpenAI SDK didn't define `audio` field in `ChoiceDelta`.
https://github.com/openai/openai-python/blob/fc4fce30aff7620c32e57301d3c40b4883810825/src/openai/types/chat/chat_completion_chunk.py#L59-L78
But, actual response has `audio` dict.
So, I added audio and audio_transcript in `OpenAICallResponseChunk` by using dict().get()

The field is not even defined in their openapi.
If you should not use non-public objects, remove this one.

https://github.com/openai/openai-python/blob/main/.stats.yml

docs/learn/calls.md
```diff
@@ -161,6 +161,72 @@ While Mirascope provides a consistent interface, you can also always access the
...
+- `audio`: Configuration for the audio output (voice, format, etc.)
+- `modalities`: List of output modalities to receive (e.g. `["text", "audio"]`)
+
+For providers that support audio outputs (currently only OpenAI), you can receive both text and audio responses from your calls:
+
+!!! mira ""
+
+    {% for method, method_title in zip(prompt_writing_methods, prompt_writing_method_titles) %}
+    === "{{ method_title }}"
+        {% for provider in ["OpenAI"] %}
```
koxudaxi, line 195: Instead of placing many python files marked `#Not Support`, I set only OpenAI.

examples/learn/calls/multi_modal_outputs/openai/base_message_param.py
```diff
@@ -0,0 +1,36 @@
...
+
+response = recommend_book(genre="fantasy")
+
+print(response.audio_transcript)
+
+if audio := response.audio:
+    audio_io = io.BytesIO(audio)
+
+    with wave.open(audio_io, "rb") as f:
+        audio_segment = AudioSegment.from_raw(
```
willbakst, line 29: can we not just do `AudioSegment.from_raw(audio_io)`?

examples/learn/prompts/multi_modal/audio/messages.py
```diff
@@ -1 +1,23 @@
-# Coming soon...
+import wave
+from mirascope.core import prompt_template, Messages
+
+
+@prompt_template()
+def identify_book_prompt(audio_wav: wave.Wave_read) -> Messages.Type:
```
willbakst, line 6: what about shorthand for non-wave audio files? like mp3?

it seems like we should either use `pydub.AudioSegment` or we should add additional tabs showing both `pydub.AudioSegment` and `wave.Wave_read`?

mirascope/core/base/_utils/_convert_messages_to_message_params.py
```diff
@@ -49,12 +63,32 @@ def _convert_message_sequence_part_to_content_part(
...
+        )
     else:
         raise ValueError(f"Invalid message sequence type: {message_sequence_part}")
 
 
 def convert_message_content_to_message_param_content(
-    message_sequence: Sequence[str | Image.Image | TextPart | ImagePart | AudioPart],
+    message_sequence: Sequence[
+        str | Image.Image | TextPart | ImagePart | AudioPart | AudioSegment | Wave_read
```
willbakst, line 90: nit: here and below, can we reorder so `Image.Image` is after `ImagePart` so the different input types are colocated better?

mirascope/core/base/_utils/_convert_messages_to_message_params.py
```diff
@@ -75,7 +109,9 @@ def _is_base_message_params(
 
 def convert_messages_to_message_params(
     messages: str
-    | Sequence[str | Image.Image | TextPart | ImagePart | AudioPart]
+    | Sequence[
+        str | Image.Image | TextPart | ImagePart | AudioPart | AudioSegment | Wave_read
```
willbakst, line 113: ditto

mirascope/core/base/messages.py
```diff
@@ -2,26 +2,45 @@
...
+from .types import AudioSegment
 
 
 class Messages:
     Type = (
         str
-        | Sequence[str | Image.Image | TextPart | ImagePart | AudioPart]
+        | Sequence[
+            str
+            | Image.Image
```
willbakst, line 20: nit: same there and and below, colocate `Image.Image` by putting it right after `ImagePart`

mirascope/core/base/types.py
```diff
@@ -0,0 +1,22 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pydub import AudioSegment
```
willbakst, line 4: won't this fail if pydub isn't installed?

mirascope/core/base/types.py
willbakst: i think it's fine, we should also move the `Image.Image` thing here too then right?

mirascope/core/openai/__init__.py
```diff
@@ -14,7 +16,9 @@
 from .stream import OpenAIStream
 from .tool import OpenAITool, OpenAIToolConfig
 
-OpenAIMessageParam: TypeAlias = ChatCompletionMessageParam | BaseMessageParam
+OpenAIMessageParam: TypeAlias = (
+    ChatCompletionMessageParam | BaseMessageParam | AudioSegment | Wave_read
```
willbakst, line 20: this seems wrong? `OpenAIMessageParam` should only be `ChatCompletionMessageParam | BaseMessageParam` since audio/image etc. inputs/outputs should be converted to message parameters by the time this type is every used

mirascope/core/openai/call_params.py
```diff
@@ -11,17 +13,36 @@
 
 from ..base import BaseCallParams
 
+if TYPE_CHECKING:
+    from openai.types.chat.chat_completion_audio_param import (  # pyright: ignore [reportMissingImports]
```
willbakst, line 17: again, won't this fail if it's not the right version?

mirascope/core/openai/call_response.py
```diff
@@ -173,3 +188,23 @@ def tool_message_params(
             )
             for tool, output in tools_and_outputs
         ]
+
+    @computed_field
+    @property
+    def _audio(self) -> ChatCompletionAudio | None:
```
willbakst, line 194: I don't quite see the value of having this as a computed field since it's just access and not constructing or doing anything.

I think instead we should just call this directly in the `audio` and `audio_transcript` properties, and we should make both of those properties computed fields so they appear in the model dump

mirascope/core/openai/call_response_chunk.py
willbakst: hmm, have you tested that it actually streams the audio back? if it does, then I think this is fine?

same thing here though, update the `audio` and `audio_transcript` properties as mentioned in my comment on the standard call response

mirascope/core/base/types.py
koxudaxi: OK, I forgot `from __future__ import annotations` in this file.

koxudaxi: If TYPE_CHECKING: is only evaluated during type checking, so Python runtime will never throw an error. (Because it is not evaluated forever )
We did same thing for PIL.Image. But, it works fine without gemini instlation option.

https://github.com/Mirascope/mirascope/blob/19a6b76e487c9e2c723be4713ec49cca7c6862b8/mirascope/core/base/_utils/_convert_messages_to_message_params.py#L18-L22

mirascope/core/openai/call_params.py
koxudaxi: Same above

mirascope/core/openai/__init__.py
koxudaxi: you are right. I fixed it.

mirascope/core/openai/call_response_chunk.py
koxudaxi: Yes, I tested this code. It works fine.

### Code:
```python
import io
import wave

from pydub.playback import play

from mirascope.core import Messages, openai
from mirascope.core.openai import AudioSegment


@openai.call(
    "gpt-4o-audio-preview",
    call_params={"audio": {"voice": "alloy", "format": "pcm16"}, "modalities": ["text", "audio"],},
    stream=True)
def recommend_book(genre: str) -> Messages.Type:

audio_chunk = b''
audio_transcript_chunk = ''
for chunk, _ in recommend_book("fantasy"):
    if chunk.audio:
        audio_chunk += chunk.audio
    if chunk.audio_transcript:
        audio_transcript_chunk += chunk.audio_transcript

print(audio_transcript_chunk)

SAMPLE_WIDTH = 2
FRAME_RATE = 24000
CHANNELS = 1

audio_segment = AudioSegment.from_raw(
        io.BytesIO(audio_chunk),
        sample_width=SAMPLE_WIDTH,
        frame_rate=FRAME_RATE,
        channels=CHANNELS,
    )
play(audio_segment)
```

### Outputs:
```bash
I highly recommend "The Name of the Wind" by Patrick Rothfuss. It's a beautifully written novel about a young man named Kvothe who grows up to become a legendary figure in the world of magic. With rich world-building and memorable characters, it's the first book in a planned trilogy called "The Kingkiller Chronicle."
Input #0, wav, from '/var/folders/1c/6xfc8h09781fp0zdndc007xh0000gn/T/tmpuk8lql66.wav':
  Duration: 00:00:22.55, bitrate: 384 kb/s
  Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 24000 Hz, 1 channels, s16, 384 kb/s
```

examples/learn/calls/multi_modal_outputs/openai/base_message_param.py
koxudaxi: We can't play without parameters like `sample_width`

```python
    audio_io = io.BytesIO(audio)

    with wave.open(audio_io, "rb") as f:
        audio_segment = AudioSegment.from_raw(audio_io)
```

```bash
  File "/Users/koudai/PycharmProjects/mirascope/examples/learn/calls/multi_modal_outputs/openai/messages.py", line 29, in <module>
    audio_segment = AudioSegment.from_raw(audio_io)
  File "/Users/koudai/PycharmProjects/mirascope/.venv/lib/python3.10/site-packages/pydub/audio_segment.py", line 812, in from_raw
    return cls.from_file(file, 'raw', sample_width=kwargs['sample_width'], frame_rate=kwargs['frame_rate'],
KeyError: 'sample_width'
```

examples/learn/calls/multi_modal_outputs/openai/messages.py
```diff
@@ -26,11 +26,6 @@ def recommend_book(genre: str) -> Messages.Type:
     audio_io = io.BytesIO(audio)
 
     with wave.open(audio_io, "rb") as f:
```
willbakst, line 28: i thought you said this doesn't work unless you provide sample width etc?

also seems weird to be using both the `wave` package and `AudioSegment` and to be importing `AudioSegment` from `mirascope.core.openai` instead of from pydub?

mirascope/core/openai/call_response.py
```diff
@@ -189,22 +189,16 @@ def tool_message_params(
...
-
     @property
     def audio(self) -> bytes | None:
         """Returns the audio data of the response."""
-        if audio := self._audio:
+        if audio := getattr(self.response.choices[0].message, "audio", None):
             return base64.b64decode(audio.data)
         return None
 
     @property
```
willbakst, line 199: shouldn't these both be computed fields so they show up in the dump?

mirascope/core/openai/call_response_chunk.py
```diff
@@ -93,22 +92,18 @@ def output_tokens(self) -> int | None:
             return self.usage.completion_tokens
         return None
 
-    @computed_field
-    @property
-    def _audio(self) -> dict[str, Any] | None:
-        """Returns the audio of the response."""
-        return getattr(self.chunk.choices[0].delta, "audio", None)
-
     @property
```
willbakst, line 95: same here

docs/learn/prompts.md
```diff
@@ -137,7 +137,7 @@ While Mirascope provides a consistent interface, support varies among providers:
 |---------------|-----------|--------|--------|------|---------|--------|
 | text          | ✓         | ✓      | ✓      | ✓    | ✓       | ✓      |
 | image         | ✓         | -      | ✓      | ✓    | ✓       | ✓      |
-| audio         | -         | -      | ✓      | -    | -       | -      |
+| audio         | -         | -      | ✓      | -    | -       | ✓      |
```
willbakst, line 140: i think we should add a little blurb here that the shorthand can handle both `pydub.AudioSegment` and `wave.Wave_read`? since we only show the `pydub.AudioSegment` example I worry users won't know they can use wave directly

either that or we should add tabs for wave vs. pydub as i mentioned before so they can see both examples

examples/learn/calls/multi_modal_outputs/openai/messages.py
koxudaxi: > i thought you said this doesn't work unless you provide sample width etc?

It works for only wave file.

> also seems weird to be using both the wave package and AudioSegment and to be importing AudioSegment from mirascope.core.openai instead of from pydub?

I fixed it

docs/learn/calls.md
willbakst, line 1: you'll need to make similar updates to the `streams.md` documentation to show how we stream audio from OpenAI with examples (same thing re: table)

willbakst (approved):

PR #633, opened 2024-10-24
chore: added mirascope_call to all call decorators

brenkao:


mirascope/core/base/_create.py
```diff
@@ -187,6 +187,7 @@ async def inner_async(
                 output._model = model
                 return output if not output_parser else output_parser(output)
 
+            inner_async.__mirascope_call__ = True  # pyright: ignore [reportAttributeAccessIssue]
```
willbakst, line 190: instead of doing this for both sync and async, just set this on the fn (see `fn._model` for reference)

same for all other cases

brenkao: done

willbakst (approved):

PR #638, opened 2024-10-25
Add workflow

koxudaxi:


mirascope/beta/workflows/__init__.py
willbakst, line 1: realization while reading through this:

what about joins that come from separate steps? For example, what if we want something like `def final_step(first_output: str, second_output: str) -> str: ...` where `first_output` is from one step and `second_output` is from another step that run in parallel?

willbakst:
Closing this for now as further discussion indicates this is likely not the best place for workflow stuff to live.

Issue #639, opened 2024-10-26
Functionality to Convert LangChain Tools to Mirascope Tools


minhquan23102000:
### Question

**Description:**

Currently, developers who want to use tools  within Mirascope need to reimplement them. This can be time-consuming and inefficient. It would be beneficial to have a feature or utility that automatically converts LangChain tools into Mirascope-compatible tools.

Langchain support wide range of tools and toolkit: https://python.langchain.com/docs/integrations/tools/

**Motivation:**

This feature would significantly simplify the development process for users who want to leverage the extensive collection of tools available in LangChain within the Mirascope framework. It would reduce development time and effort, allowing developers to focus on building their applications rather than reimplementing existing tools.

**Examples:**

* A function or script that takes a LangChain tool object as input and outputs a Mirascope-compatible tool object.
* A documentation page outlining the process of converting LangChain tools to Mirascope tools manually, if automatic conversion is not feasible.

**Benefits:**

* Reduced development time and effort.
* Increased accessibility of LangChain tools within Mirascope.
* Improved developer experience.

**Alternatives:**

Currently, the only alternative is to manually reimplement tools in Mirascope. This is a time-consuming and potentially error-prone process.

willbakst:
There are two things here that I think are worth digging into:

__1. Pre-made Tools__

From your question/comment, it seems worthwhile for Mirascope to finally implement the `mirascope.tools` module with a bunch of pre-made ready-to-go tools for users so they can get up and running quickly without having to write their own tools.

Is there a list of specific tools you would like to see added to this module first? That way we can prioritize those tools in the upcoming work. You can also create individual GitHub Issue Feature Requests for any particular tools you want to see in the library and we (or other community members) can work to include them :)

__2. How to use existing LangChain tools with Mirascope__

To be honest, I looked into implementing at least something makeshift for conversion, but their tools are all structured and implemented differently (and poorly imo), so the conversion script would be quite time consuming (and break if they make any changes to any tools).

I would much rather focus on (1) and implement the `mirascope.tools` module / suite of pre-made tools for our community, as I think this will ultimately be much better in the long run.

In the meantime, here's the best way to manually use a LangChain tool with Mirascope:

1. Create a Mirascope `BaseTool` subclass
2. Instantiate the LangChain tool as a class variable of your Mirascope tool
3. Set `__name__` equal to the LangChain tool's `.__class__.__name__` value
4. Override the `_name` and `_description` class methods using the LangChain tool's name/description field values
[...]

PR #642, opened 2024-10-28
feat: Add Mirascope tools

koxudaxi:
Progress on #639


mirascope/tools/search/duckduckgo.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Enter search queries to get URLs from search results.
+    """
+
+    queries: list[str] = Field(..., description="List of search queries")
+
+    def call(self) -> str:
+        try:
+            urls = []
+            for query in self.queries:
```
willbakst, line 44: this is just getting the URLs, but there is other useful information contained in the search result such as the title and body of the result. I would include those, or you could enable a config option around what to include in the results

mirascope/tools/base.py
```diff
@@ -0,0 +1,89 @@
...
+from pydantic import BaseModel, ConfigDict, create_model
+
+from mirascope.core import BaseTool
+
+
+class _ToolConfig(BaseModel, ABC):
+    """Base configuration for tools"""
+
+    @classmethod
+    @abstractmethod
```
willbakst, line 15: seems like rather than making this abstract it should just return `cls()` by default? then tools that need to override this can, but those that don't then don't have to implement this method

mirascope/tools/base.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @classmethod
+    def _config(cls) -> ConfigT:
+        """Get tool configuration"""
+        return cast(ConfigT, cls.__config__)
+
+    @classmethod
+    def from_config(cls, config: _ToolConfig) -> type[ConfigurableTool]:
+        """Create tool class with custom configuration"""
+        config = cls.__config__.override_defaults(config)
```
willbakst, line 50: hmm, seems like this is the only place we use this, I wonder if having a method for this is a bit overkill, why not just call `config = cls.model_validate(config)` here instead?

mirascope/tools/base.py
```diff
@@ -0,0 +1,89 @@
...
+            cls.__name__,
+            __base__=cls,
+            __module__=cls.__module__,
+        )
+        NewModel.__prompt_usage_description__ = cls.__prompt_usage_description__
+        NewModel.__config__ = config
+        return NewModel
+
+    @classmethod
+    def _get_base_system_prompt(cls) -> list[str]:
```
willbakst, line 61: these methods are providing too much of the prompt.

The idea here is really just that this contains instructions for how to use the tool (and that's it). The rest is up to the user. For example, they may write their own system prompt and then inject the usage instructions for multiple tools (which would make these methods not work well together in the multiple tools case).

Instead, there should just be a `usage_description` classmethod that returns just the usage instructions for this tool (e.g. "You can use this tool to search the web...")

mirascope/tools/base.py
```diff
@@ -0,0 +1,89 @@
...
+
+from mirascope.core import BaseTool
+
+
+class _ToolConfig(BaseModel, ABC):
+    """Base configuration for tools"""
+
+    @classmethod
+    @abstractmethod
+    def from_env(cls) -> _ToolConfig:
```
willbakst, line 16: in line with my other comment, could you make this generic? e.g. 
```python
def from_env(cls: type[_ToolConfigT]) -> _ToolConfigT:
    return cls()
```

mirascope/tools/base.py
```diff
@@ -0,0 +1,89 @@
...
+        """Create config with overridden default values"""
+        return cls.model_validate(config)
+
+
+ConfigT = TypeVar("ConfigT", bound=_ToolConfig)
+
+_ToolSchemaT = TypeVar("_ToolSchemaT")
+
+
+class ConfigurableTool(BaseTool[_ToolSchemaT], Generic[ConfigT, _ToolSchemaT], ABC):
```
willbakst, line 30: why make a whole new class? Why not just extend `BaseTool` to have this functionality?

mirascope/tools/requests.py
```diff
@@ -0,0 +1,48 @@
...
+
+    timeout: int = Field(default=5, description="Request timeout in seconds")
+
+    @classmethod
+    def from_env(cls) -> "RequestsConfig":
+        return cls()
+
+
+class Requests(ConfigurableTool):
+    """Tool for making HTTP requests to fetch content from URLs."""
```
willbakst, line 20: this docstring will be inserted into the prompt as the tool's description, so it's important that we do a good job of making this clear while also minimizing tokens. The more clear this docstring, the better the tool will likely work with the LLM.

For example, you say `fetch content...` but the `method` field below includes POST, PUT, and DELETE, which seem like not just fetch? The language here needs to be clear so the LLM uses the tool properly, so just want to make sure this point is clear and we write really good docstrings / descriptions for our pre-made tools

mirascope/tools/web/__init__.py
willbakst, line 1: hmm, what's the difference between "search" and "web" here? Seems like parsing url content would still fall under search for the most part? And the search tools are web search tools, so they could fall under web? The separation seems unclear. Also, `requests.py` isn't included in these subfolders, I'm wondering if (at least for now) we just make all the modules private (e.g. `_duckduckgo.py`) and make it clear that users should only ever import tools directly from `mirascope.tools`? That way we can organize them into submodules however we want later? WDYT?

mirascope/tools/search/base.py
```diff
@@ -0,0 +1,29 @@
...
+
+class SearchToolBase(
+    ConfigurableTool[_ConfigT, _ToolSchemaT], Generic[_ConfigT, _ToolSchemaT], ABC
+):
+    """
+    Abstract base class for search-related tools.
+    Includes search history management and date-aware prompts.
+    """
+
+    @classmethod
```
willbakst, line 19: given my previous comment, this base search tool is not needed.

each tool should get it's own usage instructions (at least for now)

we can always refactor later if we find we're duplicating a lot of the same instructions

mirascope/tools/web/parse_url_content.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class ParseURLContent(ConfigurableTool[ParseURLConfig, _ToolSchemaT]):
+    """Tool for parsing and extracting main content from URLs.
+    Fetches content from URL, removes unnecessary elements like scripts, styles, navigation, etc.,
+    and returns clean text content from the webpage's main body.
+    """
+
+    __config__ = ParseURLConfig()  # pyright: ignore [reportCallIssue]
+    __prompt_usage_description__: ClassVar[str] = """
```
willbakst, line 42: this reads like usage description for a human reader, but it's supposed to be a usage description for an LLM. Also, we want to make sure that it is succinct and clear (since more tokens cost more money, so we want to do as good a job as possible in as few tokens as possible).

willbakst, line 42: nit: this docstring contains a bunch of additional indentation that should be removed with `inspect.cleandoc` right?

mirascope/tools/search/duckduckgo.py
```diff
@@ -0,0 +1,61 @@
...
+                results = DDGS(proxies=None).text(
```
willbakst, line 45: realizing here that we somehow need to support both sync and async, particularly for tools with I/O bound operations.

Generally with `BaseTool` the user would just define `call` as async and call it a day, so for pre-made tools it seems like the best option would be to implement an async version of the tools (e.g. `AsyncDuckDuckGoSearch`)

WDYT?

mirascope/tools/base.py
koxudaxi: From my understanding, I'm already extending (inheriting from) the existing `BaseTool`. While I'm indeed introducing new Generic variables due to typing constraints, isn't "extend" referring to class inheritance in this context?

mirascope/tools/search/duckduckgo.py
koxudaxi: Makes sense.I will add it soon.

mirascope/tools/web/parse_url_content.py
koxudaxi: I wrap the description with `inspect.cleandoc`
```python
    @classmethod
    def usage_description(cls) -> str:
        """Returns instructions for using this tool."""
        return inspect.cleandoc(cls.__prompt_usage_description__)

```

mirascope/tools/base.py
willbakst: sorry, to phrase my comment better, I'm wondering why we aren't adding this functionality directly to `BaseTool`? I don't necessarily see the value in having two separate classes that we have to maintain?

mirascope/tools/web/_httpxs.py
```diff
@@ -0,0 +1,158 @@
...
+class AsyncHTTPXConfig(_ToolConfig):
+    """Configuration for async HTTPX requests"""
+
+    timeout: int = Field(
+        default=5,
+        description="Request timeout in seconds. When None, no timeout will be applied",
+    )
+
+
+class AsyncHTTPX(ConfigurableTool):
```
willbakst, line 92: nit: same thing here, what about a `_BaseHTTPX` class with a shared config etc.?

For example, the usage line about "performed asynchronously" doesn't really matter to the LLM

mirascope/tools/base.py
willbakst: actually, scratch this. I think you're right to have the separate `ConfigurableTool` class.

We can also push this into `BaseTool` later if we want for whatever reason, but likely better to keep them separate for now

mirascope/tools/base.py
```diff
@@ -0,0 +1,66 @@
...
+):
+    """
+    Abstract base class for configurable tools.
+
+    Subclasses must define a `__prompt_usage_description__` class variable
+    and __config__ class variable with a subclass of _ToolConfig.
+    """
+
+    __config__: ClassVar[_ToolConfig]
+    __prompt_usage_description__: ClassVar[str]
```
willbakst, line 38: I think we need to decide on some form of "standard" for the prompt usage descriptions so that all pre-made tools can be used the same way. Right now I'm noticing that the usage descriptions don't necessarily follow a standard structure or format.

What do you think about something like this:

```
"""
- `{TOOL_NAME}`: {short_description}
   - {bullets describing use} 
"""
```

For example, I imagine the ParseURLContent usage description would look something like this:

```
"""
- `ParseURLContent`: Returns the given URL's main content as clean text
    - Fetches the raw HTML content for the given URL
    - Removes all unwanted elements (scripts, styles, navigation, etc.)
    - Attempts to find the main content section
    - If no main content section is found, extracts all text content
    - Returns cleaned and formatted text content
    - If any errors occur during fetching or parsing, and error message is returned.
"""
```

WDYT?

mirascope/tools/web/_parse_url_content.py
```diff
@@ -0,0 +1,104 @@
...
+        description="parser to use for parsing HTML content in BeautifulSoup",
+        examples=["html.parser", "lxml", "lxml-xml", "html5lib"],
+    )
+    timeout: int = Field(
+        default=5,
+        description="Timeout in seconds for URL request",
+    )
+
+    @classmethod
+    def from_env(cls) -> ParseURLConfig:
```
willbakst, line 27: same thing here, for these default ones we can remove this method because the base is generic

mirascope/tools/web/_duckduckgo.py
```diff
@@ -0,0 +1,101 @@
...
+    max_results_per_query: int = Field(
+        default=2, description="Maximum number of results per query"
+    )
+
+    @classmethod
+    def from_env(cls) -> _ToolConfig:
+        return cls()
+
+
+_ToolSchemaT = TypeVar("_ToolSchemaT")
```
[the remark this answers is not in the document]
koxudaxi: OK, I remove `_ToolSchemaT`. I think we don't need the type for ConfigurableTool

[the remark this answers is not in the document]
koxudaxi:
```python
../mirascope/tools/base.py:24: in <module>
    class ConfigurableTool(BaseTool, Generic[_ToolConfigT], ABC):
../.venv/lib/python3.10/site-packages/pydantic/_internal/_model_construction.py:175: in __new__
    raise TypeError(error_message)
E   TypeError: All parameters must be present on typing.Generic; you should inherit from typing.Generic[~_ToolSchemaT, ~_ToolConfigT].
```

I've tried everything, but it seems I need to define all Generic types.

willbakst (approved):

Issue #643, opened 2024-10-28
Option to reuse custom client with a call decorator


tux-type:
### Description

As shown in the example below, to use a custom client one has to create the instance at the module level, which is then reused within the module. However, when importing this module elsewhere this also triggers the `client = OpenAI()` call.

```
from mirascope.core import openai
from openai import OpenAI

client = OpenAI(
    api_key="ollama",
    base_url="http://localhost:11434/v1",
)

@openai.call("MODEL_NAME_HERE", client=client)
def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book"
```

\
\
It would be good to have an option to create or otherwise specify the custom client without triggering the `OpenAI()` call on import. One option is to use partial:
```
from mirascope.core import openai
from openai import OpenAI
from functools import partial

client = partial(
    OpenAI, 
    api_key="ollama",
    base_url="http://localhost:11434/v1",
)

@openai.call("MODEL_NAME_HERE", client=client())
def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book"
```

However, this ends up calling the creation of the client on every function call. This might get somewhat expensive with many calls (?).

It would be nice to be able to lazy load the client once on first call, then reuse subsequent times. Bonus if this custom client is easily reusable between modules.

willbakst:
I think the best way to solve this would be to implement a feature for setting the client through the dynamic config.

```python
from mirascope.core import BaseDynamicConfig, Messages, openai
from openai import OpenAI

@openai.call("gpt-4o-mini")
@prompt_template("Recommend a {genre} book")
def recommend_book(genre: str) -> BaseDynamicConfig:
    return {
        "messages": [Messages.User(f"Recommend a {genre} book")],
        "client": OpenAI(...),
    }
```

This would enable passing the client to the function as an argument, which I believe would satisfy all of your needs:

```python
from mirascope.core import BaseDynamicConfig, Messages, openai
from openai import OpenAI

@openai.call("gpt-4o-mini")
@prompt_template("Recommend a {genre} book")
def recommend_book(genre: str, *, client: OpenAI | None = None) -> BaseDynamicConfig:
    if not client:
        client = OpenAI()
    return {
        "messages": [Messages.User(f"Recommend a {genre} book")],
        "client": client,
    }
```

If this works, let me know and we'll work to get this included in the `v1.7` release this Friday :)

tux-type:
Yes, this looks good. Impressed by how fast this is being worked on!

[...]

PR #644, opened 2024-10-29
feat: Support dynamic client

koxudaxi:
Implements #643


mirascope/core/base/_create.py
```diff
@@ -156,10 +156,15 @@ async def inner_async(
                 *args: _P.args, **kwargs: _P.kwargs
             ) -> TCallResponse | _ParsedOutputT:
                 fn_args = get_fn_args(fn, args, kwargs)
-                dynamic_config = await get_dynamic_configuration(fn, args, kwargs)
+                if (
+                    dynamic_config := await get_dynamic_configuration(fn, args, kwargs)
+                ) and isinstance(dynamic_config, dict):
```
koxudaxi, line 161: pyright request `isinstance` for async

mirascope/core/base/_create.py
```diff
@@ -196,10 +201,15 @@ def inner(
                 *args: _P.args, **kwargs: _P.kwargs
             ) -> TCallResponse | _ParsedOutputT:
                 fn_args = get_fn_args(fn, args, kwargs)
-                dynamic_config = get_dynamic_configuration(fn, args, kwargs)
+                if (
+                    dynamic_config := get_dynamic_configuration(fn, args, kwargs)
+                ) and "client" in dynamic_config:
```
koxudaxi, line 206: pyright requests "client" in dynamic_config for TypedDict

mirascope/core/base/_create.py
```diff
@@ -156,10 +161,15 @@ async def inner_async(
                 *args: _P.args, **kwargs: _P.kwargs
             ) -> TCallResponse | _ParsedOutputT:
                 fn_args = get_fn_args(fn, args, kwargs)
-                dynamic_config = await get_dynamic_configuration(fn, args, kwargs)
+                if (
+                    dynamic_config := await get_dynamic_configuration(fn, args, kwargs)
+                ) and isinstance(dynamic_config, dict):
+                    dynamic_client = dynamic_config.get("client")
+                else:
```
willbakst, line 168: why bother with the else clause here? why not just set `client = dynamic_config.get("client")` in the if statement?

mirascope/core/base/_create.py
willbakst: why do we check this here but check for dict instance in async? why not the same?

mirascope/core/base/_create.py
```diff
@@ -196,10 +206,15 @@ def inner(
                 *args: _P.args, **kwargs: _P.kwargs
             ) -> TCallResponse | _ParsedOutputT:
                 fn_args = get_fn_args(fn, args, kwargs)
-                dynamic_config = get_dynamic_configuration(fn, args, kwargs)
+                if (
+                    dynamic_config := get_dynamic_configuration(fn, args, kwargs)
+                ) and "client" in dynamic_config:
+                    dynamic_client = dynamic_config.get("client")
+                else:
```
willbakst, line 213: ditto here on the else

mirascope/core/base/stream.py
willbakst, line 1: you need to make the same updates here as you did for `_create` and `_extract` to get the potential dynamic client

mirascope/core/base/_extract.py
```diff
@@ -126,6 +129,12 @@ async def inner_async(
                 fields_from_call_args = get_fields_from_call_args(
                     response_model, fn, args, kwargs
                 )
+                if (
```
willbakst, line 132: is this necessary? even if the client is set in the decorator, if it's set in the dynamic config it'll get overwritten, so wouldn't `create_decorator` handle this for us already?

mirascope/core/base/structured_stream.py
```diff
@@ -262,6 +265,12 @@ async def inner_async(
                 fields_from_call_args = get_fields_from_call_args(
                     response_model, fn, args, kwargs
                 )
+                if (
```
willbakst, line 268: same comment here as `_extract`, shouldn't `stream_decorator` handle this for us?

willbakst (commented):
We should also add additional documentation examples of setting the client dynamically (namely the Custom Client sections in Calls and Async).

Can you please add these examples? If you're not certain what to do on this front just let me know and I'll clarify.

mirascope/core/base/_create.py
koxudaxi: The way will raise an error of pyright.
```pyhton
                if (
                    dynamic_config := await get_dynamic_configuration(fn, args, kwargs)
                ) and isinstance(dynamic_config, dict):
                    client = dynamic_config.get("client")
```

```
mirascope/core/base/_create.py:171:28 - error: "client" is possibly unbound (reportPossiblyUnboundVariable)
1 error, 0 
```

willbakst: hmm, how can the client be unbound if it exists as an argument and this would just be overwriting it?

mirascope/core/base/stream.py
koxudaxi: @willbakst 
which part is it? I applied the changes to `stream_factory()`

mirascope/core/base/_create.py
koxudaxi: I don't know if this is a bug, but I think it is because the client exists in a scope outside of the function.

mirascope/core/base/_create.py
koxudaxi: The code works fine.
```python
dynamic_config = await get_dynamic_configuration(fn, args, kwargs)
                if dynamic_config is not None:
                    dynamic_client = dynamic_config.get("client")
                else:
                    dynamic_client = None
```

But, the code got a type error.
```python
                dynamic_config = await get_dynamic_configuration(fn, args, kwargs)
                if dynamic_config:
                    dynamic_client = dynamic_config.get("client")
                else:
                    dynamic_client = None
```

```
/Users/koudai/PycharmProjects/mirascope/mirascope/core/base/_create.py
  /Users/koudai/PycharmProjects/mirascope/mirascope/core/base/_create.py:166:53 - error: "get" is not a known attribute of "None" (reportOptionalMemberAccess)
1 error, 0 warnings, 0 informations 

```

I think `if dynamic_config` is `True` guarantee `dynamic_config` is not None. But, the result is not same :(
pyright said this is `None`

mirascope/core/base/_create.py
koxudaxi: I fixed it

mirascope/core/base/stream.py
willbakst: i applied the fix

mirascope/core/base/_create.py
willbakst: i updated this to use `nonlocal` and just set `client` so it's a little cleaner / easier to read

mirascope/core/base/_create.py
willbakst: i figured this out and applied the fix

willbakst (approved):

PR #651, opened 2024-10-31
Add mirascope toolkit

koxudaxi:


mirascope/tools/_computer_use.py
```diff
@@ -0,0 +1,132 @@
...
+        try:
+            with self._container_context() as container:
+                exec_result = container.exec_run(
+                    cmd=["sh", "-c", command],
+                )
+            return exec_result.output.decode("utf-8")
+        except Exception as e:
+            return f"Error executing command: {str(e)}"
+
+    @contextmanager
```
willbakst, line 125: wait so if i'm understanding correctly, this will spin up a new container for every python/shell execution? It seems like instead the better thing to do would be to spin up a container at the beginning that the tool can then use? For example, what if the LLM wants to do something in one python call that relies on the outputs of a previous python call?

mirascope/tools/_filesystem.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @toolkit_tool
+    def read_file(self, path: str) -> str:
+        """Reads content from a file.
+
+        base_directory: {self.base_directory}
+
+        Args:
+            path: Path to the file
+        Returns:
```
willbakst, line 76: nit: missing newline here before "Returns:"

make sure to fix this anywhere else too

mirascope/tools/_filesystem.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def _validate_extension(self, path: str) -> str | None:
+        """Validates file extension and returns error message if invalid."""
+        extension = Path(path).suffix.lstrip(".")
+        if extension not in self._config().allowed_extensions:
+            return f"Error: Invalid file extension. Allowed: {self._config().allowed_extensions}"
+        return None
+
+    @toolkit_tool
+    def read_file(self, path: str) -> str:
```
willbakst, line 69: I feel like best practice would be to implement this as a class instead (i.e. `ReadFile`) so that the user can potentially do something like `if isinstance(tool, FileSystemToolkit.ReadFile): ...` right?

mirascope/tools/base.py
```diff
@@ -37,28 +39,75 @@ class ConfigurableTool(
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     @classmethod
-    def _config(cls) -> _ToolConfigT:
+    def _get_config(cls) -> _ToolConfigT:
```
willbakst, line 42: I understand we need this method to get the correct type, but any particular reason to update this to `_get_config`? I generally think of using "get" when there's also a setter, but in this case there should never be a setter (only the `from_config` method).

Also, something I've been thinking about is the fact that `BaseTool` already has `tool_config` as a ClassVar, and I'm starting to think that `__config__` and `tool_config` as ClassVars may become confusing?

WDYT? Any idea here around how to remove confusion?

I think `from_config` and `config` vs. `tool_config` is clear enough, but we type `config: _ToolConfigT` which is where the confusion arises for me (because that has type `ToolConfig`).

maybe we could rename this to something like `_ConfigurableToolConfig`?

mirascope/tools/base.py
```diff
@@ -37,28 +39,75 @@ class ConfigurableTool(
...
+
+
+class ConfigurableToolKit(BaseToolKit, Generic[_ToolConfigT], ABC):
+    """Abstract base class for configurable toolkit.
+
+    Subclasses must define a `__prompt_usage_description__` class variable
+    and __config__ class variable with a subclass of _ToolConfig.
+    """
+
+    __config__: ClassVar[_ToolConfig]
```
willbakst, line 79: hmm, realizing that `__config__` shouldn't be a `ClassVar` (both for `ConfigurableTool` and `ConfigurableToolKit`). The configuration is not something to be defined at the class level but something that should be defined for every instance (e.g. you can have two of the same tools or toolkits with different configurations, which doesn't really work well with class vars).

What if instead we did `_config: _ConfigurableToolConfigT`? We can still set the default variable, and we can still construct the class from config, but now it's clear we can have a different config for each tool. We could also do `__config__: PrivateAttr[_ConfigurableToolConfigT]`. Which do you prefer?

mirascope/tools/system/_computer_use.py
```diff
@@ -0,0 +1,153 @@
...
+from mirascope.core import toolkit_tool
+from mirascope.tools.base import (
+    ConfigurableTool,
+    ConfigurableToolKit,
+    _ToolConfig,
+    _ToolSchemaT,
+)
+
+
+class ComputerUseToolkitConfig(_ToolConfig):
```
willbakst, line 22: hmm, I think with Anthropic's new "Computer Use" release, this term is too overloaded, so we should be more specific. I think instead of "computer_use" and "ComputerUse" we should make it specific to Docket, so something like `DockerOperationToolKitConfig` and `BaseDockerOperationTool` and `DockerOperationToolKit`.

WDYT?

mirascope/tools/system/_computer_use.py
```diff
@@ -0,0 +1,153 @@
...
+    __container__: Container | None = None
+
+    @property
+    def _container(self) -> Container:
+        """Returns the Docker container for code execution."""
+        if self.__container__ is None:
+            self.__container__ = self._setup_container()
+        return self.__container__
+
+    def __del__(self) -> None:
```
willbakst, line 45: this is going to stop the docker container when this tool gets deleted, but that's not what we want. We want the container to persist until the LLM is done using it so that multiple calls can use the same docker instance.

We probably want the user to have a little more control over the container, right? Maybe something like a `DockerContainer` class that is required as a field of the toolkit? That way the user can create the toolkit wherever and supply it with the same docker container, and then when they are done and the container goes out of scope we'll shut it down. Thoughts?

mirascope/tools/system/_computer_use.py
```diff
@@ -0,0 +1,153 @@
...
+        default="python:3.13-slim", description="Docker image for code execution"
+    )
+    max_memory: str = Field(default="512m", description="Maximum memory allocation")
+    allow_network: bool = Field(default=True, description="Allow network access")
+
+
+class DockerOperation(ConfigurableTool[ComputerUseToolkitConfig, _ToolSchemaT], ABC):
+    """Base class for Docker operations."""
+
+    __config__ = ComputerUseToolkitConfig()
```
willbakst, line 35: hmm, so each of these tools has a different config, and it's not the same as the parent toolkit's configuration? It seems like each tool should have it's own `_ConfigurableToolConfig` if necessary? How does the user configure each individual tool? Is that even useful?

More so I care about figuring out how to propagate the toolkit's config down into the tools that are part of the toolkit. This way the user could set configuration options for the toolkit and each tool can access that same config.

This is also inline with `__config__` not making sense as a class variable (see other comment)

mirascope/tools/system/_filesystem.py
willbakst, line 1: nit: instead of `filesystem` shouldn't all of these files be names `file_system` with an underscore?

mirascope/tools/base.py
koxudaxi: > I understand we need this method to get the correct type, but any particular reason to update this to _get_config? I generally think of using "get" when there's also a setter, but in this case there should never be a setter (only the from_config method).

By not making the name a verb, we found it difficult to recognize whether it is callable or not.

>　Also, something I've been thinking about is the fact that BaseTool already has tool_config as a ClassVar, and I'm starting to think that __config__ and tool_config as ClassVars may become confusing?

I agree with this. It is indeed difficult to understand.

> maybe we could rename this to something like _ConfigurableToolConfig?

Sounds good, let's give it a try.

mirascope/tools/system/_computer_use.py
koxudaxi: I agree with you

mirascope/tools/system/_computer_use.py
koxudaxi: When from_config is called, the contents of the config are communicated to `tool`.
The problem is that it is difficult to access higher level classes due to the placement of classes within classes. Let's think about it some more.a

mirascope/tools/base.py
willbakst: Ok, `_get_config` is fine then, that makes sense

mirascope/tools/base.py
willbakst: I think instead what we should do is:
1. Define toolkit configuration as fields of the class
2. We should define a tool config for the individual tools inside of the toolkit
3. We should set the individual tool configs using the toolkit's fields on creation (i.e. `create_tools`)

docs/learn/tools.md
```diff
@@ -422,6 +422,11 @@ Mirascope provides several pre-made tools to help you get started quickly:
         - [`mirascope.tools.web.HTTPX`](../api/tools/web/httpx.md)
         - [`mirascope.tools.web.ParseURLContent`](../api/tools/web/parse_url_content.md)
         - [`mirascope.tools.web.Requests`](../api/tools/web/requests.md)
+    - System Toolkits
+        - [`mirascope.tools.system.FileSystemToolkit`](../api/tools/system/filesystem.md)
+        - [`mirascope.tools.system.ComputerUseToolkit`](../api/tools/system/computer_use.md)
```
willbakst, line 427: note to self: update computer use to docker operation anywhere unchanged

mirascope/tools/base.py
```diff
@@ -6,59 +6,95 @@
...
 ):
     """Abstract base class for configurable tools.
 
     Subclasses must define a `__prompt_usage_description__` class variable
-    and __config__ class variable with a subclass of _ToolConfig.
+    and _configurable_tool_config class variable with a subclass of _ToolConfig.
     """
 
-    __config__: ClassVar[_ToolConfig]
+    _configurable_tool_config: ClassVar[_ConfigurableToolConfig]
```
willbakst, line 38: note to self: remove underscore

willbakst (approved):

Issue #646, opened 2024-10-30
Add `text` format specifier for fields in prompts to create separate text content part for field


jimkring:
### Description

Continuing [this thread](https://mirascope-community.slack.com/archives/C072S9P1R38/p1730259117983329), I would like to have the input data of my user messages kept separate from the prompt template string itself.  My use case is that I am storing all of the messages and message parts, and having the structured (JSON) data for the various input fields kept separate from the unstructured prompt string data is helpful, since it makes it easier to extract and work with later.  For example, we have an Assistant API we are implementing and need to be able to traverse the message/conversation history.  We don't want to have to parse inside of unstructured text content to find structured data -- that's why keeping the structured separate is very helpful.

willbakst:
I think adding the format specifier should be easy. Something like this:

```python
from mirascope.core import prompt_template


@prompt_template(
    """
    First Part {text:text} Third Part
    """
)
def prompt(text: str): ...


print(prompt("Second Part")
# > [BaseMessageParam(role='user', content=[TextPart(type='text', text='First Part'), TextPart(type='text', text='Second Part'), TextPart(type='text', text='Third Part')])]
```

We could even add the analogous `{:texts}`:

```python
from mirascope.core import prompt_template


@prompt_template("{texts:texts}")
def prompt(texts: str): ...

print(prompt(["First Part", "Second Part"])
# > [BaseMessageParam(role='user', content=[TextPart(type='text', text='First Part'), TextPart(type='text', text='Second Part')])]

willbakst:
I've added this to the `v1.7` milestone and will try to get this out for the release on Friday :)

willbakst:
Unfortunately didn't get to this in time for the `v1.7` release, but will shoot to get this out as part of the next release hopefully.

willbakst:
This is released in `v1.8`!

jimkring:
Fantastic!  Thanks for the goodies @willbakst.

PR #665, opened 2024-11-06
feat: Support text format specifier

koxudaxi:
Related Issues: https://github.com/Mirascope/mirascope/issues/646

### DIFF
--- a/mirascope/core/base/prompt.py
+++ b/mirascope/core/base/prompt.py
@@ -76,8 +76,10 @@ class BasePrompt(BaseModel):
         prompt_template = (
             prompt_template.replace(":images}", "}")
             .replace(":image}", "}")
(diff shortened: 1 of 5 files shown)


mirascope/core/base/prompt.py
```diff
@@ -76,8 +76,10 @@ def __str__(self) -> str:
         prompt_template = (
             prompt_template.replace(":images}", "}")
             .replace(":image}", "}")
-            .replace(":audios", "}")
+            .replace(":audios}", "}")
             .replace(":audio}", "}")
+            .replace(":text}", "}")
```
willbakst, line 81: note to self: switch order here so plural comes first and doesn't get lost

willbakst (approved):

PR #666, opened 2024-11-06
feat: Include usage tracking for streaming by default

koxudaxi:


mirascope/core/openai/_utils/_setup_call.py
```diff
@@ -121,6 +124,8 @@ def setup_call(
                 UserWarning,
             )
         call_kwargs["tool_choice"] = "required"
+    if stream:
```
willbakst, line 127: note to self: check if stream options already exist and overwrite include usage if not present

willbakst (approved):

PR #669, opened 2024-11-07
feat: Add merge_decorators

koxudaxi:


mirascope/core/base/merge_decorators.py
```diff
@@ -0,0 +1,59 @@
...
+_R = TypeVar("_R")
+
+_WP = ParamSpec("_WP")
+_WR = TypeVar("_WR")
+
+
+def merge_decorators(
+    decorator: Callable[[Callable[_P, _R]], Callable[_WP, _WR]],
+    *additional_decorators: Callable[[Callable], Callable],
+) -> Callable[[], Callable[[Callable[_P, _R]], Callable[_WP, _WR]]]:
```
koxudaxi, line 15: Due to current Python's type system limitations, it's challenging to apply precise type annotations to variable-length arguments. 

This compromise allows for proper type checking of both the wrapped and wrapper functions in merge_decorators' arguments, making it a pragmatic implementation. The first decorator determines the type constraints, while additional decorators use more flexible typing.

docs/api/core/base/merge_decorators.md
```diff
@@ -0,0 +1,3 @@
+# mirascope.core.base.merge_decorators
+
+::: mirascope.core.base.merge_decorators
```
koxudaxi, line 3: Do we need separate documentation for this?
I've only added API documentation as I couldn't find an appropriate section in the existing documentation.

willbakst: just api is fine I think

PR #670, opened 2024-11-07
feat: Support error handler for middleware factory

koxudaxi:


mirascope/integrations/_middleware_factory.py
```diff
@@ -51,6 +51,12 @@ def middleware_factory(
         [SyncFunc | AsyncFunc], AbstractContextManager[_T]
     ] = default_context_manager,
     custom_decorator: Callable | None = None,
+    handle_error: Callable[[Exception, SyncFunc | AsyncFunc, _T | None], Any]
```
willbakst, line 54: note to self: this is a breaking change since we don't currently require named arguments, so this has to be moved to the end of the definition

docs/learn/extensions/middleware.md
```diff
@@ -9,7 +9,7 @@ We will be creating an example decorator `with_saving` that saves some metadata
     --8<-- "examples/learn/middleware/with_saving.py:6:7"
 
 
-    --8<-- "examples/learn/middleware/with_saving.py:17:27"
+    --8<-- "examples/learn/middleware/with_saving.py:17:29"
```
willbakst, line 12: note to self: double check docs with other change since line highlighting will likely change

willbakst (approved):

Issue #667, opened 2024-11-06
Add mechanism for setting common call params across providers


barapa:
### Description

When calling the same prompt across multiple providers, if you want to set model config values such as temperature, max_output_tokens, etc, each provider's types are different. Additionally, different providers have different values that need to be passed with no defaults.

For example:
```py
from google.generativeai.types.generation_types import GenerationConfig
from mirascope.core import anthropic, gemini, openai, prompt_template


@prompt_template(
    """
    SYSTEM:
    Do what you are told.

    USER:
    {what_to_do}
    """
)
def prompt(
    what_to_do: str,
): ...


def main():
    prompt_fn = prompt

    openai_call = openai.call("gpt-4o", call_params={"temperature": 0.7})(prompt_fn)
    openai_result = openai_call(what_to_do="Do something cool")

    # Anthropic requires max_tokens to be set
    anthropic_call = anthropic.call(
        "claude-3-opus-latest",
        call_params={
            "temperature": 0.7,
            "max_tokens": 2048,
        },
    )(prompt_fn)
    anthropic_result = anthropic_call(what_to_do="Do something cool")

    # Gemini requires its own specific type
    gemini_call = gemini.call(
        model="gemini-1.5-pro-002",
        call_params={
            "generation_config": GenerationConfig(temperature=0.7, max_output_tokens=2048)
        },
    )(prompt_fn)
    gemini_result = gemini_call(what_to_do="Do something cool")


if __name__ == "__main__":
    main()


```

It might be useful to have some common way to generate the equivalent sets of params for each provider.

willbakst:
Totally! We originally implemented things this way because each provider has different call parameters, but there are certain call parameters that are common, so convenience around that would certainly be nice.

I'm thinking something like this would be best:

```python
openai_call_params = openai.OpenAICallParams.from_common(temperature=0.7)
openai_call = openai.call("gpt-4o", call_params=openai_call_params)(prompt_fn)
openai_result = openai_call(what_to_do="Do something cool")

anthropic_call_params = anthropic.AnthropicCallParams.from_common(temperature=0.7)
anthropic_call = anthropic.call("claude-3-opus-latest", call_params=anthropic_call_params)(prompt_fn)
anthropic_result = anthropic_call(what_to_do="Do something cool")

gemini_call_params = gemini.GeminiCallParams.from_common(temperature=0.7)
[...]

PR #674, opened 2024-11-11
feat: Add CommonCallParams

koxudaxi:
Related Issue: https://github.com/Mirascope/mirascope/issues/667


koxudaxi:
@willbakst 
Your suggested idea was to add a from_common method to `{Provider}CallParams`.
However, since `{Provider}CallParams` is a `TypedDict`, we cannot add methods to it.
Therefore, I've prepared separate functions `get_{provider}_call_params_from_common` instead.
The implementation uses Unpack with `TypedDict` to provide type-safe keyword arguments while maintaining the same convenient interface you envisioned with the original `from_common` approach.

mirascope/core/base/call_params.py
```diff
@@ -1,6 +1,78 @@
...
+    temperature: NotRequired[float | None]
+    max_tokens: NotRequired[int | None]
+    top_p: NotRequired[float | None]
+    frequency_penalty: NotRequired[float | None]
+    presence_penalty: NotRequired[float | None]
+    seed: NotRequired[int | None]
+    stop: NotRequired[str | list[str] | None]
+
+
+def convert_params(
```
willbakst, line 40: this whole method seems like overkill? it seems like we could just return a dictionary comprehension with a few if/else checks for each provider?

For example, OpenAI would just be this:

```python
mapping = {...}
return {
    mapping[key]: value
    for key, value in common_params.items()
    if key in mapping and mapping[key] is not None
}
```

For gemini we would have this:

```python
mapping = {...}
return {
    mapping[key]: [value] if key == "stop" and isinstance(value, str) else value
    for key, value in common_params.items()
    if key in mapping and mapping[key] is not None
}
```

Seems better also to colocate the provider-specific functionality with that provider's implementation. For example, the `convert_stop_to_list` method is only used by gemini/vertex, but now lives as a base method. Same thing for `wrap_in_config` which is for gemini/vertex. It's not clear why this lives in the `convert_params` function.

mirascope/core/bedrock/call_params.py
```diff
@@ -36,3 +37,46 @@ class BedrockCallParams(BaseCallParams):
     guardrailConfig: NotRequired[GuardrailConfigurationTypeDef]
     additionalModelRequestFields: NotRequired[Mapping[str, Any]]
     additionalModelResponseFieldPaths: NotRequired[list[str]]
+
+
+def _create_inference_config(params: CommonCallParams) -> InferenceConfigurationTypeDef:
```
willbakst, line 42: this can just live inside of the conversion function as it's not very long

mirascope/core/anthropic/call_params.py
```diff
@@ -39,3 +40,19 @@ class AnthropicCallParams(BaseCallParams):
     top_k: NotRequired[int | None]
     top_p: NotRequired[float | None]
     timeout: NotRequired[float | Timeout | None]
+
+
+def get_anthropic_call_params_from_common(
```
willbakst, line 45: i know we can't attach methods to a typed dict, but I don't love the current structure of the implementation.

What if the user didn't have to do anything at all? Could we detect `CommonCallParams` under the hood and automatically convert? And then type hints / lint errors can take care of the rest?

What about something like this:

```python
from mirascope.core import anthropic

@anthropic.call(
    "claude-3-5-sonnet-latest",
    call_params={"temperature": 0.7},
)
```

The above example will fail because Anthropic requires `max_tokens`. However, if we allow `call_params` to also take `CommonCallParams` as input, we could detect that this needs to be converted and convert it under the hood.

Then all of our conversion methods can live in each provider's respective `_utils` folder and the user never needs to worry about them.

mirascope/core/anthropic/_utils/_convert_common_params.py
willbakst, line 1: nit: all of these should be named `_convert_common_call_params.py` and the methods should be `convert_common_call_params` so it's clear what we're converting (namely call params)

willbakst (approved):

PR #704, opened 2024-11-16
add logfire error handling

CharlesOural:
My first PR !

I added error handling for the logfire middleware.

I added the option to properly handle _response if the error propagate it, I use it in my use case.

### DIFF
--- a/mirascope/integrations/logfire/_utils.py
+++ b/mirascope/integrations/logfire/_utils.py
@@ -29,7 +29,7 @@ def custom_context_manager(
         yield logfire_span
 
 
-def _get_call_response_span_data(result: BaseCallResponse) -> dict:
+def _get_call_response_span_data(result: BaseCallResponse) -> dict[str, Any]:
     output: dict[str, Any] = {}
     if cost := result.cost:
         output["cost"] = cost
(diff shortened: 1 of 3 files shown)


mirascope/integrations/logfire/_utils.py
```diff
@@ -186,3 +186,55 @@ async def handle_structured_stream_async(
     span_data = get_structured_stream_span_data(result)
     span_data["async"] = True
     logfire_span.set_attributes(span_data)
+
+
+async def handle_error(
```
willbakst, line 191: we should refactor these to use a shared `_get_error_span_data` utility method

also, if you'd be so kind as to update all of the similar utilities to be private as well (e.g. `_get_call_response_span_data`) that would be awesome :)

mirascope/integrations/logfire/_utils.py
willbakst, line 1: you'll need to add tests for 100% coverge

you can run `uv run pytest tests --cov=./ --cov-report-html` to run the tests and generate the coverage report locally.

mirascope/integrations/logfire/_utils.py
CharlesFarhat: Yes it's done 💯

CharlesFarhat:
Ok all done,  also bumped the logfire version. Version 0.53 will crash with litellm and span data. they added checks in 1.0.0. I change <2 to <3 because 2.3.0 implement asyncpg :)

PR #710, opened 2024-11-21
feat: concept for streaming tool calls

njpearman:
Hi, I've been trying out mirascope with my current project. I'm loving the SDK and interface so far but I need tool call streaming to return chunk-by-chunk, which `mirascope` doesn't seem to support currently (Referencing https://github.com/Mirascope/mirascope/blob/main/mirascope/core/openai/_utils/_handle_stream.py#L70C5-L70C18, `yield` only happens when there is a finished tool call).

This PR is a proposal for an approach that I understand could solve my problem. However, I've only just started looking at mirascope so don't have full context of implications.


njpearman:
Closing this as #711 has been created.

PR #717, opened 2024-11-26
handle empty tool call function arguments

jimkring:
This pull request includes changes to improve the handling of tool calls in the `mirascope/core` package. The updates ensure that the `model_json` is properly initialized even when `tool_call.function.arguments` is empty.

**Background:**
LLMs sometimes use pass an empty string for tool call function args, which was causing an unhandled exception when deserializing the function arguments string to json with jiter.

**Solution:**
Now, we detect that `tool_call.function.arguments`  is empty and simply initialize an empty dictionary (instead of trying to deserialize an empty string).

Changes to tool call handling:

* [`mirascope/core/azure/tool.py`](diffhunk://#diff-eda7cd9a429405537fae526d49224241b435b8b5b732dfd6f120a2b282fbd29eR90-R93): Added a check to initialize `model_json` to an empty dictionary if `tool_call.function.arguments` is empty in the `from_tool_call` method.
* [`mirascope/core/groq/tool.py`](diffhunk://#diff-11161bc2a645bb46031b9fb51034df763f83b3249b81132a2343142206836cb7R77-R80): Added a check to initialize `model_json` to an empty dictionary if `tool_call.function.arguments` is empty in the `from_tool_call` method.
* [`mirascope/core/openai/tool.py`](diffhunk://#diff-816e8c0e5aeca294a4f157000998421180c9a28022c93ef85b6d570e53b7717eR93-R96): Added a check to initialize `model_json` to an empty dictionary if `tool_call.function.arguments` is empty in the `from_tool_call` method.

### DIFF
--- a/mirascope/core/azure/tool.py
+++ b/mirascope/core/azure/tool.py
@@ -87,9 +87,7 @@ class AzureTool(BaseTool):
         Args:
             tool_call: The Azure tool call from which to construct this tool instance.
         """
-        if tool_call.function.arguments:
-            model_json = jiter.from_json(tool_call.function.arguments.encode())
-        else:
-            model_json = {}
-        model_json["tool_call"] = tool_call
+        model_json = {"tool_call": tool_call}
+        if args := tool_call.function.arguments:
+            model_json |= jiter.from_json(args.encode())
         return cls.model_validate(model_json)
--- a/mirascope/core/azure/tool.py
+++ b/mirascope/core/azure/tool.py
@@ -87,6 +87,9 @@ class AzureTool(BaseTool):
         Args:
             tool_call: The Azure tool call from which to construct this tool instance.
         """
-        model_json = jiter.from_json(tool_call.function.arguments.encode())
+        if tool_call.function.arguments:
+            model_json = jiter.from_json(tool_call.function.arguments.encode())
+        else:
+            model_json = {}
         model_json["tool_call"] = tool_call
         return cls.model_validate(model_json)
--- a/mirascope/core/groq/tool.py
+++ b/mirascope/core/groq/tool.py
@@ -74,6 +74,9 @@ class GroqTool(BaseTool):
         Args:
             tool_call: The Groq tool call from which to construct this tool instance.
         """
-        model_json = jiter.from_json(tool_call.function.arguments.encode())
+        if tool_call.function.arguments:
+            model_json = jiter.from_json(tool_call.function.arguments.encode())
+        else:
+            model_json = {}
         model_json["tool_call"] = tool_call.model_dump()
         return cls.model_validate(model_json)
--- a/mirascope/core/groq/tool.py
+++ b/mirascope/core/groq/tool.py
@@ -74,9 +74,7 @@ class GroqTool(BaseTool):
         Args:
             tool_call: The Groq tool call from which to construct this tool instance.
         """
-        if tool_call.function.arguments:
-            model_json = jiter.from_json(tool_call.function.arguments.encode())
-        else:
-            model_json = {}
-        model_json["tool_call"] = tool_call.model_dump()
+        model_json = {"tool_call": tool_call}
+        if args := tool_call.function.arguments:
+            model_json |= jiter.from_json(args.encode())
         return cls.model_validate(model_json)
--- a/mirascope/core/mistral/tool.py
+++ b/mirascope/core/mistral/tool.py
@@ -72,7 +72,9 @@ class MistralTool(BaseTool):
         Args:
             tool_call: The Mistral tool call from which to construct this tool instance.
         """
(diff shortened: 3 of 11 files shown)


mirascope/core/azure/tool.py
```diff
@@ -87,6 +87,9 @@ def from_tool_call(cls, tool_call: ChatCompletionsToolCall) -> AzureTool:
         Args:
             tool_call: The Azure tool call from which to construct this tool instance.
         """
-        model_json = jiter.from_json(tool_call.function.arguments.encode())
+        if tool_call.function.arguments:
```
willbakst, line 90: I think we should update this to the following here and elsewhere:

```python
model_json = {"tool_call": tool_call}
if args := tool_call.function.arguments:
    model_json |= jiter.from_json(args.encode())
return cls.model_validate(model_json)
```

willbakst:
@jimkring good catch!

could you also update the remaining 7 providers?

mirascope/core/azure/tool.py
willbakst: note: this will also maintain 100% test coverage so you won't need to make any other changes :)

jimkring:
@willbakst I'll take a look. I did a search and only found those three identical bits of code. Maybe the other providers are subtly different.

mirascope/core/azure/tool.py
jimkring: @willbakst I agree with this style recommendation (will work in python 3.9+ and we're on 3.10+)

willbakst:
> @willbakst I'll take a look. I did a search and only found those three identical bits of code. Maybe the other providers are subtly different.

oh you're right because of the issue being with `jiter`

can you update `mistral` to match the new recommended style from my other comment?

also, i just noticed that `gemini` and `vertex` have an issue where it will not allow empty args (when it should), would be great if you could update those as well as part of this

mirascope/core/mistral/tool.py
```diff
@@ -72,8 +72,7 @@ def from_tool_call(cls, tool_call: ToolCall) -> MistralTool:
         Args:
             tool_call: The Mistral tool call from which to construct this tool instance.
         """
-        model_json = tool_call.function.arguments
-        if isinstance(model_json, str):
-            model_json = jiter.from_json(model_json.encode())
-        model_json["tool_call"] = tool_call.model_dump()
+        model_json = {"tool_call": tool_call.model_dump()}
+        if args := tool_call.function.arguments and isinstance(args, str):
```
willbakst, line 76: you'll have to wrap this:

```python
if (args := tool_call.function.arguments) and isinstance(args, str):
```

jimkring:
@willbakst I've updated several other files/providers. Please take a look at how the differ in subtle ways. In some cases the `arguments` from the tool call are coming from the provide as a string, in other cases it's a Pydantic model.  I don't have a way to test all these special cases, so we'll want to be thoughtful about our review and scope of changes.

mirascope/core/mistral/tool.py
willbakst: oh wait actually this is wrong, cause if `tool_call.function.arguments` is not a string I believe it could be a dictionary, in which case we'll need to handle it differently from string (like we did before by setting `model_json = tool_call.function.arguments` directly)

jimkring: @willbakst I'm still making friends with the walrus :)

mirascope/core/groq/tool.py
```diff
@@ -74,6 +74,7 @@ def from_tool_call(cls, tool_call: ChatCompletionMessageToolCall) -> GroqTool:
         Args:
             tool_call: The Groq tool call from which to construct this tool instance.
         """
-        model_json = jiter.from_json(tool_call.function.arguments.encode())
-        model_json["tool_call"] = tool_call.model_dump()
+        model_json = {"tool_call": tool_call}
```
willbakst, line 77: this was previously `tool_call.model_dump()` so we should make sure to keep that

that's my mistake in my previous suggestion

willbakst:
> so we'll want to be thoughtful about our review and scope of changes.

happy to provide comments, also happy to take this over the finish line, just let me know :)

mirascope/core/mistral/tool.py
jimkring: @willbakst updated it. lmk how that looks.

jimkring:
@willbakst it's all yours to finish up, nopw. Thanks for the style suggestions and fixes -- I learned a couple things :+1:

mirascope/core/groq/tool.py
jimkring: I'll let you tidy this one up. Thanks @willbakst

willbakst (approved):

willbakst:
this is now released in `v1.10.1`

thank you for the contribution!!

Issue #711, opened 2024-11-21
Streaming Partial Tools


willbakst:
> Originally proposed by @njpearman in #710 as part of the chunks. Discussions in the slack community resulted in this current proposal

# Description

Right now, streaming with tools only ever returns the full tool call; however, it can be useful to receive the partial tool and the tool call argument deltas (particularly for showing the streaming tools on a UI, for example).

---

# Proposal

```python
from mirascope.core import openai

def my_tool(arg: str) -> str:
    return "dummy tool"

@openai.call("gpt-4o-mini", stream={"partial_tools": True}, tools=[my_tool])
def my_stream() -> str:
    return "Call my_tool"

stream = my_stream()
for chunk, tool in stream:
    if tool:
        if tool.delta is not None:
            # handle partial tool call case
        else:
            # handle full tool call case
    else:
        # handle no tool call case
```

The idea is that `tool` will be a partial tool (constructed using partial validation), and we can identify that it is a partial tool through the presence of the `delta` property, which will only be `None` for the final, full tool.

---

# Implementation Details

1. We'll need to update [`call_factory`](https://github.com/Mirascope/mirascope/blob/5a5473cb07b28972c8f2420d00ff0d42d9335e84/mirascope/core/base/_call_factory.py#L126) so that `stream` accepts either a `bool` or a `StreamOptions` typed dict.
2. Then we'll need to update `stream_factory` to pass something like `allow_partial` based on the stream options to the [`handle_stream`](https://github.com/Mirascope/mirascope/blob/5a5473cb07b28972c8f2420d00ff0d42d9335e84/mirascope/core/base/stream.py#L425) and [`handle_stream_async`](https://github.com/Mirascope/mirascope/blob/5a5473cb07b28972c8f2420d00ff0d42d9335e84/mirascope/core/base/stream.py#L376) callbacks.
3. We'll need to update `handle_stream` and `handle_stream_async` for every supported provider to accept `allow_partial` and to yield the partial tool for each chunk where a tool call is present on the chunk. We should use our [`partial`](https://github.com/Mirascope/mirascope/blob/main/mirascope/core/base/_partial.py) decorator like we do for structured streaming (at least for now since Pydantic's new partial validation is limited, experimental, and would require bumping the minimum version).
[...]

PR #718, opened 2024-11-26
feat: support-streaming-partial-tools

koxudaxi:
closes: https://github.com/Mirascope/mirascope/issues/711


mirascope/core/base/stream.py
```diff
@@ -374,7 +379,9 @@ async def generator() -> (
                     ]
                 ):
                     async for chunk, tool in handle_stream_async(
-                        await create(stream=True, **call_kwargs), tool_types
+                        await create(stream=True, **call_kwargs),
+                        tool_types,
+                        allow_partial_tool=partial_tools,
```
willbakst, line 384: nit: i don't like that the stream handlers call this `allow_partial_tool` and then the argument here is `partial_tools` we should make this consistent

let's just make them all `partial_tools`? or make them all `allow_partial_tools` up to you

mirascope/core/base/tool.py
```diff
@@ -84,6 +85,17 @@ def call(self) -> str:
     __custom_name__: ClassVar[str] = ""
     tool_config: ClassVar[ToolConfig] = ToolConfig()
     model_config = ConfigDict(arbitrary_types_allowed=True)
+    _delta: str | None = None
+
+    @property
+    def delta(self) -> str | None:
+        return self._delta
```
willbakst, line 92: nit: have this return `None` and remove `_delta`

then in the partial tool construction overwrite the entire property with a lambda to return the delta (so that `_delta` doesn't show up in autocomplete)

mirascope/core/base/tool.py
```diff
@@ -84,6 +85,17 @@ def call(self) -> str:
...
+
+    @classmethod
+    def _dict_from_json(cls, json: str, allow_partial: bool = False) -> dict[str, Any]:
+        return jiter.from_json(
```
willbakst, line 96: we should return `{}` if `json` is the empty string (just to be safe)

mirascope/core/base/_call_factory.py
```diff
@@ -205,6 +206,8 @@ def base_call(
                     json_mode=json_mode,
                     client=client,
                     call_params=call_params,
+                    partial_tools=isinstance(stream, dict)
```
willbakst, line 209: why are we passing `partial_tools` to extract factory?

mirascope/core/base/tool.py
koxudaxi: The pydantic structure does not allow for property overrides.
```python
/Users/koudai/PycharmProjects/mirascope/.venv/bin/python /Users/koudai/Library/Application Support/JetBrains/PyCharm2024.2/scratches/scratch_119.py 
sagemaker.config INFO - Not applying SDK defaults from location: /Library/Application Support/sagemaker/config.yaml
sagemaker.config INFO - Not applying SDK defaults from location: /Users/koudai/Library/Application Support/sagemaker/config.yaml
  File "/Users/koudai/Library/Application Support/JetBrains/PyCharm2024.2/scratches/scratch_119.py", line 11, in <module>
    for chunk, tool in stream:
  File "/Users/koudai/PycharmProjects/mirascope/mirascope/core/openai/stream.py", line 76, in __iter__
    for chunk, tool in super().__iter__():
  File "/Users/koudai/PycharmProjects/mirascope/mirascope/core/base/stream.py", line 146, in __iter__
    for chunk, tool in self.stream:
  File "/Users/koudai/PycharmProjects/mirascope/mirascope/core/base/stream.py", line 432, in generator
    yield from handle_stream(
  File "/Users/koudai/PycharmProjects/mirascope/mirascope/core/openai/_utils/_handle_stream.py", line 99, in handle_stream
    tool, current_tool_call, current_tool_type = _handle_chunk(
  File "/Users/koudai/PycharmProjects/mirascope/mirascope/core/openai/_utils/_handle_stream.py", line 73, in _handle_chunk
    partial_tool.delta = cast(str, tool_call.function.arguments)
  File "/Users/koudai/PycharmProjects/mirascope/.venv/lib/python3.10/site-packages/pydantic/main.py", line 918, in __setattr__
    attr.__set__(self, value)
AttributeError: can't set attribute 'delta'

Process finished with exit code 1
```

koxudaxi:
@willbakst 
One more thing. mistral apparently always returns the complete tools in stream. In other words, we will not provide this option for mistral, since it will never be delta.

mirascope/core/base/tool.py
willbakst: Ah ok, what if we just create a field `delta` instead of a property and exclude it from the tool schema generation (like we do for the `tool_call` attribute)?

koxudaxi:
@willbakst 
Also groq seems to behave the same way as mistral.
In other words, this seems to be a special behavior only for openai.

willbakst:
> @willbakst 
> 
> Also groq seems to behave the same way as mistral.
> 
> In other words, this seems to be a special behavior only for openai.

I know for sure that Anthropic also supports this

koxudaxi:
OK, I have confirmed Anthropic return partial tools.

mirascope/core/anthropic/_utils/_handle_stream.py
```diff
@@ -59,31 +60,55 @@ def _handle_chunk(
...
+        # Return partial tool if enabled
+        if partial_tools and current_tool_type:
+            partial_tool_call = ToolUseBlock(
+                id=current_tool_call.id,
+                input=buffer,
+                name=current_tool_call.name,
+                type="tool_use",
+            )
+            partial_tool = current_tool_type.from_tool_call(partial_tool_call)
+            if hasattr(chunk.delta, "partial_json"):
```
willbakst, line 72: why do we need this check here? we access `chunk.delta.partial_json` above just fine?

mirascope/core/anthropic/tool.py
```diff
@@ -87,6 +88,13 @@ def from_tool_call(cls, tool_call: ToolUseBlock) -> AnthropicTool:
             tool_call: The Anthropic tool call from which to construct this tool
                 instance.
         """
-        model_json = copy.deepcopy(tool_call.input)
```
willbakst, line 90: nit: update this to the following:

```python
@classmethod
def from_tool_call(
    cls, tool_call: ToolUseBlock, allow_partial: bool = False
) -> AnthropicTool:
    """Constructs an `AnthropicTool` instance from a `tool_call`.

    Args:
        tool_call: The anthropic tool call from which to construct this tool
            instance.
    """
    model_json = {"tool_call": tool_call}
    model_json |= (
        cls._dict_from_json(tool_call.input, True)
        if isinstance(tool_call.input, str)
        else tool_call.input
    )
    if allow_partial:
        return partial(cls, {"tool_call", "delta"}).model_validate(model_json)
    return cls.model_validate(model_json)
```

willbakst (approved):

PR #732, opened 2024-12-05
feat: Add MCP client

koxudaxi:


mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+from pydantic import AnyUrl, BaseModel, ConfigDict, Field, create_model
+
+from mirascope.core import BaseMessageParam, BaseTool
+from mirascope.core.base import AudioPart, DocumentPart, ImagePart
+from mirascope.core.base._utils import convert_base_model_to_base_tool
+
+_BaseToolT = TypeVar("_BaseToolT", bound=BaseTool)
+
+
+class MCPClientTool(BaseTool):
```
koxudaxi, line 28: I noticed that Client and Server handle Tool differently, so I need a dedicated tool for each.

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+        parsed_results = []
+        for content in result.content:
+            if isinstance(content, TextContent):
+                parsed_results.append(content.text)
+            else:
+                parsed_results.append(content)
+        return [
+            content.text if isinstance(content, TextContent) else content
+            for content in result.content
+        ]
```
koxudaxi, line 47: Access MCP Server and run tool remotely.

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+            annotation = field_type
+        else:
+            default = Field(None, description=field_schema.get("description"))
+            annotation = field_type | None
+
+        fields[field_name] = (annotation, default)
+
+    return create_model(
+        snake_to_pascal(tool.name), __config__=ConfigDict(extra="allow"), **fields
+    )
```
koxudaxi, line 206: The BaseModel is dynamically created from the JSON Schema here. Of course, this is a minimal implementation; it does not support json shcmea with complex nested structure of arguments types or references.
However,  I think this implementation is sufficient for the current situation.

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+        if arg.required:
+            default = Field(..., description=arg.description)
+            annotation = str
+        else:
+            default = Field(None, description=arg.description)
+            annotation = str | None
+
+        fields[arg.name] = annotation, default
+
+    return create_model(prompt.name, __base__=MCPClientTool, **fields)
```
koxudaxi, line 224: If I am not mistaken, if we want to reproduce the behavior of `@prompt_template`, we will need to create a prompt model from the definition.

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+class MCPClient:
```
koxudaxi, line 227: Eventually, an `async contextmanager` is prepared that returns this client and manages the session as a context.

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+    def __init__(self, session: ClientSession) -> None:
+        self.session: ClientSession = session
+
+    async def list_resources(self) -> list[Resource]:
```
koxudaxi, line 231: Here, the `type` of the mcp library is returned as it is. it would be ideal to define a mcp type specifically for mirascope, but I don't know how valuable this would be in the current situation, so this is the implementation.
What do you think?

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+        result = await self.session.list_resources()
+        return result.resources
+
+    async def read_resource(
+        self, uri: str | AnyUrl
+    ) -> list[str | BlobResourceContents]:
```
koxudaxi, line 237: Here, too, the type of the mcp library is returned as it is, but for text, it is expanded to str for easier access.

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+    ) -> Callable[..., Awaitable[list[BaseMessageParam]]]:
+        async def async_prompt(**kwargs: str) -> list[BaseMessageParam]:
+            result = await self.session.get_prompt(name, kwargs)
+
+            return [
+                _convert_prompt_message_to_base_message_params(prompt_message)
+                for prompt_message in result.messages
+            ]
+
+        return async_prompt
```
koxudaxi, line 264: In this method, create a async function `Callable[..., Awaitable[list[BaseMessageParam]]]`

mirascope/mcp/client.py
```diff
@@ -0,0 +1,280 @@
...
+        for tool in list_tool_result.tools:
+            model = create_model_from_tool(tool)
+
+            class CustomMCPTool(MCPClientTool):
+                __call_tool__ = self.session.call_tool
+                __custom_name__ = tool.name
+
+            converted_tools.append(
+                convert_base_model_to_base_tool(model, CustomMCPTool)
+            )
```
koxudaxi, line 279: This is the best I can come up with, can you think of any other good approaches?

mirascope/mcp/client.py
willbakst: @brenkao do we still have our JSON schema to Pydantic model code somewhere? I feel like we wrote a pretty comprehensive version of this function that we could re-use here I'm just blanking on where that lives right now

mirascope/mcp/client.py
willbakst: why is that? what are we using the resulting model for?

mirascope/mcp/client.py
willbakst: hmm yeah I think using the mcp lib types for now is good unless we have a clear reason to wrap

mirascope/mcp/client.py
willbakst: why not just push the client tool functionality into the server tool so that the tool returned by the server is just ready to go?

mirascope/mcp/client.py
koxudaxi: Oh, really, this is a function I forgot to delete that I already don't use anywhere 😅  I will delete it.

mirascope/mcp/client.py
koxudaxi: @willbakst 
Sorry, I’m a bit confused. Since the data returned by the server is already predetermined, I believe we need to construct the tool on the client side based on the interface definition provided by the server’s tool.

willbakst (approved):

PR #737, opened 2024-12-09
blog: reduce amount of spacing between subheading and body text

orlando21:
seems to be more space than usual

### DIFF
--- a/docs/blog/posts/engineers-should-handle-prompting-llms.md
+++ b/docs/blog/posts/engineers-should-handle-prompting-llms.md
@@ -31,7 +31,6 @@ In fact, **trying to write out that logic in text would be even more complicated
 Below, we outline how we arrived at this truth, and how the solution we’ve developed (a Python-based LLM development library) helps developers manage prompts in the codebase easily and efficiently, making, in our experience, LLM app development faster and more enjoyable.
 
 ## **Our Frustrations with Developer Tools for Prompt Engineering**
-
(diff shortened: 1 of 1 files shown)


docs/blog/posts/engineers-should-handle-prompting-llms.md
```diff
@@ -31,7 +31,6 @@ In fact, **trying to write out that logic in text would be even more complicated
 Below, we outline how we arrived at this truth, and how the solution we’ve developed (a Python-based LLM development library) helps developers manage prompts in the codebase easily and efficiently, making, in our experience, LLM app development faster and more enjoyable.
 
 ## **Our Frustrations with Developer Tools for Prompt Engineering**
-
```
willbakst, line 34: this won't actually change the spacing on the rendered post since we convert the markdown to html

what's the issue with the spacing here? it should be the same for every header (see header below on line 50 that has the same whitespace on line 51 that you're removing here)

orlando21: On second look it appears the formatting of the subheadings in the blog article containing the pull request (see engineers.png) is different from the rest of the articles (see challenges.png)
engineers
challenges

willbakst: ah I see the issue, it's that the header is bolded when it shouldn't be. If we remove the bolding from the headers it renders properly

willbakst (approved):

PR #740, opened 2024-12-11
feat: add provider agnostic call decorator

koxudaxi:


mirascope/core/base/_call_factory.py
```diff
@@ -87,6 +87,7 @@ def call_factory(  # noqa: ANN202
     handle_stream_async: HandleStreamAsync[
         _AsyncResponseChunkT, _BaseCallResponseChunkT, _BaseToolT
     ],
+    provider: str | None = None,
```
koxudaxi, line 90: In order to hand over the provider name to Response, we decided to pass it 
from call_factory with the argument provider.
We can create new class from ClassResponse.
```python
class OpenAIBaseCallResponse(CallResponse):.
    _provider: str = "openai"
```
but I don't want to create new one for providers, the policy was to pass the name to call_factory.

mirascope/llm/_base_provider_converter.py
```diff
@@ -0,0 +1,50 @@
...
+    """Number of tokens in the generated completion."""
+
+    prompt_tokens: int
+    """Number of tokens in the prompt."""
+
+    total_tokens: int
+    """Total number of tokens used in the request (prompt + completion)."""
+
+
+class BaseProviderConverter(ABC):
```
koxudaxi, line 22: I decided to aggregate the specific logic for each provider into this class.
What do you think?

mirascope/llm/call_response.py
```diff
@@ -0,0 +1,138 @@
...
+        if (provider := values.get("provider")) and (provider_converter := _get_provider_converter(provider)):
+            provider_call_response = (
+                provider_converter.get_call_response_class().model_validate(values)
+            )
+            return {
+                **values,
+                "provider_call_response": provider_call_response,
+                "provider": provider,
+            }
+        raise ValueError("Invalid provider")
```
koxudaxi, line 66: inject provider specific response.

mirascope/core/base/types.py
```diff
@@ -35,3 +37,19 @@ def set_channels(self, channels: int) -> AudioSegment: ...
             def set_sample_width(self, sample_width: int) -> AudioSegment: ...
             def export(self, format: str) -> FileIO: ...
             def read(self) -> bytes: ...
+
+
+FinishReason: TypeAlias = Literal[
+    "stop", "length", "tool_calls", "content_filter", "function_call"
```
willbakst, line 43: nit: i think we can drop "function_call" since that is deprecated?

mirascope/core/bedrock/call_response_chunk.py
```diff
@@ -87,3 +93,21 @@ def output_tokens(self) -> int | None:
...
+
+    @property
+    def common_finish_reasons(self) -> list[types.FinishReason] | None:
+        return _convert_finish_reasons_to_common_finish_reasons(
+            cast(list[str], self.finish_reasons)
+        )
+
+    @property
+    def common_usage(self) -> Usage | None:
+        if self.input_tokens is None and self.output_tokens is None:
```
willbakst, line 105: why have a shared utility for the finish reason conversion but not he usage creation?

is there any reason not to just have shared utilities for all of these? Even if it's just a simple cast line so that any changes in the future happen only in one place?

mirascope/core/gemini/call_response.py
```diff
@@ -178,3 +186,16 @@ def tool_message_params(
...
             }
         ]
+
+    @property
+    def common_finish_reasons(self) -> list[FinishReason] | None:
+        return _convert_finish_reasons_to_common_finish_reasons(self.finish_reasons)
+
+    @property
+    def common_message_param(self) -> BaseMessageParam:
+        return _convert_parts_to_base_message_param(self.response.parts)
```
willbakst, line 196: why have a utility here for gemini and not the other providers?

mirascope/core/cohere/call_response_chunk.py
```diff
@@ -101,3 +108,27 @@ def output_tokens(self) -> float | None:
...
+    def common_finish_reasons(self) -> list[types.FinishReason] | None:
+        if not self.finish_reasons:
+            return None
+        return _convert_finish_reasons_to_common_finish_reasons(
+            cast(list[str], self.finish_reasons)
+        )
+
+    @property
+    def common_message_param(self) -> BaseMessageParam:
+        raise NotImplementedError
```
willbakst, line 122: are we planning on implementing this?

mirascope/core/groq/call_response.py
```diff
@@ -162,3 +166,24 @@ def tool_message_params(
             )
             for tool, output in tools_and_outputs
         ]
+
+    @property
+    def common_finish_reasons(self) -> list[FinishReason] | None:
+        return cast(list[FinishReason], self.finish_reasons)
+
+    @property
+    def common_message_param(self) -> BaseMessageParam:
```
willbakst, line 175: realizing that we need to update `BaseMessageParam` to account for tool calls. This means we'll need to add support for `role="tool"` (including whatever other metadata tool messages might have and converting these into provider-specific tool messages, which we currently don't do because we assume the user is using provider-specific tool messages right now).

mirascope/core/mistral/_utils/_convert_assistant_message_content_to_base_message_param.py
willbakst, line 1: we should try our best to match the utilities across all providers so that the structure of the codebase remains consistent.

For example, if we're going to have a utility for one provider, that means we should have the same utility for all providers. This will also help to ensure that we maintain consistent interfaces both externally and internally.

mirascope/llm/call_response.py
willbakst, line 1: in line with my previous comment, this is missing a provider-agnostic version of `tool_message_params`

mirascope/llm/llm_call.py
```diff
@@ -0,0 +1,172 @@
...
+                    final = await result
+                    return _wrap_result(final)
+
+                return async_wrapper()
+            else:
+                return _wrap_result(result)
+
+        return inner
+
+    def _wrap_result(result: BaseCallResponse | Stream) -> CallResponse | Stream:
```
willbakst, line 162: nit: define this above `wrapper` since `inner` uses this method (or pull it out into a private utility so we don't define it every time?)

mirascope/llm/stream.py
willbakst, line 1: this is also missing the provider-agnostic `tool_message_params` method.

mirascope/core/cohere/call_response_chunk.py
koxudaxi: no, my mistake

mirascope/core/mistral/_utils/_convert_assistant_message_content_to_base_message_param.py
koxudaxi: I have created ToolCallPart and ToolResultPart and added processing for them.
However, currently there are no use cases for ToolCallPart.

koxudaxi: I realized that while we plan to convert ToolResultPart to provider-specific Messages, there's a constraint that tool_results can only be sent to the provider that requested the tool because a tool_call_id is required.

mirascope/llm/_call_protocol.py
```diff
@@ -0,0 +1,481 @@
...
+    )
+except ImportError:
+    AsyncOpenAIDynamicConfig = OpenAITool = OpenAIDynamicConfig = None
+    AsyncAzureOpenAI = AsyncOpenAI = AzureOpenAI = OpenAI = None
+    ChatCompletion = ChatCompletionChunk = ChatCompletionMessageParam = (
+        ChatCompletionUserMessageParam
+    ) = None
+
+
+class Call(Protocol):
```
koxudaxi, line 55: The Protocol is generated by the script

scripts/llm_call_protocol/render.py
```diff
@@ -0,0 +1,59 @@
...
+        base_stream="Stream[ChatCompletion, ChatCompletionChunk]",
+        base_tool="OpenAITool",
+        base_dynamic_config="OpenAIDynamicConfig",
+        async_base_dynamic_config="AsyncOpenAIDynamicConfig",
+        base_call_param="BaseCallParams",
+        sync_base_client="OpenAI | AzureOpenAI",
+        async_base_client="AsyncOpenAI | AsyncAzureOpenAI",
+        same_sync_and_async_base_client="None",
+    ),
+]
```
koxudaxi, line 40: I expect we will add other providers

docs/learn/calls.md
```diff
@@ -97,6 +101,30 @@ Let's take a look at a basic example using Mirascope to call both OpenAI and Ant
         ```
     {% endfor %}
 
+The `llm.call` decorator accepts a `provider` and `model` arguments and returns a provider-agnostic `CallResponse` instance that provides a consistent interface regardless of the underlying provider.
+
+You can override provider settings at runtime using `llm.override`. This takes a function decorated with `llm.call` and lets you specify:
+
+- `provider_override`: Change the provider being called
```
willbakst, line 108: since we're using `llm.override` we should drop "override" from the args

for example, it would just be `llm.override(recommend_book, call_params={...})`

docs/learn/calls.md
```diff
@@ -97,6 +101,30 @@ Let's take a look at a basic example using Mirascope to call both OpenAI and Ant
...
+The `llm.call` decorator accepts a `provider` and `model` arguments and returns a provider-agnostic `CallResponse` instance that provides a consistent interface regardless of the underlying provider.
+
+You can override provider settings at runtime using `llm.override`. This takes a function decorated with `llm.call` and lets you specify:
+
+- `provider_override`: Change the provider being called 
+- `model_override`: Use a different model
+- `call_params_override`: Override call parameters like temperature
+- `client_override`: Use a different client instance
+
+When using `provider_override`, you must also specify either `model_override` or `call_params_override`.
```
willbakst, line 113: how do we properly type this? if the user specifies `provider="openai"` in the original decorator and then calls `llm.override` with `model="claude-3-5-sonnet-latest"` how do we let the user know this is invalid?

it seems like the `model="openai:gpt-4o-mini"` structured string approach requires that both provider and model be set together and match and enables proper type hinting.

what do you think?

docs/learn/calls.md
```diff
@@ -465,6 +493,49 @@ When making LLM calls, it's important to handle potential errors. Mirascope pres
 
 By catching provider-specific errors, you can implement appropriate error handling and fallback strategies in your application. You can of course always catch the base Exception instead of provider-specific exceptions (which we needed to do in some of our examples due to not being able to find the right exceptions to catch for those providers...).
 
+## Provider-Agnostic Call Interface
```
willbakst, line 496: this is duplicated a bunch with the previous usage section? seems like this can be entirely removed?

mirascope/core/anthropic/tool.py
```diff
@@ -52,7 +52,7 @@ def recommend_book(genre: str) -> str:
     __provider__ = "anthropic"
     __tool_config_type__ = AnthropicToolConfig
 
-    tool_call: SkipJsonSchema[ToolUseBlock]
+    tool_call: SkipJsonSchema[ToolUseBlock | None] = None
```
willbakst, line 55: this cannot be none as the tool cannot be constructed without it, why have you added none typing and a default value?

mirascope/core/base/tool.py
```diff
@@ -87,6 +90,7 @@ def call(self) -> str:
     tool_config: ClassVar[ToolConfig] = ToolConfig()
     model_config = ConfigDict(arbitrary_types_allowed=True)
     delta: SkipJsonSchema[str | None] = None
+    tool_call: SkipJsonSchema[_ToolMessageParamT | None] = None
```
willbakst, line 93: same thing here, why add none?

mirascope/core/mistral/tool.py
```diff
@@ -40,8 +39,6 @@ def recommend_book(genre: str) -> str:
 
     __provider__ = "mistral"
 
-    tool_call: SkipJsonSchema[ToolCall]
```
willbakst, line 43: remember `BaseTool` cannot have any generics, but you've updated this to use generics. not sure why you've done that? we need to define the field/attribute on each provider-specific model so that the user can define `BaseTool` without any generics and not see type errors for not providing the generics

mirascope/llm/tool.py
```diff
@@ -22,16 +24,19 @@ def __new__(
...
+
+class Tool(BaseTool[_ToolMessageParamT], metaclass=_DelegateAbstractMethodsForTool):
     """
     A provider-agnostic Tool class.
     - No BaseProviderConverter
     - Relies on _response having `common_` methods/properties if needed.
     """
 
-    _tool: BaseTool
+    _tool: BaseTool[_ToolMessageParamT]
```
willbakst, line 37: instead of using a generic here, we should put this generic on the `Tool` class and then properly type any child functions as necessary (see other comment on why we can't use generics here)

mirascope/core/base/message_param.py
```diff
@@ -77,18 +77,58 @@ class DocumentPart(BaseModel):
...
+
+
 class BaseMessageParam(BaseModel):
     """A base class for message parameters.
 
     usage docs: learn/prompts.md#prompt-templates-messages
 
     Attributes:
-        role: The role of the message (e.g. "system", "user", "assistant")
+        role: The role of the message (e.g. "system", "user", "assistant", "tool")
```
willbakst, line 117: i believe messages with `tool` roles also have a tool id that we should include

mirascope/core/base/message_param.py
```diff
@@ -77,18 +77,58 @@ class DocumentPart(BaseModel):
...
     content: (
         str
-        | Sequence[TextPart | ImagePart | AudioPart | CacheControlPart | DocumentPart]
+        | Sequence[
+            TextPart
+            | ImagePart
+            | AudioPart
+            | CacheControlPart
+            | DocumentPart
+            | ToolCallPart
```
willbakst, line 131: this is weird because i'm not sure what a `ToolCallPart` or `ToolResultPart` are. is that something that exists across all providers? what happens if i specify a `ToolCallPart` with OpenAI, which I don't believe has anything like that?

mirascope/core/openai/_utils/_convert_message_params.py
```diff
@@ -60,6 +60,17 @@ def convert_message_params(
                             "type": "input_audio",
                         }
                     )
+                elif part.type == "tool_result":
```
willbakst, line 63: this doesn't make sense to me since we're updating `converted_message_params` by adding a new message param except we're inside an iteration loop on parts not messages. Shouldn't we check for the `tool` role earlier and handle it accordingly?

same it true for any of the other conversions.

we also need to make sure we properly handle tool calls both for converting to messages with the role "tool" as well as assistant messages that contain tool calls.

if any of this is confusing let me know and we can discuss further

mirascope/core/anthropic/_utils/_convert_message_param_to_base_message_param.py
willbakst, line 1: I think for clarity here we should merge this utility function and the `convert_message_params` utility function into a single class that operates like a serializer.

For example, something like `BaseMessageParamConverter` with subclass `AnthropicMessageParamConverter` that implements the abstract methods `convert(message_params: ...)` with overloads so that calling it with `list[BaseMessageParam]` returns Anthropic's specific message param but calling it with Anthropic's message param converts it to `list[BaseMessageParam]`

The overload paths would then just call the original conversion utilities we have now except updated to properly handle tool calls

willbakst: it could also operate on a single message param at a time and then the user of the class can convert their messages themselves in a loop.

mirascope/core/anthropic/tool.py
koxudaxi: The type definition here became unnecessary as it's now generic, so I will remove it.

mirascope/core/base/tool.py
koxudaxi: I have removed the line

mirascope/core/openai/_utils/_convert_message_params.py
koxudaxi: Thank you for pointing these issues out. I see that handling part.type == "tool_result" inside the loop on parts may cause confusion about the final message role. Also, dealing with tool calls and results in a single pass can be tricky if we do not separate the logic for role="tool".

mirascope/core/base/message_param.py
koxudaxi: That makes sense. For tool-role messages, I’ll include an optional “tool id” attribute in the message param class. That way if the provider or the usage scenario requires an ID or name for the tool, we can add it in.

mirascope/core/base/message_param.py
koxudaxi: You're right, it can be confusing for a provider like OpenAI that doesn’t have a dedicated “tool” concept at the raw message level. I will handle those parts in a specialized path so that if we're using a provider that doesn’t support explicit “tool result,” we skip or transform it in a default way (e.g., fallback to normal assistant content or function calls).

koxudaxi:
@willbakst 
I've created a converter and organized the types for Tool and related components.
I'll check the messages around Tool for individual providers next.

mirascope/core/anthropic/tool.py
willbakst: `BaseTool` cannot have the generic because we expect users to define subclasses of `BaseTool` without providing the generic type.

`tool_call` can never be `None` as it must be present to construct / instantiate the tool instance. This is why we provide `tool_call` for every provider typed with that provider's specific tool call type.

mirascope/core/base/message_param.py
willbakst: Right, we just need to define our own Mirascope-specific message structure, and then we need to be able to convert back and forth between that message structure and the provider-specific message structure 1:1

For example, if we maintain tool calls as content parts (like Anthropic), we can then choose to include them as `tool_calls` instead of `content` when converting back into OpenAI.

Does this make sense?

mirascope/core/openai/_utils/_convert_message_params.py
willbakst: The core idea is that by this point in the code we've confirmed that we have a `BaseMessageParam` instance. This means that so long as we structure the Mirascope messages properly, we should always be able to convert that structure back into the provider-specific one.

For example, we may want to store tool calls as part of the `content` sequence in a `BaseMessageParam` with role `"assistant"` but when converting into OpenAI we actually inject those content parts into the `tool_calls` key of the message (whereas Anthropic will have it as part of the content sequence).

Does this make sense?

mirascope/core/anthropic/tool.py
koxudaxi: OK, I have reverted it.

mirascope/core/base/message_param.py
koxudaxi: I agree with you.

mirascope/core/base/message_param.py
willbakst: In replying to other comments, I'm realizing that we need to first define how we're going to maintain Mirascope `BaseMessageParam` objects so that we can convert back into provider-specific messages.

One thing I'm noticing is that if we store tool calls as a `tool_call` attribute of the message parameter (like OpenAI does for assistant messages) then we won't be able to mimic Anthropic which interweaves tool calls with text content.

However, if we store `ToolCallPart` like you had or something (that somewhat matches Anthropic) then we can properly interweave with content but still separate them into `tool_calls` when matching OpenAI.

What do you think? Does storing it on our end in this structure make sense to you wrt. being able to convert back?

mirascope/core/openai/_utils/_convert_message_params.py
koxudaxi: Yes, I agree we need to review how we handle tool-related messages that work on a per-message basis, like in OpenAI.

mirascope/core/openai/_utils/_convert_message_params.py
```diff
@@ -60,12 +62,49 @@ def convert_message_params(
                             "type": "input_audio",
                         }
                     )
+                elif part.type == "tool_call":
+                    if converted_content:
```
willbakst, line 66: what is happening here? if there's converted content at this point, wouldn't the content be part of the assistant message with the tool calls?

mirascope/core/base/tool.py
```diff
@@ -60,7 +60,10 @@ def generate(
         return json_schema
 
 
-class BaseTool(BaseModel):
+_ToolMessageParamT = TypeVar("_ToolMessageParamT")  # noqa: F821
```
willbakst, line 63: nit: can we remove this now?

mkdocs.yml
```diff
@@ -371,6 +371,8 @@ nav:
           - Middleware: "api/integrations/middleware.md"
           - OpenTelemetry: "api/integrations/otel.md"
           - Tenacity: "api/integrations/tenacity.md"
+      - LLM:
```
willbakst, line 374: nit: put this above tools

mirascope/core/openai/_utils/_convert_message_params.py
koxudaxi: I have improved the logic when tool_call has content.

docs/learn/calls.md
koxudaxi: I've set up the type definitions so that if either model or provider is specified, the other must also be specified to avoid type errors.

docs/learn/calls.md
```diff
@@ -105,12 +105,12 @@ The `llm.call` decorator accepts a `provider` and `model` arguments and returns
...
-- `model_override`: Use a different model
-- `call_params_override`: Override call parameters like temperature
-- `client_override`: Use a different client instance
+- `provider`: Change the provider being called 
+- `model`: Use a different model
+- `call_params`: Override call parameters like temperature
+- `client`: Use a different client instance
 
-When using `provider_override`, you must also specify either `model_override` or `call_params_override`.
+When using `provider`, you must also specify either `model` or `call_params`.
```
willbakst, line 113: if using `provider` then `model` is required, right? otherwise as discussed previously you could have the model as `gpt-4o-mini` by default but override `provider="anthropic"` and have a failure?

furthermore, if the user is overriding the provider, it's almost certain that they'll have to override `model` as well unless they provide the same provider as the default.

as mentioned before, we should either require both `provider` and `model` or we should merge them

willbakst: given your other resolved comment about proper type-checking, it seems like this just needs a language update?

How about "When using `provider`, you must also specify `model` and vice-versa."

koxudaxi: Thanks for catching this! The documentation is incorrect - when specifying a provider, model is always required (this is enforced by our type system). I'll update the docs to fix this.

willbakst (approved):

Issue #716, opened 2024-11-25
Documentation around using Mirascope with local models (e.g. vLLM, Ollama, etc.)


willbakst:
### Description

Many local model providers such as vLLM and Ollama have support for OpenAI compatibility, which means you can use Mirascope with local models by setting the client.

This is not explicitly clear, and I'm sure there are also other limitations (e.g. tools) that are worth mentioning in more detail.

We should include a "Learn" section page specifically around how to use local models (i.e. "Local Models") that makes usage with Mirascope explicitly clear.

petri:
Ollama just released functions-as-tools support: https://ollama.com/blog/functions-as-tools

PR #744, opened 2024-12-14
docs: add docs & example for local model usage

Kigstn:
Heyo!

I was looking at miracast to replace my llamaindex usage, as I find llamaindex increasingly annoying to work with for various reasons.

For our workflow, it is important that we have local model support via ollama. After some digging I found out that this is supported via the openai integration, but that there is no official ollama integration.

After reading the ticket asking for increased documentation #716, I went ahead and wrote some basic docs & examples.

Feedback is more then welcome :)


Kigstn:
Quick screenshot of the docs page
image

docs/learn/local_models.md
```diff
@@ -0,0 +1,34 @@
+---
+search:
+  boost: 3
+---
+
+# Local Models
+
+Many local model host (like [Ollama](https://github.com/ollama/ollama) & [vLLM](https://github.com/vllm-project/vllm)) have compatibility to the OpenAI API.
```
willbakst, line 8: nit: host -> hosts

docs/learn/local_models.md
```diff
@@ -0,0 +1,34 @@
...
+This allows you to use all of their features via the OpenAi integration, but not all OpenAi features are available.
```
willbakst, line 9: nit: OpenAi -> OpenAI (two places)

examples/learn/local_models/ollama.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+
+class Book(BaseModel):
+    title: str
+    author: str
+
+
+@openai.call("llama3.2", response_model=Book, client=ollama_client)
+async def extract_book(text: str) -> str:
```
willbakst, line 21: nit: these examples should be synchronous (only the Async section should have async)

examples/learn/local_models/ollama.py
willbakst, line 1: the examples across providers should be exactly the same for each section.

If we want to show a basic call, we should show it for both providers. If we want to show structured outputs, we should show it for both providers.

docs/learn/local_models.md
```diff
@@ -0,0 +1,34 @@
+---
+search:
+  boost: 3
+---
+
+# Local Models
```
willbakst, line 6: I think we should title this "Open-Source Models" instead of local (since they can still be served in the cloud using e.g. vLLM). WDYT?

Kigstn:
```suggestion
# Open-Source Models
```

docs/learn/local_models.md
Kigstn:
```suggestion
Many local open-source model hosts (like [Ollama](https://github.com/ollama/ollama) & [vLLM](https://github.com/vllm-project/vllm)) have compatibility to the OpenAI API.
```

Kigstn:
Thanks for the comments! Adapted it to hopefully suit them :)

willbakst:
> Thanks for the comments! Adapted it to hopefully suit them :)

Sorry this slipped through with the holidays. I will work to get this merged alongside a few other docs updates. Thank you for the work here!

willbakst (approved):

PR #760, opened 2025-01-02
docs: remember to import prompt_template in Text Classification Tutorial

RiptideStar:
In the Text Classification tutorial, there is a tiny code error because forgetting to import prompt_template. 

While prompt_template is imported in the Binary Classification section, it is unclear whether or not prompt_template should be imported because "openai" is _reimported_ without prompt_template, so to clear all confusion, just adding prompt_template back to match the same import sentence from the first section above.


willbakst:
It's already imported, so Ruff removes it, which is now causing lint errors, so I'm closing this since there isn't an issue when run in sequence.

Issue #762, opened 2025-01-13
Add support for parts type in prompt template


CharlesFarhat:
### Description

Add support to do :

```
from mirascope.core import BaseDynamicConfig, prompt_template


@prompt_template(
    """
    USER: First part. {middle_parts:parts}. Last part.
    """
)
def example() -> BaseDynamicConfig:
    return {"computed_fields": {"middle_parts": ["Second part", image]}}

```

CharlesFarhat:
Linked to https://github.com/Mirascope/mirascope/pull/763

willbakst:
Released in `v1.15.0` with [docs](https://mirascope.com/learn/prompts/#format-specifiers)!

PR #763, opened 2025-01-13
added support for parts in prompt template

CharlesFarhat:
linked to https://github.com/Mirascope/mirascope/issues/762


mirascope/core/base/_utils/_parse_content_template.py
```diff
@@ -22,7 +22,7 @@
 from ._pil_image_to_bytes import pil_image_to_bytes
 
 _PartType = Literal[
-    "text", "texts", "image", "images", "audio", "audios", "cache_control"
+    "text", "texts", "image", "images", "audio", "audios", "cache_control", "parts"
```
willbakst, line 25: why only "parts" and not also the single "part"?

feels weird only having the plural here but the singular for all other types

mirascope/core/base/_utils/_parse_content_template.py
```diff
@@ -22,7 +22,7 @@
 from ._pil_image_to_bytes import pil_image_to_bytes
 
 _PartType = Literal[
```
willbakst, line 24: noticed this is missing "document" and "documents", can you add those please?

CharlesFarhat: yes ! will do

mirascope/core/base/_utils/_parse_content_template.py
CharlesFarhat: How would you see the part type ? just a way to send any BaseMessageParam ? I could add it quickly if so

CharlesFarhat: yes ! i add this right away

CharlesFarhat: ok done

willbakst (dismissed):

willbakst (approved):

PR #776, opened 2025-01-24
feat: Add provider Google

koxudaxi:
I cannot merge this PR because of dependency issue with `python-genai package`.
The package uses `typing.TypedDict` instead of `typing_extensions.TypedDict`, which causes Pydantic error in Python <3.12.

I created issue for this problem here: googleapis/python-genai#189

We need to wait for fix for this issue.


mirascope/core/google/_utils/_convert_message_params.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def convert_message_params(
+    message_params: list[BaseMessageParam | ContentDict],
+) -> list[ContentDict]:
+    converted_message_params = []
+    for message_param in message_params:
+        if not isinstance(message_param, BaseMessageParam):
+            converted_message_params.append(message_param)
+        elif (role := message_param.role) == "system":
```
willbakst, line 15: I just noticed that there is a `system_instruction` argument for the generation config. This likely also exists for the `gemini` and `vertex` module integrations.

We should update the code to use the actual system instructions rather than this pseudo system message in all places.

For reference, we do something similar for Anthropic, which also has a separate system message argument.

willbakst: Given this is blocked by an external issue, let's actually make the gemini/vertex updates as a separate PR so that change can be released

koxudaxi: OK,  I have updated the `system_instruction`

mirascope/core/google/_utils/_convert_message_params.py
```diff
@@ -0,0 +1,157 @@
...
+                            raise ValueError(
+                                f"Unsupported image media type: {media_type}. "
+                                "Google currently only supports JPEG, PNG, WebP, HEIC, "
+                                "and HEIF images."
+                            )
+                    else:
+                        media_type = "image/unknown"
+                    converted_content.append(
+                        PartDict(
+                            file_data=FileDataDict(
```
willbakst, line 88: have you actually tested this? I believe file data required a `gs://` uploaded uri and will fail on `http://` or `https://`

this means we'll have to take the downloaded image and use their `genai.Client().files.upload` method to get a `file_ref` that we can insert? So we should only pass `part.url` if it's already a `file_uri`

willbakst: Also, if we're already downloading the image, we should just send the downloaded image as an image part.

We should determine after conversion if the total package size would be > 20mb and if so iterate through all data parts and upload them using their client and provide the refs.

see here: https://ai.google.dev/gemini-api/docs/vision?lang=python#large-images

koxudaxi: I believe it worked when I tested it last night (I remember verifying the 'image/unknown' behavior), but I'll fix it using the method described in that documentation.

willbakst: If this works then that should be fine. We should still test the 20mb limit to see if this handles that case or not

koxudaxi: OK, I tested the 30mb file.
I'm working on vertext mode

willbakst: ok that's awesome then, less work for us is great, feel free to resolve this.

i will approve once coverage is met

willbakst: > I'm working on vertext mode

Shouldn't vertex just work if you provide the custom client properly?

mirascope/core/google/_utils/_calculate_cost.py
```diff
@@ -0,0 +1,8 @@
+"""Calculate the cost of a Google API call."""
+
+
+def calculate_cost(
```
willbakst, line 4: why is this `None`? Aren't the prices published?

koxudaxi: I have updated the table.

willbakst (approved):

Issue #785, opened 2025-01-29
Missing computed fields (cost, input_tokens, output_tokens, usage


ionflow:
### Description

I'm noticing some missing computed fields in some of the model responses expected according to the documentation: https://mirascope.com/learn/calls/#handling-responses

Here's what I've looked at that affects what I'm doing, but may not be exhaustive:

| model | cost | input_tokens | output_tokens | usage |
|---------|------|--------------|---------------|---------|
| gpt-4o-2024-11-20 | ❌ | ✅ | ✅ | ✅ |
| gpt-4o-2024-08-06 | ✅ | ✅ | ✅ | ✅ |
| gpt-4o-2024-05-13 | ✅ | ✅ | ✅ |
| gpt-4o-mini-2024-07-18 | ✅ | ✅ | ✅ | ✅ |
| gemini-2.0-flash-exp | ❌ | ❌ | ❌ | ❌ |
| gemini-1.5-flash | ❌ | ❌ | ❌ | ❌ |
| gemini-1.5-flash-8b | ❌ | ❌ | ❌ | ❌ |
| claude-3-5-sonnet-20241022 | ❌ | ✅ | ✅ | ✅ |
| claude-3-5-sonnet-20240620 | ✅ | ✅ | ✅ |
| claude-3-5-haiku-20241022 |  ❌ | ✅ | ✅ | ✅ |

```
from mirascope import llm
from mirascope.core import base
from mirascope.integrations.otel import configure, with_otel
from pydantic import BaseModel

class BookRecommendation(BaseModel):
    title: str
    author: str
    genre: str

@llm.call(provider="anthropic", model="claude-3-5-sonnet-20240620", json_mode=True, response_model=BookRecommendation)
async def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book."

async def test_call(
    provider: Providers,
    genre: str,
    model: Models = None,
) -> BookRecommendation:
    try:
        book = await llm.override(
            recommend_book,
            provider=provider.value.lower(),
            model=model.value,
        )(genre)

        response = cast(base.BaseCallResponse, book._response)
        print("cost:", response.cost)
        return book
    except Exception as e:
        print(e)
        raise HTTPException(status_code=500, detail=str(e))
```

### Python, Mirascope & OS Versions, related packages (not required)

```TOML

```

willbakst:
Solution here should just be adding the models and pricing to the calculation tables.

willbakst:
[...]

PR #794, opened 2025-01-31
feat: Update cost tables

koxudaxi:
Closes: https://github.com/Mirascope/mirascope/issues/785


koxudaxi:
@willbakst @brenkao 

I noticed that while there's no cost calculation function implemented for Gemini, the response seems to include token counts. 

I don't recall the details, so could you clarify - can't we use these values for cost calculation? 

Is there a specific reason it hasn't been implemented, such as issues with the pricing structure?

https://ai.google.dev/pricing#1_5flash

```python
from mirascope.core.gemini import gemini_call


@gemini_call("gemini-1.5-flash")


response = recommend_book("fantasy")
print(response.response.usage_metadata)
# > prompt_token_count: 5
# > candidates_token_count: 190
# > total_token_count: 195
```

willbakst:
> @willbakst @brenkao
> 
> I noticed that while there's no cost calculation function implemented for Gemini, the response seems to include token counts.
> 
> I don't recall the details, so could you clarify - can't we use these values for cost calculation?
> 
> Is there a specific reason it hasn't been implemented, such as issues with the pricing structure?
> 
> https://ai.google.dev/pricing#1_5flash
> 
> ```python
> from mirascope.core.gemini import gemini_call
> 
> 
> @gemini_call("gemini-1.5-flash")
> def recommend_book(genre: str) -> str:
>     return f"Recommend a {genre} book"
> 
> 
> response = recommend_book("fantasy")
> print(response.response.usage_metadata)
> # > prompt_token_count: 5
> # > candidates_token_count: 190
> # > total_token_count: 195
> ```

I don't remember why we didn't, be we should implement this so cost tracking for gemini works

willbakst:
@koxudaxi we should also update the code to mark all of those properties as computed fields so they show up in the model dump (i.e. cost, input tokens, etc.)

willbakst (approved):

willbakst:
I added cost tracking for Gemini, further updated OpenAI pricing with all listed models, and updated call response objects to include additional information in the dump through computed fields

Issue #811, opened 2025-02-10
Update documentation to use `llm.call` as the default everywhere that makes sense


willbakst:
I think it makes the most sense only the show the provider-specific `mirascope.core` modules when discussing provider-specific features. For example, using provider-specific messages, call parameters, features (e.g. OpenAI `strict`).

We could also remove a lot of the tabbing across the documentation by showing just a single `llm.call` method example. This would make it a lot easier to maintain the documentation long term as we add more features.

- [X] WELCOME
- [ ] Why
- [ ] Tutorials (there are so many oof)
- [x] Calls
- [x] Streams
- [x] Chaining
- [X] Response Models
- [X] JSON Mode
- [X] Output Parsers
- [x] Tools
- [x] Agents
- [x] Evals
- [x] Async
- [x] Retries
- [ ] Middleware
- [ ] MCP Client
- [ ] Integrations

teamdandelion:
I'd be happy to take this on, too, if @koxudaxi doesn't mind. It aligns well with my intention to read through and grok all the docs. 🙂

willbakst:
Sure! Happy to re-assign.

There are really two tasks that are part of this issue:

1. Replace all provider-specific usage examples with `llm.call` examples and update line highlighting as necessary.
2. Add a new section to the `Calls` documentation calls "Supported Providers" that shows all of the supported providers.
3. Add a new section called "Provider-Specific Calls" that showcases using provider specific modules (e.g. `openai`) for each provider and a brief explanation around why one might use the specific module (e.g. access to provider-specific messages etc.)
4. We can then remove all of the provider-based tabbing across the documentation and reduce down to a single `llm.call` example using e.g. `provider="openai", model="gpt-4o-mini"`
This should be a fairly rote task since all the examples already exist we're just updating them to use `llm.call`.

I leave it to you how you want to complete all four tasks involved. Feel free to send PRs piecewise for review / merge / release. Smaller PRs will be easier to review :)

willbakst:
With #817 I've added documentation for using Vertex (for the Google example using a custom client).

I'm not sure if this is sufficient for someone looking for how to use Vertex, so we may still want to mention this or something in the "Supported Providers" section.

willbakst:
After reviewing #825 that makes progress on this issue, I think we should keep the provider-specific tabs since we can dynamically generate them from a single `llm.call` example.

[...]

PR #825, opened 2025-02-17
feat: provider-agnostic Get Started docs

teamdandelion:
This commit converts the get started docs to show provider-agnostic API usage (via @llm.call), while still showing the specific code usage for each provider.

In the UI, I kept showing provider-specific code, both for consistency with the "Official SDK" comparison that's already included, but also because it's a very direct way of showing off Mirascope's wide range of supported providers.

Since the code is now nearly identical for each provider, we can unify to a single implementation (using openai gpt-4o-mini by default), and then substitute in provider-specific logic at docs build time. I've added logic for doing so (`generate_provider_examples.py`), and a pre-build hook that invokes it. The substitution logic is tested, and ensures that we get a build time error if the provider substitution fails (e.g. because an example didn't use openai/gpt-4o-mini as expected). Also, `mkdocs` is updated to watch the examples directory for file changes.

I haven't deleted any of the provider-specific implementations, since they're also used for `response_model.md`, and I wanted to get a minimal conversion out for review without leaving `response_model` in a half-switched state. However, if we move forward with this approach, we can delete most of the provider example files, leaving only cases like the sdk usage.

For cases where we don't have any provider-specific code to compare to, we could keep the per-provider tabbed structure, or we could drop it and just use openai or anthropic as consistent examples. I can see benefits to either approach, and I'm curious which you prefer.

Along the way, I made one API change, to expose `BaseCallResponse` in `mirascope.core`. This avoids bloating the example code with an extra import from `mirascope.core.base`, and seems very consistent with the other exports in `mirascope.core`. I think it'd be nice if we could get down to one import line in the standard case of using llm call, though I'm not sure if there's a principled way to do it. (Maybe expose `call` and `override` as `llm_call` and `llm_override` inside core?) 

Progress on #811

### DIFF
--- a/docs/generate_provider_examples.py
+++ b/docs/generate_provider_examples.py
@@ -31,7 +31,7 @@ PROVIDER_INFO: dict[Provider, ProviderInfo] = {
     "bedrock": ProviderInfo(
         provider="bedrock",
         title="Bedrock",
-        model="anthropic.claude-3-haiku-20240307-v1:0",
+        model="amazon.nova-lite-v1:0",
     ),
 }
 
@@ -45,8 +45,8 @@ def substitute_provider_import(content: str, provider: Provider) -> str:
     )
 
 
-def substitute_provider_cast(content: str, provider: Provider) -> str:
-    """Substitute provider in cast statement."""
+def substitute_provider_type(content: str, provider: Provider) -> str:
+    """Substitute provider in type names (e.g. openai.OpenAICallResponse)."""
     provider_name = provider.capitalize()
 
     if provider == "litellm":
@@ -55,8 +55,8 @@ def substitute_provider_cast(content: str, provider: Provider) -> str:
         provider_name = "OpenAI"
 
     return re.sub(
-        r"cast\(openai\.OpenAICallResponse,",
-        f"cast({provider}.{provider_name}CallResponse,",
+        r"openai\.OpenAI",
+        f"{provider}.{provider_name}",
         content,
     )
 
@@ -103,7 +103,7 @@ def substitute_provider_specific_content(content: str, provider: Provider) -> st
 
     content = substitute_llm_call_decorator(content, provider)
     content = substitute_provider_import(content, provider)
-    content = substitute_provider_cast(content, provider)
+    content = substitute_provider_type(content, provider)
     return content
 
 
--- a/docs/generate_provider_examples.py
+++ b/docs/generate_provider_examples.py
@@ -1,0 +1,150 @@
+import re
+from pathlib import Path
+
+from mirascope.llm import Provider
+from pydantic import BaseModel
+
+EXAMPLES_ROOT = Path("examples")
+SNIPPETS_DIR = Path("build/snippets")
+
+
+class ProviderInfo(BaseModel):
+    """Provider information for UI and substitutions."""
+
+    provider: Provider
+    title: str
+    model: str
+
+
+PROVIDER_INFO: dict[Provider, ProviderInfo] = {
+    "openai": ProviderInfo(provider="openai", title="OpenAI", model="gpt-4o-mini"),
+    "anthropic": ProviderInfo(
+        provider="anthropic", title="Anthropic", model="claude-3-5-sonnet-20240620"
+    ),
+    "mistral": ProviderInfo(
+        provider="mistral", title="Mistral", model="mistral-large-latest"
+    ),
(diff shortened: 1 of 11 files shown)


docs/generate_provider_examples.py
```diff
@@ -0,0 +1,115 @@
+import re
+from pathlib import Path
+
+from mirascope.llm._protocols import Provider
```
teamdandelion, line 4: I'm wondering if there's a reason for not exposing `Provider` through a public module? Seems nice to have especially as its part of the type signature for `llm.call`

teamdandelion: Decided to expose it in #826

docs/generate_provider_examples.py
```diff
@@ -0,0 +1,150 @@
+import re
+from pathlib import Path
+
+from mirascope.llm import Provider
+from pydantic import BaseModel
+
+EXAMPLES_ROOT = Path("examples")
+SNIPPETS_DIR = Path("build/snippets")
```
willbakst, line 8: since this is ignored for git, there's a chance the folder doesn't exist.

i tried running mkdocs serve and it failed. we should create this path if it does not already exist

docs/generate_provider_examples.py
```diff
@@ -0,0 +1,150 @@
...
+
+    provider: Provider
+    title: str
+    model: str
+
+
+PROVIDER_INFO: dict[Provider, ProviderInfo] = {
+    "openai": ProviderInfo(provider="openai", title="OpenAI", model="gpt-4o-mini"),
+    "anthropic": ProviderInfo(
+        provider="anthropic", title="Anthropic", model="claude-3-5-sonnet-20240620"
```
willbakst, line 22: nit: we should should update this to `claude-3-5-sonnet-latest`

docs/generate_provider_examples.py
```diff
@@ -0,0 +1,150 @@
...
+    ),
+    "mistral": ProviderInfo(
+        provider="mistral", title="Mistral", model="mistral-large-latest"
+    ),
+    "google": ProviderInfo(provider="google", title="Google", model="gemini-1.5-flash"),
```
willbakst, line 27: nit: we should update this to `gemini-2.0-flash`

docs/generate_provider_examples.py
teamdandelion: Great - updated

docs/generate_provider_examples.py
teamdandelion: Updated

teamdandelion:
Pushed fixes, including tests that verify we properly create the snippets directory and bring code coverage back to 100%

examples/learn/response_models/basic_usage/messages.py
```diff
@@ -0,0 +1,26 @@
...
+@llm.call(provider="openai", model="gpt-4o-mini", response_model=Book)
+def extract_book(text: str) -> Messages.Type:
+    return Messages.User(f"Extract {text}")
+
+
+book = extract_book("The Name of the Wind by Patrick Rothfuss")
+print(book)
+# Output: title='The Name of the Wind' author='Patrick Rothfuss'
+
+response = cast(openai.OpenAICallResponse, book._response)  # pyright: ignore[reportAttributeAccessIssue]
```
teamdandelion, line 24: @willbakst — do you think it's a better example to casting it as an OpenAICallResponse here, or do you think we can further simplify by casting it as `llm.CallResponse`? They both extend BaseModel so model_dump should work either way. And if we're guiding towards provider-agnosticism, probably better to guide towards llm.CallResponse in the process.

willbakst: I think casting it as `llm.CallResponse` is the better cast (re: provider-agnosticism)

teamdandelion: Sounds good - done.

teamdandelion:
Ok, I made a few more tweaks:
- Changed to provider-agnostic llm.CallResponse, but left a generalized version of the import and casting substitution logic, in case we have other examples where we do want to keep a bit of provider-specific code.
- Changed default bedrock model to `amazon.nova-lite-v1:0`, because it's a lot shorter than `anthropic.claude-3-haiku-20240307-v1:0` and will let us avoid inconsistent line-wrapping across providers

If you're okay with the defaults change, I think this should be good to merge.

willbakst (approved):

PR #826, opened 2025-02-18
feat: expose more imports through mirascope, core, and llm

teamdandelion:
This commit adds support for the following imports:

```py
from mirascope import Messages
from mirascope.core import BaseCallResponse, BaseStream
from mirascope.llm import CallResponse, Provider
```

Rationale: 

First, including Messages in the top level scope allows us to use only `from mirascope` imports for most of the example code, without needing to reference `mirascope.core`. This is particularly nice when we refactor to use `llm.call` by default (#811), so that most examples will only need a single line of imports from mirascope.

I didn't include `FromCallArgs` or `ResponseModelConfigDict`, out of conservativism since they aren't frequently used in example code, and it's breaking to remove them after they've been added. However I'm open to including them.

I added `BaseCallResponse` and `BaseStream` to the core module, but didn't promote them to top level because `BaseCallResponse` is maybe redundant to `llm.CallResponse`, and `BaseStream` isn't seeing a lot of usage in example code.

Adding `CallResponse` to `llm` seems readily appropriate. I added `Provider` too, as I've wanted it a few times as an API consumer, and it's part of the signature for the `llm.call` decorator.

### DIFF
--- a/mirascope/__init__.py
+++ b/mirascope/__init__.py
@@ -32,5 +32,6 @@ __all__ = [
     "integrations",
     "prompt_template",
     "retries",
+    "Messages",
     "__version__",
 ]
--- a/mirascope/core/__init__.py
+++ b/mirascope/core/__init__.py
@@ -4,9 +4,11 @@ from contextlib import suppress
 
 from . import base
 from .base import (
+    BaseCallResponse,
     BaseDynamicConfig,
     BaseMessageParam,
     BasePrompt,
+    BaseStream,
     BaseTool,
     BaseToolKit,
     FromCallArgs,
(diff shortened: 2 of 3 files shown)


teamdandelion:
Oops, this is broken right now. I'll re-open when fixed.

teamdandelion:
Actually, the PR is fine. I was seeing type errors in vscode (pylance extension) in `core/base/stream.py`, however they pre-exist this commit, and they aren't reproducible via pyright. I'm not sure why they are only appearing in pylance and not pyright 🤔

image

willbakst:
> Actually, the PR is fine. I was seeing type errors in vscode (pylance extension) in `core/base/stream.py`, however they pre-exist this commit, and they aren't reproducible via pyright. I'm not sure why they are only appearing in pylance and not pyright 🤔
> 
> image

Could be that pylance is using a newer version of pyright?

willbakst (approved):

willbakst:
I've approved and will push to merge this in. We can address the pylance vs. pyright issue as a separate PR

Issue #824, opened 2025-02-17
Send as File when payload limit exceeded for Google


CharlesOural:
### Description

It would be great to check if the payload exeed 20MB and then send images or part files as google cloud file and then pass the uri.
See : https://ai.google.dev/gemini-api/docs/vision?lang=python#large-images

CharlesOural:
Some context for the docs : 

- It will add any images or audio that will make the total payload size greater than the google cloud limit of 20 MB
- The hard limit is 15 Mb to have some room for the 1M token window
- Be careful, file will be stored 48h on the  google cloud and you have a free 20GB max limit

willbakst:
Released in `v1.18.3`! I considered this more of a "fix" and less of a "feature" since the interface doesn't change it just handles cases that would've otherwise not worked

PR #828, opened 2025-02-18
Add google max payload size limit

CharlesOural:
Closes https://github.com/Mirascope/mirascope/issues/824

### DIFF
--- a/mirascope/core/google/_utils/_convert_message_params.py
+++ b/mirascope/core/google/_utils/_convert_message_params.py
@@ -1,11 +1,19 @@
 """Utility for converting `BaseMessageParam` to `ContentsType`"""
 
+import asyncio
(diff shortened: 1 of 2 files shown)


CharlesOural:
I included the async logic because the google upload is vert slow... On my benchmark more than 2s per file and if not using async on a 200 pages document we need more than 10mn to just upload it (only 20MB)....

And I think mirascope should handle this logic because this can't be handle by the user as it would break the provider agnostic mirascope logic...

CharlesOural:
But the current async logic won't work... I think the function should be transformed async or threading should be used... But changing the function async is a bit too much core logic to me haha letting @willbakst take a look :)

willbakst (approved):

PR #838, opened 2025-02-19
set max payload limit to 10mb for google because of base64 33% overhead

CharlesOural:
Small fix to handle very large number of pictures, because the base64 bring a 33% overhead + some 1M text


mirascope/core/google/_utils/_convert_message_params.py
```diff
@@ -52,8 +52,8 @@ def _check_audio_media_type(media_type: str) -> None:
 
 
 def _over_file_size_limit(size: int) -> bool:
-    """Check if the total file size exceeds the limit (15mb)."""
-    return size > 15 * 1024 * 1024  # 15MB
+    """Check if the total file size exceeds the limit (10mb). Google limit is 20MB but base64 adds 33% to the size. And there is also text..."""
```
willbakst, line 55: nit: docstring is too long. Make sure the short description is shorter than the line length limit and then put the long description after with a double newline separation

mirascope/core/google/_utils/_convert_message_params.py
```diff
@@ -52,8 +52,8 @@ def _check_audio_media_type(media_type: str) -> None:
 
 
 def _over_file_size_limit(size: int) -> bool:
```
willbakst, line 54: If text is also included in the payload limit, we should include it in the total size calculation

CharlesOural: humm I tried some ways to also include the text size but nothing was quite working properly.... If you had something in mind ?

mirascope/core/google/_utils/_convert_message_params.py
CharlesOural: done

CharlesOural:
Going to sleep, i am letting you take over ;) Happy to continue if needed tomorrow

mirascope/core/google/_utils/_convert_message_params.py
willbakst: I can't find any documentation on what is counted against the payload beyond the 20mb limit for files.

I will merge this in for now with the 10mb limit and we can adjust later with more clarity

willbakst (approved):

PR #846, opened 2025-02-19
fix: incorrect exclusion of custom titles for fields in base models

### LINKED ISSUE #845: Include `title` field for tools when it's provided by the user
### Description

The custom title field is missing from the resulting schema because we strip them. We should only strip them when title is not provided by the user.

```python
from pydantic import Field

from mirascope.core import BaseTool, openai


class MyTool(BaseTool):
    """My Tool."""

    field: str = Field(title="Custom Title")


print(openai.OpenAITool.type_from_base_model_type(MyTool).tool_schema())
# {
#     "function": {
#         "name": "MyTool",
#         "description": "My Tool.",
#         "parameters": {
#             "properties": {"field": {"type": "string"}},
#             "required": ["field"],
#             "type": "object",
#         },
#     },
#     "type": "function",
# }
```

```TOML
mirascope==1.18.3
```

willbakst:
Closes #845

### DIFF
--- a/examples/learn/response_models/few_shot_examples/google/base_message_param.py
+++ b/examples/learn/response_models/few_shot_examples/google/base_message_param.py
@@ -1,5 +1,5 @@
 # ============================================================
-#                         NOT SUPPORTED                       
+#                         NOT SUPPORTED
 # ============================================================
 # This example isn't supported for this provider.
 #
--- a/examples/learn/response_models/few_shot_examples/google/messages.py
+++ b/examples/learn/response_models/few_shot_examples/google/messages.py
@@ -1,5 +1,5 @@
 # ============================================================
-#                         NOT SUPPORTED                       
+#                         NOT SUPPORTED
 # ============================================================
 # This example isn't supported for this provider.
 #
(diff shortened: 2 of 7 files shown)


examples/learn/response_models/few_shot_examples/google/base_message_param.py
```diff
@@ -1,5 +1,5 @@
 # ============================================================
-#                         NOT SUPPORTED
```
teamdandelion, line 2: I'll update my scripts so this stops happening :)

teamdandelion:
Bit of a late comment as I see this already merged but—it seems like a bit of a smell that a change to the behavior of the base tool was tested only via adding a field to an existing test of the openai module. I'm a fan of tests as de facto documentation, I'm wondering whether it would make sense to add a test to tests/core/base/test_tool.py?

willbakst:
> Bit of a late comment as I see this already merged but—it seems like a bit of a smell that a change to the behavior of the base tool was tested only via adding a field to an existing test of the openai module. I'm a fan of tests as de facto documentation, I'm wondering whether it would make sense to add a test to tests/core/base/test_tool.py?

`BaseTool` doesn't have a `tool_schema` method implemented, so I opted for testing it through a tool that does have it implemented, but you're right this can be done better.

I will submit a follow-up that fixes this

willbakst:
https://github.com/Mirascope/mirascope/pull/847

PR #869, opened 2025-02-23
feat: (mostly) migrate calls.md to provider-agnostic usage

teamdandelion:
This one is a bit of a doozy, and requires rewriting the prose because the original documentation was pretty oriented on provider-specific calls. @willbakst, I encourage taking a close read through of the docs, and I welcome you pushing commits with fixes / changes directly to my branch (I have edits by maintainers turned on).

One thing to note is I didn't change the custom messages / custom client sections, since I wasn't exactly sure how it should turn out (for CustomMessages I don't really understand what the example is trying to show, for CustomClient I don't know if there is a generic Client class). I did make a tweak to the Google example to use the `google` module not deprecated `gemini`, and made a tweak there to fix a type issue that came up on doing so. 

I updated the generator so that the override examples are generated automatically, for provider-specific usage I could have done the same, however I felt it made sense to keep them checked in for type checking purposes. 

Progress on #811


willbakst (approved):

willbakst:
Thanks for doing this! I pushed a commit with some fixes / reorganization. Let me know if you have any questions / comments on the changes I made.

I will merge this, and if there are any additional updates we can merge another PR.

Issue #870, opened 2025-02-24
Expose cost tracking as a public module


willbakst:
### Description

Currently we implement `calculate_cost` methods in `_calculate_cost.py` files for each provider. Instead, we should implement a single `calculate_cost` method in the `llm` module that supports all providers.

For example, the user would do something like `llm.calculate_cost(...)`, which would require:

1. Implementing the new `llm.calculate_cost` method by refactoring existing functionality
2. Updating `BaseCallResponse` to implement the `cost` property (rather than abstract property implemented for each provider)
3. Updating `llm.calculate_cost` to be fully accurate (e.g. claude cache write tokens are missing, openai audio token costs are missing, realtime costs are missing, etc.)

This is generally useful beyond other mirascope usage (e.g. Lilypad could use this for non-mirascope calls to calculate cost without having to duplicate the implementation in the lilypad codebase, which is what we currently do).

We'll probably want to do something like `mirascope.llm.costs` where we implement `calculate_cost.py` and then some form of `_openai_calculate_cost.py` sub-modules for each provider. We can then import `calculate_cost` into the `llm` init.

willbakst:
This will be released in the next release once we ensure correctness with #875

willbakst:
Released in `v1.20`!

PR #874, opened 2025-02-25
feat: Implement unified cost tracking API across all providers

koxudaxi:
Progress on #870


mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Args:
+        provider: The LLM provider (e.g., "openai", "anthropic")
+        model: The model name (e.g., "gpt-4", "claude-3-opus")
+        prompt_tokens: Number of input/prompt tokens
+        completion_tokens: Number of output/completion tokens (if applicable)
+        stream: Whether the call was streamed
+        **kwargs: Additional provider-specific parameters
+            - audio_seconds: For audio models like Whisper
+            - cached: Whether the response was cached (affects cost for some providers)
```
willbakst, line 43: nit: it's more than just cached or not, it's how many tokens were cached

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,107 @@
...
+]
+
+
+def calculate_cost(
+    provider: Provider,
+    model: str,
+    prompt_tokens: int,
+    completion_tokens: int | None = None,
+    stream: bool = False,
+    **kwargs: Any,  # noqa: ANN401
```
willbakst, line 28: will this ultimately be the metadata object you defined in base.types?

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,107 @@
...
+            - audio_seconds: For audio models like Whisper
+            - cached: Whether the response was cached (affects cost for some providers)
+            - images: Number or size of images for multimodal models
+            - context_length: For providers that price differently based on context window size
+
+    Returns:
+        The calculated cost in USD or None if unable to calculate
+    """
+    if provider == "openai":
+        from ._openai_calculate_cost import calculate_cost as openai_calculate_cost
```
willbakst, line 51: will these modules (which I can't find?) have anything specific for each provider?

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,107 @@
...
+
+        return cohere_calculate_cost(model, prompt_tokens, completion_tokens, **kwargs)
+    elif provider == "gemini":
+        from ._gemini_calculate_cost import calculate_cost as gemini_calculate_cost
+
+        return gemini_calculate_cost(model, prompt_tokens, completion_tokens, **kwargs)
+    elif provider == "google":
+        from ._google_calculate_cost import calculate_cost as google_calculate_cost
+
+        return google_calculate_cost(model, prompt_tokens, completion_tokens, **kwargs)
```
willbakst, line 81: we should double check if `google` has different pricing for AI Studio vs. Vertex and if so handle that here.

We can generally ignore the `gemini` and `vertex` modules since we've deprecated them.

willbakst (commented):
took a brief look, looking good so far, left some comments that came to mind as I was looking.

koxudaxi:
I have considered about interface and design a bit, and organized it in following way. I would like to confirm if this approach is acceptable.

- To prevent regression, I copied existing cost calculation functions to `llm/costs`
- Changed `CostMetadata` from TypedDict to `BaseModel`
- I was uncertain about using inheritance classes for provider-specific attributes, because decided to include all attributes, including provider-specific ones, in `CostMetadata` (we will ablet to add provider-specific metadata mdoel if we have many metadata)
- Provider-specific metadata will be implemented in `cost_metadata` property of each provider's stream and callResponse classes.

willbakst:
> @willbakst I have considered about interface and design a bit, and organized it in following way. I would like to confirm if this approach is acceptable.
> 
> * To prevent regression, I copied existing cost calculation functions to `llm/costs`
> * Changed `CostMetadata` from TypedDict to `BaseModel`
> * I was uncertain about using inheritance classes for provider-specific attributes, because decided to include all attributes, including provider-specific ones, in `CostMetadata` (we will ablet to add provider-specific metadata mdoel if we have many metadata)
> * Provider-specific metadata will be implemented in `cost_metadata` property of each provider's stream and callResponse classes.

Yeah I think we should have a single `CostMetadata` object, and for fields that a provider doesn't have, we'll just leave them empty/blank so calling `cost_metadata` is consistent right?

koxudaxi:
OK, I will proceed with the implementation.

mirascope/llm/costs/calculate_cost.py
koxudaxi: I added `cached_tokens`

mirascope/llm/costs/calculate_cost.py
koxudaxi: I change it to `metadata: CostMetadata`

mirascope/core/base/call_response.py
```diff
@@ -225,12 +226,42 @@ def output_tokens(self) -> int | float | None:
...
-        """Should return the cost of the response in dollars.
+    def cost_metadata(self) -> CostMetadata:
+        """Get metadata required for cost calculation.
 
-        If there is no cost, this method must return None.
+        Returns:
+            A dictionary of metadata relevant to cost calculation
         """
-        ...
+        raise NotImplementedError("Subclasses must implement cost_metadata property")
```
willbakst, line 235: nit: this can just be `...` since it's an abstract methods (like the others)

mirascope/core/base/call_response.py
```diff
@@ -225,12 +226,42 @@ def output_tokens(self) -> int | float | None:
...
+
+    @computed_field
+    @property
+    def cost(self) -> float | None:
+        """Calculate the cost of this API call using the unified calculate_cost function."""
+        from mirascope.llm.costs import calculate_cost
```
willbakst, line 241: any reason to lazy import this?

mirascope/core/base/call_response.py
```diff
@@ -225,12 +226,42 @@ def output_tokens(self) -> int | float | None:
...
+
+        if self.input_tokens is None or self.output_tokens is None:
+            return None
+
+        return calculate_cost(
+            provider=self.provider,
+            model=model,
+            input_tokens=self.input_tokens,
+            output_tokens=self.output_tokens,
+            cached_tokens=self.cached_tokens,
```
willbakst, line 255: shouldn't we push `cached_tokens` and other things into cost metadata? I'm wondering if maybe we should consider also putting even `input_tokens` and `output_tokens` inside of the metadata object so that it contains everything needed to calculate cost. We should still leave the properties though

mirascope/core/base/call_response.py
```diff
@@ -225,12 +226,42 @@ def output_tokens(self) -> int | float | None:
...
+            metadata=self.cost_metadata,
+        )
+
+    @property
+    def provider(self) -> Provider:
+        """Get the provider used for this API call."""
+        from ...llm import Provider
```
willbakst, line 262: same here, any reason to lazy import?

mirascope/core/base/stream.py
```diff
@@ -210,12 +215,35 @@ def _update_properties(self, chunk: _BaseCallResponseChunkT) -> None:
         if chunk.finish_reasons is not None:
             self.finish_reasons = chunk.finish_reasons
 
+    @property
+    def provider(self) -> "Provider":
```
willbakst, line 219: you put "Provider" here but didn't in the other files where you implement the same property?

mirascope/core/base/types.py
```diff
@@ -68,3 +68,94 @@ class Usage(BaseModel):
...
+    cache_write: Annotated[
+        bool | None,
+        Field(default=None, description="[Anthropic] Whether cache write occurred"),
+    ] = None
+    tool_use_tokens: Annotated[
+        int | None,
+        Field(default=None, description="[Anthropic] Tokens used for tool calls"),
+    ] = None
+
+    # Vertex/Google-specific fields
```
willbakst, line 145: nit: since we've deprecated `gemini`/`vertex`, do we need this still if we're only supporting `google` going forward?

mirascope/core/base/types.py
```diff
@@ -68,3 +68,94 @@ class Usage(BaseModel):
...
+    character_count: Annotated[
+        int | None,
+        Field(
+            default=None,
+            description="[Vertex/Google] Character count for character-based pricing models",
+        ),
+    ] = None
+
+    # Gemini-specific fields
```
willbakst, line 154: nit: wouldn't this be "Google-specific fields"?

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,127 @@
+"""Cost calculation utilities for LLM API calls."""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from mirascope.core.base.types import CostMetadata
+
+Provider = Literal[
```
willbakst, line 9: don't we already have `Provider` defined in the types module?

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,127 @@
...
+    if cached_tokens is None:
+        cached_tokens = 0
+
+    # Initialize empty metadata if none provided
+    if metadata is None:
+        metadata = CostMetadata()
+
+    # Route to provider-specific implementations
+    if provider == "openai":
+        from ._openai_calculate_cost import calculate_cost as openai_calculate_cost
```
willbakst, line 59: is there any reason to lazy import these? I don't see anything provider specific for any of the functions?

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+
+    elif provider == "xai":
+        from ._xai_calculate_cost import calculate_cost as xai_calculate_cost
+
+        return xai_calculate_cost(input_tokens, cached_tokens, output_tokens, model)
+
+    elif provider == "litellm":
+        # LiteLLM currently does not support cost calculation
+        return None
```
willbakst, line 125: LiteLLM does support cost calculation. See the original LiteLLM cost property and how we did it so we can support it here

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,127 @@
...
+    "groq",
+    "litellm",
+    "mistral",
+    "openai",
+    "vertex",
+    "xai",
+]
+
+
+def calculate_cost(
```
willbakst, line 25: we should update `calculate_cost` here to take only `provider`, `model`, and `metadata` where `metadata` contains all the info we may need to calculate cost

mirascope/llm/costs/calculate_cost.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Initialize empty metadata if none provided
+    if metadata is None:
+        metadata = CostMetadata()
+
+    # Route to provider-specific implementations
+    if provider == "openai":
+        from ._openai_calculate_cost import calculate_cost as openai_calculate_cost
+
+        return openai_calculate_cost(input_tokens, cached_tokens, output_tokens, model)
```
willbakst, line 61: we should update all of the provider-specific cost calculation methods to take `model` and `metadata` so that we can later add things to metadata and update cost calculation without updating the signature

mirascope/llm/stream.py
```diff
@@ -178,3 +179,7 @@ def tool_message_params(  # pyright: ignore [reportIncompatibleMethodOverride]
                 message parameters should be constructed.
         """
         return CallResponse.tool_message_params(tools_and_outputs)
+
+    @property
+    def cost_metadata(self) -> CostMetadata:
```
willbakst, line 184: I don't believe this property is necessary. We pass through all properties to `self._stream`, so the only reason to implement the property is if we need to do something different / provider-agnostic

mirascope/core/base/call_response.py
koxudaxi: To avoid circular reference errors during import.

mirascope/core/base/call_response.py
koxudaxi: To avoid circular reference errors during import.

mirascope/core/anthropic/call_response.py
```diff
@@ -202,3 +194,9 @@ def common_user_message_param(self) -> BaseMessageParam | None:
         return AnthropicMessageParamConverter.from_provider(
             [(self.user_message_param)]
         )[0]
+
+    @computed_field
+    @property
+    def cost_metadata(self) -> CostMetadata:
+        """Get metadata required for cost calculation."""
+        return CostMetadata()
```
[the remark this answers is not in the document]
koxudaxi: I implemented existing attributes such as input_tokens in the parent class. 
There are no plans to implement new attributes in this PR.

mirascope/core/base/call_response.py
willbakst: what causes it to be circular? we're not importing anything from here in the costs module, are we?

if circular imports are an issue, should we just move things to `mirascope.costs` to more easily avoid this?

mirascope/core/base/stream.py
```diff
@@ -211,10 +212,33 @@ def _update_properties(self, chunk: _BaseCallResponseChunkT) -> None:
...
     def cost(self) -> float | None:
-        """Returns the cost of the stream."""
-        ...
+        """Calculate the cost of this streaming API call."""
+        from mirascope.llm.costs import calculate_cost
+
+        if self.input_tokens is None or self.output_tokens is None:
+            return None
+
+        from ...llm import Provider
```
willbakst, line 235: nit: same here, import `from ...llm.types import Provider` should work without issue right?

willbakst: should we move `Provider` into `mirascope.core`? i feel like that might make more sense? and then we can export it in the `llm` module from core for easy import (and so we don't break current imports)

mirascope/core/anthropic/call_response.py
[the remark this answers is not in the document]
willbakst: why not? i feel like that's somewhat important for a unified cost tracking API that is correct/accurate?

if not as part of this PR, we should post another issue and work on that next

mirascope/core/base/call_response.py
willbakst: see other comments

mirascope/core/base/stream.py
koxudaxi: I moved the Provider to `mirascope.core`

mirascope/llm/costs/calculate_cost.py
koxudaxi: I moved the implementation to here

examples/learn/agents/state_management/provider_agnostic/base_message_param.py
```diff
@@ -15,7 +16,7 @@ def _call(self, query: str) -> list[BaseMessageParam]:
 
     def run(
         self,
-        provider: llm.Provider,
```
willbakst, line 18: ideally this would remain unchanged (by exporting `Provider` from the `llm` module).

We should also definitely export `Provider` from the base init so that we can do `from mirascope import Provider`

mirascope/core/base/call_response.py
```diff
@@ -248,20 +252,15 @@ def cost(self) -> float | None:
             return None
 
         return calculate_cost(
-            provider=self.provider,
+            provider=cast(Provider, self.provider),
```
willbakst, line 255: why cast here? why not instead just update `self.provider` to return `Provider`?

mirascope/core/base/stream.py
```diff
@@ -216,31 +213,28 @@ def _update_properties(self, chunk: _BaseCallResponseChunkT) -> None:
...
     def cost(self) -> float | None:
         """Calculate the cost of this streaming API call."""
-        from mirascope.llm.costs import calculate_cost
 
         if self.input_tokens is None or self.output_tokens is None:
             return None
 
         return calculate_cost(
-            provider=self.provider,
+            provider=cast(Provider, self.provider),
```
willbakst, line 236: same here

mirascope/costs/_anthropic_calculate_cost.py
```diff
@@ -1,10 +1,10 @@
 """Calculate the cost of a completion using the Anthropic API."""
 
+from mirascope.core.base.types import CostMetadata
```
willbakst, line 3: why not relative import here?

mirascope/core/base/call_response.py
```diff
@@ -252,15 +252,15 @@ def cost(self) -> float | None:
...
             model=model,
             metadata=self.cost_metadata,
         )
 
     @property
-    def provider(self) -> str:
+    def provider(self) -> Provider:
         """Get the provider used for this API call."""
-        return self._provider
+        return cast(Provider, self._provider)
```
willbakst, line 263: why cast here and not just make `_provider: Provider`?

examples/learn/agents/state_management/provider_agnostic/base_message_param.py
koxudaxi: OK, I have reverted it.

mirascope/core/base/stream.py
koxudaxi: same changes

mirascope/core/base/call_response.py
koxudaxi: same changes

mirascope/core/base/types.py
```diff
@@ -142,20 +163,28 @@ class CostMetadata(BaseModel):
...
 
-    # Vertex/Google-specific fields
-    character_count: Annotated[
-        int | None,
-        Field(
-            default=None,
-            description="[Vertex/Google] Character count for character-based pricing models",
-        ),
+    # LiteLLM-specific fields
+    response: Annotated[
```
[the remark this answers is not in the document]
koxudaxi: I have chnaged it

mirascope/costs/_anthropic_calculate_cost.py
koxudaxi: I found many absolute imports. I will fix it soon

willbakst (approved):

PR #877, opened 2025-02-28
feat: check that all docs code snippets resolve

teamdandelion:
There are a lot of docs, and a lot of code snippets in those docs - over 2300. They are split between checked-in and autogenerated files, and often parameterized by provider, by API style, and others.

As a result, it's really easy to add missing imports (or break existing ones)—especially when adding a new provider. As a testament to that, there are currently 95 missing snippets. Most are related to the addition of the xai provider, but there are also broken google examples in the integrations docs.

This PR doesn't fix all of the missing snippets (it makes sense to wrap up #811 first as that will resolve many but not all missing xai snippets on its own). Instead, it adds a tool for enumerating all missing snippets, and an automated test to ensure no new missing snippets are added. I grandfathered in existing missing snippets so that the build passes.

Note: Doing this "properly" would require parsing the jinja2 templates to figure out what variables are in scope, for the cases were we have non-standard variables like `{{ tool_method }}` in `docs/learn/tools.md`. This seemed really hard, so I took the expedient approach of hardcoding in the non-standard variables. Whenever new nonstandard variables are used in snippet path definitions in documentation, this file will need to be updated, too... which will happen organically due to test failures.

### DIFF
--- a/docs/find_missing_snippets.py
+++ b/docs/find_missing_snippets.py
@@ -1,0 +1,141 @@
+"""Scope:
+Finds every code import in docs/learn/**/*.md, and checks that the target file exists.
+
+Usage: python -m docs.find_missing_snippets
+
+Note: The script depends on hand-maintained substitutions
+in FILE_SPECIFIC_SUBSTITUTUTIONS, which may require updates as documents
+are changed or updated. There is automated testing to ensure that all new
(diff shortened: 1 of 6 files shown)


willbakst:
this is a great step, agree that the full on jinja impl would be overkill

willbakst (approved):

Issue #875, opened 2025-02-27
Update `CostMetadata` to capture all information so cost tracking is fully accurate


willbakst:
### Description

Right now we only capture input, output, and cached tokens. We should capture all cost metadata to ensure we're properly calculating cost across all providers and various token types (e.g. audio, cache writes, etc.)

willbakst:
We should be mindful to cover all cost calculation. For example, we will need to update streaming. See [here](https://github.com/Mirascope/mirascope/blob/9e55c19842f9356ff6be4cd24cd7442ff3085f1e/mirascope/core/base/stream.py#L185)

willbakst:
This is merged and will be released shortly

willbakst:
Released in `v1.20`!

PR #878, opened 2025-02-28
Update llm costs

koxudaxi:
Related Issues: https://github.com/Mirascope/mirascope/issues/875


mirascope/core/google/call_response.py
```diff
@@ -222,4 +222,5 @@ def common_user_message_param(self) -> BaseMessageParam | None:
 
     @property
     def cost_metadata(self) -> CostMetadata:
+        # TODO: Add VertxAI flag to the cost metadata
```
willbakst, line 225: confused about this TODO in a PR that's supposed to implement up-to-date cost tracking. I'm assuming Vertex results in different pricing? Why not handle it as part of this PR?

mirascope/core/openai/_utils/_convert_message_params.py
```diff
@@ -83,8 +83,8 @@ def convert_message_params(
                     audio = _load_media(part.url)
                     audio_type = get_audio_type(audio)
                     if audio_type not in [
```
willbakst, line 85: below for "format" we split assuming the slash is there. since it isn't we should update that below as well to just use audio type as-is

mirascope/core/groq/call_response.py
```diff
@@ -189,4 +189,21 @@ def common_user_message_param(self) -> BaseMessageParam | None:
 
     @property
     def cost_metadata(self) -> CostMetadata:
-        return super().cost_metadata
+        cost_metadata = super().cost_metadata
```
willbakst, line 192: why call super here and not just return `CostMetadata(...)`?

mirascope/core/costs/_openai_calculate_cost.py
```diff
@@ -2,6 +2,83 @@
...
 from ..base.types import CostMetadata
 
+# Constants for image token calculation
+LOW_DETAIL_IMAGE_TOKENS = 85
+HIGH_DETAIL_TILE_TOKENS = 170
+HIGH_DETAIL_BASE_TOKENS = 85
+TILE_SIZE = 512
+
+
+def calculate_image_tokens(metadata: CostMetadata) -> int | None:
```
willbakst, line 12: nit: make this private to this module

mirascope/core/costs/_cohere_calculate_cost.py
```diff
@@ -12,8 +12,8 @@ def calculate_cost(
     https://cohere.com/pricing
 
     Model              Input               Cached     Output
-    command-r          $0.5 / 1M tokens	              $1.5 / 1M tokens
-    command-r-plus     $3 / 1M tokens	              $15 / 1M tokens
+    command-r          $0.15 / 1M tokens	          $0.6 / 1M tokens
```
willbakst, line 15: nit: weird spacing?

mirascope/core/costs/_anthropic_calculate_cost.py
```diff
@@ -14,6 +14,7 @@ def calculate_cost(
     Model                                      Input                 Cached                Output
     claude-3-5-haiku                           $0.80  / 1M tokens    $0.08  / 1M tokens    $4.00  / 1M tokens
     claude-3-5-haiku-20241022                  $0.80  / 1M tokens    $0.08  / 1M tokens    $4.00  / 1M tokens
+    claude-3-7-sonnet                          $3.00  / 1M tokens    $0.30  / 1M tokens    $15.00 / 1M tokens
```
willbakst, line 17: I believe all of these need `-latest` (if you try running it with this model name it fails)

also, what about date specific tags like we have for the other models?

mirascope/core/google/call_response.py
willbakst: it looks like you do handle it in the cost calculation, is this simple missing the flag for whether or not it's using vertex?

koxudaxi: I was thinking about how to pass the flag and forgot to implement it.

mirascope/core/costs/_cohere_calculate_cost.py
koxudaxi: image

It's good in my window.

koxudaxi: it was `tab`. I changed it to space

mirascope/core/groq/call_response.py
koxudaxi: I updated it.

mirascope/core/openai/_utils/_convert_message_params.py
koxudaxi: OK , I updated it.

mirascope/core/costs/_anthropic_calculate_cost.py
koxudaxi: I fixed the model names

willbakst (approved):

PR #880, opened 2025-02-28
feat: migrate chaining.md to llm.call api usage

teamdandelion:
More progress on #811.

The import testing in #877 immediately paid off—I would otherwise have broken imports in WHY.md without noticing it.

I removed the "shorthand" api examples for computed fields, since those examples were actually just copy+pastes of the messages api style. It appears that there is no way to provide computed fields in shorthand style, so I made a note to that effect, though please correct me if I'm missing something.

Number of missing xai examples went down from 79 to 65. :)


docs/learn/chaining.md
```diff
@@ -75,28 +75,32 @@ We recommend using nested chains for better observability when using tracing too
...
                 ```python hl_lines="16 23-24"
                 {% endif %}
-                --8<-- "examples/learn/chaining/computed_fields/{{ provider | provider_dir }}/{{ method }}.py"
+                --8<-- "build/snippets/learn/chaining/computed_fields/{{ provider | provider_dir }}/{{ method_not_shorthand }}.py"
                 ```
             {% endfor %}
+        {% endif %}
         {% endfor %}
 
+    Since specifying computed fields requires returning additional fields, they are not supported in the shorthand API style.
```
willbakst, line 102: for such cases I have generally just matched the shortand and messages examples 1:1 (rather than calling it unsupported). personally i find this makes more sense so that if i select shorthand and am going through the docs I still see the "shortest"-hand version of a given feature

docs/find_missing_snippets.py
```diff
@@ -40,6 +40,13 @@
         "{{ audio_pkg }}": ["pydub", "wave"],
         "{{ filename }}": ["lists", "texts", "parts"],
     },
+    "docs/learn/chaining.md": {
```
willbakst, line 43: re: other comment, this should become unnecessary

docs/learn/chaining.md
teamdandelion: Makes sense, re-added

willbakst (approved):

PR #885, opened 2025-03-02
feat: migrate retries.md to llm.call

teamdandelion:
#811


mkdocs.yml
```diff
@@ -112,6 +112,10 @@ extra:
     - learn/response_models/json_mode
     - learn/response_models/streaming
     - learn/response_models/validation
+    - learn/retries/calls
+    - learn/retries/error_reinsertion
+    - learn/retries/streams
+    - learn/retries/tools
```
willbakst, line 118: what about fallbacks? are we not including that because the examples use two providers?

docs/learn/retries.md
willbakst, line 1: while reviewing a noticed a separate typo in the second fallback example that has `gpt-4o-minis` with an extra "s"

if you could fix that as part of this PR that would be great

mkdocs.yml
teamdandelion: This config is used to track where we use the generate_provider_examples code to produce per-provider example code for the cross-provider tabbing. As written, the fallback examples don't have cross-provider tabbing, just a fixed example per method that shows two providers (openai and anthropic). That example was already written using `llm.call`. so I didn't see a need to change anything about it.

docs/learn/retries.md
teamdandelion: Fixed :)

willbakst (approved):

Issue #884, opened 2025-03-02
`llm.context`: ContextManager for overriding llm call parameters


teamdandelion:
### Description

Consider a case where you want to compare outputs from a particular call across multiple providers, while writing the code in a provider-agnostic way. For example, we want to compare openai and anthropic. There are a few ways to do this - for example by using `llm.override`, or a directly invoking `llm.call` on a prompt template. 

### `llm.override`
```python
@llm.call(provider="openai", model="gpt-4o-mini")
def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book"


# OpenAI
openai_response = recommend_book("fantasy")
print(openai_response.content)

# Anthropic
anthropic_response = llm.override(
    recommend_book, provider="anthropic", model="claude-3-5-sonnet-latest"
)("fantasy")
print(anthropic_response.content)
```

### defining multiple calls
```python
@prompt_template()
def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book"

# OpenAI
openai_call = llm.call(provider="openai", model="gpt-4o-mini")
openai_recommend_book = openai_call(recommend_book)
openai_response = openai_recommend_book("fantasy")
print(openai_response.content)

# Anthropic
anthropic_call = llm.call(provider="anthropic", model="claude-3-5-sonnet-latest")
anthropic_recommend_book = anthropic_call(recommend_book)
anthropic_response = anthropic_recommend_book("fantasy")
print(anthropic_response.content)
```

These both feel a little awkward, and require breaking out of the nice "just use decorators" approach and using higher order functions just for the basic task of runtime provider specification.

I'd like to suggest `llm.context` as a context-manager based approach for overriding llm parameters, as in:

```python
@llm.call()
def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book"

# OpenAI
with llm.context(provider="openai", model="gpt-4o-mini"):
    openai_response = recommend_book("fantasy")
    print(openai_response.content)

# Anthropic
with llm.context(provider="anthropic", model="claude-3-5-sonnet-latest"):
    anthropic_response = recommend_book("anthropic")
    print(anthropic_response.content)
```

Ideally, this `with llm.context(...):` approach would be able to modify all arguments to `llm.call` which don't require changing the type signature, and without needing to manually track the context manager variable. 

Per discussion with @willbakst, when the type signature would change (ie. to turn on streaming, or add a response model), a higher-order function approach is likely needed, like `llm.override`. 

```python
[...]

PR #909, opened 2025-03-10
fix: make llm.context work for async functions

teamdandelion:
As presently implemented, the new `llm.context` api (#884) does not consistently work for async functions. Consider the following two examples:

```python
import asyncio

from mirascope import llm


@llm.call(provider="openai", model="gpt-4o-mini")
async def introspect_model() -> str:
    return f"What model are you?"


async def example1():
    print("example1:")
    openai_response = await introspect_model()
    print(openai_response.content)

    with llm.context(provider="anthropic", model="claude-3-5-sonnet-latest"):
        anthropic_response = await introspect_model()
        print(anthropic_response.content)


async def example2():
    print("example2:")
    openai_response_future = introspect_model()

    with llm.context(provider="anthropic", model="claude-3-5-sonnet-latest"):
        anthropic_response_future = introspect_model()

    openai_response, anthropic_response = await asyncio.gather(
        openai_response_future, anthropic_response_future
    )

    print(openai_response.content)
    print(anthropic_response.content)


async def main():
    await example1()
    await example2()


if __name__ == "__main__":
    asyncio.run(main())
```

Right now example 1 works but example 2 doesn't:

```
example1:
I am based on OpenAI's GPT-3 model. How can I assist you today?
I'm Claude, an AI assistant created by Anthropic. I aim to be direct and honest in my communications.

example2:
I am based on OpenAI's GPT-3 model, specifically designed for providing information and answering questions across a wide range of topics. If you have any specific questions or need assistance, feel free to ask!
I am based on OpenAI's GPT-3.5 architecture. How can I assist you today?
```

The issue being in the second example, when we actually execute the code, we're no longer inside the context of the context manager.

This commit fixes the issue by capturing the context into a closure at call time for the decorated function. I've added tests: specifically two tests to `test_call.py`, of which only the first one passes under the current implementation, and both pass with the fixed implementation. 

Note, in the present implementation `llm.override` works fine. However, I added a similar test case to `test_override` for caution.

### DIFF
--- a/mirascope/llm/_call.py
+++ b/mirascope/llm/_call.py
@@ -20,7 +20,11 @@ from ..core.base import (
 from ..core.base._utils import fn_is_async
 from ..core.base.stream_config import StreamConfig
 from ..core.base.types import LocalProvider, Provider
-from ._context import CallArgs, apply_context_overrides_to_call_args
+from ._context import (
+    CallArgs,
+    apply_context_overrides_to_call_args,
+    get_current_context,
+)
 from ._protocols import (
     AsyncLLMFunctionDecorator,
     CallDecorator,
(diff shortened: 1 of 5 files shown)


mirascope/llm/_call.py
```diff
@@ -235,55 +239,66 @@ def wrapper(
         | Awaitable[(_ResponseModelT | CallResponse)],
     ]:
         if fn_is_async(fn):
-
+            # Create a wrapper function that captures the current context when called
             @wraps(fn)
-            async def inner_async(
```
willbakst, line 240: why switch this from async to Awaitable?

willbakst (approved):

willbakst:
Great catch, thank you for submitting the fix!

One question, otherwise approved! If it requires a change I'll approve again if it's stale.

mirascope/llm/_call.py
teamdandelion: The key is when we evaluate `get_current_context()`. If we have a sync function that returns an awaitable, then we get the context at call time (when the future is created), which fixes the issue. If we have an async function, we defer execution until later when the function is awaited, but by that time we are potentially no longer in the correct `llm.context` scope.

willbakst: aha I see thank you for explaining, that makes sense. I was misunderstanding the bug initially, but this makes it make sense (and why I missed it originally)

willbakst:
Just released this in `v1.21.4`

Issue #896, opened 2025-03-08
Mirascope `v2.0` Roadmap


willbakst:
## Why `v2.0` major version bump?

There are some items on my laundry list of TODOs for Mirascope that are breaking changes, so I've been ignoring them / pushing them off until the time felt right to do a big push and release a new major version (which is a lot of work).

Namely:
- Clean up `core` in favor of `llm.call` as the default
- Big focus on performance including all non-API call benchmarks
    - Mirascope should (ideally) not really exist when it comes to the clock.
    - That may be true today. We should know for sure (and make it so).
- Shift the package more towards only having to learn Mirascope
    - Essentially this means the "Learn" section should only ever import from Mirascope
    - Anything provider-specific should only be necessary for truly provider-specific features
- Clean up class / function naming to remove `Base` from anything where users are not subclassing
    - Users are using `BaseMessageParam` but not subclassing it, so `MessageParam` should suffice
    - Same is true for `BaseDynamicConfig` etc.
- Make `pydantic` an optional dependency
    - While Pydantic is great for many things, it introduces a lot of unnecessary overhead
        - (namely in places where we use it internally and don't have to)
    - This breaks the above point about "only Mirascope" since Response Models require Pydantic (but don't have to)
    - For users who want Pydantic, they can and should be able to use it.

## Why push for `llm` as the default?

Originally we implemented Mirascope such that it would be easy to switch providers, but the interface was not truly agnostic. You would need to use higher-order functions (decorators) dynamically on `@prompt_template` decorated functions. This is not great.

We are making strides towards a truly provider-agnostic interface with `llm.call` and `llm.override` etc. but still rely on the original provider-specific calls followed by conversion when requested or on construction. This results in a bunch of unnecessary compute time spent on all of the provider-specific class creation rather than just always converting everything to a common type.

**If we're hoping to build the standardized base interface for building with LLMs, then everything must be provider-agnostic by default.**

Accessing provider-specific features should be possible with minimal changes to existing code and only necessary when Mirascope does not natively support such a feature (e.g. OpenAI releases a new feature that only they support).

To me, this means that for almost all use-cases, `llm.call` and other `llm` module methods (e.g. `override`) should be sufficient (and in fact be the right solution). We can then implement new/different interfaces that work with `llm` or are separate for things that are provider-specific.

[...]

PR #930, opened 2025-03-21
feat: add `llm.prompt`

teamdandelion:
This is a first stab at implementing the `llm.prompt` api, as discussed in #884, particularly https://github.com/Mirascope/mirascope/issues/884#issuecomment-2701750473.

It is basically a sibling to `llm.call` that doesn't require providing the model and provider upfront, but rather assumes that they will be present via the `llm.context` context manager. Rather than re-implement the logic, under the hood it has some minimal wrapping and then calls `llm.call`. Once we implement performance benchmarking (as discussed in #896), we may find it's better not to wrap `llm.call` here, but for a proof of concept for the interface, this seemed fine.

I copied a lot of the (complex) type signature for `llm.call`, including making a `PromptDecorator` class, which may be cargo cult-y.

Initially, rather than taking a simple decorator approach, I wanted to return a callable `Prompt` class. This would have enabled nice APIs like chaining function calls on the prompt in order to change its structural parameters; e.g. if `prompt.stream(True)` would return a new `Prompt` that has streaming enabled, or `prompt.response_model(Book)` would make a new version of the same prompt with a response model configured. However, I found this to be a nightmare from a typing perspective, so I backed off to just implementing a decorator a-la call.

I think there's value in having a way to set up decorated llm calls while not providing the provider or model upfront, hence advocating for this API. Though, I wonder if having a separate `llm.prompt` api distinct from `llm.call` is worth it, versus just making the provider and model optional on `llm.call`, with an error thrown at runtime if no default model/provider are present, and it was called outside of a context.

I didn't update docs or examples in this PR, however here's a code snippet which both works and puts `llm.prompt` a bit through the paces:

```python
from mirascope import llm


# Simple example using the prompt decorator
@llm.prompt()
    """Recommend a book in the given genre."""
    return f"Recommend a {genre} book."


# Example with response model
class Book(BaseModel):
    """An extracted book recommendation."""

    title: str
    author: str
    year: int
    description: str


@llm.prompt(response_model=Book)
def extract_book(genre: str) -> str:
    """Extract a structured book recommendation."""
    return f"""
    Recommend a {genre} book. Respond with a JSON object with the following fields:
    - title: the title of the book
    - author: the author of the book
    - year: the year the book was published
    - description: a brief description of the book
    """


# Example with streaming
@llm.prompt(stream=True)
def stream_book_review(book_title: str) -> str:
    """Generate a streaming review for the given book."""
    return f"Write a detailed review of the book '{book_title}'."


def main():
    print("Provider-agnostic LLM prompts with mirascope.llm.prompt\n")

    # Example 1: Simple prompt with OpenAI
    print("Example 1: Simple book recommendation with OpenAI")
    with llm.context(provider="openai", model="gpt-4o-mini"):
        openai_response: llm.CallResponse = recommend_book("fantasy")
        print(f"OpenAI Response: {openai_response.content}")

    # Example 2: Same prompt with Anthropic
    print("\nExample 2: Same prompt with Anthropic")
    with llm.context(provider="anthropic", model="claude-3-haiku-20240307"):
        anthropic_response: llm.CallResponse = recommend_book("science fiction")
        print(f"Anthropic Response: {anthropic_response.content}")

    # Example 3: Using a structured response model
    print("\nExample 3: Structured response with OpenAI")
    with llm.context(provider="openai", model="gpt-4o-mini"):
        book: Book = extract_book("mystery")
        print(f"Extracted Book: {book}")
        print(f"Book Title: {book.title}")
        print(f"Book Author: {book.author}")

    # Example 4: Streaming response
    print("\nExample 4: Streaming response")
    with llm.context(provider="openai", model="gpt-4o-mini"):
        stream: llm.Stream = stream_book_review("The Lord of the Rings")
        print("Streaming review:")
        for chunk, _ in stream:
            # Print without newline and flush to show streaming effect
            print(chunk.content, end="", flush=True)
        print()  # Add a newline at the end


if __name__ == "__main__":
    main()
```

### DIFF
--- a/mirascope/llm/_prompt.py
+++ b/mirascope/llm/_prompt.py
@@ -119,7 +119,7 @@ def _prompt(
                 | (_ResponseModelT | CallResponse)
             ]:
                 # Capture the context at call time, not await time
-                current_context = get_current_context()
+                context = get_current_context()
 
                 # Define an async function that uses the captured context
                 async def context_bound_inner_async() -> (
@@ -129,26 +129,24 @@ def _prompt(
                     | _ParsedOutputT
                     | (_ResponseModelT | CallResponse)
                 ):
-                    if current_context is None:
+                    if (
+                        context is None
+                        or context.provider is None
+                        or context.model is None
+                    ):
                         raise ValueError(
-                            "Prompt can only be called within a llm.context"
+                            "Prompt can only be called within a llm.context with a provider and model."
+                            "Use `llm.context(provider=..., model=...)` to specify the provider and model."
                         )
 
-                    provider = current_context.provider
-                    model = current_context.model
-                    if provider is None:
-                        raise ValueError("Provider must be specified in the context")
-                    if model is None:
-                        raise ValueError("Model must be specified in the context")
-
                     # Get provider-specific overrides from context
-                    client_override = current_context.client or client
-                    call_params_override = current_context.call_params or call_params
+                    client_override = context.client or client
+                    call_params_override = context.call_params or call_params
 
                     # Create a dynamically decorated version of our function
                     decorated = call(  # type: ignore[reportCallIssue]
-                        provider=cast(Provider, provider),
-                        model=model,
+                        provider=cast(Provider, context.provider),
+                        model=context.model,
                         stream=stream,  # type: ignore[reportArgumentType]
                         tools=tools,
                         response_model=response_model,  # type: ignore[reportArgumentType]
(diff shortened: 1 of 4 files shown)


mirascope/llm/_prompt.py
```diff
@@ -0,0 +1,258 @@
...
+                *args: _P.args, **kwargs: _P.kwargs
+            ) -> (
+                CallResponse
+                | Stream
+                | _ResponseModelT
+                | _ParsedOutputT
+                | (_ResponseModelT | CallResponse)
+            ):
+                context = get_current_context()
+                if context is None or context.provider is None or context.model is None:
```
willbakst, line 179: nit: refactor this into a private utility with the above duplicate (just so the messages will always match)

mirascope/llm/_protocols.py
```diff
@@ -498,3 +498,334 @@ def __call__(
     Any,
     Any,
 ]
+
+
+class PromptDecorator(
```
willbakst, line 503: I see your point here now about whether we need `llm.prompt` or if `llm.call` should just make `provider/model` optional and raise the error if not provided at runtime, especially since the return types aren't different so the only way to know you need context is to see it has the `llm.prompt` decorator (at which point you could see that it doesn't have model/provider). Could also potentially update to return something like `RequiresContext[...]` as a type wrapper passthrough or something if model or provider aren't provided.

mirascope/llm/_prompt.py
teamdandelion: Done

mirascope/llm/_protocols.py
willbakst: Hmm, I think this will actually be moot somewhat when we merge provider and model.

In the short term, we could provide the following type aliases:

```python
RequiresContext[T]: TypeAlias = T
RequiresProvider[T]: TypeAlias = T
RequiresModel[T]: TypeAlias = T
```

We would then remove `RequiresProvider` and `RequiresModel` as part of the merge.

I imagine these do little but let someone know that it requires context right where they are in the code.

I think this would be helpful/useful even without explicit lint errors.

PR #933, opened 2025-03-26
feat: verify uv.lock is up-to-date in CI

teamdandelion:
We recently [bumped the version #](https://github.com/Mirascope/mirascope/commit/7591658aba56fbd518698aa83356613db59ec1d7) without updating `uv.lock`, with the result that the lockfile is out of date, and any usage of uv will tend to produce a somewhat confusing extra diff to a checked in file.

This PR adds `uv lock --check` to the CI lint workflow so build will fail if `uv.lock` is outdated. I'm intentionally committing just the workflow change first, so we can verify that CI correctly fails and with a helpful error message. Afterwards, I'll update `uv.lock`, at which point this PR will be good to merge.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,7 @@ jobs:
         run: uv lock --check
 
       - name: Set Up Python Environment
-        run: uv python install 3.11
+        run: uv python install 3.10
 
       - name: Restore uv cache
         uses: actions/cache@v4
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,8 +21,12 @@ jobs:
       - uses: codespell-project/actions-codespell@v2
         with:
           exclude_file: "docs/tutorials/evals/evaluating_web_search_agent.ipynb,docs/tutorials/agents/web_search_agent.ipynb,docs/tutorials/more_advanced/knowledge_graph.ipynb"
+
       - name: Set up uv
-        run: curl -LsSf https://astral.sh/uv/0.5.6/install.sh | sh
+        run: curl -LsSf https://astral.sh/uv/0.6.10/install.sh | sh
+
+      - name: Check uv lock file is up-to-date
+        run: uv lock --check
 
       - name: Set Up Python Environment
         run: uv python install 3.11
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,4 @@
 version = 1
-revision = 1
 requires-python = ">=3.10"
 resolution-markers = [
     "python_full_version < '3.11' and sys_platform == 'linux'",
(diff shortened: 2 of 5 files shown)


.github/workflows/lint.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         with:
           exclude_file: "docs/tutorials/evals/evaluating_web_search_agent.ipynb,docs/tutorials/agents/web_search_agent.ipynb,docs/tutorials/more_advanced/knowledge_graph.ipynb"
       - name: Set up uv
-        run: curl -LsSf https://astral.sh/uv/0.5.6/install.sh | sh
+        run: curl -LsSf https://astral.sh/uv/0.5.31/install.sh | sh
```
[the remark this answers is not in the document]
teamdandelion: Sure, done. I also made the change across other CI tasks that install uv.

Btw, not sure if it really matters, but releases.yml uses python 3.10 and lint.yml uses python 3.11

teamdandelion:
I got the [clean CI failure](https://github.com/Mirascope/mirascope/actions/runs/14089840929/job/39463462055?pr=933) I was hoping for, so I've now pushed the fixed commit.

uv.lock
```diff
@@ -3324,7 +3323,6 @@ requires-dist = [
     { name = "typing-extensions", specifier = ">=4.10.0" },
     { name = "websockets", marker = "extra == 'realtime'", specifier = ">=13.1,<14" },
 ]
-provides-extras = ["anthropic", "azure", "bedrock", "cohere", "gemini", "google", "groq", "tenacity", "hyperdx", "langfuse", "litellm", "logfire", "mistral", "openai", "opentelemetry", "vertex", "realtime", "mcp", "xai"]
```
teamdandelion, line 3327: Not sure why this line came out, @willbakst does it look concerning?

.github/workflows/lint.yml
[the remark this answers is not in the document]
willbakst: oh hmm they should probably all use 3.10 since that's the lowest version we support

uv.lock
willbakst: Could be part of the `0.5` -> `0.6` bump where this is no longer needed?

teamdandelion: I was getting this diff on 0.5 too

willbakst: have you been running `uv sync --all-extras --dev`?

teamdandelion: I haven't, however running this doesn't change the generated lockfile

willbakst: then i think this should be fine

willbakst (dismissed):

.github/workflows/lint.yml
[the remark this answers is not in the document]
teamdandelion: Ok, I added a nit commit changing lint.yml to use 3.10

willbakst (approved):

Issue #936, opened 2025-03-28
Error when calling google with call_params


ionflow:
### Description

Getting this error when calling llm.override to google with call_params value:
```
  File "/Users/ionflow/Sites/hui/backend/.venv/lib/python3.13/site-packages/mirascope/core/base/_extract.py", line 135, in inner_async
    call_response = await create_decorator(
                    ^^^^^^^^^^^^^^^^^^^^^^^
        fn=fn, **create_decorator_kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )(*args, **kwargs)
    ^^^^^^^^^^^^^^^^^^
  File "/Users/ionflow/Sites/hui/backend/.venv/lib/python3.13/site-packages/mirascope/core/base/_create.py", line 188, in inner_async
    response = await create(stream=False, **call_kwargs)
                     ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ionflow/Sites/hui/backend/.venv/lib/python3.13/site-packages/mirascope/core/base/_utils/_get_create_fn_or_async_create_fn.py", line 78, in create_or_stream
    return cast(Awaitable[_NonStreamedResponse], async_func(**kwargs))
                                                 ~~~~~~~~~~^^^^^^^^^^
TypeError: AsyncModels.generate_content() got an unexpected keyword argument 'generation_config'
```

Can be reproduced with this code:
```
from mirascope import llm
from pydantic import BaseModel
from dotenv import load_dotenv
import os
import asyncio

load_dotenv()

class Book(BaseModel):
    title: str
    author: str

os.environ["GOOGLE_API_KEY"] = os.getenv("GOOGLE_API_KEY")

@llm.call(
    provider="anthropic",
    model="claude-3-5-sonnet-latest",
    json_mode=True,
    response_model=Book,
)
async def recommend_book(genre: str) -> str:
    return f"Recommend a {genre} book"

async def main():
    override_response = await llm.override(
        recommend_book,
        provider="google",
        model="gemini-2.0-flash",
        call_params={"temperature": 0.7},
    )("fantasy")

    print(override_response)

if __name__ == "__main__":
    asyncio.run(main())
```

### Python, Mirascope & OS Versions, related packages (not required)

```TOML
[project]
name = "project-name"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
[...]

PR #937, opened 2025-03-28
Fix google call params error 936

ionflow:
# Fix: Improve handling of configuration parameters in Google API calls

Issue: [Error when calling google with call_params #936](#936)

## Description
This PR fixes an issue where the Google API would fail with an `AsyncModels.generate_content() got an unexpected keyword argument 'generation_config'` error when using call_params.

The root problem was in how common parameters were being transferred from the conversion function to the actual API call configuration. By addressing this at the source, we've created a more robust solution that follows the codebase's established patterns.

## Changes
- Implemented a cleaner approach to handling parameters in `_setup_call.py` by properly transferring generation_config parameters to the API's required config object
- Ensured stop sequences are always correctly formatted as arrays in `_convert_common_call_params.py` to match Google API requirements

## Tests
- Fixed the existing test file `tests/core/google/test_call_params.py` to properly type-cast parameters
- Verified API compatibility with careful parameter handling

This code now doesn't fail:
```
from mirascope import llm
from mirascope.core import google
import os
import asyncio

load_dotenv()

class Book(BaseModel):
    title: str
    author: str

os.environ["GOOGLE_API_KEY"] = os.getenv("GOOGLE_API_KEY")

@llm.call(
    provider="google",
    model="gemini-2.0-flash",
    json_mode=True,
    response_model=Book,
)
async def recommend_book(genre: str) -> str:

async def main():
    override_response = await llm.override(
        recommend_book,
        provider="google",
        model="gemini-2.0-flash",
        call_params={"max_tokens": 100, "temperature": 0.7},
    )("fantasy")
    
    print(override_response)

if __name__ == "__main__":
    asyncio.run(main())
```


mirascope/core/google/_utils/_setup_call.py
```diff
@@ -131,17 +131,34 @@ def setup_call(
     call_kwargs = cast(GoogleCallKwargs, base_call_kwargs)
     messages = cast(list[BaseMessageParam | ContentDict], messages)
 
+    # Handle CommonCallParams that were converted to generation_config
```
willbakst, line 134: hmm, if there's a problem with this, we should fix it in the conversion so this gets the correct thing, right?

here: https://github.com/Mirascope/mirascope/blob/7591658aba56fbd518698aa83356613db59ec1d7/mirascope/core/google/_utils/_convert_common_call_params.py#L18

mirascope/core/google/_utils/_setup_call.py
```diff
@@ -131,17 +131,28 @@ def setup_call(
...
```
willbakst, line 134: do we still need this given the other fix?

ionflow: The `convert_common_call_params` function returns `generation_config` which makes sense to me since google api is expecting a config instead of top level args. Would you rather we keep them top level and then merge them into `config` later?

willbakst: ideally the `conver_common_call_params` method would convert the common call params into the exact object needed here in setup call?

willbakst (dismissed):

willbakst (approved):

PR #941, opened 2025-04-15
fix: Tool usage with llm.call

### LINKED ISSUE #940: Tools not working with `llm.call`
### Description

Reproduce Steps:

1. Run this code:

```python
from mirascope import BaseMessageParam, llm, prompt_template
from mirascope.tools import DuckDuckGoSearch, ParseURLContent


@llm.call(
    "google",
    "gemini-2.5-pro-exp-03-25",
    tools=[DuckDuckGoSearch, ParseURLContent],
)
@prompt_template(
    """
    SYSTEM:
    You are an agent that can search the web.

    MESSAGES: {messages}
    USER: {query}
    """
)
def web_agent(query: str | None, messages: list[BaseMessageParam]) -> None: ...


if __name__ == "__main__":
    response = web_agent("What is Mirascope?", [])
    messages = [response.user_message_param, response.message_param]
    while tools := response.tools:
        print("[CALLING TOOLS]")
        tools_and_outputs = [(tool, tool.call()) for tool in tools]
        messages += response.tool_message_params(tools_and_outputs)
        response = web_agent(None, messages)
    print("[FINAL RESPONSE]")
    print(response.content)
```

2. View error output:

```plaintext
pydantic_core._pydantic_core.ValidationError: 2 validation errors for CallResponse
tool_types.0
  Input should be a subclass of Tool [type=is_subclass_of, input_value=<class 'mirascope.core.ba...on_to_base_tool.search'>, input_type=ModelMetaclass]
    For further information visit https://errors.pydantic.dev/2.11/v/is_subclass_of
tool_types.1
  Input should be a subclass of Tool [type=is_subclass_of, input_value=<class 'mirascope.core.ba...ion_to_base_tool.parse'>, input_type=ModelMetaclass]
    For further information visit https://errors.pydantic.dev/2.11/v/is_subclass_of
```

```TOML
mirascope==1.22.2
```

teamdandelion:
This updates the llm.CallResponse class so it accepts BaseCallResponses with BaseTools. Since the tool() and tools() methods on llm.CallResponse are implemented in terms of BaseCallResponse.common_tools() (which converts to llm.Tool), no other changes are needed to maintain llm.CallResponse's API contract.

Added tests, which fail prior to this change.

Fixes #940

### DIFF
--- a/mirascope/llm/call_response.py
+++ b/mirascope/llm/call_response.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from functools import cached_property
 from typing import Any
 
@@ -133,12 +134,12 @@ class CallResponse(
     @classmethod
     @transform_tool_outputs
     def tool_message_params(
-        cls, tools_and_outputs: list[tuple[Tool, str]]
+        cls, tools_and_outputs: Sequence[tuple[BaseTool, str]]
     ) -> list[BaseMessageParam]:
         """Returns the tool message parameters for tool call results.
 
         Args:
-            tools_and_outputs: The list of tools and their outputs from which the tool
+            tools_and_outputs: The sequence of tools and their outputs from which the tool
                 message parameters should be constructed.
         """
 
--- a/mirascope/llm/call_response.py
+++ b/mirascope/llm/call_response.py
@@ -26,7 +26,7 @@ from .tool import Tool
 class CallResponse(
     BaseCallResponse[
         Any,
-        Tool,
+        BaseTool,
         Any,
         BaseDynamicConfig[Any, Any, Any],
         BaseMessageParam,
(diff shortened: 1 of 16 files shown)


mirascope/llm/call_response.py
```diff
@@ -132,7 +132,7 @@ def tool(self) -> Tool | None:  # pyright: ignore [reportIncompatibleVariableOve
 
     @classmethod
     @transform_tool_outputs
-    def tool_message_params(
+    def tool_message_params(  # pyright: ignore [reportIncompatibleMethodOverride]
```
teamdandelion, line 135: Note: There's an alternative to this `pyright: ignore` which is to refactor CallResponse to take a sequence rather than a list; since sequence is covariant, that is a more "principled" fix. Adding this override seems idiomatic to the codebase so I went with the simpler approach. However, let me know if you prefer the other (I have the commit)

willbakst: I think better not to ignore and use `Sequence` since that shouldn't break anything

willbakst (approved):

PR #943, opened 2025-04-16
Fix `llm.call` tool looping with anthropic

willbakst:


teamdandelion (approved):
approved, but please regenerate the uv lockfile before merging

PR #977, opened 2025-05-22
Fix: tool calls with bedrock

### LINKED ISSUE #972: Bedrock tool calls fail.
### Description

I am using the exact example from the [documentation](https://mirascope.com/docs/mirascope/learn/agents#integrating-tools).

I am using `bedrock` as the backend.

The example always fails when a tool call is made with the following error:

```
line 59, in <module>
    Librarian().run()
line 55, in run
    step_output = self._step(query)
                  ^^^^^^^^^^^^^^^^^
 line 45, in _step
    return self._step("")
           ^^^^^^^^^^^^^^
line 37, in _step
    response = self._call(query)
               ^^^^^^^^^^^^^^^^^
 line 346, in inner
    result = decorated(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
mirascope\core\base\_create.py", line 232, in inner
    response = create(stream=False, **call_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
mirascope\core\base\_utils\_get_create_fn_or_async_create_fn.py", line 135, in create_or_stream
    return cast(_NonStreamedResponse, sync_func(**kwargs))
                                      ^^^^^^^^^^^^^^^^^^^
botocore\client.py", line 569, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
botocore\client.py", line 980, in _make_api_call
    request_dict = self._convert_to_request_dict(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
botocore\client.py", line 1047, in _convert_to_request_dict
    request_dict = self._serializer.serialize_to_request(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
botocore\validate.py", line 381, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter messages[2].content[0].toolResult.toolUseId, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```

I have tested with with `sonnet-3.5`, `haiku` and `sonnet-3.7` with the same result. Have not tested with any other models from amazon.

```TOML
`mirascope[bedrock]==1.24.0`
```

alexanderlazarev0:
Resolves #972.


mirascope/llm/call_response.py
```diff
@@ -159,7 +150,7 @@ def _get_tool_call_id(_tool: BaseTool) -> str | None:
                         type="tool_result",
                         name=tool._name(),
                         content=output,
-                        id=_get_tool_call_id(tool),
+                        id=getattr(tool, "id", None),
```
willbakst, line 153: why not just use `tool.id` here since it's now a guaranteed property?

mirascope/llm/tool.py
```diff
@@ -36,6 +36,10 @@ class Tool(BaseTool, metaclass=_DelegateAbstractMethodsForTool):
 
     _tool: BaseTool
 
+    @property
+    def id(self) -> str | None:
```
willbakst, line 40: i don't believe this is necessary and can be safely removed (since we forward properties / methods to the original tool by default)

mirascope/llm/tool.py
```diff
@@ -52,6 +56,7 @@ def __getattribute__(self, name: str) -> Any:  # noqa: ANN401
             "__pydantic_extra__",
             "__pydantic_private__",
             "__class_getitem__",
+            "id",
```
willbakst, line 59: similarly here, we shouldn't have to remove this since it will just use the original tool's id property, right?

tests/core/bedrock/_utils/test_message_param_converter.py
```diff
@@ -201,16 +199,16 @@ def test_tool_use_with_text():
     }
     results = BedrockMessageParamConverter.from_provider([message_param])  # pyright: ignore [reportArgumentType]
     assert results == [
-        BaseMessageParam(role="user", content=[TextPart(type="text", text="Hello")]),
         BaseMessageParam(
```
willbakst, line 202: not sure if this is an issue with the code or the way the test is written, but we should structure things such that the tool call part is part of an assistant message (so that we can confirm things are working as expected).

tests/llm/test_call_response.py
```diff
@@ -189,6 +189,10 @@ class ToolCallWithID:
     class ToolWithID(Tool):
         tool_call: ClassVar[Any] = ToolCallWithID()
 
+        @property
+        def id(self) -> str | None:
```
willbakst, line 193: is this necessary? see other comment

mirascope/llm/tool.py
alexanderlazarev0: I did it this way because otherwise you get the following behaviour:
```python
class ToolCallWithID:
    id = "tool_call_with_id"

class ToolWithID(Tool):
    tool_call: ClassVar[Any] = ToolCallWithID()

    def call(self): ...
    @property
    def model_fields(self): ...  # pyright: ignore [reportIncompatibleVariableOverride]

    field1: str = "tool_field"
    
class MockTool(BaseTool):
    def call(self): ...

tool_with_id = ToolWithID(MockTool)
print(tool_with_id.id) # == None
```
Because instead of getting `tool_with_id.tool_call.id` it will get `tool_with_id._tool.tool_call.id`.

tests/llm/test_call_response.py
alexanderlazarev0: WIll remove.

mirascope/llm/tool.py
willbakst: My hope here is that we would get `tool_with_id._tool.id` which would hit the Bedrock specific `tool.id` property you've overridden to do the right thing

willbakst: In the test the issue then is that `MockTool` is not properly mocking the tool we actually expect to receive, which would be a provider-specific tool, right?

alexanderlazarev0:
I'll do it )

willbakst (approved):

PR #983, opened 2025-05-30
Implement native JSON mode with schema for Gemini

### LINKED ISSUE #472: Add support for Gemini Structured Outputs
### Description

Recently Google released Structured Outputs for Gemini (much like OpenAI).

Should be fairly simple to enable this by updating `setup_call` for `gemini` and updating the warning to include `gemini` as a module that supports `strict` in the response model schema (although it seems like structured outputs for tools is not quite supported).

See: https://ai.google.dev/gemini-api/docs/json-mode?lang=python#supply-schema-in-config

Only difficulty I can see right now is that the only two supported inputs are either raw type annotations (so no `BaseModel`) or their `genai.protos.Schema` type (which will require conversion from `BaseModel`). Should have most of this implemented already I believe, so still shouldn't be terribly difficult.

ipriyaaanshu:
## Implement native JSON mode with schema for Gemini

Closes #472 

This pull request enhances the Google Gemini integration within Mirascope to support native, schema-constrained JSON outputs, aligning with the latest capabilities of the Gemini API.

**Key Changes & Motivations:**

1.  **Native Gemini JSON Mode (`response_schema`)**:
    *   When using `@llm.call(provider="google", ...)` (or the older `@google_call`) with `json_mode=True` and a Pydantic `response_model`, Mirascope now directly passes the Pydantic model to the `generation_config.response_schema` parameter of the Google Gemini API.
    *   This leverages Gemini's built-in mechanism for enforcing that the LLM's output strictly adheres to the provided schema, offering a more direct and potentially more reliable method for structured JSON generation compared to relying solely on prompting or tool calling as a fallback for this specific mode.
    *   If `json_mode=True` is used without a `response_model` (and no tools are specified), a generic prompt instructing the model to output JSON is still appended to the messages, maintaining existing behavior for that scenario.

2.  **Updated Warning for Strict Outputs**:
    *   The warning message in `BaseTool` (triggered when `model_config={"strict": True}` is used on a `response_model`) has been updated to include `"google"` in its list of providers that support such strict structured outputs when `json_mode=True`. This reflects the new capability.

3.  **Unit Tests**:
    *   New unit tests have been added to `tests/core/google/_utils/test_setup_call.py` to specifically verify the new logic:
        *   Ensures `response_schema` is correctly set with the Pydantic model.
        *   Ensures `response_mime_type` is set to `"application/json"`.
        *   Confirms that the generic JSON prompting mechanism (via `_utils.json_mode_content`) is *not* invoked when `response_schema` is active.
    *   The new tests align with the existing testing structure and conventions in the file.

4.  **Minor Docstring Fix**:
    *   Corrected a minor typo from `json_modem` to `json_mode` in the docstring of `google_call` in `mirascope/core/google/_call.py`.

**How to Test:**

Reviewers can test this by:
1.  Checking out this branch.
2.  Ensuring Mirascope is installed in editable mode with the `google` extra (`uv pip install -e ".[google]"`).
3.  Setting the `GOOGLE_API_KEY` environment variable.
4.  Running a test code that uses `@llm.call(provider="google", response_model=MyModel, json_mode=True)`.
5.  Observing that the Gemini API call successfully returns structured data conforming to `MyModel`.
6.  Running `pytest` to ensure the new and existing unit tests pass, particularly `tests/core/google/_utils/test_setup_call.py`.

This change aims to provide a more robust and idiomatic integration with Gemini's evolving features for structured data extraction.

### DIFF
--- a/mirascope/core/google/_utils/_setup_call.py
+++ b/mirascope/core/google/_utils/_setup_call.py
@@ -147,11 +147,16 @@ def setup_call(
 
     if json_mode:
         with _generate_content_config_context(call_kwargs) as config:
-            if not tools:
-                config.response_mime_type = "application/json"
-        messages[-1]["parts"].append(  # pyright: ignore [reportTypedDictNotRequiredAccess, reportOptionalMemberAccess, reportArgumentType]
-            PartDict(text=_utils.json_mode_content(response_model))
-        )  # pyright: ignore [reportTypedDictNotRequiredAccess, reportOptionalMemberAccess, reportArgumentType]
+            config.response_mime_type = "application/json"
+            if response_model:
+                config.response_schema = response_model
+
+            elif not tools:
+                messages[-1][
+                    "parts"
+                ].append(  # pyright: ignore [reportTypedDictNotRequiredAccess, reportOptionalMemberAccess, reportArgumentType]
+                    PartDict(text=_utils.json_mode_content(None))
+                )  # pyright: ignore [reportTypedDictNotRequiredAccess, reportOptionalMemberAccess, reportArgumentType]
     elif response_model:
         assert tool_types, "At least one tool must be provided for extraction."
         with _generate_content_config_context(call_kwargs) as config:
--- a/mirascope/core/base/tool.py
+++ b/mirascope/core/base/tool.py
@@ -205,7 +205,11 @@ class BaseTool(BaseModel, ABC):
                 UserWarning,
             )
 
-        if "strict" in cls.model_config and cls.__provider__ not in ["openai", "azure"]:
+        if "strict" in cls.model_config and cls.__provider__ not in [
+            "openai",
+            "azure",
+            "google",
+        ]:
             warnings.warn(
                 f"{cls.__provider__} does not support strict structured outputs, but "
                 "you have configured `strict=True` in your `ResponseModelConfigDict`. "
--- a/mirascope/core/google/_call.py
+++ b/mirascope/core/google/_call.py
@@ -56,7 +56,7 @@ Args:
     output_parser (Callable[[GoogleCallResponse | ResponseModelT], Any]): A function
         for parsing the call response whose value will be returned in place of the
         original call response.
-    json_modem (bool): Whether to use JSON Mode.
+    json_mode (bool): Whether to use JSON Mode.
     client (object): An optional custom client to use in place of the default client.
     call_params (GoogleCallParams): The `GoogleCallParams` call parameters to use in the
         API call.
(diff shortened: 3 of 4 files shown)


mirascope/core/google/_utils/_setup_call.py
```diff
@@ -147,11 +147,16 @@ def setup_call(
 
     if json_mode:
```
willbakst, line 148: shouldn't we also for the strict mode setup if strict is True?

willbakst: In the v1 interface, if strict is not set, we should use either JSON mode as before or tools as before.

The update would be to detect that we're using strict and force the structured output through this new path.

ipriyaaanshu: Ah, so something like this?
`if json_mode or (response_model and getattr(response_model, "model_config", {}).get("strict", False)):`

ipriyaaanshu: Cool! I'll push the changes.

ipriyaaanshu:
@willbakst can you please confirm if my recent commits look good?

mirascope/core/google/_utils/_setup_call.py
```diff
@@ -145,13 +145,17 @@ def setup_call(
                     Part.model_validate(part) for part in message_parts
                 ]
 
-    if json_mode:
+    if json_mode or (response_model and getattr(response_model, "model_config", {}).get("strict", False)):
```
willbakst, line 148: the null check on `response_model` should guarantee the presence of `model_config`, so you can do something like this:

```python
if json_mode or (response_model and response_model.model_config.get("strict", False)): ...
```

mirascope/core/google/_utils/_setup_call.py
```diff
@@ -145,13 +145,17 @@ def setup_call(
...
         with _generate_content_config_context(call_kwargs) as config:
-            if not tools:
-                config.response_mime_type = "application/json"
-        messages[-1]["parts"].append(  # pyright: ignore [reportTypedDictNotRequiredAccess, reportOptionalMemberAccess, reportArgumentType]
-            PartDict(text=_utils.json_mode_content(response_model))
-        )  # pyright: ignore [reportTypedDictNotRequiredAccess, reportOptionalMemberAccess, reportArgumentType]
+            config.response_mime_type = "application/json"
```
willbakst, line 150: this was previously inside of the `if not tools: ...` check because it's possible to set both `response_model` and `tools`. We need to make sure that if both are set we don't force the response into `application/json` since the model may want to respond with a tool call.

mirascope/core/google/_utils/_setup_call.py
```diff
@@ -145,13 +145,17 @@ def setup_call(
...
+            if response_model:
+                config.response_schema = response_model
+            elif not tools:
```
willbakst, line 153: why move this inside of this check for `not tools`? Previously we've always included this json mode content message.

willbakst:
Opting for the `v2` implementation and closing this

Issue #872, opened 2025-02-24
Formally support Anthropic Claude 3.7 Thinking (without type errors)


willbakst:
### Description

Newest Claude 3.7 Sonnet release adds support for reasoning through a thinking config. While users can still use this through `call_params`, it will result in a type error. We should add this field directly to avoid this.

willbakst:
We should also add support for the new [ThinkingBlock](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/thinking_block.py)

willbakst:
Thinking on this more, we should probably implement support for this in a way that will later support other providers that enable reasoning. Something provider-agnostic like we do for tools etc. would be best. This should include support for streaming as well.

teamdandelion:
FYI, Claude 3.7 thinking is [not compatible with forced tool use](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking):

<img width="785" alt="Image" src="https://github.com/user-attachments/assets/26d250ff-52df-4715-9ac8-f0f429224087" />

This means that when using thinking, it will break our standard implementation for response_model. Which means likely we'd need some logic like "if thinking is enabled, then use json_mode to get response model" (like with response model + tools). For me this is another nudge to move response model away from the forced tool approach across the board.

willbakst:
Oh this is a great catch.

We may want to push off our implementation of thinking as part of Mirascope 2.0 (or at least mark early interfaces experimental)

willbakst:
This is now released in `v1.25.0`!

PR #984, opened 2025-06-02
feat: add Anthropic support for thinking

teamdandelion:
This PR adds support for [Anthropic extended thinking](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking).

- Pins Anthropic dependency to at least [0.47.0](https://github.com/anthropics/anthropic-sdk-python/releases/tag/v0.47.0), at which point thinking apis were added. Updates tests to reflect the presence of the Citations property
- Updates `AnthropicCallParams` to optionally include the ThinkingConfigParam
- Changes `AnthropicCallResponse` so the `content` property contains the text content of the first *text* block, not the first block overall. This is necessary for intuitive and correct behavior when using thinking, as the first block will be the thinking block, not a text block. This is also better aligned with the stated docstring. All tests continued to pass. I view this as essentially a bugfix, and unlikely to break any user code.
- Updates `AnthropicCallResponse` to expose the thinking and signature field. 
- Updates `AnthropicCallResponseChunk` to expose thinking and signatures on the appropriate delta chunks, and to track thinking and signature when reconstructing the `AnthropicCallResponse`
- Update tests
- Gives Claude Code some additional permissions (checking Anthropic docs, using `ls` and `rg`).

In addition to the unit tests, I used the following script to test thinking support using the real anthropic API.

Note that currently thinking with response models is not supported (cf https://github.com/Mirascope/mirascope/issues/872#issuecomment-2746764449). The script below demonstrates that behavior (we get a response code 400 error from the Anthropic API for the response model section).

Note this adds thinking support as a mostly undocumented feature. 

Addresses #872. 

```py
"""
Demo of Anthropic thinking support in Mirascope
"""

import os

from enum import Enum

from mirascope.core.anthropic import AnthropicCallParams, AnthropicTool, anthropic_call

# Load environment variables
load_dotenv()


# Example 1: Regular thinking
@anthropic_call(
    "claude-3-7-sonnet-latest",
    call_params=AnthropicCallParams(
        max_tokens=2048, thinking={"type": "enabled", "budget_tokens": 1024}
    ),
)
def solve_math_problem(problem: str):
    return f"Solve this step by step: {problem}"


# Example 2: Streaming thinking
@anthropic_call(
    "claude-3-7-sonnet-latest",
    call_params=AnthropicCallParams(
        max_tokens=2048, thinking={"type": "enabled", "budget_tokens": 1024}
    ),
    stream=True,
)
def stream_thinking_response(question: str):
    return f"Think through this carefully: {question}"


# Example 3: Thinking with tools
class MathOperation(str, Enum):
    """Supported mathematical operations."""
    ADD = "add"
    MULTIPLY = "multiply"
    DIVIDE = "divide"


class Calculator(AnthropicTool):
    """A simple calculator tool that performs basic mathematical operations.
    
    Use this tool to calculate mathematical expressions. Supported operations:
    - add: Addition of two numbers
    - multiply: Multiplication of two numbers  
    - divide: Division of two numbers
    """

    operation: MathOperation
    a: float
    b: float

    def call(self) -> str:
        """Perform the calculation."""
        if self.operation == MathOperation.ADD:
            return f"{self.a} + {self.b} = {self.a + self.b}"
        elif self.operation == MathOperation.MULTIPLY:
            return f"{self.a} * {self.b} = {self.a * self.b}"
        elif self.operation == MathOperation.DIVIDE:
            if self.b == 0:
                return "Error: Division by zero"
            return f"{self.a} / {self.b} = {self.a / self.b}"
        else:
            return f"Unknown operation: {self.operation}"


@anthropic_call(
    "claude-3-7-sonnet-latest",
    call_params=AnthropicCallParams(
        max_tokens=2048, thinking={"type": "enabled", "budget_tokens": 1024}
    ),
    tools=[Calculator],
)
def thinking_with_tools(problem: str):
    return f"Use the calculator tool to help solve: {problem}"


# Example 4: Thinking with response model
class Book(BaseModel):
    """An extracted book with title and author."""
    
    title: str
    author: str


@anthropic_call(
    "claude-3-7-sonnet-latest",
    call_params=AnthropicCallParams(
        max_tokens=2048, thinking={"type": "enabled", "budget_tokens": 1024}
    ),
    response_model=Book,
)
def extract_book_with_thinking(text: str):
    return f"Extract the book title and author from this text: {text}"


def demo_regular_thinking():
    """Demo regular thinking API usage."""
    print("=== Regular Thinking Demo ===")

    response = solve_math_problem("What is 234 * 567?")

    print(f"Thinking: {response.thinking}")
    print(f"Signature: {response.signature}")
    print(f"Content: {response.content}")
    print()


def demo_streaming_thinking():
    """Demo streaming thinking API usage."""
    print("=== Streaming Thinking Demo ===")

    stream = stream_thinking_response("What are the pros and cons of renewable energy?")

    print("Thinking:")

    for chunk, _ in stream:
        if chunk.thinking:
            print(chunk.thinking, end="", flush=True)

        if chunk.signature:
            print(f"\n[SIGNATURE] {chunk.signature}\n Content:", end="\n", flush=True)

        if chunk.content:
            print(chunk.content, end="", flush=True)

    print("\n")

    call_response = stream.construct_call_response()
    print(f"Final thinking: {call_response.thinking}")
    print(f"Final signature: {call_response.signature}")
    print(f"Final content: {call_response.content}")
    print()


def demo_thinking_with_tools():
    """Demo thinking with tool usage."""
    print("=== Thinking with Tools Demo ===")

    response = thinking_with_tools(
        "I need to calculate 123.45 * 67.89 and then add 100 to the result"
    )

    print(f"Thinking: {response.thinking}")
    print(f"Content: {response.content}")

    if response.tools:
        print(f"Tools used: {len(response.tools)}")
        for i, tool in enumerate(response.tools):
            result = tool.call()
            print(f"Tool {i + 1}: {result}")

    print()


def demo_thinking_with_response_model():
    """Demo thinking with response model."""
    print("=== Thinking with Response Model Demo ===")
    
    book = extract_book_with_thinking(
        "I just finished reading 'The Hobbit' written by J.R.R. Tolkien, and it was amazing!"
    )
    
    # Access the structured response
    print(f"Title: {book.title}")
    print(f"Author: {book.author}")
    
    # Access the thinking from the original call response
    call_response = book._call_response  # Access to original response
    print(f"Thinking: {call_response.thinking}")
    
    print()


def main():
    """Run all thinking demos."""
    if not os.getenv("ANTHROPIC_API_KEY"):
        print("ERROR: ANTHROPIC_API_KEY environment variable not set")
        return

    try:
        demo_regular_thinking()
        demo_streaming_thinking()
        demo_thinking_with_tools()
        demo_thinking_with_response_model()
    except Exception as e:
        print(f"ERROR: {e}")
        import traceback

        traceback.print_exc()


if __name__ == "__main__":
    main()
```

### DIFF
--- a/mirascope/core/anthropic/_thinking.py
+++ b/mirascope/core/anthropic/_thinking.py
@@ -1,0 +1,70 @@
+"""Centralized thinking-related imports and compatibility handling for Anthropic."""
+
+from typing import TYPE_CHECKING, Literal
+
+from pydantic import BaseModel
+from typing_extensions import TypedDict
+
+
+# Always define the stubs with underscore names
+class _ThinkingConfigParam(TypedDict):
+    type: Literal["enabled"]
+    budget_tokens: int
+
+
+class _ThinkingBlock(BaseModel):
+    signature: str
+    thinking: str
+    type: Literal["thinking"]
+
+
+class _ThinkingDelta(BaseModel):
+    thinking: str
+    type: Literal["thinking_delta"]
+
+
+class _SignatureDelta(BaseModel):
+    signature: str
+    type: Literal["signature_delta"]
+
+
+HAS_THINKING_SUPPORT = True
+
+# Define the public names based on what's available
+if TYPE_CHECKING:
+    # For static analysis, always use our stubs so types are consistent
+    ThinkingConfigParam = _ThinkingConfigParam
+    ThinkingBlock = _ThinkingBlock
+    ThinkingDelta = _ThinkingDelta
+    SignatureDelta = _SignatureDelta
+else:
+    # At runtime, use real types if available, otherwise stubs
+    try:
+        from anthropic.types import (  # pyright: ignore [reportAttributeAccessIssue]
+            ThinkingBlock,
+            ThinkingDelta,
+        )
+        from anthropic.types.signature_delta import (
+            SignatureDelta,  # pyright: ignore [reportMissingImports]
+        )
+        from anthropic.types.thinking_config_param import (
+            ThinkingConfigParam,  # pyright: ignore [reportMissingImports]
+        )
+
+        HAS_THINKING_SUPPORT = True
+
+    except ImportError:  # pragma: no cover
+        ThinkingConfigParam = _ThinkingConfigParam
+        ThinkingBlock = _ThinkingBlock
+        ThinkingDelta = _ThinkingDelta
+        SignatureDelta = _SignatureDelta
+        HAS_THINKING_SUPPORT = False
+
+
+__all__ = [
+    "HAS_THINKING_SUPPORT",
+    "SignatureDelta",
+    "ThinkingBlock",
+    "ThinkingConfigParam",
+    "ThinkingDelta",
+]
--- a/mirascope/core/anthropic/call_params.py
+++ b/mirascope/core/anthropic/call_params.py
@@ -7,11 +7,11 @@ from collections.abc import Iterable
 from anthropic.types import TextBlockParam
 from anthropic.types.completion_create_params import Metadata
 from anthropic.types.message_create_params import ToolChoice
-from anthropic.types.thinking_config_param import ThinkingConfigParam
 from httpx import Timeout
 from typing_extensions import NotRequired
 
 from ..base import BaseCallParams
+from ._thinking import ThinkingConfigParam
 
 
 class AnthropicCallParams(BaseCallParams):
--- a/mirascope/core/anthropic/stream.py
+++ b/mirascope/core/anthropic/stream.py
@@ -10,7 +10,6 @@ from anthropic.types import (
     Message,
     MessageParam,
     TextBlock,
-    ThinkingBlock,
     ToolParam,
     ToolUseBlock,
     Usage,
@@ -24,6 +23,7 @@ from ..base.call_kwargs import BaseCallKwargs
 from ..base.metadata import Metadata
 from ..base.stream import BaseStream
 from ..base.types import CostMetadata
+from ._thinking import ThinkingBlock
 from .call_params import AnthropicCallParams
 from .call_response import AnthropicCallResponse
 from .call_response_chunk import AnthropicCallResponseChunk
@@ -152,7 +152,7 @@ class AnthropicStream(
         # Add thinking block first if we have thinking content
         if hasattr(self, "thinking") and self.thinking:
             content_blocks.append(
-                ThinkingBlock(
+                ThinkingBlock(  # pyright: ignore [reportArgumentType]
                     type="thinking",
                     thinking=self.thinking,
                     signature=getattr(self, "signature", ""),
--- a/mirascope/core/anthropic/_utils/_setup_call.py
+++ b/mirascope/core/anthropic/_utils/_setup_call.py
@@ -19,6 +19,7 @@ from ...base import BaseMessageParam, BaseTool, _utils
 from ...base._utils import AsyncCreateFn, CreateFn
 from ...base.stream_config import StreamConfig
 from .._call_kwargs import AnthropicCallKwargs
+from .._thinking import HAS_THINKING_SUPPORT
 from ..call_params import AnthropicCallParams
 from ..dynamic_config import AnthropicDynamicConfig, AsyncAnthropicDynamicConfig
 from ..tool import AnthropicTool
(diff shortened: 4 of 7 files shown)


willbakst:
A few high level notes before I dig deeper into the implementation:

> Pins Anthropic dependency to at least [0.47.0](https://github.com/anthropics/anthropic-sdk-python/releases/tag/v0.47.0), at which point thinking apis were added. Updates tests to reflect the presence of the Citations property

Generally we try to maintain the lowest version possible so that users can upgrade without having to upgrade their version of the SDK. We do something similar for OpenAI with Audio where we allow users to have versions without audio and only fail in the event that they try to use those features on the wrong version. For `v2` we can definitely bump the project versions to the minimum versions for each provider that support the `v2` feature set.

> Note that currently thinking with response models is not supported (cf https://github.com/Mirascope/mirascope/issues/872#issuecomment-2746764449).

Technically it's forced tool use that doesn't work with thinking, not response models. If we detect that thinking is set during setup for Anthropic, we should be able to get a (slightly less reliable) response model using tools by skipping the forced tool use. Response Models with thinking should also work with `json_mode=True`. We could also log a warning or something indicating what's happening under the hood, and we could also potentially update the prompt to indicate we want forced tool use without the explicit setting (although I'm always in favor of leaving the prompt alone if possible).

> Note this adds thinking support as a mostly undocumented feature.

Should we update the documentation to mention we support thinking with a table for the providers we currently support? We would likely want to indicate which providers support thinking + which ones we support. Could also just wait to update the documentation until we add support for all providers that currently support thinking. WDYT?

.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
+  "permissions": {
+    "allow": [
+      "WebFetch(domain:docs.anthropic.com)",
```
willbakst, line 4: should we just add all of the providers we support?

mirascope/core/anthropic/call_params.py
```diff
@@ -7,6 +7,7 @@
 from anthropic.types import TextBlockParam
 from anthropic.types.completion_create_params import Metadata
 from anthropic.types.message_create_params import ToolChoice
+from anthropic.types.thinking_config_param import ThinkingConfigParam
```
willbakst: see top-level comment + OpenAI audio implementation for reference of how to do this without forcing a version bump

willbakst (commented):
Noting that this only supports using `anthropic.call` since there is no `llm` module message conversion.

Happy to leave that out of this PR, work to add thinking support for OpenAI and Google etc., and then add `llm` module support once we have thinking support for all providers we support that currently support thinking. WDYT?

.claude/settings.local.json
teamdandelion: It'd be good to have, but I see that as out of scope for this PR. For now I'm happy to just add permissions to sites when I direct Claude to fetch them.

mirascope/core/anthropic/call_params.py
teamdandelion: Done, in https://github.com/Mirascope/mirascope/pull/984/commits/d835f3041e55d95ebe4906212e7f06eedc0b41b9

mirascope/core/anthropic/stream.py
```diff
@@ -102,6 +149,16 @@ def construct_call_response(self) -> AnthropicCallResponse:
 
         content_blocks: list[ContentBlock] = []
 
+        # Add thinking block first if we have thinking content
+        if hasattr(self, "thinking") and self.thinking:
+            content_blocks.append(
+                ThinkingBlock(  # pyright: ignore [reportArgumentType]
```
willbakst, line 155: is this ignore because we're using a stub rather than the true type?

seems dangerous to ignore because of that since we'll now lose proper checking on the actual type

mirascope/core/anthropic/_thinking.py
willbakst, line 1: note: we should definitely think through our policy here when we work on v2. my thinking is that starting in v2 and going forward if we add a feature in a new release we should fix the min version to the version of the underlying sdk that has said feature so we can avoid this super annoying back support stuff. I don't really see an issue with saying "if you want to upgrade mirascope you'll also have to upgrade package x"

mirascope/core/anthropic/stream.py
teamdandelion: The ignore is so that it will still typecheck with anthropic <= 0.47 - since prior to 0.47 ThinkingBlock is not a valid ContentBlock.

Here's ContentBlock (in Anthropic 0.52)
```py
ContentBlock: TypeAlias = Annotated[
    Union[TextBlock, ToolUseBlock, ServerToolUseBlock, WebSearchToolResultBlock, ThinkingBlock, RedactedThinkingBlock],
    PropertyInfo(discriminator="type"),
]
```

mirascope/core/anthropic/_thinking.py
teamdandelion: Agreed. I don't see a lot of benefit for end users in upgrading Mirascope but sticking to an outdated version of the underlying LLM provider.

Issue #957, opened 2025-04-30
Logfire span for llm.call and when errors


CharlesOural:
### Description

When using this code with llm.call :
```python
from dotenv import load_dotenv
import os

dot_env_path = './../../docker/dev/.env.local'
load_dotenv(dotenv_path=dot_env_path)

from mirascope.llm import call
from mirascope.core import prompt_template, google, openai
from mirascope.integrations.logfire import with_logfire
import logfire

logfire.configure()

@with_logfire()
@llm.call(
    provider="openai",
    model="gemini-2.5-flash-preview-04-17",
    stream=True,
)
@prompt_template(
    """
    SYSTEM: {system_message}

    USER: Hello ! What is the author of the book "The Name of the Wind" ?
    """
)
async def test_gemini(
    system_message: str,
):
    pass

stream = test_gemini("You're a helpful assistant.")
async for chunk, _ in await stream:
    print(chunk.content, end="", flush=True)
```

Then span generated won't have any details or data (for both openai and gemini).

Same, if there is any errors when processing this (before fixing the google streaming problem) if the call crashes, there will have no span in the logfire mirascope call. Same if there is a 503 or a server error from the provider the call send won't be shown in the span.

### Python, Mirascope & OS Versions, related packages (not required)

```TOML
latest mirascope
```

willbakst:
Ok I'll take a look when I find some time.

It's tough to keep up with external tools and all of their changes. Main reason why we're dropping integrations in v2 in favor of tools implementing integrations themselves so that the integration properly keeps up with their own tool's changes.

willbakst:
Given we are deprecating integrations and completely removing them in v2, I'm opting to close this issue. I'll leave it open for a short period in case there are complaints with this, and if not I will close after that period.

I will also reach out to the logfire team about taking over the integration if they are interested.

CharlesOural:
Ok I understand, but is it possible in the mean time to use OTEL exporter or something to catch spans ?

willbakst:
You should be able to just use logfire's SDK directly. It won't be pretty in the UI but it should capture the spans/traces

CharlesOural:
do you think it would be a lot of work to create this implementation ? So it would work like the fastapi instrumentalisation from the logfire team but it would be for mirascope ?

I guess you export OTEL span and it's a matter of catching them and properly show the details ?

CharlesOural:
[...]

PR #987, opened 2025-06-04
Fix bug tool uses with langfuse

ThierryNapspirit:
Bug appears with langfuse used with tools, when the LLM uses a tool.

Example :
```
@with_langfuse()
@llm.call(provider="openai",model="gpt-4o-mini", tools=search_toolkit.get_tools())
def agent_software_description_tools(query: str, history: List[BaseMessageParam | str], variables: dict):
   return "...prompt..."
```
Error : "AttributeError: 'BaseMessageParam' object has no attribute 'get'"

Code copied from "logfire/_utils.py".

### DIFF
--- a/mirascope/integrations/langfuse/_utils.py
+++ b/mirascope/integrations/langfuse/_utils.py
@@ -18,18 +18,54 @@ class ModelUsage(BaseModel):
     unit: str
 
 
+def _get_tool_calls(result: BaseCallResponse) -> list[dict[str, Any]] | None:
+    if tools := result.tools:
+        tool_calls = [
+            {
+                "function": {
+                    "arguments": tool.args,
+                    "name": tool._name(),
+                }
+            }
+            for tool in tools
+        ]
+        return tool_calls
+    return None
+
+
 def _get_call_response_observation(
     result: BaseCallResponse, fn: Callable
 ) -> dict[str, Any]:
     metadata = get_metadata(fn, {})
     tags = metadata.get("tags", [])
+    output = None
+    if content := result.content:
+        output = content
+    if tools := result.tools:
+        tool_calls = [
+            {
+                "function": {
+                    "arguments": tool.args,
+                    "name": tool._name(),
+                }
+            }
+            for tool in tools
+        ]
+        if output is not None:
+            output = {
+                'content': output,
+                'tools' : tool_calls
+            }
+        else:
+            output = tool_calls
+
     return {
         "name": f"{fn.__name__} with {result.model}",
         "input": result.messages,
         "metadata": result.response,
         "tags": tags,
         "model": result.model,
-        "output": result.message_param.get("content", None),
+        "output": output,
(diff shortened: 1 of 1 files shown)


teamdandelion:
Hi @ThierryNapspirit, thank you for the pull request! Great to see a fix from a first time contributor. :)

It looks like your commit breaks some existing tests. Would you be willing to do the following?

1. Fix the failing test
2. Add a test specific to the bug you're fixing — one that fails before your change, but passes after it
3. Expand the example in your PR description to something self-contained and copy+pasteable (with imports etc)

ThierryNapspirit:
The test is currently failing, but the core question is: **should we update the code or the test?**

For Langfuse, do we expect the output field to:

    Contain only the LLM response text from the "content" field (e.g., {"role": "assistant", "content": "test_content"}),
    or
    Include both text content and tool call definitions (similar to how Logfire handles this)?

    If (1) is the intended behavior, I will update the code to prevent result.message_param from being None.
    If (2) is correct, then I’ll need to update the test instead.

Would appreciate clarification before proceeding.

willbakst:
I believe it would be (2) in your description that we would want (and thus a change to the tests)?

It's also worth noting that in `v2` we will be completely removing integrations with external libraries / products due to the maintenance burden (it's fairly difficult to keep up with changes in their products), and we are not currently maintaining these integrations actively (due to the burden + we use our own tool Lilypad, which doesn't have a direct Mirascope integration itself and instead integrates with Mirascope on its own separately like we expect other tools to do going forward).

However, we are more than happy to merge and release fixes from contributors in the meantime (like this PR).

Ultimately I'm hopeful that companies like Langfuse will be willing to take on the burden for their own users who want to use Mirascope. I'll reach out to them now to see what they say.

For reference, you can see my recommendation wrt. Logfire [here](https://github.com/Mirascope/mirascope/issues/957#issuecomment-2908081994) for what to do when we fully drop support should third party companies decide not to take on the burden.

ThierryNapspirit:
I've been really enjoying working with Mirascope — it's a fantastic tool! 
The decorators are both powerful and elegantly designed, which makes building with it a real pleasure.

As I look ahead to version 2, I'm seriously considering moving away from Langfuse and migrating my codebase to Lilypad. 
It's a big decision, especially since I've already built a full agent + team + FSM framework on top of mirascope — so switching would involve quite a few changes.

Alternatively, I might just hold out and see if you release your own FSM framework… 👀

willbakst:
> Alternatively, I might just hold out and see if you release your own FSM framework… 👀

We have something in the works :)

Biggest focus right now for v2 is making sure we support everything in v1 with all breaking change improvements + additional first-class support for things like thinking etc.

Agents will have a decorator, and we'll have a graph module as well. We actually already have an early experimental `FiniteStateMachine` interface.

However, given these are new they will remain experimental until we feel confident in the interface design and implementation.

PR #990, opened 2025-06-10
fix langfuse integration bug?

hyusap:
# our error, cursor's analysis:
## Summary
The issue you encountered was an AttributeError: 'BaseMessageParam' object has no attribute 'get' error occurring in the Mirascope-Langfuse integration. This was happening specifically with the dialectic_call and dialectic_stream functions that use the Anthropic provider.
## Root Cause:
The Langfuse integration code in Mirascope was trying to call .get("content", None) on a BaseMessageParam object
BaseMessageParam is a Pydantic model, not a dictionary, so it doesn't have a .get() method
This is a compatibility issue between Mirascope 1.25.0 and Langfuse 2.60.8

# fix
change the get to a getattr.

### DIFF
--- a/mirascope/integrations/langfuse/_utils.py
+++ b/mirascope/integrations/langfuse/_utils.py
@@ -29,7 +29,7 @@ def _get_call_response_observation(
         "metadata": result.response,
         "tags": tags,
         "model": result.model,
-        "output": getattr(result.message_param, "content", None),
+        "output": result.content,
     }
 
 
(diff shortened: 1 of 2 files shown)


mirascope/integrations/langfuse/_utils.py
```diff
@@ -29,7 +29,7 @@ def _get_call_response_observation(
         "metadata": result.response,
         "tags": tags,
         "model": result.model,
-        "output": result.message_param.get("content", None),
+        "output": getattr(result.message_param, "content", None),
```
willbakst: can you implement a test that fails before this change as passes after it?

hyusap:
updated to proper content, and updated the test_get_call_response_observation test to properly check for the content param rather than the message_param's content param.

willbakst (approved):

PR #991, opened 2025-06-12
Setup documentation site for v2

teamdandelion:
- Adds a docs/API viewer for the site, using `bun run dev` and `bun run build`
- Adds a wrangler.jsonc for deployment to v2.mirascope.com
- Re-organizes the v2 __init__.py files so that every submodule is exported in __all__, and so that symbols referenced in the API docs (like BaseCall or BaseResponse) are also exported
- Adds a very simple code example (with type checking) which is directly injected into a simple mdx document as proof of concept. The code is present in the copy+pastable mdx and the llms text.
- API documentation is re-generated on each invocation of `bun run dev`, and operates by automatically exploring the public api. Implementation is forked out of mirascope/website to mirascope/api2mdx. It also prints to console all "missing api links" (eg for DepsT)
- Currently there is no live reloading of docs content, you need to restart the server.


package.json
willbakst, line 1: why not put this inside of the docs folder and make `"name": "mirascope-docs"` or something?

teamdandelion:
BTW: In this PR I add all the modules to llm/__init__, but I don't take out any symbols. I actually do want to clean up that file and remove some of the aliases, but I expect to do so in a separate PR.

wrangler.jsonc
willbakst, line 1: same thing here re: inside the docs folder

scripts/merge-api-meta.ts
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bun
+
+import fs from "fs";
+import path from "path";
+import type {
+  SectionSpec,
+  DocSpec,
+  FullDocsSpec,
+} from "../node_modules/@mirascope/docs-viewer/src/lib/content/spec";
```
willbakst: this feels super weird having `"../node_modules/@mirascope/..."` rather than just `@mirascope/...` can you explain why this is the case?

pyproject.toml
```diff
@@ -203,6 +204,7 @@ exclude = [
     "examples/learn/tools/basic_usage/official_sdk/bedrock_sdk.py",
     "examples/learn/response_models/basic_usage/official_sdk/bedrock_sdk.py",
     ".venv",
+    "node_modules",
```
willbakst: this feels so wrong lol

pyproject.toml
```diff
@@ -123,6 +123,7 @@ dev-dependencies = [
     "notebook>=7.2.2",
     "nbconvert>=7.16.4",
     "docker>=7.1.0",
+    "api2mdx>=0.2.2",
```
willbakst, line 130: I can't find this package, not sure what the source is

package.json
```diff
@@ -0,0 +1,19 @@
...
+    "docs:clean": "rm -rf .docs-content dist",
+    "docs:copy": "mkdir -p .docs-content && cp -r mirascope/experimental/v2/docs .docs-content/ && cp -r mirascope/experimental/v2/examples .docs-content/",
+    "docs:api": "PYTHONPATH=. uv run api2mdx --source-path . --package mirascope.experimental.v2.llm --output .docs-content/docs/api",
+    "docs:merge-api": "bun scripts/merge-api-meta.ts",
+    "docs:serve": "bun node_modules/@mirascope/docs-viewer/scripts/start.ts --content-dir .docs-content --working-dir dist",
+    "docs:prepare": "bun docs:clean && bun docs:copy && bun docs:api && bun docs:merge-api",
+    "docs:build": "bun node_modules/@mirascope/docs-viewer/scripts/build.ts --content-dir .docs-content --output-dir dist"
+  },
+  "devDependencies": {
+    "@mirascope/docs-viewer": "0.4.9",
```
willbakst: why do we have a docs-viewer package rather than just using `Mirascope/ui`?

bun.lock
willbakst, line 1: ideally this would also be colocated inside of the docs fold with the other related elements

mirascope/experimental/v2/examples/basic_call.py
```diff
@@ -0,0 +1,10 @@
+from mirascope.experimental.v2 import llm
+
+
+@llm.call("anthropic:claude-3-5-sonnet-latest")
+def answer_question(question: str) -> str:
+    return f"Answer this question: ${question}"
```
willbakst, line 6: lol this isn't valid python f-string (it's typescript, drop the `$`)

mirascope/experimental/v2/docs/index.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Mirascope v2
+description: Mirascope v2 Documentation
+---
+
+# Welcome to Mirascope v2
+
+## Basic Usage
+
+<CodeExample file="@/examples/basic_call.py" />
```
willbakst, line 10: what is the benefit of injecting code from separate files rather than just having them present here?

if we can extract the examples and run lint / tests, this seems like it only adds indirection? one benefit I can think of is that a code editor might show you lint errors before actually running the linter, but that's really it

pyproject.toml
teamdandelion: https://github.com/Mirascope/api2mdx 
https://pypi.org/project/api2mdx/

willbakst: future reference, there's no need to actually release a package to use it unless you plan to maintain it for external contributors / expect other people to want to download said package. In the `pyproject.toml` we can just point the package source to the repo

another concern I have with the package is that it depends on a completely separate typescript package, so it feels weird to have as a released pypi package.

willbakst: to be clear, i'm not against / opposed to having the `api2mdx` package, mostly just noting that there are other means of dependency before we're ready to release something for whatever reason

package.json
teamdandelion: mirascope-ui consists of (mostly) frontend component code that we want to standardize across our frontend UIs. created in the style of a shadcn registry, and with the assumption that the components are relatively small blocks that are committed into the consuming repo, and individually updated.

docs-viewer is a fork of mirascope/website, including the backend for processing mdx content files, vite configuration for bundling and building the site, mdx-based route generation using tanstack. it's not a component, more like a framework — a la nextra but focused on our specific style and use cases.

I don't think docs-viewer should ever live in mirascope/ui, as the scope is much larger, and the sync/update strategy in mirascope/ui is intended for modular components, not frameworks. However, this is an occasion to extract a lot more of the UI into mirascope-ui — eg. all of the mdx components used in api rendering 

In the long run, it'd be good to deduplicate the shared parts of docs-viewer and mirascope/website. Assuming we don't want mirascope/mirascope (or other docs repos) to actually own setting up tanstack routes etc, we'll want to keep docs-viewer as a framework. However, we could refactor mirascope/website to depend on it, so the website becomes "just another docs site".

mirascope/experimental/v2/docs/index.mdx
teamdandelion: I found that doing all the examples inline was a worse development experience, for several reasons:

1. Lack of editor support, including syntax highlighting and typing integration, is friction. The iteration cycle of "get type errors displayed directly in editor" is much nicer than "run a script that regenerates snippets and re-runs typechecking". Much higher likelihood of writing code that doesn't typecheck, and not realizing until you get to CI.
2. Worse tooling support. Tools like `ruff format --fix` will fix the generated files, not the source. So being able to use auto code formatters or otherwise maintain consistent styles in the example code requires building additional infrastructure, or manually copying the changes back into the file.
3. Consistently updating the inline examples is painful, especially trying to make systematic changes to files containing multiple examples. E.g. if we have a guide with 4 distinct examples in two styles, and want to refactor them all consistently. It's at lot easier to refactor (& review) a small change to 8 distinct files , vs applying the change to 8 locations within a large mdx file. 
4. We avoid the general "two sources of truth" issue, where the snippet files confuse contributors and LLMs who may try to edit them rather than the versions in the mdx. 

Overall, the .py representation is a lot more powerful and has great tooling support, and keeping them well maintained inside of mdx is expensive.

package.json
teamdandelion: Working on it

mirascope/experimental/v2/docs/index.mdx
willbakst: Let's make sure that all code highlighting has to happen in the code example itself through the comment syntax we've been using (i.e. remove any meta highlighting functionality from the `<CodeExample />` block)

scripts/merge-api-meta.ts
teamdandelion: It's a bit of cruft, fixed it to import from @mirascope.

package.json
teamdandelion: Done. This means `bun run dev` or `bun run build` at top level will no longer work (you'd need to cd into `mirascope/experimental/v2/docs_config`), I added `uv run docs-dev` and `uv run docs-build` instead.

pyproject.toml
teamdandelion: Heh well moved it under the docs_config dir now if that feels better

pyproject.toml
teamdandelion: Having talked about it this morning, I think we should rename docs-viewer to docs-framework or something, and absorb api2mdx into it. You're right that it's tightly coupled to that system in particular, also we could absorb mergeApiMeta.ts into docs-framework too. 

Good point about pointing pyproject.toml to the repo. That would have been cleaner.

bun.lock
teamdandelion: Done

mirascope/experimental/v2/examples/basic_call.py
teamdandelion: lol fixed. Just to be nitpicky, I think it is technically a valid f-string, just one with an unintended $ inside it :P

willbakst (approved):

PR #992, opened 2025-06-13
chore: cleanup v2 branch to only contain v2 code

teamdandelion:
This commit removes ~everything except for the v2 code so that we can have a minimal / clean slate for developing Mirascope v2.

How I tested it:
- `uv run docs-dev` and `uv run docs-build && npx serve --single docs_config/dist` both work
- `uv run pyright .` has no type errors

This PR is hard to review because so much is getting removed; I recommend looking at pyproject.toml and readme.md in particular.

Note: This is merging into the v2-docs branch; once v2-docs merges we can re-target this to merge into v2.


teamdandelion:
Moved it to docs/config.

Issue #994, opened 2025-06-22
Remove custom file size detection google provider


CharlesOural:
### Description

I made a contribution some time ago to automatically detect if the payload we send to google is larger than 10MB. But some update of the google sdk later, it seems that it's bugged and not working when changing to vertex ai (as it don't support remote file upload).

I suggest to remove this special behavior and let the user decide if it needs to upload the images. And if using vertex ai you don't need (it accept up to 500MB payloads).

For now it does not work for both gemini and vertex because of a bug inside the google sdk... I pushed a bug report but they don't seems to be responsive...

If you are ok with that I can go ahead and propose a pull request. If you think it's better to keep it I can add a switch to only use with gemini and deactivate for vertex.

willbakst:
Seems like the easiest fix here would be to disable it for vertex until the bug is fixed?

CharlesOural:
true... I'll do it !

Also I noticed that vertex is a separate provider but you just need to use :
`export GOOGLE_GENAI_USE_VERTEXAI=True` and use the google provider. Is it something you use ?

willbakst:
Yes, if you use the `google` provider it uses their `genai` packages which support this (you can also do it through a custom client with the same settings)

CharlesOural:
perfect ! So you are going to deprecate vertex and gemini provider ?

working on the fix

willbakst:
They are already deprecated

CharlesOural:
amazing, done the pull request here : https://github.com/Mirascope/mirascope/pull/995

willbakst:
Merged and released in `v1.25.2`

CharlesOural:
let's rock !

PR #995, opened 2025-06-22
fix upload images if payload size exeed 5MB when using vertex ai

CharlesOural:
see https://github.com/Mirascope/mirascope/issues/994#issuecomment-2994501542

### DIFF
--- a/mirascope/core/google/_utils/_convert_message_params.py
+++ b/mirascope/core/google/_utils/_convert_message_params.py
@@ -89,7 +89,10 @@ async def _convert_message_params_async(
                     converted_content.append(PartDict(inline_data=blob_dict))
                     image_size = len(part.image)
                     total_payload_size += image_size
-                    if _over_file_size_limit(total_payload_size):
+                    if (
+                        _over_file_size_limit(total_payload_size)
+                        and not client.vertexai
+                    ):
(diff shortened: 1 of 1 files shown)


mirascope/core/google/_utils/_convert_message_params.py
```diff
@@ -89,7 +89,10 @@ async def _convert_message_params_async(
                     converted_content.append(PartDict(inline_data=blob_dict))
                     image_size = len(part.image)
                     total_payload_size += image_size
-                    if _over_file_size_limit(total_payload_size):
+                    if (
+                        _over_file_size_limit(total_payload_size)
```
willbakst, line 93: why not just update `_over_file_size_limit` to be the vertex limit if using vertex and accept the bug above 500mb until they fix it?

CharlesOural: vertex ai don't support the `client.aio.files.upload` you can't send files to the vertex ai api. They have a hard limit of 500MB for their api.

CharlesOural: This is why I just removed the feature of file upload if exceed a certain limit when using vertex ai client...

PR #1026, opened 2025-06-23
chore: remove Text content dataclass

teamdandelion:
It is redundant with directly specifying content as str. It's cleaner if we only have one canonical way to represent str content.


willbakst (approved):

mirascope/llm/responses/base_response.py
```diff
@@ -48,7 +48,7 @@ class BaseResponse(Generic[ResponseContentT, T]):
     content: ResponseContentT | Sequence[ResponseContentT]
     """The content generated by the LLM."""
 
-    texts: Sequence[Text]
+    texts: Sequence[str]
```
teamdandelion, line 51: Note: there was a prior inconsistency here where response.text is `str` not `Text`, but response.texts was a sequence of `Text` not sequence of `str`. IMO this validates that we're better off having just one way to represent it.

PR #1033, opened 2025-06-23
Add messages.mdx with examples

teamdandelion:


docs/_meta.json
```diff
@@ -9,7 +9,11 @@
         "children": [
           {
             "slug": "index",
-            "label": "Mirascope V2"
+            "label": "Welcome to Mirascope V2"
+          },
+          {
+            "slug": "messages",
```
willbakst: this should be inside a "Learn" section (like we currently have for v1 that separates from Welcome etc.)

docs/_meta.json
```diff
@@ -9,7 +9,11 @@
         "children": [
           {
             "slug": "index",
-            "label": "Mirascope V2"
+            "label": "Welcome to Mirascope V2"
```
willbakst: I think I prefer just "Welcome" here

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
+---
+title: Messages
+description: Learn how to use Messages to communicate with Large Language Models in Mirascope.
```
willbakst: nit: add (LLM) here

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+---
+
+# Messages
+
+Messages are the fundamental building blocks of Large Language Model (LLM) interactions. Understanding messages is essential for building any LLM-powered application, as they define how you communicate with an AI model and how it responds to you.
```
willbakst: nit: I wonder if it would read better to say "Messages are the fundamental building blocks for interacting with Large Language Models (LLMs). Understanding..."

willbakst: nit: I find the usage of "AI" a little weird, feels like it should be "LLM" given context, i.e. "...communicate with an LLM and how it responds to you."

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+## What are Messages?
+
+In the context of LLMs, a "message" represents a single unit of communication in a conversation. Think of it like a chat conversation where each participant takes turns speaking. Each message has two key components:
```
willbakst: nit: "In the context of LLMs" seems redundant given intro paragraph?

I might also do something like "A `Message` represents..." where `Message` links to the API reference.

This could result in removing the final sentence of this section.

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+1. **Role**: Who is speaking (system, user, or assistant)
```
willbakst: nit: what do you think about having this be `role` instead of **Role**? Same for content

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+2. **Content**: What is being said (text, images, audio, etc.)
+
+This message-based structure allows LLMs to maintain context across a conversation and understand the flow of dialogue between different participants.
+
+In Mirascope, each message is represented by a [Message class](../api/messages#message).
+
+## Why Messages Matter
+
+Understanding messages is crucial because they are the fundamental mechanism for communicating with large language models. Properly using messages enables many capabilities:
```
willbakst: Seems redundant here, it's the second time we've said "fundamental..."

I feel like maybe we should drop the "What are Messages?" section header and merge its content with the content directly beneath the title?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+1. **Context Preservation**: Messages allow LLMs to maintain conversation history and context
```
willbakst: likely worth mentioning tools and agent stuff here in some capacity

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+2. **Multimodal Capabilities**: You can include text, images, audio, and other content types
+3. **Role Clarity**: Different roles help the LLM understand the conversation structure
+4. **Conversation Flow**: Messages enable back-and-forth dialogue rather than single isolated queries
+
+When you call an LLM through Mirascope, you'll pass an array of messages that represents the entire conversation context.
+
+## Message Structure Fundamentals
+
+Every message in Mirascope follows the same structure: it has a role, indicating who the message is from, and it contains the content of the message.
```
willbakst: again this feels redundant with the information above on role and content

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+These shorthand functions (`llm.system`, `llm.user`, `llm.assistant`) are the most convenient way to construct messages.
+
+<Note title="Manual Message Construction" collapsible={true} defaultOpen={false}>
+
+For more control, you can construct messages directly using the `Message` class:
+
+<CodeExample file="examples/messages/basic.py" lines="7-13" />
+
+Both approaches create identical message objects. We recommend the shorthand functions for convenience.
```
willbakst: nit: update to "objects, so we recommend..."

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+</Note>
+
+## Message Roles Deep Dive
+
+Every message must have a role that identifies who is speaking. There are three role options: `"system"`, `"user"`, and `"assistant"`.
```
willbakst: realizing that technically OpenAI has a "developer" role that we should likely support

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+For more control, you can construct messages directly using the `Message` class:
+
+<CodeExample file="examples/messages/basic.py" lines="7-13" />
+
+Both approaches create identical message objects. We recommend the shorthand functions for convenience.
+
+</Note>
+
+## Message Roles Deep Dive
```
willbakst: nit: drop "Deep Dive", I think "Message Roles" is sufficient

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+3. **Role Clarity**: Different roles help the LLM understand the conversation structure
+4. **Conversation Flow**: Messages enable back-and-forth dialogue rather than single isolated queries
+
+When you call an LLM through Mirascope, you'll pass an array of messages that represents the entire conversation context.
+
+## Message Structure Fundamentals
+
+Every message in Mirascope follows the same structure: it has a role, indicating who the message is from, and it contains the content of the message.
+
+### Basic Message Creation
```
willbakst: I think we should remove this entire section and spread it out.

I think we should put a single example earlier as part of the intro, something like this:

```python
from mirascope import llm

user_message = llm.Message(role="user", content="Hello!")
```

We should put this right before mentioning `role` and `content` so that they have a code example as context.

Then I think we should mention the roles available in the "Message Roles" section where we have the info reversed to say "Message Construction Shorthand" where we mention the `llm.user` and other such messages, immediately followed by explanations of each role where the code examples use the shorthand.

WDYT?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+</Note>
+
+## Message Roles Deep Dive
+
+Every message must have a role that identifies who is speaking. There are three role options: `"system"`, `"user"`, and `"assistant"`.
+
+### System Messages
+
+System messages set the context and instructions for the AI assistant. They're like giving the AI a job description or personality before the conversation begins.
```
willbakst: qq: when do we choose "LLM" vs. "AI?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+<CodeExample file="examples/messages/basic.py" lines="3-3" />
+
+System messages typically define the AI's role or persona, provide instructions or constraints, and give context about the task or domain. Often times prompt engineering focuses on improving the system message, for example by adding detailed examples of representative queries and the desired outputs.
```
willbakst: I prefer the user/assistant message content structure below where it's a single paragraph followed by the code example. The system message section feels a little redundant, and I think we could merge the two paragraphs into a single one for more consistency

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+<CodeExample file="examples/messages/basic.py" lines="4-4" />
+
+### Assistant Messages
+
+Assistant messages are responses from the AI. You typically don't create these yourself - they're generated when the LLM responds to your conversation. However, you might manually create them when building conversation history or providing examples.
+
+<CodeExample file="examples/messages/basic.py" lines="5-5" />
+
+### Role Interaction Patterns
```
willbakst: same thing here, start with the paragraph explanation followed by the code example

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+Here's how roles work together in a typical conversation:
+
+<CodeExample file="examples/messages/basic.py" lines="15-19" />
+
+The system message sets the context, user messages drive the conversation forward, and assistant messages provide responses that maintain the established context.
```
willbakst: we should also mention tools here I think?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+<CodeExample file="examples/messages/content.py" lines="4-9" />
+
+<Info>
+Images can be provided as:
+- File paths to local images
+- URLs to online images
+- Base64-encoded image data
+- Raw bytes data
+
+You can optionally provide the id field as a unique identifier for that image, which can be useful for tracking and referencing generated images.
```
willbakst: if we mention id, should we include it in the example with an optional comment like we do for transcript for audio below?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+<CodeExample file="examples/messages/content.py" lines="11-16" />
+
+<Info>
+Audio files can be provided as:
+- File paths to local audio files
+- URLs to online files
+- Base64-encoded strings
+- Raw bytes data
+
+Audio files may also have an optional transcript, which is useful for accessibility. As with images, you may also provide an id as a unique identifier.
```
willbakst: same question here

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+</Info>
+
+### Content Combinations
+
+You can combine different content types in a single message:
+
+<CodeExample file="examples/messages/content.py" lines="18-24" />
+
+This is particularly powerful for creating rich, multimodal interactions where you want to provide both textual context and visual or audio information.
```
willbakst: nit: update language to "...want to provide text, visual, and/or audio context"

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+## Building Conversations
```
willbakst: hmm, seeing this now I feel like the earlier "Role Interaction Patterns" section feels a bit redundant? Seems like we can just roll that into this section?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+Messages work together to create conversations. The order and structure of messages determines how the AI understands and responds to your requests.
+
+### Conversation Structure
+
+A conversation is simply a list of messages in chronological order:
+
+<CodeExample file="examples/messages/conversations.py" lines="4-11" />
+
+This conversation structure maintains context, allowing the AI to remember what was discussed earlier and provide relevant responses.
```
willbakst: nit: I would maybe replace "remember" with "see" and put it in quotes?

willbakst: nit: I would update to "...what was discussed earlier in order to generate a more relevant response"

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+### Conversation Patterns
```
willbakst: this feels a little bit weird because all cases include the generated response, but in most cases when writing the messages array the user will stop short of the final assistant message.

Maybe we should update this as such? Something like "Single Query" rather than "Single Exchange"?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+**Single Exchange:**
+
+<CodeExample file="examples/messages/conversations.py" lines="14-17" />
+
+**Multi-turn Conversation:**
+
+<CodeExample file="examples/messages/conversations.py" lines="20-26" />
+
+**Context Building:**
```
willbakst: what's the difference between this and the "mult-turn" section above?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+Each message in a conversation adds to the context that the AI uses to generate responses. The AI can:
+
+- Reference information from earlier messages
+- Maintain consistent personality and tone
+- Build on previous topics and decisions
+- Understand implied context from the conversation flow
+
+### Practical Example
+
+Here's a complete conversation from start to finish:
```
willbakst: nit: update to "Here's an example of a complete..."

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+Message names are useful for:
+
+- Multi-user conversations
+- Different AI assistants with specialized roles
+- Tracking conversation participants
+- Creating more structured dialogue flows
+
+### Advanced Content Types
+
+Mirascope v2 supports additional specialized content types for advanced use cases:
```
willbakst: nit: drop `v2` since it'll just be "Mirascope" on release

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+### Practical Example
+
+Here's a complete conversation from start to finish:
+
+<CodeExample file="examples/messages/conversations.py" lines="44-64" />
+
+This conversation demonstrates how system instructions, user queries, and assistant responses work together to create a coherent, helpful interaction.
+
+## Advanced Features
```
willbakst: I see Claude do this a lot (break things down into Basic vs. Advanced), but I don't know that I actually like that patter. I think I would rather just have all of these content types explained alongside all of the other content types.

In particular the video, document, etc. content types feel weirdly disjointed even though they are all just content types. Also I expect the document content type to be one of the more common use-cases and not necessarily any more advanced than image or audio

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+#### Document Content
+
+For processing documents like PDFs, Word files, or other structured content:
+
+<CodeExample file="examples/messages/advanced.py" lines="15-16" />
+
+#### Tool Interactions
+
+Tool content is relevant when the LLM wants to call external functions, also known as "Tools". The `ToolCall` content allows the LLM to specify that it wants to call a function, and the `ToolOutput` content allows the LLM to work with the result of calling the tool. For more info, see our guide on [tools](./tools).
```
willbakst: should we link "Tools" to it's (eventual) docs page?

willbakst: nit: what do you think about updating "with the result of calling the tool" to "with the output of the executed tool"?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+<CodeExample file="examples/messages/advanced.py" lines="19-36" />
+
+#### Thinking Content
+
+Some models support thinking, in which case they spend time reasoning before generating their output content. The Thinking content gives insight into the model's thinking process.
```
willbakst: noticing this here and elsewhere, should it be "`Thinking` content" and similar for other mentions of content?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+<CodeExample file="examples/messages/advanced.py" lines="39-48" />
```
willbakst: I think the `id` field will be something like a hash generated by the model but I'm not sure?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+### Dynamic Message Construction
```
willbakst: this is just a prompt template lol, let's maybe save this for the very next section?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+You can build messages programmatically based on your application's logic:
+
+<CodeExample file="examples/messages/advanced.py" lines="52-68" />
+
+## Next Steps
+
+Now that you understand messages, you're ready to learn how to use them with Mirascope's `llm.call` decorator to actually interact with LLM providers. Messages form the foundation for all LLM interactions, whether you're making simple queries, building complex conversations, or creating AI agents.
```
willbakst: Technically `llm.call` only accepts Prompt Templates, so I think we should just mention that the next step is to learn about prompt templates and that this is how we'll eventually call the LLM?

docs/messages.mdx
```diff
@@ -0,0 +1,256 @@
+---
+title: Messages
+description: Learn how to use Messages to communicate with Large Language Models in Mirascope.
+---
+
+# Messages
```
willbakst: Another note: recently the term "Context Engineering" has been becoming more popular. The idea is that it's not "prompt engineering" so much as engineering the correct context for the LLM / Agent can do the right thing.

We might want to use this term here as the core of what "Messages" really are -- Context. This is in line with the `llm.Context` we'll introduce, which at it's core manages the history of messages i.e. the context

docs/_meta.json
teamdandelion: Done

docs/_meta.json
teamdandelion: Done

docs/messages.mdx
teamdandelion: I think we can translate "system" to "developer" at the openai compat layer. As they don't have a "system" role and the definition of the developer role seems basically equivalent to the system role.
https://platform.openai.com/docs/guides/text#message-roles-and-instruction-following

docs/messages.mdx
teamdandelion: Done

docs/messages.mdx
teamdandelion: Done

docs/messages.mdx
teamdandelion: Lol good point.

docs/messages.mdx
teamdandelion: Updated for realism.

docs/messages.mdx
teamdandelion: Seems reasonable, updated.

docs/messages.mdx
teamdandelion: Yeah it's kind of cruft, I"ll take it out.

docs/messages.mdx
teamdandelion: Yeah, on reflection I agree. I collapsed them back together.

docs/messages.mdx
teamdandelion: Rewritten

docs/messages.mdx
teamdandelion: Sure

docs/messages.mdx
willbakst: Ok I'm actually reneging on this. Since we're defining a spec, and `system` is most common, let's just stick with `system` and no `developer` at least for now.

docs/messages.mdx
teamdandelion: I updated it to use LLM consistently.

teamdandelion (commented):
Addressed all the comments (I think). Also generally simplified/streamlined/removed fluff

docs/messages.mdx
teamdandelion: I added the "context" framing subtly in the opening paragraph, and the term "context engineering" to the System messages paragraph.

docs/learn/messages.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+_TODO: Replace this with centralized compatibilty table (#1031)_
+
+### Text Content
+
+The most common type of content is text, represented as plain strings:
+
+<CodeExample file="examples/messages/content.py" lines="3-3" />
+
+### Image, Audio, and Video Content
```
willbakst: nit: i'm torn if this remain as is or change to "Media Content", would love to hear your reasoning on the choice so I can ease my mind lol

docs/learn/messages.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+### Document Content
+
+`Document` content allows you to embed structured document files, such as PDFs. Like media content, it can be specified via a file path, a URL, a base64-encoded string, or raw bytes.
+
+<CodeExample file="examples/messages/content.py" lines="28-32" />
+
+### Tool Interactions
+
+`ToolCall` and `ToolOutput` content is relevant when the LLM is calling external functions, also known as [Tools](./tools). The `ToolCall` content allows the LLM to specify that it wants to use a tool, and the `ToolOutput` content gives the LLM access to the output from that computation.
```
willbakst: in the render I'm seeing tools in quotes, but here I'm not. I prefer not, so assuming the render is lagging, this can be resolved.

willbakst: nit: do you think we should make it slightly more explicit here around the role? i.e. "The `ToolCall` content in an assistant message allows the LLM to specify that it wants to use a tool, and the `ToolOutput` content gives the LLM access to the output from that computation through a following user message."

I don't necessarily love the way I'm phrasing it, it's mostly an example around how we might be more explicit. Not necessary, just curious what you think

docs/learn/messages.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+Some models support thinking, in which case they create intermediate reasoning content that informs their final response. The `Thinking` content gives insight into the model's thinking process. Depending on the provider, this may be the raw thinking text, or a generated summary.
+
+<CodeExample file="examples/messages/content.py" lines="47-55" />
+
+### Content Sequences
+
+A message can contain a sequence of content objects, potentially of multiple types:
+
+<CodeExample file="examples/messages/content.py" lines="57-63" />
```
willbakst: nit: maybe rather than `image` put an actual `llm.Image` here?

docs/learn/messages.mdx
teamdandelion: Yeah, I had a followon commit (maybe I force pushed that one as a tiny nit) where I removed the quotes.

docs/learn/messages.mdx
teamdandelion: I was on the fence too. I think I actually slightly prefer Media content, so I'll push a change.

docs/learn/messages.mdx
teamdandelion: Great call, I think that's definitely worth documenting explicitly.

docs/learn/messages.mdx
teamdandelion: Done (in commit I"m about to push)

willbakst (approved):

PR #1034, opened 2025-06-24
feat: initial v2 typescript structure

willbakst:
- Initial package structure
- Port of the `content` module
- Port of the `messages` module

The typescript example would look something like this:

```typescript
import { llm } from 'mirascope';

const msgs: llm.Message[] = [
    llm.system('You are a helpful assistant.'),
    llm.user([
        'I just read this book:',
        {'type': 'image', 'data': 'path/to/book.png', 'mimeType': 'image/png'},
        'What should I read next?'
    ]),
];
```


typescript/src/llm/content/thinking.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @fileoverview The `Thinking` content class.
+ */
+
+/**
+ * Thinking content for a message.
+ * 
+ * Represents the thinking or thought process of the assistant. This is part
+ * of an assistant message's content.
+ */
+export type Thinking = {
```
teamdandelion, line 11: Note: We'll probably want to add a signature field for compatibility with Anthropic (here & in python)

typescript/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "mirascope",
+  "version": "2.0.0-alpha.0",
```
teamdandelion, line 3: Should we make this exactly consistent with the python version? `2.0.0-alpha-0`

teamdandelion (approved):

typescript/package.json
willbakst: the python version should be `2.0.0-alpha.0`

typescript/src/llm/content/thinking.ts
willbakst: my thinking was that `id` would represent the provider-agnostic "signature"

docs/learn/messages.mdx
```diff
@@ -5,11 +5,11 @@ description: Learn how to use Messages to communicate with Large Language Models
 
 # Messages
 
-Messages are the fundamental building blocks of Large Language Model (LLM) interactions. They are a shared abstraction underpinning LLM usage: you provide context to the LLM via messages, and it responds by generating messages of its own.
+Messages are the fundamental building blocks of Large Language Model (LLM) interactions.They are a shared abstraction underpinning LLM usage: you provide context to the LLM via messages, and it responds by generating messages of its own.
```
teamdandelion: Nit: missing space after `interactions.`

typescript/.eslintrc.json
```diff
@@ -0,0 +1,22 @@
+{
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 2020,
```
teamdandelion: Is there a particular reason for using ecmaVersion 2020 rather than "latest" or something more recent? eslint.org recommends using latest. I notice the tsconfig.build.json targets ES2022. 

https://eslint.org/docs/latest/use/configure/language-options

typescript/README.md
```diff
@@ -0,0 +1,15 @@
...
+bun install
+```
+
+To run:
+
+```bash
+bun run index.ts
+```
+
+This project was created using `bun init` in bun v1.2.7. [Bun](https://bun.sh) is a fast all-in-one JavaScript runtime.
```
teamdandelion: I don't believe `bun run index.ts` will do anything? Maybe update to include the command to run for typechecking. (I like setting up `bun run typecheck` in package.json)

typescript/examples/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
```
teamdandelion: Same question as in the other tsconfig.

typescript/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    // Environment setup & latest features
+    "lib": ["ESNext"],
+    "target": "ESNext",
+    "module": "ESNext",
+    "moduleDetection": "force",
+    "jsx": "react-jsx",
+    "allowJs": true,
+
+    // Bundler mode
+    "moduleResolution": "bundler",
```
teamdandelion: Claude thinks that given this is a library not an application, we may be better off using "node16" for module resolution — that way downstream consumers don't need a bundler.

typescript/.eslintrc.json
willbakst: I just went with what claude generated lol, I can update

typescript/examples/tsconfig.json
willbakst: I believe this is for backwards compat?

typescript/README.md
willbakst: For now just updating to `bun run build` and will address this later as part of the general README updates

typescript/examples/tsconfig.json
teamdandelion: I like the change to "latest" in the main tsconfig, I think we should probably do the same here for consistency

willbakst: Updating to "ESNext"

Issue #997, opened 2025-06-23
Mirascope `v2.0` Python Prompts Interface & Documentation


willbakst:

willbakst:
A few notes from an offline discussion:

* Change `prompt_template` -> `prompt` that takes an optional `template: str` argument. If the `template` is provided, it should conform to our updated custom templating language format and the decorated function must return `None` (i.e. empty function).
* Update `template: str` to use a custom templating language format, something like this:

  ```python
  """
  [SYSTEM] You are a helpful assistant
  [USER] Answer this question: {{ question }}
  """
  ```
* Remove the `:list` and `:lists` specifiers. These feel more like filters than specifiers given that they operate on the template variable rather than specify what it is. For example, `{{ var:image }}` indicates that the `var` variable is in fact an `llm.Image`, whereas `{{ var:list }}` transforms `var` into a newlines separated list, which feels more like a filter/operation on `var` than a specifier. 
* Remove `DynamicConfig` entirely. It feels unnecessarily confusing. It's also completely useless unless using the `template: str` approach, so at that point might as well just use the functional approach or implement better dynamic configuration support directly inside the template itself so it's truly language agnostic.
* No longer support format specifiers in prompt templates since that is specific to Python. Only support it through formatted strings in the functional approach.
* Instead of calling it "Shorthand" vs. "Template" in the tabs for differentiating between the two paths to writing prompts, we should call it "Functional" vs. "Template"

The hope here is that we are extremely clear on what the language-agnostic prompt templating system supports and looks like while keeping language-specific functionality tightly coupled to the language-specific implementation path (namely writing a python function).

PR #1071, opened 2025-06-25
v2 Prompt cleanup

teamdandelion:
This PR implements a prompt interface improvements for #997, including discussion [here](https://github.com/Mirascope/mirascope/issues/997#issuecomment-3002184276). Specifically:

- Remove the StringReturns (and variants) as it is redundant with with ContentReturns
- Remove DynamicConfig and associated support
- Rename "prompt_template" to "prompt" (likewise for related types, and modules)
- Make the template argument optional for prompt, update decorator types to reflect this
- Update prompt docstring to explain that template argument is optional, and also to specify the template substitution rules in more depth
- Adds some basic usage examples, to verify that the typing works

A change I'm considering and interested in feedback on, is reorganizing the protocols in prompt/decorator.py. Consider the following snippet from decorator.py:

```python
class ContentReturn(Protocol[P]):
    """Protocol for a prompt function that returns a single content part."""

    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Content: ...

class ContentSequenceReturn(Protocol[P]):
    """Protocol for a prompt function that returns a content parts sequence."""

    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Sequence[Content]: ...

class MessagesReturn(Protocol[P]):
    """Protocol for a prompt function that returns a list of messages."""

    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> list[Message]: ...

Prompt: TypeAlias = ContentReturn[P] | ContentSequenceReturn[P] | MessagesReturn[P]
"""A prompt function.

A `Prompt` function takes input arguments `P` and returns one of:
- A single `Content` part that will be rendered as a single user message
- A sequence of `Content` parts that will be rendered as a single user message
- A list of `Message` objects that will be rendered as-is
"""

class PromptDecorator(Protocol[DepsT]):
    """Protocol for the `prompt` decorator."""
    @overload
    def __call__(self, fn: Callable[P, None]) -> MessagesReturn[P]:
        """Decorator for creating a prompt."""
        ...
    @overload
    def __call__(self, fn: Callable[P, Content]) -> MessagesReturn[P]:
        """Decorator for creating a prompt from a content function."""
        ...

    @overload
    def __call__(self, fn: Callable[P, Sequence[Content]]) -> MessagesReturn[P]:
        """Decorator for creating a prompt from a content sequence function."""
        ...

    @overload
    def __call__(self, fn: Callable[P, list[Message]]) -> MessagesReturn[P]:
        """Decorator for creating a prompt from a messages function."""
        ...

    @overload
    def __call__(self, fn: Callable[P, None]) -> MessagesReturn[P]:
        """Decorator for creating a prompt from a template function."""
        ...   
```

I think it's rather counter-intuitive that the "Prompt" type describes the union `ContentReturn[P] | ContentSequenceReturn[P] | MessagesReturn[P]`. As I currently understand it, it would make more sense for this type to be called "Promptable" (or maybe PromptFunction). This type is the type that _can_ be made into a prompt, but actually the MessagesReturn is the prompt (as can be seen by the fact that it's the one getting returned by calling the prompt decorator). Planning to sketch out that change in a separate PR.

### DIFF
--- a/mirascope/llm/prompts/decorator.py
+++ b/mirascope/llm/prompts/decorator.py
@@ -314,11 +314,11 @@ def prompt(
     - [MESSAGES] indicates the next variable contains a list of messages to include
     - `{{ variable }}` injects the variable as a string, unless annotated
     - Annotations: `{{ variable:annotation }}` where annotation is one of:
-      image, images, audio, audios, video, videos, document, documents
+        image, images, audio, audios, video, videos, document, documents
     - Single content annotations (image, audio, video, document) expect a file path,
-      URL, base64 string, or bytes, which becomes a content part with inferred mime-type
+        URL, base64 string, or bytes, which becomes a content part with inferred mime-type
     - Multiple content annotations (images, audios, videos, documents) expect a list
-      of strings or bytes, each becoming a content part with inferred mime-type
+        of strings or bytes, each becoming a content part with inferred mime-type
 
     Examples:
         ```python


mirascope/llm/calls/base_call.py
```diff
@@ -8,12 +8,12 @@
...
 
 from ..models import LLM
-from ..prompt import AsyncPrompt, Prompt
+from ..prompt import AsyncPromptable, Promptable
 from ..tools import ToolDef
 from ..types import Jsonable
 
 P = ParamSpec("P")
-PromptT = TypeVar("PromptT", bound=Prompt | AsyncPrompt)
+PromptT = TypeVar("PromptT", bound=Promptable | AsyncPromptable)
```
willbakst: I don't believe any of these changes across all of the calls are correct now.

Consider this example:

```python
from mirascope import llm

@llm.call("openai:gpt-4o-mini")
def answer_question(question: str) -> str:
    return f"Answer this question: {question}"
```

This would work because

```python
def answer_question(question: str) -> str:
    return f"Answer this question: {question}"
```

is `Promptable`. But now let's consider this example:

```python
@llm.call("openai:gpt-4o-mini")
@llm.prompt()
def answer_question(question: str) -> str:
    return f"Answer this question: {question}"
```

Now

```python
@llm.prompt()
def answer_question(question: str) -> str:
    return f"Answer this question: {question}"
```

is no longer `Promptable` but instead `Prompt`.

We need to be careful here to ensure that all of these calls accept not only `Promptable | AsyncPromptable` because `llm.call` will automatically handle them, but also `Prompt | AsyncPrompt`.

I believe updating

```python
PromptT = TypeVar("PromptT", bound=Prompt | Promptable | AsyncPrompt | AsyncPromptable)
```

everywhere should be a sufficient fix?

teamdandelion: Bear in mind that Prompt is a valid Promptable. Promptable is types that can be promoted to prompt, it just so happens that for Prompt the promotion is a no-op.

`Promptable: TypeAlias = ContentPromptable[P] | ContentSequencePromptable[P] | Prompt[P]`

Does this address your concern?

mirascope/llm/prompt/__init__.py
willbakst, line 1: rename this module to `prompts` not just `prompt`

mirascope/llm/prompt/decorator.py
```diff
@@ -0,0 +1,354 @@
...
+        template: A string template with placeholders for variables using the
+            format `{{ variable_name }}` and optional section markers like [SYSTEM],
+            [USER], and [ASSISTANT].
+
+    Returns:
+        A decorator function that converts the decorated function into a prompt.
+
+    Template substitution rules:
+    - If [USER], [ASSISTANT], or [SYSTEM] is present, it demarcates the start of a new
+    message bearing that role.
```
willbakst: weird indentation here should be fixed

mirascope/llm/prompt/decorator.py
```diff
@@ -0,0 +1,354 @@
...
+    ...
+
+
+def prompt(
+    template: str | None = None,
+) -> PromptableDecorator | PromptTemplateDecorator:
+    '''Prompt decorator for turning functions into prompts.
+
+    This decorator always takes a function and transforms it into a function that
+    returns a list of llm.Messages - i.e. a prompt. Its exact behavior depends on
```
willbakst: nit: noticing this should be `list[llm.Message]`

mirascope/llm/prompt/decorator.py
```diff
@@ -0,0 +1,354 @@
...
+
+    If called with a template string, then the decorated function should have arguments
+    matching variables needing substitution in the template, and an empty
+    function body. In this case, the resulting prompt will generate messages as
+    specified in the template string.
+
+    If called without a template string, it should decorate a function that returns
+    either a single content piece (including strings), or a sequence of content pieces,
+    or a list of messages. In this case, the resulting prompt will return either
+    a single user message including the supplied content, or (if the function provides
```
willbakst: nit: technically it returns a list with a single user message

mirascope/llm/prompt/decorator.py
```diff
@@ -0,0 +1,354 @@
...
+    a list of messages), it will return those messages.
+
+    Args:
+        template: A string template with placeholders for variables using the
+            format `{{ variable_name }}` and optional section markers like [SYSTEM],
```
willbakst: nit: section -> role

mirascope/llm/calls/base_call.py
willbakst: Ah ok I missed that. In that case, update `PromptT` -> `PromptableT`

mirascope/llm/prompts/decorator.py
```diff
@@ -283,72 +286,52 @@ def prompt(
...
-    variable is expected to be a list of strings or bytes that specify individual pieces
-    of content, each of which will be added to the current message (with automatically
-    inferred mime type).
-
-    Example using a template:
+    - [USER], [ASSISTANT], [SYSTEM] demarcate the start of a new message with that role
+    - [MESSAGES] indicates the next variable contains a list of messages to include
+    - `{{ variable }}` injects the variable as a string, unless annotated
+    - Annotations: `{{ variable:annotation }}` where annotation is one of:
+      image, images, audio, audios, video, videos, document, documents
```
willbakst: nit: these should be indented one more time than the first line for readability

PR #1081, opened 2025-06-26
fix: improve typescript typechecking

teamdandelion:
Examples still don't typecheck, but the main library does

Running `bun run typescript:typecheck` from root level produces only the errors below.

Fixing this involves either moving the examples directory out of typescript and to root (as typescript-examples or as part of unified examples dir), or updating the examples to import directly from source file as upposed to from "mirascope". I think the former option is clearly better. The current approach doesn't work because trying to do `bun i` inside the examples directory triggers a loop of endlessly installing a recursive dependency tree

```
examples/messages/content.ts:1:21 - error TS2307: Cannot find module 'mirascope' or its corresponding type declarations.

1 import { llm } from 'mirascope';
                      ~~~~~~~~~~~

examples/messages/manual.ts:1:21 - error TS2307: Cannot find module 'mirascope' or its corresponding type declarations.

1 import { llm } from 'mirascope';
                      ~~~~~~~~~~~

examples/messages/prompts.ts:1:21 - error TS2307: Cannot find module 'mirascope' or its corresponding type declarations.

1 import { llm } from 'mirascope';
                      ~~~~~~~~~~~

examples/messages/shorthand.ts:1:21 - error TS2307: Cannot find module 'mirascope' or its corresponding type declarations.

1 import { llm } from 'mirascope';
                      ~~~~~~~~~~~


Found 4 errors in 4 files.

Errors  Files
     1  examples/messages/content.ts:1
     1  examples/messages/manual.ts:1
     1  examples/messages/prompts.ts:1
     1  examples/messages/shorthand.ts:1
     ```

### CODE REVIEW

File: typescript/src/llm/content/index.ts [hunk 1]
```diff
@@ -2,13 +2,13 @@
...
  */
 
-import type { Audio } from './audio';
-import type { Document } from './document';
-import type { Image } from './image';
-import type { Thinking } from './thinking';
-import type { ToolCall } from './tool-call';
-import type { ToolOutput } from './tool-output';
-import type { Video } from './video';
+import type { Audio } from './audio.js';
```
willbakst: at the very least wouldn't this be `.ts`? same everywhere else unless i'm missing something?

teamdandelion: The issue is we're using node16 module resolution, which is recommended for libraries as it's the most widely usable by potential dependents (doesn't assume they are using a bundler, or typescript). This means we're compiling to .js files and .d.ts files. In TS5 there is a "allowImportingTsExtensions" feature, which you would think fixes this, but really doesn't—if we try to use it in build we get:

`error TS5096: Option 'allowImportingTsExtensions' can only be used when either 'noEmit' or 'emitDeclarationOnly' is set.`

Basically I think this is the way we have to do it, unless we decide to use bundler module resolution. But that would force all mirascope-ts dependents to use a bundler (vs "just throwing together a simple CLI tool or experiment" and running it as raw JS), thus adding a bit "obstruction" in our abstraction

willbakst: I genuinely don't know if the library will work if the end user doesn't use a bundled.

Unlike Python, TypeScript doesn't have the same access to runtime information, and we'll likely be depending on types to make the library work at all (for example with tools where we need the type info to build the schema)

I'd also argue we're building a TypeScript library, not a JavaScript library, so I wouldn't necessarily call it an "obstruction" to require the bundler if it's expected

teamdandelion: Ok, that makes sense to me. I changed the module resolution back to bundler. And I fixed the dependency issue by making `typescript/mirascope` and `typescript/examples` into two workspaces of the `typescript/` package. The `typescript/mirascope/package.json` describes the library itself while `typescript/package.json` just sets up the multi workspace configuration.

willbakst (approved):

PR #1090, opened 2025-06-26
v2: Setup lint-staged precommit hooks

teamdandelion:
This adds precommit hooks for both Typescript and Python code verifying typechecking, formatting, and lint rules.
It's pretty fast in both cases (around 0.5s for Typescript, 1.5s for Python since pyright is slower, faster when we someday use ty)

It's implemented via a `lint-staged.ts` in the python and ts repos, needed to set it up that way to avoid cd breaking lint-staged, and to strip file arguments so they don't confuse the type checkers.

The only downside is getting node_modules and bun.lock at root level, but I think it's worth it for lint staged.


willbakst:
Should we start using `ty`?

Happy to merge this in and address using `ty` in a separate PR, but I see teams using it? Might be worth giving it a try.

willbakst (approved):

teamdandelion:
Ty runs in 0.15s on the v2 python codebase, which is 10x faster, and awesome. Currently I'm seeing 75 type errors. Here's a quick Claude Code analysis of them:

>   Main Error Categories:
> 
>   1. ParamSpec issues (~25 errors): Many errors related to using ParamSpec with
>   Generic and Protocol, which ty doesn't seem to support the same way as pyright
>   2. Invalid type forms (~15 errors): Issues with Callable[P, R] where P is a
>   ParamSpec - ty expects explicit parameter lists
>   3. Invalid assignments (~10 errors): Type mismatches, especially with ContextVar
>   generics
>   4. Invalid return types (~15 errors): Protocol methods returning None instead of
>   expected types
>   5. Invalid argument types (~10 errors): Similar ParamSpec issues in different
>   contexts
> 
>   Assessment:
> 
>   These errors would be moderately difficult to address because:
> 
>   - Core issue: ty appears to have stricter/different rules around ParamSpec usage
>   compared to pyright
>   - Systematic changes needed: Many errors stem from the same patterns used throughout
>    the codebase
>   - Breaking changes required: Would need to refactor generic type patterns,
>   potentially affecting API signatures
> 
>   Recommendation: Since the codebase already typechecks under pyright (the standard)
>   and ty is experimental/alpha, it may not be worth addressing these errors unless you
>    specifically need ty compatibility. The errors suggest fundamental differences in
>   how ty handles advanced typing features.

My 2c: This early stage is a great time to look into understanding what's not working, and to either adapt our interfaces to work with Ty, or (if not readily feasible) submit issues upstream and see what Astral thinks. I think 10x faster typechecking is worth putting in some effort towards at least understanding what's going on here. Plus if I do it I will wind up learning a lot more about the finer points of Python typing in the Mirascope codebase. Maybe I can put 2-4 hours into exploring this next week and we can see what comes up.

willbakst:
I think I should take a look first before you spend time on `ty`.

From the points you've brought up, my guess is that it's just super early and doesn't handle stuff like `ParamSpec` well enough yet, in which case the 2-4 hours you might spend would be a complete waste of time.

While slower, `pyright` is generally quite correct. Also we're talking about ~1 second here, so unless `ty` is correct where `pyright` is wrong, I'd argue not worth for now.

PR #1092, opened 2025-06-27
TypeScript Prompt Templates Interface & Documentation

willbakst:


typescript/mirascope/src/llm/prompt-templates/decorator.ts
```diff
@@ -0,0 +1,218 @@
...
+ *   - A sequence of `Content` parts that will be rendered as a single user message
+ *   - A list of `Message` objects that will be rendered as-is
+ */
+type AsyncContextPromptable<P extends PromptParams, DepsT = undefined> =
+  | AsyncContextContentPromptable<P, DepsT>
+  | AsyncContextContentSequencePromptable<P, DepsT>
+  | AsyncContextPromptTemplate<P, DepsT>;
+
+/**
+ * Prompt decorator for turning functions into prompts.
```
teamdandelion: Does it actually make sense to have this file called decorator.ts / a prompt decorator function, if we aren't actually using TS decorators?

typescript/mirascope/src/llm/prompt-templates/decorator.ts
```diff
@@ -0,0 +1,218 @@
...
+ *   `)
+ *   static domainQuestion(domain: string, question: string): {} {
+ *     return {};
+ *   }
+ *
+ *   @promptTemplate()
+ *   static answerQuestion(question: string): string {
+ *     return `Answer: ${question}`;
+ *   }
+ * }
```
teamdandelion: Needs update

typescript/examples/prompt-templates/functional.ts
```diff
@@ -0,0 +1,91 @@
...
+const recommendGenrePromptTemplate = llm.promptTemplate(
+  ({ genre }: { genre: string }) => `Please recommend a ${genre} book`
+);
+
+const analyzeImagePromptTemplate = llm.promptTemplate(
+  ({ image }: { image: llm.Image }) => [
+    'Please recommend a book, based on the themes in this image:',
+    image,
+  ]
+);
```
teamdandelion: I'm kind of wondering how relevant these patterns will be to how we wind up implementing the call interface, and whether we should be aiming to follow the python docs 1:1 here, or just present the prompt template documentation a little bit differently in TS vs in Python

typescript/mirascope/src/llm/prompt-templates/decorator.ts
willbakst: Done!

typescript/mirascope/src/llm/prompt-templates/decorator.ts
willbakst: Done!

typescript/examples/prompt-templates/functional.ts
willbakst: What do you mean by "these patterns" specifically?

As implemented, the `promptTemplate` method implemented the language-agnostic spec of promoting a promptable into a prompt template, which is how I expect the call interface to work. For example, possibly something like this:

```typescript
import { llm } from '@mirascope/mirascope';

const analyzeImage = llm.call({
  model: 'openai:gpt-4o-mini',
  template: ({ image }: { image: llm.Image }) => [
    'Please recommend a book, based on the themes in this image:',
    image,
  ],
});
```

For the spec-style approach it might look like this:

```typescript
import { llm } from '@mirascope/mirascope';

const analyzeImage = llm.call({
  model: 'openai:gpt-4o-mini',
  template: llm.promptTemplate`
    Please recommend a book, based on the themes in this image: {{ image:image }}
  `,
});
```

The thing to notice is that the `template` argument for `llm.call` accepts a `Promptable` (or its variants), and the only time we'll ever really use `llm.promptTemplate` with `llm.call` is when we're using the spec-style templating. This is exactly the same as in Python (where we drop `@llm.prompt_template` unless using spec-style templating).

I think the libraries will be exactly the same in terms of functionality / structure other than the usage of decorators in Python vs. builders in TypeScript, so on the docs side I think we'll just need to be mindful of the usage of those terms.

Curious your thoughts here.

typescript/examples/prompt-templates/intro.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const recommend_genre_prompt_template = (genre: string) => [
+  llm.messages.user(`Recommend a ${genre} book`),
+];
+
+const recommend_genre_prompt_template_from_spec = llm.definePromptTemplate<{
+  genre: string;
+}>`Recommend a {{ genre }} book`;
+
+const recommend_genre_prompt: llm.Message[] =
```
willbakst, line 23: Curious what you think about this (and the Python equivalent) addition.

My thinking here is that we previously did not have a single example in the entire doc showing usage of actually calling the function. For the non-promoted functional case, this is somewhat clear, but the type hinting is not explicitly clear for any of the other `@llm.prompt_template` or `llm.definePromptTemplate` examples, so I added this to have at least one example showing usage.

typescript/examples/prompt-templates/functional.ts
teamdandelion: Thanks, these examples answer my question. This general pattern seems fine to me, with some notes:

- In example one, it would be cleaner (and equivalent) to wrtie:

```typescript
import { llm } from '@mirascope/mirascope';

const analyzeImage = llm.call({
  model: 'openai:gpt-4o-mini',
  template: ( image: llm.Image) => [
    'Please recommend a book, based on the themes in this image:',
    image,
  ],
});
```

- In example 2, I think we would need the template type variables to properly type analyzeImage:

```typescript
import { llm } from '@mirascope/mirascope';

const analyzeImage = llm.call({
  model: 'openai:gpt-4o-mini',
  template: llm.promptTemplate<{image: llm.Image}>`
    Please recommend a book, based on the themes in this image: {{ image:image }}
  `,
});
```

typescript/examples/prompt-templates/intro.ts
teamdandelion: I'm happy with adding the output here. In my ideal world, when we add the automated example testing, we integrate this kind of expected output logging so these aren't just comments depicting the results from print statements, but are actually validated output.

teamdandelion (commented):
I think we can greatly simplify the type declaration for prompt_template by removing the Positional/Non Positional duplication, and using a union type instead, as follows:

```typescript
type Promptable<T> = T extends readonly unknown[]
  ? (...args: T) => Content | Content[] | Message[]
  : (params: T) => Content | Content[] | Message[];
```

I pushed a full commit that fleshes out this approach, has no type errors, and produces type errors if I break the examples. Take a look and tell me what you think?

https://github.com/Mirascope/mirascope/commit/975b32962a15efa266781b9ca98bd4965805b044

typescript/examples/prompt-templates/functional.ts
willbakst: When I originally wrote the example for (1) I hadn't yet added support for positional arguments. Now that that exists, yes I agree. For (2), you are correct that the typing is missing from my example.

It's also worth noting that since we're not using decorators, I'm opting for `definePromptTemplate` and `defineCall` and `defineTool` as the method names to be more explicit that these are definition and not function calls.

willbakst:
> I think we can greatly simplify the type declaration for prompt-template.ts by removing the Positional/Non Positional duplication, and using conditional types instead, as follows:
> 
> ```ts
> type Promptable<T> = T extends readonly unknown[]
>   ? (...args: T) => Content | Content[] | Message[]
>   : (params: T) => Content | Content[] | Message[];
> ```
> 
> I pushed a full commit that fleshes out this approach, has no type errors, and produces type errors if I break the examples. Take a look and tell me what you think?
> 
> [975b329](https://github.com/Mirascope/mirascope/commit/975b32962a15efa266781b9ca98bd4965805b044)

Yeah this is much preferred. I tried this and it wasn't working for whatever reason, so I went explicit, but I much prefer the simplification here given it works.

Feel free to push the change directly to this branch / resolve comments / merge it in when it's ready

PR #1094, opened 2025-07-01
Add calls.mdx, iterate on Response api

teamdandelion:
Major changes in this PR:

- Response.model is changed to a full model specifier string (eg `"openai:gpt-4o-mini"`) for consistency with the call interface
- Rename: BaseResponse -> Response, and Response -> SimpleResponse. Move tool and tool properties to BaseResponse. Maintain type safety by having SImpleResponse / ContextResponse refine the type of tool returned. (Future improvement: Rename BaseTool -> Tool, Tool -> SimpleTool) 

Rationale:
It's a bit confusing to center "BaseResponse" as the type that you get back from calls — much nicer to say you get a `llm.Response`. However, that is inaccurate unless we change the class names. Note that unless you are worried about tools, you don't actually care about SimpleResponse vs ContextResponse - and in that case you can refine the type you are assigning to. 

- Simplify ResponseContent type implementation while maintaining type safety, differentiate SimpleResponseContent and ContextResponseContent

- Modify `response.text` so that it gives back all text content, concatenated, and an empty string if empty. This makes it slightly less consistent with the other content-type accessors, however text is already special in multiple respects: in being a raw string rather than a dataclass, and in having a very natural empty representation ("") and a very natural way to concatenate multiple text parts. IMO having it return the first text piece (but potentially not all of the text) is leaking an implementation detail in a way that will confuse users.

- Make `response.content` always a sequence of content. This is simpler and more consistent with the other accessors.
- Make Response no longer an iterable. There is no need when you can just iterate over `response.content`.
- Add thinking content to responses.
- Add unified exception classes, rather than passing through provider-specific exceptions. This is a much better fit for Mirascope v2 being unified and holistically abstracting over the provider APIs.
- Add an initial draft of calls.mdx
- Enable custom clients, add to docs
- Enable params (moving temperature and max_tokens to generic params across all providers), add to docs

Another note: I wrote the intro content for calls.mdx to show the spec and functional style, and then wrote the subsequent examples in the functional style. I propose this as a pattern going forward for the coming guides. I think this strikes a nice balance between consistently showing the users that both styles are available and they can choose their preference, and not needing to duplicate every single example, particularly ones where there is no added educational value in showing both functional and spec styles. Even if it doesn't clutter the user view, it saves maintenance burden for us, and it reduces wasted tokens in the LLM text (the LLM does not need to see every single example in both styles)


docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+# Calls
+
+<Note>
+  If you haven't already, we recommend reading the sections on [Messages](/docs/learn/messages) and [Prompt Templates](/docs/learn/prompt_templates).
+</Note>
+
+Mirascope calls are how you actually execute your prompts and get responses from LLMs. They provide a provider-agnostic interface that works consistently across OpenAI, Anthropic, Google, [and other providers](/docs/compatibility). As with [prompt templates](/docs/learn/prompt_templates), Mirascope approaches calls through a parameterized, functional interface.
+
+Let's dive into an actual example! Here's a LLM call that gets a genre-aware book recommendation:
```
willbakst: nit: "a LLM" -> "an LLM"

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+
+</Tab>
+</TabbedSection>
+
+<Note title="Functional vs Spec Style" collapsible={true} defaultOpen={false}>
+
+By clicking the tabs, you can choose between the [functional](/docs/learn/prompt_templates#functional-style-templates) and [spec](/docs/learn/prompt_templates#spec-style-templates) styles for prompt templating. Both approaches produce the same result.
+
+We're showing both styles in the intro; however for the rest of the doc, we'll use whichever style is most convenient for the example at hand.
```
willbakst: nit: missing comma after "however, "

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+<Tab value="Spec">
+<CodeExample file="examples/calls/intro_spec.py"/>
+
+
+</Tab>
+</TabbedSection>
+
+<Note title="Functional vs Spec Style" collapsible={true} defaultOpen={false}>
+
+By clicking the tabs, you can choose between the [functional](/docs/learn/prompt_templates#functional-style-templates) and [spec](/docs/learn/prompt_templates#spec-style-templates) styles for prompt templating. Both approaches produce the same result.
```
willbakst, line 30: we should either show functional and spec everywhere or pick one and run with that everywhere except prompt templates.

my strong preference is to show both so that a user who prefers spec (or functional) can see spec (or functional) for every example once they select spec (or functional); however, given the maintenance burden I would understand the choice to only show functional.

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+description: Learn how to call LLMs and get responses, learning how to move from context engineering to LLM powered results.
+---
+
+# Calls
+
+<Note>
+  If you haven't already, we recommend reading the sections on [Messages](/docs/learn/messages) and [Prompt Templates](/docs/learn/prompt_templates).
+</Note>
+
+Mirascope calls are how you actually execute your prompts and get responses from LLMs. They provide a provider-agnostic interface that works consistently across OpenAI, Anthropic, Google, [and other providers](/docs/compatibility). As with [prompt templates](/docs/learn/prompt_templates), Mirascope approaches calls through a parameterized, functional interface.
```
willbakst: There's language from the current docs that I think is good that would work well here.

What if this first paragraph read like this?

"A Mirascope "call" is a powerful tool that enables you to transform [prompt templates](/docs/learn/prompt_templates) into LLM API calls with minimal boilerplate code while providing type safety and consistency across OpenAI, Anthropic, Google, [and other providers we support](/docs/compatibility)."

docs/content/learn/calls.mdx
willbakst: nit: "gets" -> "generates"

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+</TabbedSection>
+
+<Note title="Functional vs Spec Style" collapsible={true} defaultOpen={false}>
+
+By clicking the tabs, you can choose between the [functional](/docs/learn/prompt_templates#functional-style-templates) and [spec](/docs/learn/prompt_templates#spec-style-templates) styles for prompt templating. Both approaches produce the same result.
+
+We're showing both styles in the intro; however for the rest of the doc, we'll use whichever style is most convenient for the example at hand.
+</Note>
+
+As you can see, `recommend_book` is a function which, when called, calls the LLM for us. We choose a model and define the prompt, and we get back a `llm.Response` object that has the actual response content from the LLM.
```
willbakst: nit: this reads a bit weird.

How about something like this?

"As you can see, `recommend_book` is a prompt template that's been transformed into an LLM function where calling it returns an `llm.Response` object generated by the LLM you select."

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+
+## Model and Provider Selection
+
+Mirascope makes it easy to switch to a different LLM model, or even an entirely different provider; you can just update the model choice string. See our [compatibility page](/docs/compatibility) to see all the available model choices, as well as what features they support.
```
willbakst: nit: structure here is overly complicated.

how about this?

"Mirascope makes switching to a different LLM, or even an entirely different provider, as easy as updating a single string. See our [compatibility page](/docs/compatibility) to see all of the available `provider:model` choices and features they support."

willbakst: what if we added a followup sentence here like this?

"Try switching the provider in the right bar to see examples with your preferred provider"

Also, we should highlight the `llm.call` decorator line in the above example (and ensure provider swaps happen, which isn't right now)

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+## Working with Response Objects
+
+Regardless of which provider you use, you'll get back a consistent object: an [llm.Response](/docs/api/responses#response). Let's explore how to work with responses, starting with the most common use cases.
+
+### Getting Response Content
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `text` | str | Returns all text content concatenated with newlines, or empty string if none. This is usually what you want for text-only responses. |
+| `image` | [Image](/docs/api/content#image) \| None | Returns the first image in the response content, if any. |
```
willbakst: why not colocate `image` with `images` etc.? I would find that a lot easier to grok

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+| `texts` | Sequence[str] | All text content as individual strings (empty sequence if none). |
+| `images` | Sequence[[Image](/docs/api/content#image)] | All image content (empty sequence if none). |
+| `audios` | Sequence[[Audio](/docs/api/content#audio)] | All audio content (empty sequence if none). |
+| `videos` | Sequence[[Video](/docs/api/content#video)] | All video content (empty sequence if none). |
+| `thinkings` | Sequence[[Thinking](/docs/api/content#thinking)] | All thinking content (empty sequence if none). |
+| `tools` | Sequence[[Tool](/docs/api/tools#tool)] \| Sequence[[ContextTool](/docs/api/tools#contexttool)] | All tools called by the LLM (empty sequence if none). See our [Tools guide](/docs/learn/tools) for details. |
+
+### Response Metadata
+
+Information about the response quality and cost:
```
willbakst: nit: I think we can drop this (also not sure where "response quality" comes into play)

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `finish_reason` | [FinishReason](/docs/api/responses#finishreason) | The reason why the LLM finished generating a response (e.g., "stop", "length"). |
+| `usage` | [Usage](/docs/api/responses#usage) \| None | The token usage statistics for the request to the LLM, if available. |
+| `cost` | float \| None | The cost of the request to the LLM, if available. |
+
+### Call Information
+
+Details about how the call was made:
```
willbakst: same here

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `provider` | str | The provider of the LLM (e.g. "openai", "anthropic"). |
+| `model` | str | The model identifier that generated the response (e.g. "openai:gpt-4", "anthropic:claude-3-5-sonnet-latest"). |
```
willbakst, line 84: I would either keep this as just the model or remove `provider` as a property. My preference would be to remove `property` for consistency with the `model` param to `llm.call` as you mentioned

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+| `args` | dict[str, Any] | The arguments used to generate the response. |
+| `spec` | str \| None | The string spec used to define the messages array, if a spec-style prompt was used. |
+| `messages` | list[[Message](/docs/api/messages#message)] | The messages used to generate the response. This will not include the system message. |
```
willbakst: if we're not going to include the system message here, I would argue we should include a `system` property so it's accessible.

However, originally we excluded the system message so you would easily take `response.messages` and update/insert it for future calls. With the new addition of `llm.Context` this is no longer necessary since the context will handle this. No harm in including the system message here (if any) since it's technically part of the messages used to generate the response and is more likely the expectation.

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+| `raw` | Any | The raw response from the LLM. This contains the original, unprocessed response object from the provider. |
+
+### Response Formatting
+
+For structured outputs, you can format responses according to a schema:
```
willbakst, line 92: nit: this should link (to the non-existent but will exist page)?

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+| Method | Type | Description |
+|--------|------|-------------|
+| `format()` | T | Format the response according to a response format parser. This method is only available if the call was created with a response format type. See our [Response Models guide](/docs/learn/response_models) for details. |
```
willbakst: nit: guide will be called either "Structured Outputs" or "Response Format". Given earlier comment, you could also drop the link here in favor of above (my preference).

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+- [`llm.PermissionError`](/docs/api/exceptions#permissionerror): Permission/authorization failures (403)
+- [`llm.BadRequestError`](/docs/api/exceptions#badrequesterror): Malformed requests (400, 422)
+- [`llm.NotFoundError`](/docs/api/exceptions#notfounderror): Requested resource not found (404)
+- [`llm.RateLimitError`](/docs/api/exceptions#ratelimiterror): Rate limits exceeded (429)
+- [`llm.ServerError`](/docs/api/exceptions#servererror): Server-side errors (500+)
+- [`llm.TimeoutError`](/docs/api/exceptions#timeouterror): Request timeouts or deadline exceeded
+
+### Basic Error Handling
+
+<CodeExample file="examples/calls/exceptions.py" />
```
willbakst, line 119: why do some of these have `e.provider` but others don't?

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+## Call Params
```
willbakst: I would move this section way up right after "Model and Provider Selection" before "Working with Response Objects"

docs/content/learn/calls.mdx
willbakst, line 1: noticing some weirdness on the sidebar where "Call" appears even though it's the title and everything else is nested. Mostly just a note since I know we can't fix this here and will do so during website updates.

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Mirascope supports common parameters like `temperature` and `max_tokens` that work consistently across all providers:
+
+<CodeExample file="examples/calls/params.py" />
+
+These parameters are part of the base interface, so you can use them with any model without worrying about provider-specific differences.
+
+## Handling Token Limits
```
willbakst: IMO if we're going to show handling token limits here, we should either implement an interface for continuation utilities or hold off and remove this section until we implement said interface.

One reason behind this is that handling token limits is more complex than just text responses. What do you do when generating audio? What about structured outputs?

I think the scope for continuation is large enough that we should hold off for a future feature at least until we have the initial experimental interface in users' hands.

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+<CodeExample file="examples/calls/token_continuation.py" />
+
+This pattern is useful for generating longer content while respecting token limits and avoiding truncated responses.
+
+## Custom Clients
+
+If you want to access provider-specific client configuration, you may pass a custom client tied to that specific provider, as below:
+
+<TabbedSection>
```
willbakst: shouldn't these be handled by the provider switch? Or no because it's custom client and thus not just a provider:model swithc?

python/mirascope/llm/responses/simple_response.py
```diff
@@ -0,0 +1,21 @@
...
+from ..tools import Tool
+from ..types import Jsonable
+from .content import SimpleResponseContent
+from .response import Response
+
+T = TypeVar("T", bound=object | None, default=None)
+
+
+@dataclass
+class SimpleResponse(Response[SimpleResponseContent, Tool[..., Jsonable], T]):
```
willbakst: I really really don't like this. I don't understand what's wrong with having `BaseResponse` which `Response` and `ContextResponse` subclass with proper typing?

To put it another way, we could completely remove `BaseResponse` and just have `Response` and `ContextResponse`. The only reason we have `BaseResponse` at all is DRY for all the shared properties (which to me is exactly why I originally used "Base" in the first place).

I also have an extremely strong distaste for the term "Simple" here becuase it's not "simple" it's just a response that is not a context response (but it still has tools, structure, etc.)

python/mirascope/llm/responses/content.py
```diff
@@ -1,10 +1,16 @@
 from typing import TypeAlias
 
-from ..content import Audio, Image, Video
+from ..content import Audio, Image, Thinking, Video
 from ..tools import ContextTool, Tool
 
-ResponseContent: TypeAlias = str | Image | Audio | Video | Tool
-"""Content types that can be returned in a model response."""
+BaseResponseContent: TypeAlias = str | Image | Audio | Video | Thinking
```
willbakst: this really confuses me. Why are you against `BaseResponse` but not against `BaseResponseContent`?

python/examples/calls/custom_client_anthropic.py
```diff
@@ -0,0 +1,16 @@
+from mirascope import llm
+
+custom_client = llm.models.AnthropicClient()
```
willbakst: there's something about this that I don't like. I think it's the fact that we're accessing a "client" from the "models" module (where `llm.models.OpenAI` makes sense but `llm.models.OpenAIClient` feels wrong).

I think I would prefer something like this:

```python
from mirascope import llm

custom_client = llm.clients.OpenAIClient()
...
```

This leads to my next question / hot take: what if we have a clients` module?

My idea is that rather than having to duplicate the overloads in `LLM` for every provider (e.g. `OpenAI`), we instead make `LLM` and actual class with an implementation (rather than an `ABC`) that uses `ClientT` under the hood to implement things. As written now, `LLM` would require that we implement each individual provider-specific model, which would incur tons of duplicated overloads (and I'm not really sure what the clients would do in this case anyway).

We can then implement each specific method explicitly with no overloads for each client. For example, `call` and `context_call` and `structured_call` and `structured_context_call` etc. that we then use internally in `LLM` to implement `LLM`. We can have a `BaseClient` abstract class with all of the individual methods that need implementation and reduce the number of overloads we have to maintain.

We can then still have `llm.models.OpenAI` which is just `LLM` with the correct settings. This way users can still access provider-specific features through provider-specific models and still have the interface convenience of `LLM`.

Does this make sense?

docs/content/learn/calls.mdx
teamdandelion: I like it.

docs/content/learn/calls.mdx
teamdandelion: Great — I thought it was weird that the system message wasn't included. Updating here & in the base response docstring.

docs/content/learn/calls.mdx
teamdandelion: No good reason. Dropped it, also changed the exception to have model instead of provider for consistency with call and response.

python/mirascope/llm/responses/simple_response.py
teamdandelion: Discussed offline; reverting to the original structure.

docs/content/learn/calls.mdx
teamdandelion: Good question. I haven't quite worked out the details of provider substitution in this codebase (the old approach was to have a sentinel value like $PROVIDER but that won't typecheck). Given that this code is actually generic except for substituting the client we use, I think we can fold it into provider substitution. Deduplicated.

python/mirascope/llm/responses/content.py
teamdandelion: Discussed offline, it's really about how the api docs render subclasses.

docs/content/learn/calls.mdx
teamdandelion: Took it out of this doc.

python/examples/calls/custom_client_anthropic.py
teamdandelion: Yes, I think this makes sense. I implemented it, along with a refactor to move the provider implementations into provider specific modules (as in llm.providers.openai.Client, llm.providers.openai.Model, etc). This made more sense to me than splitting the provider specific code across multiple modules.

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,126 @@
...
+</TabbedSection>
+
+<Note title="Functional vs Spec Style" collapsible={true} defaultOpen={false}>
+
+By clicking the tabs, you can choose between the [functional](/docs/learn/prompt_templates#functional-style-templates) and [spec](/docs/learn/prompt_templates#spec-style-templates) styles for prompt templating. Both approaches produce the same result.
+
+We're showing both styles in the intro; however, for the rest of the doc, we'll use whichever style is most convenient for the example at hand.
+</Note>
+
+As you can see, `recommend_book` is a prompt template that's been transformed into an LLM function. Calling it returns a `llm.Response` object generated by your chosen LLM.
```
willbakst, line 35: nit: should be "returns an `llm.Response`" ("a" -> "an")

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+## Model and Provider Selection
+
+Mirascope makes switching to a different LLM, or even an entirely different provider, as easy as updating a single string. See our [compatibility page](/docs/compatibility) to see all the `provider:model` choices, as well as what features they support. Try switching the provider in the right sidebar to update examples with your preferred provider.
+
+## Call Params
+
+Mirascope supports common parameters like `temperature` and `max_tokens` that work consistently across all providers:
+
+<CodeExample file="examples/calls/params.py" />
```
willbakst, line 46: i'm still not seeing examples update the provider?

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,126 @@
...
+As you can see, `recommend_book` is a prompt template that's been transformed into an LLM function. Calling it returns a `llm.Response` object generated by your chosen LLM.
+
+
+## Model and Provider Selection
+
+Mirascope makes switching to a different LLM, or even an entirely different provider, as easy as updating a single string. See our [compatibility page](/docs/compatibility) to see all the `provider:model` choices, as well as what features they support. Try switching the provider in the right sidebar to update examples with your preferred provider.
+
+## Call Params
+
+Mirascope supports common parameters like `temperature` and `max_tokens` that work consistently across all providers:
```
willbakst, line 44: noticing we don't really mention or link to any table of the available call params with descriptions

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,126 @@
...
+| `messages` | list[[Message](/docs/api/messages#message)] | The messages used to generate the response. |
+| `raw` | Any | The raw response from the LLM. This contains the original, unprocessed response object from the provider. |
+
+### Response Formatting
+
+For structured outputs, you can format responses according to a schema:
+
+| Method | Type | Description |
+|--------|------|-------------|
+| `format()` | T | Format the response according to a response format parser. This method is only available if the call was created with a response format type. See our [Response Models guide](/docs/learn/structured_outputs) for details. |
```
willbakst, line 96: it's no longer Response Models. I would prefer just to link above for my nit comment and remove this

docs/content/learn/calls.mdx
```diff
@@ -0,0 +1,126 @@
...
+- [`llm.RateLimitError`](/docs/api/exceptions#ratelimiterror): Rate limits exceeded (429)
+- [`llm.ServerError`](/docs/api/exceptions#servererror): Server-side errors (500+)
+- [`llm.TimeoutError`](/docs/api/exceptions#timeouterror): Request timeouts or deadline exceeded
+
+### Basic Error Handling
+
+<CodeExample file="examples/calls/exceptions.py" />
+
+
+## Custom Clients
```
willbakst, line 122: nit: missing next steps section?

python/examples/basic_call.py
```diff
@@ -6,5 +6,5 @@ def answer_question(question: str) -> str:
     return f"Answer this question: ${question}"
 
 
-response: llm.Response = answer_question("What is the capital of France?")
+response: llm.BaseResponse = answer_question("What is the capital of France?")
```
willbakst: hmm, this should just be `llm.Response` no?

python/examples/calls/custom_client.py
```diff
@@ -0,0 +1,16 @@
...
+
+@llm.call(
+    "openai:gpt-4o-mini",
+    client=custom_client,
+)
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book"
+
+
+response: llm.BaseResponse = recommend_book("fantasy")
```
willbakst, line 14: same here, we should never use `BaseResponse` we should only ever use `Response` or `ContextResponse`

python/mirascope/llm/clients/anthropic.py
```diff
@@ -0,0 +1,202 @@
...
+class AnthropicClient(BaseClient[AnthropicMessage, BaseParams, ANTHROPIC_REGISTERED_LLMS]):
+    """The client for the Anthropic LLM model."""
+    
+    def call(
+        self,
+        *,
+        model: ANTHROPIC_REGISTERED_LLMS,
+        messages: Sequence[AnthropicMessage],
+        tools: Sequence[ToolDef] | None = None,
+        params: BaseParams | None = None,
```
willbakst: shouldn't this be the specific anthropic params defined below?

python/mirascope/llm/clients/google.py
```diff
@@ -0,0 +1,202 @@
...
+class GoogleClient(BaseClient[GoogleMessage, BaseParams, GOOGLE_REGISTERED_LLMS]):
+    """The client for the Google LLM model."""
+    
+    def call(
+        self,
+        *,
+        model: GOOGLE_REGISTERED_LLMS,
+        messages: Sequence[GoogleMessage],
+        tools: Sequence[ToolDef] | None = None,
+        params: BaseParams | None = None,
```
willbakst: same here

python/mirascope/llm/clients/openai.py
```diff
@@ -0,0 +1,202 @@
...
+class OpenAIClient(BaseClient[OpenAIMessage, BaseParams, OPENAI_REGISTERED_LLMS]):
+    """The client for the OpenAI LLM model."""
+
+    def call(
+        self,
+        *,
+        model: OPENAI_REGISTERED_LLMS,
+        messages: Sequence[OpenAIMessage],
+        tools: Sequence[ToolDef] | None = None,
+        params: BaseParams | None = None,
```
willbakst: same here

python/mirascope/llm/responses/base_response.py
```diff
@@ -1,46 +1,51 @@
...
-    """The model used to generate the response (e.g. "gpt-4", "claude-3-5-sonnet-latet")."""
+    model: REGISTERED_LLMS
+    """The model identifier that generated the response (e.g. "openai:gpt-4", "anthropic:claude-3-5-sonnet-latet")."""
 
     args: dict[str, Any]
     """The arguments used to generate the response."""
 
-    template: str | None
-    """The string template used to define the messages array, if any."""
+    spec: str | None
```
willbakst, line 42: note: figure you'll update this in follow-on, just a note, feel free to resolve

python/mirascope/llm/responses/base_response.py
```diff
@@ -55,18 +60,29 @@ class BaseResponse(Generic[ResponseContentT, T]):
...
+
+    @property
+    def tool(self) -> ToolT | None:
+        """Returns the first tool used in the response, if any."""
+        raise NotImplementedError()
+
     @property
-    def text(self) -> str | None:
-        """Returns the first text in the response content, if any."""
+    def text(self) -> str:
```
willbakst: i thought we discussed keeping this as the first text part and having a different method for the concat that has some `__repr__` or something?

python/mirascope/llm/calls/decorator.py
```diff
@@ -361,8 +372,8 @@ def call(
     | Sequence[ToolDef | ContextToolDef[..., Any, DepsT]]
     | None = None,
     response_format: type[T] | None = None,
-    # client: Client | None = None,
-    # **params: Unpack[Params],
+    client: BaseClient | None = None,
+    **kwargs: Any,
```
willbakst: kinda confused by this, why is it not `**params`?

python/mirascope/llm/clients/base.py
```diff
@@ -0,0 +1,242 @@
...
+    Stream,
+    StructuredStream,
+)
+from ..tools import ContextToolDef, ToolDef
+
+T = TypeVar("T", bound=object | None)
+MessageT = TypeVar("MessageT")
+ParamsT = TypeVar("ParamsT", bound="BaseParams")
+DepsT = TypeVar("DepsT")
+ModelT = TypeVar("ModelT", bound=str)
```
willbakst: nit: should this be `LLMT`? I don't like how that looks lol but it feels more accurate?

one thing to note is that at some point we're going to introduce `EmbeddingModel`, so worth taking that into account where "Model" may not just mean "LLM"

python/examples/calls/custom_client_anthropic.py
willbakst: Resolving given offline discussion + related change

docs/content/learn/calls.mdx
teamdandelion: Ack. I'm not going to change it as the re-write changes this section entirely.

docs/content/learn/calls.mdx
teamdandelion: It needs re-implementing in docs-framework in a way that's compatible with Mirascope v2. Tracked it in linear with more context.

docs/content/learn/calls.mdx
teamdandelion: Updated this in the rewrite.

docs/content/learn/calls.mdx
teamdandelion: Ack, addressed in the rewrite.

docs/content/learn/calls.mdx
teamdandelion: Added in the rewrite.

python/examples/basic_call.py
teamdandelion: Ack. Removed this example, it was a placeholder from earlier when I just wanted to make sure that *something* typechecked.

python/examples/calls/custom_client.py
teamdandelion: Ack. In the re-write these lines get dropped so not going to change it here.

python/mirascope/llm/clients/anthropic.py
teamdandelion: Updated, thanks

python/mirascope/llm/responses/base_response.py
teamdandelion: Yup, fixed in the followon

python/mirascope/llm/responses/base_response.py
teamdandelion: Good catch, fixed. Added __repr__ in the followon.

python/mirascope/llm/calls/decorator.py
teamdandelion: Good catch. I blame claude lol. Fixed.

python/mirascope/llm/clients/base.py
teamdandelion: Changed. Could also do `LLM_T` maybe less confusing? Also changed the bound to REGISTERED_LLMs

willbakst: Another option would be `LlmT` but I really don't like that lol. I think `LLMT` is fine (at least for now).

willbakst (approved):

PR #1100, opened 2025-07-03
add streams.mdx

teamdandelion:
Notes:

- Refactored the Stream inheritance structure to approximately match Response, with the nuance that we have a BaseSyncStream and BaseAsyncStream that act as intermediaries between BaseStream and Stream/ContextStream/AsyncStream/AsyncContextStream
- I left structured output classes untouched (as with Response)
- I left out the audio streaming content from the streams guide. I think it's an advanced subject that is better suited by its own guide, so I put a deadlink in next steps as a placeholder. It has special dependencies and a special use case. If we want to really support this in the best way, we should add support for the OpenAI realtime api / Gemini live api, which would support bidirectional streaming (inputs and outputs); however I'm not sure how the output streaming would fit into the @llm.call paradigm. For now I suggest we let it be—what do you think?


willbakst:
Streaming audio does not necessarily require the realtime API.

See: https://platform.openai.com/docs/api-reference/responses/create (slick the "Streaming" tab).

We should have a provider-agnostic way to handle TTS and SST models through the `Audio` content type. We may want to introduce content chunks (e.g. `AudioChunk`) or some means of knowing that what we have is not the full content yet.

I'm not necessarily opposed to having separate documentation for how to handle multi-media, but we should be careful about multi-media vs. realtime, which are not 1:1. For realtime we'll have a different API entirely (we have a beta version of this for openai in v1 you can check out)

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+description: Learn how to stream LLM call responses for a more real-time interaction
+---
+
+# Streams
+
+<Note>
+  If you haven't already, we recommend first reading about [Calls](/docs/learn/calls).
+</Note>
+
+Streaming is a powerful feature when using LLMs that allows you to process chunks of an LLM response in real-time as they are generated. This can be particularly useful for long-running tasks, providing immediate feedback to users, or implementing more responsive applications.
```
willbakst: nit: drop "real-time" due to term overload, just say "process chunks of an LLM response as they are generated rather than waiting for the entire response to be generated first" or something?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+1. **Immediate feedback**: Users can see responses as they're being generated, creating a more 
+interactive experience.
+2. **Incremental processing**: Applications can process and act on partial results as they arrive.
+3. **Early termination**: If the desired information is found early in the response, processing can be
+  stopped without waiting for the full generation.
+
+## Basic Usage and Syntax
+
+To use streaming, first create a prompt and transform it into a [call](/docs/learn/calls), just as you would with a non-streaming response. Then, instead of directly invoking the call, you call `.stream()` on it to get a [`llm.Stream`](python/mirascope/llm/responses/stream.py) instead.
```
willbakst: nit: again "a" -> "an" before `llm.Stream` because `llm` starts with the "eh" sound

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+<TabbedSection>
+<Tab value="Python">
+<CodeExample file="examples/streams/intro.py" />
+</Tab>
+</TabbedSection>
+
+`llm.Stream` is an iterable. When you iterate on it, you get [`llm.StreamChunk`](/docs/api/responses#streamchunk)s as they are ready. Each chunk has a `content` property that contains one piece of `llm.Content` (typically strings). In the example, we print each one with `end=""` and `flush=True` to ensure the output is displayed in real-time without line breaks.
```
willbakst: nit: "as they are ready" -> "as they are generated"

willbakst: hmmm, we may want to introduce `ContentChunk` so we can distinctly separate a full `Content` piece and a "chunk"

another option could be to include another optional field on content like `delta` that represents the delta of that content that was just generated where the value is the collected value up to and including that delta. this is worth a discussion

willbakst: again drop "real-time" here

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+Each `llm.StreamChunk` provides multiple ways to access its content. You can access the raw `content` property directly, or use type-specific accessor properties like `.text`, `.image`, `.audio`, etc. for convenience:
+
+<CodeExample file="examples/streams/chunk_content.py" />
+
+The accessor properties return `None` if that content type is not present in the chunk, making it easy to handle different content types safely. You can also use `str(chunk)` to get a string representation across all types of media.
+
+### Converting Streams to Responses
+
+You can convert a stream to a standard response using the `stream.to_response()` method. You can invoke it before or after consuming the stream. Note that calling `to_response` will iterate through all the chunks, exhausting the stream. You won't be able to iterate over it afterwards.
```
willbakst: you must call `to_response` after iterating through the stream since otherwise the response can't be reconstructed as we won't have any info to reconstruct it

willbakst: what if instead of `stream.to_response` we had `llm.Response.from_stream(stream)`? WDYT? Not sure which I like better, but figured I would toss another option out there for discussion

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+<CodeExample file="examples/streams/to_response.py" />
+
+### Monitoring Stream Usage & Cost
+
+Each chunk provides `usage` and `cost` properties that give token usage and cost for that chunk (if available). The stream also keeps track of its running total `usage` and `cost`, as shown below:
+
+<CodeExample file="examples/streams/usage_and_cost.py" />
```
willbakst, line 86: the print for `chunk.usage` has a newline but cost doesn't, which is a little weird? Or is this because if there is usage there will be cost?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+## Error Handling
+
+Error handling in streams is similar to [standard calls](/docs/learn/calls#error-handling). However, it's important to note that errors may occur during iteration rather than at the initial function call:
+
+<CodeExample file="examples/streams/error_handling.py" />
```
willbakst, line 92: nit: drop the newlines between the except and else blocks here?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+If you have a synchronous call definition, you can use the `stream_async()` method to get an async stream:
+
+<TabbedSection>
+<Tab value="Python">
+<CodeExample file="examples/streams/async_stream_method.py" />
+</Tab>
+</TabbedSection>
+
+Note that converting an async stream to a response is an async operation:
```
willbakst: why is this operation async? it doesn't need to be?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+<Tab value="Python">
+<CodeExample file="examples/streams/async_stream_response.py" />
+</Tab>
+</TabbedSection>
+
+### Async Stream Properties and Methods
+
+`llm.AsyncStream` has the same properties and methods as `llm.Stream`, including:
+
+- `finish_reason`, `usage`, and `cost` properties that are populated as the stream is consumed
```
willbakst: a little weird we didn't introduce this earlier for normal string. probably best to introduce this for streams earlier like we do for calls and then just mention everything is the same except for the specific difference?

willbakst: oh i'm now seeing you cover this in the following section. I'm thinking maybe we move async after properties and methods before next steps?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+### Formatting Methods
+
+| Method | Type | Description |
+|--------|------|-------------|
+| `format()` | T | Format the content of this chunk according to a response format parser. This method is only available if the call was created with a response format to generate [structured outputs](/docs/learn/structured_outputs). |
+
+## Stream Properties and Methods
+
+After creating a stream, you can access properties and methods on the `llm.Stream` object itself:
```
willbakst: it's a little weird the header for this has "Methods" but then there are only properties? Where are the methods like `to_response` etc.?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+## Next Steps
+
+By mastering streaming in Mirascope, you can create more responsive and efficient LLM-powered applications that provide real-time feedback to users.
+
+Next, we recommend exploring:
+
+- [Audio Streaming](/docs/learn/audio-streaming) - Learn how to handle audio outputs in streaming responses for speech-based applications
+- [Context](/docs/learn/context) - Automatically maintain conversation history and dependencies across multiple calls
+- [Tools](/docs/learn/tools) - Give LLMs access to custom tools and see how tool calls work in streaming responses
+- [Agents](/docs/learn/agents) - Build autonomous agents that can stream their reasoning and actions in real-time
```
willbakst: nit: again drop "real-time"

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,152 @@
...
+| `usage` | [Usage](/docs/api/responses#usage) \| None | The token usage statistics reflecting all chunks processed so far. Updates as chunks are consumed. |
+| `cost` | [Decimal](https://docs.python.org/3/library/decimal.html) \| None | The cost reflecting all chunks processed so far. Updates as chunks are consumed. |
+
+## Next Steps
+
+By mastering streaming in Mirascope, you can create more responsive and efficient LLM-powered applications that provide real-time feedback to users.
+
+Next, we recommend exploring:
+
+- [Audio Streaming](/docs/learn/audio-streaming) - Learn how to handle audio outputs in streaming responses for speech-based applications
```
willbakst, line 183: as I said in my top level comment, we should distinguish between streaming and real-time, which are distinct things that will have different APIs.

my preference would be to cover all things streaming here and then have a separate realtime doc (which we can reference and add later after v2 initial release when we can focus on it as a follow-on feature / interface)

willbakst: actually upon further inspection, my initial understanding might not work here.

https://platform.openai.com/docs/api-reference/audio

while there are distinct APIs for realtime and streaming audio that we should keep separate, streaming audio is inherently somewhat different from a standard generation. I also don't think images or video support streaming, so we may just not want to include those in the streaming API at all and push for separate APIs specifically focused on Audio and Realtime (still separate).

for example something like `llm.speech` or `llm.tts` and `llm.stt` or something (as well as the `Realtime` api)

willbakst: Ok actually confirming my original understanding, at least the OpenAI completions API had audio streaming, which we support in v1:

https://github.com/Mirascope/mirascope/blob/4a266179830f4edf4663b1430d562d9d1bfa6a2c/mirascope/core/openai/stream.py#L96

docs/content/learn/streams.mdx
teamdandelion: Updated

docs/content/learn/streams.mdx
teamdandelion: I think it's a little more discoverable this way, so I'd default to leaving it as is.

docs/content/learn/streams.mdx
teamdandelion: The preceding newline print is essentially "cleanup" to add a final newline after the streamed content (which is being printed with `end=""`. 

I agree this is a little confusing, I am changing the example to have a cleanup print statement with a comment separately from prepending a newline in the usage print.

docs/content/learn/streams.mdx
teamdandelion: Done. Also simplied down to just one except block (catching llm.MirascopeError)

docs/content/learn/streams.mdx
teamdandelion: Vestigal from when I was writing to_response so it iterates over the stream for you. Removed

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+### Multi-modal stream output
+
+Streams can output every modality of content, via distinct chunk types like `llm.TextChunk`, `llm.ImageChunk`, and so forth. Each chunk has a string `delta`, which is the incremental content in that chunk. The semantics of the delta vary based on chunk type. For text, the delta is the incremental text of that particular chunk. For images, each delta is a base64-encoded string with a progressive render of the image. For audio, the delta is a base64-encoded string of audio data, but there is also a `transcript_delta` with incremental transcripts.
+
+Mirascope also provides `llm.ChunkStart` and `llm.ChunkEnd` chunks, which demarcate the start and end of a particular piece of content. This is helpful to distinguish between e.g. multiple images that are being progressively rendered and streamed in sequence. When iterating over the chunks in a stream, you can use the `type` discriminator field to process each chunk type distinctly. 
+
+<TabbedSection>
+<Tab value="Python">
+<CodeExample file="examples/streams/content_type.py" />
```
willbakst, line 40: nit: you can update the check to:

```python
if chunk.type not in ["chunk_start", "chunk_end"]: ...
```

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,198 @@
...
+</Tab>
+</TabbedSection>
+
+Note that for multimodal streams, the `ChunkStart` and `ChunkEnd` will print as helpful markers showing where content begins and ends, as in the following example:
+
+<TabbedSection>
+<Tab value="Python">
+<CodeExample file="examples/streams/multimodal_print.py" />
```
willbakst, line 48: nit: the end chunk repr is `{...:done}` which feels weird since it's `ChunkEnd` and not `ChunkDone`. I would rather `{...:end}` I think? I'm assuming you chose "done" to mean "done generating" so I get it, but still feels weird.

willbakst, line 48: nit: should we highlight line 20?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,198 @@
...
+</TabbedSection>
+
+
+### Grouped Streaming API
+
+For processing multimodal streams, Mirascope offers a more advanced `groups()` API that organizes related chunks into distinct content pieces. Each group is itself an iterator, and iterates over a sequence of chunks that will combine into one piece of content. 
+
+<TabbedSection>
+<Tab value="Python">
+<CodeExample file="examples/streams/groups.py" />
```
willbakst, line 59: what if instead of `group.final` we just had `group.collect()` which would return the equivalent but also exhaust the group if necessary to fully collect the final response?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,198 @@
...
+Mirascope tries to match the reconstructed `llm.Response` with the corresponding non-streaming response. However, they might not be exactly the same.
+</Warning>
+
+<CodeExample file="examples/streams/to_response.py" />
+
+### Monitoring Stream Usage & Cost
+
+Once a stream is complete, you can check its total usage and cost, as in the example below. These properties will not be available until the stream is exhausted.
+
+<CodeExample file="examples/streams/usage_and_cost.py" />
```
willbakst, line 86: nit: this has `print("")` for trailing newline but earlier examples just have `print()`

we should make this consistent, and I think I prefer `print("")`?

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,198 @@
...
+Either way, you will get an `llm.AsyncStream` that supports async iteration. If your prompt is synchronous, then we recommend using `stream_async` for call-time flexibility.
+
+### Async Properties 
+
+`llm.AsyncStream` has the same properties and methods as `llm.Stream`, including:
+
+- `finish_reason`, `usage`, and `cost` properties that are populated as the stream is consumed
+- The same error handling patterns apply to async streams
+
+The main difference is that you use `async for` instead of `for` to iterate through the chunks. Similarly, you will use `async for` to iterate through groups, and through the chunks within each group. Once the stream is exhausted, `to_response` is still synchronous.
```
willbakst: nit: I wouldn't have thought `to_response` would be async until you mention it, I think you can drop this.

docs/content/learn/streams.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+<TabbedSection>
+<Tab value="Python">
+<CodeExample file="examples/streams/async_decorated.py" />
+</Tab>
+</TabbedSection>
+
+Either way, you will get an `llm.AsyncStream` that supports async iteration. If your prompt is synchronous, then we recommend using `stream_async` for call-time flexibility.
+
+### Async Properties
```
willbakst: I wonder if we need this section at all vs. just mentioning this as a single line at the end of the previous section, along the lines of:

"Note that `llm.AsyncStream` has the same properties and methods as `llm.Stream`. The only difference is that you use `async for` instead of just `for` to iterate through the chunks."

We can also drop the bit about groups since we cover that in the section immediately following

docs/content/learn/streams.mdx
teamdandelion: I like this in principle, in practice it's a type error.
image
image

docs/content/learn/streams.mdx
teamdandelion: I like this. Note that it means collect is async on async stream.

docs/content/learn/streams.mdx
willbakst: ah darn

PR #1101, opened 2025-07-04
feat: document async calls in calls.mdx

teamdandelion:
n.b. disabling ruff rule "B905" in examples so we can use `zip` without strict

### DIFF
--- a/docs/content/learn/calls.mdx
+++ b/docs/content/learn/calls.mdx
@@ -208,6 +208,22 @@ Mirascope defines the following exception types:
 
 <CodeExample file="examples/calls/exceptions.py" />
 
+## Async Calls
+
+Async calls are useful when you need non-blocking operations or want to make multiple LLM calls concurrently. To make a call async, simply define the function as `async` and use the `await` keyword when calling it. Here's an example, using `asyncio` to start the asynchronous event loop and run the main function:
+
+<CodeExample file="examples/calls/async.py" />
+
+If you wish, you can also define the call as usual, and then use the `call_async` method on the call. This can be helpful if you want to use the same call in synchronous and asynchronous settings.
+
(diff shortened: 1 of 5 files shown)


willbakst:
Why do we care not to use `strict`? Isn't it best practice to use it?

docs/content/learn/calls.mdx
```diff
@@ -208,6 +208,22 @@ Mirascope defines the following exception types:
 
 <CodeExample file="examples/calls/exceptions.py" />
 
+## Async Calls
+
+Async calls are useful when you need non-blocking operations or want to make multiple LLM calls concurrently. To make a call async, simply define the function as `async` and use the `await` keyword when calling it. Here's an example, using `asyncio` to start the asynchronous event loop and run the main function:
+
+<CodeExample file="examples/calls/async.py" />
+
+If you wish, you can also define the call as usual, and then use the `call_async` method on the call. This can be helpful if you want to use the same call in synchronous and asynchronous settings.
```
willbakst: nit: I'm wondering if it's worth explaining to people that the only real benefit of marking the function with `async` is when you need the prompt itself to do something asynchronously. Otherwise generally better to just use `call_async` on a sync definition since it's 1:1 equivalent. We can still keep the examples as-is, this would mostly be a note for reference for a user.

teamdandelion: Sg, added

teamdandelion:
> Why do we care not to use `strict`? Isn't it best practice to use it?

I figured leaving it out made the example slightly simpler and focused attention on the parallelism (also matches how we wrote the example in v1). However I'm fine with using strict too. So I changed it and removed the lint exception.

willbakst (approved):

PR #1103, opened 2025-07-04
refactor: wrap all str content in a llm.Text

teamdandelion:
While here: Normalize message content as always a sequence, for consistency, since all user-facing APIs need to promote the str content anyway.


docs/content/learn/messages.mdx
```diff
@@ -117,7 +117,7 @@ _TODO: Replace this with centralized compatibility table (#1031)_
 
 ### Text Content
 
-The most common type of content is text, represented as plain strings:
+The most common type of content is text. It's represented as a `llm.Text` for consistency, but our APIs will automatically wrap plain strings:
```
willbakst: nit: again "a" -> "an" since `llm.Text` starts with "eh"

python/examples/calls/prompts.py
```diff
@@ -14,7 +14,7 @@ def audio_prompt(audio: llm.Audio) -> llm.Content:
 
 def multimodal_prompt(
     audio: llm.Audio, image: llm.Image, video: llm.Video, documents: list[llm.Document]
-) -> list[llm.Content]:
+) -> list[llm.Content | str]:
```
willbakst: nit: for some reason I prefer `str | llm.Content`. not a big deal if you prefer it the way it's currently written

python/mirascope/llm/messages/message.py
```diff
@@ -35,52 +35,85 @@ class Message:
...
 
     Returns:
         A Message with the assistant role.
     """
-    return Message(role="assistant", content=content, name=name)
+
+    return Message(role="assistant", content=_promote(content), name=name)
+
+
+def _promote(input: str | Content | Sequence[str | Content]) -> Sequence[Content]:
```
willbakst: nit: always define a method before it's used (it's easier to read through the file)

also, let's make this more explicit e.g. `_promote_content`

python/mirascope/llm/prompts/decorator.py
```diff
@@ -22,21 +22,23 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> list[Message]: ...
 class ContentPrompt(Protocol[P]):
     """Protocol for a Prompt function that returns a single content part."""
 
-    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Content: ...
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Content | str: ...
```
willbakst: nit: same thing here re: `str | Content`

python/mirascope/llm/prompts/decorator.py
```diff
@@ -22,21 +22,23 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> list[Message]: ...
...
 
 
 class ContentSequencePrompt(Protocol[P]):
     """Protocol for a prompt function that returns a content parts sequence."""
 
-    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Sequence[Content]: ...
+    def __call__(
+        self, *args: P.args, **kwargs: P.kwargs
+    ) -> Sequence[Content | str]: ...
```
willbakst: ditto here and a bunch of other places, so this will be the last comment. if you agree, please change everywhere

willbakst (approved):

PR #1111, opened 2025-07-10
v2: add response.to_tool, stream.to_tool, remove ContextTool

teamdandelion:


python/mirascope/llm/responses/response.py
```diff
@@ -95,6 +96,10 @@ def thinking(self) -> Thinking | None:
         """Returns the first thinking in the response content, if any."""
         raise NotImplementedError()
 
+    def to_tool(self, tool_call: ToolCall) -> Tool:
```
willbakst: I think this is confusing with `to_response` converting `Response`. I think this should just be `tool` with a matching `tools` method that converts a `list[ToolCall]` into `list[Tool]`

python/mirascope/llm/streams/base.py
```diff
@@ -39,3 +41,7 @@ def to_response(self) -> Response:
 
         """
         raise NotImplementedError()
+
+    def to_tool(self, tool_call: ToolCall) -> Tool:
```
willbakst: same here

python/mirascope/llm/responses/response.py
teamdandelion: Ok, I changed it to suit your preferences. 

I'm not 100% sold on this interface for tools, I have another idea that I want to explore which I think will create a path for async tools too. But since this one is at the bottom of a big stack of PRs I'd rather merge it and then explore alternative interfaces in another PR.

willbakst (approved):

PR #1113, opened 2025-07-10
refactor: centralize type variables

teamdandelion:


python/mirascope/llm/prompts/decorator.py
```diff
@@ -73,6 +71,15 @@ async def __call__(
 """
 
 
+PromptT = TypeVar("PromptT", bound=Prompt | AsyncPrompt)
```
willbakst: nit: it's a little weird `PromptT` is here but `ResponseFormatT` is in `types`? Only because of circular ref, but we should try to be consistent?

python/mirascope/llm/types/type_vars.py
```diff
@@ -0,0 +1,46 @@
...
+FormatT = TypeVar("FormatT", bound=object | None, default=None)
+"""Type variable for structured response format types.
+
+This TypeVar represents the type of structured output format that LLM responses
+can be parsed into, such as Pydantic models, dataclasses, or custom classes.
+It can be None for unstructured responses and defaults to None when no specific
+format is required.
+"""
+
+ResponseFormatT = TypeVar("ResponseFormatT", bound=object)
```
willbakst: can you elaborate on the benefit here of separating this from `FormatT`? The naming is a little bit confusing and that fact that one can be `None` isn't directly clear from the naming

teamdandelion: One advantage of this refactor is it reveals the places where we had distinct types sharing the same name. In this case we had the `TypeVar("T", bound=object | None, default=None)` in most cases and `TypeVar("T", bound=object)` in  the response_formatting package particularly. So I didn't really separate them, so much as surface the pre-existing separation. I'll rename it to RequiredFormatT for better clarity.

Maybe they could be unified, but at first glance using RequiredFormatT in the response_formatting package makes sense to me.

python/mirascope/llm/prompts/decorator.py
teamdandelion: Makes sense. I pulled it into types package instead. Arguably we could take a lot of other types from decorator.py (like Prompt, MessagesPrompt, ...) into types, but that would be a much bigger reorg.

teamdandelion:
PTAL. I did another pass of type var extraction, so now every single typevar is exported from within the types package.

python/mirascope/llm/types/type_vars.py
willbakst: To ask it another way, I'm wondering what's stopping someone from having a response format that could be `None` if they wanted? In which case unifying makes the most sense

python/mirascope/llm/types/prompts.py
willbakst, line 1: there's something I really don't like about this structure, so I think we should sync on it

teamdandelion:
Merged in #1121

PR #1116, opened 2025-07-10
v2: add tools examples (except async tools)

teamdandelion:


python/examples/tools/basic.py
```diff
@@ -0,0 +1,31 @@
...
+    """List the available books in the library."""
+    return ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+
+
+@llm.call(model="openai:gpt-4o-mini", tools=[available_books])
+def librarian(genre: str):
+    return f"Recommend an available book in {genre}"
+
+
+def main():
```
willbakst, line 15: nit: what's the benefit of the `main` fn indentation?

if we eventually use these examples in our documentation, why not just remove the indentation?

python/examples/tools/basic.py
```diff
@@ -0,0 +1,31 @@
...
+
+@llm.tool()
+def available_books() -> list[str]:
+    """List the available books in the library."""
+    return ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+
+
+@llm.call(model="openai:gpt-4o-mini", tools=[available_books])
+def librarian(genre: str):
+    return f"Recommend an available book in {genre}"
```
willbakst, line 12: nit: "Recommend a {genre} book that's available to rent"

python/examples/tools/basic_async_stream.py
```diff
@@ -0,0 +1,36 @@
...
+                async for chunk in group:
+                    print(chunk)
+            if group.type == "tool_call":
+                tool_call = await group.collect()
+        if tool_call:
+            tool = stream.tool(tool_call)
+            tool_output = tool.call()
+            stream = librarian.resume_stream_async(stream.to_response(), tool_output)
+        else:
+            break
```
willbakst, line 32: i don't love this pattern. I would prefer `if not tool_call: break` (with a newline) so that the tool call handling isn't indented.

the same thing applies here and elsewhere

python/examples/tools/basic_async_context.py
```diff
@@ -0,0 +1,43 @@
+import asyncio
+from dataclasses import dataclass
+
+from mirascope import llm
+
+
+@dataclass
+class Library:
+    available_books: list[str]
```
willbakst, line 9: nit: just name this "books" as in the books in the library. then `available_books` as a tool returns the list of available books, which is `ctx.deps.books` but would be something else (for example maybe `books` is a `list[Book]` where `Book` as an attribute `available: bool`

python/examples/tools/basic_context.py
```diff
@@ -0,0 +1,42 @@
+from dataclasses import dataclass
+
+from mirascope import llm
+
+
+@dataclass
+class Library:
+    available_books: list[str]
```
willbakst, line 8: same here and elsewhere

python/examples/tools/handoff_tool.py
```diff
@@ -0,0 +1,46 @@
...
+    return ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+
+
+@llm.tool()
+def ask_user_preference(question: str) -> str:
+    """Ask the user for their preference when you need more specific information.
+    Use this when you need to gather additional details from the user to make a better recommendation.
+    """
+    # This tool doesn't directly return data - it signals a handoff to get user input
+    raise RuntimeError("Handoff tool should not be invoked.")
```
willbakst, line 16: hmm, why not just have this tool actually get the user input and have that input injected back into the flow through the tool call output?

python/examples/tools/handoff_tool_context.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@llm.tool(deps_type=Library)
+def available_books(ctx: llm.Context[Library]) -> list[str]:
+    """List the available books in the library."""
+    return ctx.deps.available_books
+
+
+@llm.tool(deps_type=Library)
+def ask_user_preference(ctx: llm.Context[Library], question: str) -> str:
```
willbakst, line 18: same here and elsewhere

python/examples/tools/mix_context_non_context.py
```diff
@@ -0,0 +1,52 @@
...
+    tools=[available_books, hardcoded_extra_book],
+)
+def librarian(ctx: llm.Context[Library], genre: str):
+    return f"Recommend an available book in {genre}"
+
+
+def main():
+    with llm.context(deps=library) as ctx:
+        response: llm.Response[Library] = librarian(ctx, "fantasy")
+        while tool_call := response.tool_call:
```
willbakst, line 38: this technically is potentially incorrect since the LLM may use parallel tool calling and call multiple tools in a single call, in which case this won't handle that

python/examples/tools/multi_tool_stream.py
```diff
@@ -0,0 +1,53 @@
...
+                    print(chunk)
+            if group.type == "tool_call":
+                tool_call = group.collect()
+                tool_calls.append(tool_call)
+        if tool_calls:
+            tools = stream.tools(tool_calls)
+            outputs = [tool.call() for tool in tools]
+            stream = librarian.resume_stream(stream.to_response(), outputs)
+        else:
+            break
```
willbakst, line 49: same thing here and elsewhere

python/examples/tools/multiple_dep_types_context.py
```diff
@@ -0,0 +1,75 @@
...
+class MetaLibrary:
+    library: Library
+    secret_library: SecretLibrary
+
+
+meta_library = MetaLibrary(library=library, secret_library=secret_library)
+
+
+# NOTE / TODO: It would be nice if we could define the tools in terms of their logical dep,
+# and have a composition pattern, rather than needing to redefine the tools in terms of
```
willbakst, line 32: the typing will be way easier if we just always type against context and a single top level dep

python/examples/tools/validation.py
```diff
@@ -0,0 +1,35 @@
...
+    while tool_call := response.tool_call:
+        tool: llm.Tool | None
+        try:
+            tool = response.tool(tool_call)
+        except llm.ToolNotFoundError:
+            print("Tool not found!")
+            exit(1)
+        try:
+            output = tool.call()
+        except Exception:
```
willbakst, line 26: i would rather validation here be explicit rather than a generic exception. what exactly are we validation, and how do we handle the issue?

for now, I would completely exclude anything related to reliability and put everything related to reliability int those examples / docs

python/examples/tools/basic.py
teamdandelion: I did it this way for consistency with the async cases, which need a main function

python/examples/tools/mix_context_non_context.py
teamdandelion: Good point, changed it.

python/examples/tools/handoff_tool.py
teamdandelion: Yeah that's a reasonable question. The goal here is really to show of the pattern of checking which tool is being used and doing some custom logic outside of the tool. However, to fully show off the power of the interface we should capture the output of the tool call in a type safe fashion. I'll write a new example.

teamdandelion:
Merged in #1121

PR #1117, opened 2025-07-10
feat: add examples of the resume API

teamdandelion:
I organized them under streams/ and calls/ examples and show alternative approaches to making a chatbot, including response.messages, context, and the resume API. I think resume API is cleanest and we should make that the default for this kind of case.

### DIFF
--- a/python/examples/calls/chatbot/chatbot_context.py
+++ b/python/examples/calls/chatbot/chatbot_context.py
@@ -1,0 +1,29 @@
+from mirascope import llm
+
+SYSTEM_MESSAGE = "You are an insightful chatbot named Sazed. You prioritize..."
+
+
+@llm.call(model="openai:gpt-4o-mini")
+def sazed(ctx: llm.Context, query: str):
+    return [
+        llm.messages.system(SYSTEM_MESSAGE),
+        *ctx.messages,  # Automatically filters out system message
+        query,
+    ]
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
+    with llm.context() as ctx:
+        response: llm.Response = sazed(ctx, user_input)
+        try:
+            while True:
+                print("[SAZED]: ", response.text)
+                user_input = input("[YOU]: ")
+                response = sazed(ctx, user_input)
+        except KeyboardInterrupt:
+            print("[SAZED]: Goodbye")
+            exit(0)
+
+
+main()
--- a/python/examples/calls/chatbot/chatbot_continuations.py
+++ b/python/examples/calls/chatbot/chatbot_continuations.py
@@ -1,0 +1,27 @@
+from mirascope import llm
+
+SYSTEM_MESSAGE = "You are an insightful chatbot named Sazed. You prioritize..."
+
+
+@llm.call(model="openai:gpt-4o-mini")
+def sazed(query: str):
+    return [
+        llm.messages.system(SYSTEM_MESSAGE),
+        query,
+    ]
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
+    response: llm.Response = sazed(user_input)
+    try:
+        while True:
+            print("[SAZED]: ", response.text)
+            user_input = input("[YOU]: ")
+            response = sazed.resume(response, user_input)
+    except KeyboardInterrupt:
+        print("[SAZED]: Goodbye")
+        exit(0)
+
+
+main()
--- a/python/examples/calls/chatbot/chatbot_response_messages.py
+++ b/python/examples/calls/chatbot/chatbot_response_messages.py
@@ -1,0 +1,29 @@
+from mirascope import llm
+
+SYSTEM_MESSAGE = "You are an insightful chatbot named Sazed. You prioritize..."
+
+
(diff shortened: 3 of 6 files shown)


python/examples/calls/chatbot/chatbot_continuations.py
willbakst, line 1: nit: rename file around the `.resume` naming rather than "continuations"?

python/examples/calls/chatbot/chatbot_continuations.py
```diff
@@ -0,0 +1,27 @@
+from mirascope import llm
+
+SYSTEM_MESSAGE = "You are an insightful chatbot named Sazed. You prioritize..."
```
willbakst, line 3: I don't love the `...` in the example, and I think we should just put this directly in the message rather than refactored?

python/examples/calls/chatbot/chatbot_response_messages.py
```diff
@@ -0,0 +1,29 @@
+from mirascope import llm
+
+SYSTEM_MESSAGE = "You are an insightful chatbot named Sazed. You prioritize..."
+
+
+@llm.call(model="openai:gpt-4o-mini")
+def sazed(messages: list[llm.Message], query: str):
```
willbakst, line 7: not in love with this pattern, and I don't know why anyone would choose this pattern over context or `.resume`? I feel like if we're going to show manual messages, we should also show manually managing the messages array in this example? rather than cleaning it in the prompt?

python/examples/streams/chatbot/chatbot_stream_context.py
```diff
@@ -0,0 +1,32 @@
...
+def sazed(ctx: llm.Context, query: str):
+    return [
+        llm.messages.system(SYSTEM_MESSAGE),
+        *ctx.messages,  # Automatically filters out system message
+        query,
+    ]
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
```
willbakst, line 16: why not put this inside the while true?

python/examples/calls/chatbot/chatbot_context.py
```diff
@@ -0,0 +1,29 @@
...
+        llm.messages.system(SYSTEM_MESSAGE),
+        *ctx.messages,  # Automatically filters out system message
+        query,
+    ]
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
+    with llm.context() as ctx:
+        response: llm.Response = sazed(ctx, user_input)
```
willbakst, line 18: why not put this inside the while true? just have user input, sazed response, loop?

python/examples/streams/chatbot/chatbot_stream_continuations.py
```diff
@@ -0,0 +1,30 @@
...
+@llm.call(model="openai:gpt-4o-mini")
+def sazed(query: str):
+    return [
+        llm.messages.system(SYSTEM_MESSAGE),
+        query,
+    ]
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
```
willbakst, line 15: same thing here with putting this inside the loop to simplify?

python/examples/streams/chatbot/chatbot_stream_messages_prompt.py
```diff
@@ -0,0 +1,33 @@
+from mirascope import llm
+
+SYSTEM_MESSAGE = "You are an insightful chatbot named Sazed. You prioritize..."
+
+
+@llm.call(model="openai:gpt-4o-mini")
+def sazed(messages: list[llm.Message], query: str):
+    non_system_messages = [m for m in messages if m.role != "system"]
```
willbakst, line 8: same thing here as other comment

python/examples/calls/chatbot/chatbot_response_messages.py
teamdandelion: Sg, I'll drop these examples and just show ctx and resume.

python/examples/streams/chatbot/chatbot_stream_context.py
teamdandelion: I can do this for the context examples, but not the continuation (now resume) examples.

python/examples/streams/chatbot/chatbot_stream_continuations.py
teamdandelion: If we move it into the loop, then we need to add an additional conditional to check whether we're in the first invocation (no response to resume) or subsequent ones. I think this way is cleaner. (Example below is written in the response case but applies for streams too.)

However in the context cases we can simplify it as youve suggested.

```python
def main():
    user_input = input("[USER]: ")
    response: llm.Response = sazed(user_input)
    while True:
        print("[SAZED]: ", response.text)
        user_input = input("[USER]: ")
        response = sazed.resume(response, user_input)
```

Versus:

```python
def main():
    response: llm.Response  | None = None
    while True:
        user_input = input("[USER]: ")
        if response:
            response = sazed(user_input)
        else:
            response = sazed.resume(response, user_input)
        print("[SAZED]: ", response.text)
```

teamdandelion:
Merged in #1121

PR #1118, opened 2025-07-11
v2: update structured output interface, add examples

teamdandelion:
I got rid of StructuredStream and AsyncStructuredStream and unified them into Stream and AsyncStream.

The new contract is `stream.format()` is equivalent to `stream.to_response().format()` and attempts to format whatever the current state of the stream is.

The docstring on StructuredStream said that iterating over the stream would directly return partial structured output, but that didn't really make sense — since `.format()` can throw a validation error, it should be something that gets called explicitly, rather than happening while iterating.

Also: cleaned up the interface so with a synchronous prompt, creating an async stream is synchronous. Prior to this PR it's inconsistent (sometimes sync sometimes async)


python/examples/structured_outputs/async_context_stream.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+async def main():
+    library = Library(available_books=["Mistborn", "Dune", "The Name of the Wind"])
+    with llm.context(deps=library) as ctx:
+        stream: llm.AsyncStream[Library, Book] = recommend_book.stream_async(
+            ctx, "fantasy"
+        )
+        async for _ in stream:
+            partial_book: Book = stream.format()
```
willbakst, line 33: the typing here should not be `Book` it should be `Partial[Book]` ideally

python/examples/structured_outputs/async_stream.py
```diff
@@ -0,0 +1,25 @@
...
+
+@llm.call("openai:gpt-4o-mini", response_format=Book)
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book."
+
+
+async def main():
+    stream: llm.AsyncStream[None, Book] = recommend_book.stream_async("fantasy")
+    async for _ in stream:
+        partial_book: Book = stream.format()
```
willbakst, line 21: same here

python/examples/structured_outputs/context_stream.py
```diff
@@ -0,0 +1,34 @@
...
+    The following books are available: {ctx.deps.available_books}
+    """
+
+
+def main():
+    library = Library(available_books=["Mistborn", "Dune", "The Name of the Wind"])
+    with llm.context(deps=library) as ctx:
+        stream: llm.Stream[Library, Book] = recommend_book.stream(ctx, "fantasy")
+        for _ in stream:
+            partial_book: Book = stream.format()
```
willbakst, line 30: same here

python/examples/structured_outputs/stream.py
```diff
@@ -0,0 +1,24 @@
...
+
+@llm.call("openai:gpt-4o-mini", response_format=Book)
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book."
+
+
+def main():
+    stream: llm.Stream[None, Book] = recommend_book.stream("fantasy")
+    for _ in stream:
+        partial_book: Book = stream.format()
```
willbakst, line 20: same here

python/examples/structured_outputs/tools.py
```diff
@@ -0,0 +1,44 @@
...
+    isbn: str
+
+
+@llm.call("openai:gpt-4o-mini", response_format=Book)
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book from the library. Ensure the isbn is accurate."
+
+
+def main():
+    response: llm.Response[None, Book] = recommend_book("fantasy")
```
willbakst, line 34: should the response format here potentially be `None` in the even that there isn't anything to format yet?

python/examples/structured_outputs/validation_call.py
```diff
@@ -0,0 +1,26 @@
...
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book with page count."
+
+
+def main():
+    response: llm.Response[None, Book] = recommend_book("fantasy")
+    try:
+        book: Book = response.format()
+        print(book)
+    except ValidationError as e:
```
willbakst, line 22: as I said in a separate PR, I would likely leave all reliability stuff to the reliability examples and docs

python/examples/structured_outputs/validation_stream.py
```diff
@@ -0,0 +1,28 @@
...
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book."
+
+
+def main():
+    stream: llm.Stream[None, Book] = recommend_book.stream("fantasy")
+
+    for _ in stream:
+        try:
+            partial_book: Book = stream.format()
```
willbakst, line 22: same here

python/mirascope/llm/calls/structured_call.py
willbakst, line 1: if we're merging structured streams and streams should we not also merge structured calls and calls?

python/mirascope/llm/streams/base.py
```diff
@@ -53,6 +53,25 @@ def to_response(self) -> Response[DepsT, FormatT]:
         """
         raise NotImplementedError()
 
+    def format(self) -> FormatT:
+        """Format the already-streamed content using the response format parser.
+
+        This method is only available if the call was created with a FormatT.
```
willbakst, line 59: nit: it is technically always available (it just will return `None` if there's no format)

python/mirascope/llm/calls/structured_call.py
teamdandelion: Done (elsewhere)

python/examples/structured_outputs/validation_stream.py
teamdandelion: Removed

python/examples/structured_outputs/validation_call.py
teamdandelion: Removed

python/mirascope/llm/streams/base.py
teamdandelion: Updated (here and in Response where I copied this from)

python/examples/structured_outputs/tools.py
teamdandelion: Hmm, yeah maybe? Let's talk about the partial response handling more offline.

python/examples/structured_outputs/async_context_stream.py
teamdandelion: I like this idea, but where do we get Partial from? It's not in typing or typing_extensions.

willbakst: It's a custom thing we implemented in v1 to actually update a Pydantic base model to allow none so it wouldn't fail on validating a partial structured output.

We'll have to implement something similar since Python doesn't have this functionality natively

teamdandelion:
Merged in #1121

PR #1119, opened 2025-07-11
v2: add agent examples

teamdandelion:


python/examples/agents/agent.py
```diff
@@ -0,0 +1,22 @@
+from mirascope import llm
+
+
+@llm.agent(model="openai:gpt-4o-mini")
+def sazed(ctx: llm.Context):
+    return "You are an insightful and helpful agent named Sazed. You prioritize..."
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
```
willbakst, line 10: again here why not put this in the while loop?

python/examples/agents/agent.py
```diff
@@ -0,0 +1,22 @@
+from mirascope import llm
+
+
+@llm.agent(model="openai:gpt-4o-mini")
+def sazed(ctx: llm.Context):
+    return "You are an insightful and helpful agent named Sazed. You prioritize..."
```
willbakst, line 6: the `...` here makes this somewhat not runnable?

python/examples/agents/agent.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def main():
+    user_input = input("What would you like to chat with Sazed about?")
+    response: llm.Response = sazed(user_input)
+    try:
+        while True:
+            print("[SAZED]: ", response.text)
+            user_input = input("[YOU]: ")
+            response = sazed(user_input)
```
willbakst, line 16: shouldn't this be in a context block? or does each call to the agent lose context from the previous run unless run inside a context block? my feeling is context should be a required kwarg

python/examples/agents/agent_context_tools.py
```diff
@@ -0,0 +1,36 @@
...
+
+@llm.agent(model="openai:gpt-4o-mini", deps_type=Knowledge, tools=[consult_knowledge])
+def sazed(ctx: llm.Context[Knowledge]):
+    return "You are an insightful and helpful agent named Sazed. You prioritize..."
+
+
+def main():
+    knowledge = Knowledge(repository="...")
+    with llm.context(deps=knowledge) as ctx:
+        user_input = input("What would you like to chat with Sazed about?")
```
willbakst, line 24: gonna be the same thing here and everywhere around pushing the initial user input inside the loop

python/examples/agents/agent_structured_validation.py
```diff
@@ -0,0 +1,45 @@
...
+    user_input = input("What would you like to chat with Sazed about? ")
+    response: llm.Response[None, AgentResponse] = sazed(user_input)
+    try:
+        while True:
+            try:
+                structured_response: AgentResponse = response.format()
+                print(
+                    f"[SAZED]: {structured_response.message} (Confidence: {structured_response.confidence}%)"
+                )
+            except ValidationError as e:
```
willbakst, line 34: same thing here re: reliability

python/examples/agents/agent_validation.py
willbakst, line 1: same thing here re: reliability

python/examples/agents/agent.py
teamdandelion: Good catch, done

python/examples/agents/agent.py
teamdandelion: Hmm good point. We _could_ make it use a consistent auto-generated default context, however I like the idea of requiring it explicitly instead.

teamdandelion:
Merged in #1121

PR #1121, opened 2025-07-12
Merge of PRs 1113-1119

teamdandelion:
This bundles together PRs #1113, #1114, #1115, #1116, #1117, #1118, #1119 as well as addressing all review feedback therein.


python/examples/calls/chatbot/chatbot_context.py
```diff
@@ -0,0 +1,23 @@
...
+        query,
+    ]
+
+
+def main():
+    with llm.context() as ctx:
+        user_input = input("[USER]: ")
+        response: llm.Response = sazed(ctx, user_input)
+        while True:
+            print("[SAZED]: ", response.text)
```
willbakst: Shouldn't we just print the repr here?

python/examples/calls/chatbot/chatbot_resume.py
```diff
@@ -0,0 +1,21 @@
...
+        llm.messages.system("You are an insightful and helpful chatbot named Sazed."),
+        query,
+    ]
+
+
+def main():
+    user_input = input("[USER]: ")
+    response: llm.Response = sazed(user_input)
+    while True:
+        print("[SAZED]: ", response.text)
```
willbakst: Same here re: repr

python/examples/calls/chatbot/chatbot_context.py
```diff
@@ -0,0 +1,23 @@
...
+    return [
+        llm.messages.system("You are an insightful and helpful chatbot named Sazed."),
+        *ctx.messages,  # Automatically filters out system message
+        query,
+    ]
+
+
+def main():
+    with llm.context() as ctx:
+        user_input = input("[USER]: ")
```
willbakst: push inside the while loop?

python/examples/structured_outputs/async_context_stream.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+async def main():
+    library = Library(available_books=["Mistborn", "Dune", "The Name of the Wind"])
+    with llm.context(deps=library) as ctx:
+        stream: llm.AsyncStream[Library, Book] = recommend_book.stream_async(
+            ctx, "fantasy"
+        )
+        async for _ in stream:
+            partial_book: Book = stream.format()
```
willbakst, line 33: Is the plan to address the discussion around partial in a follow-up PR?

python/examples/structured_outputs/tools.py
```diff
@@ -0,0 +1,44 @@
...
+    return f"Recommend a {genre} book from the library. Ensure the isbn is accurate."
+
+
+def main():
+    response: llm.Response[None, Book] = recommend_book("fantasy")
+    while tool_call := response.tool_call:
+        output = response.tool(tool_call).call()
+        response = recommend_book.resume(response, output)
+    # Now that we have text response, format it
+    # TODO: Discuss how we want to implement response.format
```
willbakst, line 39: Do we want to discuss here through comments or separately?

python/examples/tools/basic.py
```diff
@@ -0,0 +1,31 @@
...
+        print(f"Tool call: {tool_call.name}")
+        # Tool call: available_books
+        tool = response.tool(tool_call)
+        output = tool.call()
+        print(f"Tool returned: {output.value}")
+        # Tool returned: ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+        response = librarian.resume(response, output)
+
+    print(response.text)
+    "I recommend Mistborn, by Brandon Sanderson..."
```
willbakst: Should be commented?

python/examples/tools/basic_async.py
```diff
@@ -0,0 +1,33 @@
...
+        print(f"Tool call: {tool_call.name}")
+        # Tool call: available_books
+        tool = response.tool(tool_call)
+        output = tool.call()
+        print(f"Tool returned: {output.value}")
+        # Tool returned: ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+        response = await librarian.resume_async(response, output)
+
+    print(response.text)
+    "I recommend Mistborn, by Brandon Sanderson..."
```
willbakst: Same here and elsewhere

python/examples/tools/basic_async_context.py
```diff
@@ -0,0 +1,43 @@
...
+        while tool_call := response.tool_call:
+            print(f"Tool call: {tool_call.name}")
+            # Tool call: available_books
+            tool = response.tool(tool_call)
+            output = tool.call()
+            print(f"Tool returned: {output.value}")
+            # Tool returned: ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+            response = await librarian.resume_async(response, output)
+
+        print(response.text)
```
willbakst: Shouldn't we print the repr here? And elsewhere? What's the recommendation we push for here?

python/examples/tools/basic_context_stream.py
```diff
@@ -0,0 +1,45 @@
...
+        stream: llm.Stream[Library] = librarian.stream(ctx, "fantasy")
+        while True:
+            tool_call: llm.ToolCall | None = None
+            for group in stream.groups():
+                if group.type == "text":
+                    for chunk in group:
+                        print(chunk)
+                if group.type == "tool_call":
+                    tool_call = group.collect()
+            if tool_call:
```
willbakst, line 36: Same here re: indentation

python/examples/tools/basic_context.py
```diff
@@ -0,0 +1,42 @@
...
+            print(f"Tool call: {tool_call.name}")
+            # Tool call: available_books
+            tool = response.tool(tool_call)
+            output = tool.call()
+            print(f"Tool returned: {output.value}")
+            # Tool returned: ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+            response = librarian.resume(response, output)
+
+        print(response.text)
+        "I recommend Mistborn, by Brandon Sanderson..."
```
willbakst: Noting I haven't commented on all of these that need fixing

python/examples/tools/multi_tool.py
```diff
@@ -0,0 +1,44 @@
...
+    elif genre == "scifi":
+        return ["Dune", "Foundation", "The Player of Games"]
+    elif genre == "philosophy":
+        return [
+            "Gödel, Escher, Bach",
+            "Man's Search for Meaning",
+            "The Consolations of Philosophy",
+        ]
+    else:
+        raise KeyError()
```
willbakst: There's two ways to handle tools with an error like this. One would be for the tool to return a string of the error so that the LLM can see the error and resolve on the following loop. The other would be to throw an error and handle it at the user level. Since we're not handling the error here I would opt for the first in this example

python/examples/tools/multi_tool_stream.py
```diff
@@ -0,0 +1,52 @@
...
+    elif genre == "scifi":
+        return ["Dune", "Foundation", "The Player of Games"]
+    elif genre == "philosophy":
+        return [
+            "Gödel, Escher, Bach",
+            "Man's Search for Meaning",
+            "The Consolations of Philosophy",
+        ]
+    else:
+        raise KeyError()
```
willbakst: Same here and elsewhere

python/examples/tools/multiple_dep_types_context.py
```diff
@@ -0,0 +1,75 @@
...
+@dataclass
+class MetaLibrary:
+    library: Library
+    secret_library: SecretLibrary
+
+
+meta_library = MetaLibrary(library=library, secret_library=secret_library)
+
+
+# NOTE / TODO: It would be nice if we could define the tools in terms of their logical dep,
```
willbakst: I think the typing on this is either not possible or such a nightmare that it's not worth it

python/examples/calls/chatbot/chatbot_context.py
teamdandelion: Good call — that is a better pattern especially in the case where other output modalities get generated (the repr will show there was something so it doesn't get lost entirely)

python/examples/structured_outputs/async_context_stream.py
teamdandelion: Yeah, let's explore this separately from this PR.

python/examples/structured_outputs/tools.py
teamdandelion: Let's have a separate conversation about this (this PR is already large enough)

python/examples/tools/basic_async_context.py
teamdandelion: Good call, updating to print the repr (here and elsewhere)

python/examples/tools/multi_tool.py
teamdandelion: Makes sense, updated.

python/examples/tools/multiple_dep_types_context.py
teamdandelion: Ack. Removed the comment, though I'll keep letting this simmer in the back of my mind and see if I come up with anything.

python/examples/tools/basic.py
```diff
@@ -23,8 +23,8 @@ def main():
         # Tool returned: ["Mistborn", "Gödel, Escher, Bach", "Dune"]
         response = librarian.resume(response, output)
 
-    print(response.text)
-    "I recommend Mistborn, by Brandon Sanderson..."
+    print(response)
+    # "I recommend Mistborn, by Brandon Sanderson..."
```
willbakst, line 27: nit: we should do `# > ...` so it looks like console output, here and elsewhere

willbakst (approved):

python/examples/tools/basic.py
teamdandelion: Ack - since https://github.com/Mirascope/mirascope/pull/1122 adds a lot more examples in this genre, I am going to fix these all in that PR consistently.

python/examples/tools/basic_context_stream.py
teamdandelion: Ack. Since I already created https://github.com/Mirascope/mirascope/pull/1122 which adds a lot more examples and reworks how tool calling happens, I am going to merge this as-is and fixup all the examples consistently there.

PR #1122, opened 2025-07-14
Rework tools interface, add support for async tools

teamdandelion:
## Reworks tools interface. 
- Now `@llm.tool()` returns a `Tool` (which is approximately a renamed ToolDef) that doesn't have any arguments wrapped in.
- `tool.call` takes a `ToolCall` and returns a `ToolOutput`, which is intuitive
- Invoking tool calls is now done via interfaces on the Call (via Toolkit), not the Response - as in `librarian.tools.call(tool_call)`, or (if you want to check the tool itself) `librarian.get_tool(tool_call)`. I think this makes more sene than having it on the response. The tools are really part of the call specification, not the response. And it makes more sense when using the agent decorator that the response object is not trying to give you responsibility for tools, which you have handed off anyway.
- This also sets things up better for async tools, as you can have a AsyncCall return AsyncTools and there is no need to make Response generic on tool, or split AsyncResponse, or anything like that.
- We no longer have a wrapper class that wraps the context and tool call and tool definition in one, so you need to use patterns like `librarian.tools.call(ctx, tool_call)` rather than just `tool.call()`. The examples are still pretty clean, I don't see this as a bi issue.

## Add Async tools
- Now we can use `@llm.tool()` on an async function and get an AsyncTool or AsyncContextTool
- Reworked all the async tools examples to show async tools, including selective handling with async tools.
- Sync calls can use async tools (since you are responsible for calling them), but if an agent has any async tools, it must be an async agent

## Other cleanups
Fix the call interface so that generating async streams is never async (unnecessary since the AsyncStream encapsulates the async behavior)


python/examples/streams/async_decorated.py
```diff
@@ -9,7 +9,7 @@ async def recommend_book(genre: str) -> str:
 
 
 async def main():
-    stream = await recommend_book.stream("fantasy")
+    stream = recommend_book.stream("fantasy")
```
willbakst: I believe we need to ensure async `.stream` is async awaitable because some implementations will require we do async operations before we can return the stream generator

python/examples/tools/basic_async_stream.py
```diff
@@ -24,11 +25,11 @@ async def main():
                     print(chunk)
             if group.type == "tool_call":
                 tool_call = await group.collect()
-        if not tool_call:
+        if tool_call:
```
willbakst: nit: again here do `if not tool_call: break` and then remove indentation for the rest

python/examples/tools/basic_context_stream.py
```diff
@@ -33,13 +33,12 @@ def main():
                         print(chunk)
                 if group.type == "tool_call":
                     tool_call = group.collect()
-            if tool_call:
-                tool = stream.tool(tool_call)
-                tool_output = tool.call()
-                stream = librarian.resume_stream(stream, tool_output)
-            else:
+            if not tool_call:
```
willbakst, line 36: this is better and how I want the other example (resolve this, it's just a note)

python/examples/tools/mixed_sync_async.py
```diff
@@ -0,0 +1,39 @@
...
+    while tool_call := response.tool_call:
+        print(f"Tool call: {tool_call.name}")
+        # Tool call: available_books
+        output = await librarian.call_tool(tool_call)
+        print(f"Tool returned: {output.value}")
+        # Tool returned: ["Mistborn", "Gödel, Escher, Bach", "Dune"]
+        response = await librarian.resume_async(response, output)
+
+    print(response)
+    # "I recommend Mistborn, by Brandon Sanderson..."
```
willbakst: nit: here and elsewhere do `# > ...` so it's more clear it's a console output

python/examples/tools/selective_handling_mixed_sync_async.py
```diff
@@ -0,0 +1,67 @@
...
+            # This works even though we used available_books.to_async(), due to __eq__ override
+            tool == available_books
+        ):
+            output = tool.call(tool_call)
+            books: list[str] = output.value
+            print(f"Available books: {books}")
+            response = await librarian.resume(response, output)
+        else:
+            # This will never happen
+            raise ValueError(f"Unexpected tool: {tool}")
```
willbakst: I wonder if we could type `librarian.get_tool` such that a final `else` could capture all tools (i.e. first `if` catches `reserve_book` such that `else` catches `available_books`).

regardless, we can also safely drop this else block. the only error that would occur would be on `get_tool` if there's a `ToolNotFoundError` or something

same here and elsewhere

python/mirascope/llm/calls/async_call.py
```diff
@@ -28,40 +31,72 @@ async def call_async(
...
 
-    async def stream(
-        self, *args: P.args, **kwargs: P.kwargs
-    ) -> AsyncStream[None, FormatT]:
+    def stream(self, *args: P.args, **kwargs: P.kwargs) -> AsyncStream[None, FormatT]:
         """Generates a streaming response using the LLM asynchronously."""
         raise NotImplementedError()
 
-    async def stream_async(
+    def stream_async(
```
willbakst: as with earlier comment I think we need to keep these async so we can call await on the clients

python/mirascope/llm/calls/async_call.py
```diff
@@ -28,40 +31,72 @@ async def call_async(
...
+        self,
+        output: Response[None, FormatT] | BaseStream[None, FormatT],
+        content: UserContent | Sequence[UserContent],
     ) -> AsyncStream[None, FormatT]:
         """Generate a new stream by continuing from a previous output, plus new user content."""
         raise NotImplementedError()
 
-    async def resume_stream_async(
-        self, output: Response[None, FormatT] | BaseStream[None, FormatT], content: UserContent | Sequence[UserContent]
+    def resume_stream_async(
```
willbakst: same for these and elsewhere

python/mirascope/llm/clients/anthropic.py
```diff
@@ -167,7 +167,7 @@ def stream_async(
         *,
         model: ANTHROPIC_REGISTERED_LLMS,
         messages: Sequence[AnthropicMessage],
-        tools: Sequence[ToolDef] | None = None,
+        tools: Sequence[Tool] | None = None,
```
willbakst: wouldn't this as well as any of the other async methods require `AsyncTool`? same here and elsewhere

python/mirascope/llm/clients/base.py
```diff
@@ -126,7 +126,7 @@ async def call_async(
         *,
         model: LLMT,
         messages: Sequence[ProviderMessageT],
-        tools: Sequence[ToolDef] | None = None,
+        tools: Sequence[Tool] | None = None,
```
willbakst: like here for example re: `AsyncTool`

python/mirascope/llm/models/base.py
```diff
@@ -122,7 +122,7 @@ async def call_async(
         *,
         ctx: None = None,
         messages: Sequence[ProviderMessageT],
-        tools: Sequence[ToolDef] | None = None,
+        tools: Sequence[Tool] | None = None,
```
willbakst: ditto here and elsewhere

python/mirascope/llm/tools/context_tool.py
```diff
@@ -0,0 +1,39 @@
...
+    fn: Callable[Concatenate[Context[DepsT], P], JsonableCovariantT]
+    """The function that implements the tool's functionality."""
+
+    def call(
+        self, ctx: Context[DepsT], call: ToolCall
+    ) -> ToolOutput[JsonableCovariantT]:
+        """Call the tool using an LLM-provided ToolCall."""
+        raise NotImplementedError()
+
+    def to_async(self) -> AsyncContextTool[P, JsonableCovariantT, DepsT]:
```
willbakst: we will need to be very careful with how we implement this to ensure a sync process can in fact be run async safely. we may also want the reverse with async -> sync? this would be running it on a thread to avoid spinning up another event loop which would fail

python/mirascope/llm/tools/tool.py
```diff
@@ -1,49 +1,33 @@
-"""Tool interface for LLM interactions.
+"""The `ToolDef` class for defining tools that LLMs can request be called."""
```
willbakst: nit: `Tool` not `ToolDef` given change. make sure you catch this in other places I may have missed

python/mirascope/llm/agents/decorator.py
```diff
@@ -71,7 +72,7 @@ def agent(
     model: ANTHROPIC_REGISTERED_LLMS,
     *,
     deps_type: type[DepsT] = NoneType,
-    tools: Sequence[ToolDef | ContextToolDef[..., Any, DepsT]] | None = None,
+    tools: Sequence[Tool | ContextTool[..., Jsonable, DepsT]] | None = None,
```
willbakst: don't we need to update the agent decorator with `Tool` vs. `AsyncTool` the same as for call?

python/examples/streams/async_decorated.py
teamdandelion: Updated here and across codebase

python/examples/tools/selective_handling_mixed_sync_async.py
teamdandelion: I'm doubtful about getting that typing to work, however fine dropping this block. Done.

python/mirascope/llm/tools/tool.py
teamdandelion: Fixed via global find+replace.

python/mirascope/llm/clients/anthropic.py
teamdandelion: Done

python/mirascope/llm/agents/decorator.py
teamdandelion: Done

python/mirascope/llm/tools/context_tool.py
teamdandelion: Ack. I'm not sure about the async->sync conversion; it feels a little safer to guide folks this way.

python/examples/agents/agent_async_context_stream.py
```diff
@@ -25,7 +25,7 @@ async def main():
     with llm.context(deps=knowledge) as ctx:
         while True:
             user_input = input("[USER]: ")
-            stream = sazed.stream_async(user_input, ctx=ctx)
+            stream = await sazed.stream_async(user_input, ctx=ctx)
```
willbakst, line 28: note: i'm realizing that with the v2 interface this may not actually be required. depends on where we call the client to get the stream, which I guess will depend on how we implement streaming for individual providers. if we're just constructing the stream object, which then doesn't actually call the API until the iterator begins / opens, then this doesn't necessarily have to require await.

however, in the event that we want to do something inside of `.stream_async` that is async for whatever reason, I think making this async and requiring await here is likely the best move anyway

python/examples/agents/agent_async_context_tools.py
```diff
@@ -0,0 +1,36 @@
...
+
+@llm.tool(deps_type=Knowledge)
+async def consult_knowledge(ctx: llm.Context[Knowledge], subject: str) -> str:
+    """Consult knowledge about a specific subject."""
+    await asyncio.sleep(0.1)  # Simulate fetching from database
+    return f"I have found comprehensive information about {subject} from {ctx.deps.repository}."
+
+
+@llm.agent(model="openai:gpt-4o-mini", deps_type=Knowledge, tools=[consult_knowledge])
+def sazed(ctx: llm.Context[Knowledge]):
```
willbakst, line 20: shouldn't this be async? or did you intend to use `.call_async` below since the prompt doesn't have to be async?

note also that part of me wants all of our async examples to use async rather than the async methods even though we recommend using async methods. there's something about `async def` that makes it more clear the example is async than looking for a `.call_async`. WDYT?

python/examples/agents/agent_async_context_tools_stream.py
```diff
@@ -0,0 +1,35 @@
...
+
+@llm.tool(deps_type=Knowledge)
+async def consult_knowledge(ctx: llm.Context[Knowledge], subject: str) -> str:
+    """Consult knowledge about a specific subject."""
+    await asyncio.sleep(0.1)  # Simulate fetching from database
+    return f"I have found comprehensive information about {subject} from {ctx.deps.repository}."
+
+
+@llm.agent(model="openai:gpt-4o-mini", deps_type=Knowledge, tools=[consult_knowledge])
+def sazed(ctx: llm.Context[Knowledge]):
```
willbakst, line 20: same here

python/examples/agents/agent_async_tools.py
```diff
@@ -0,0 +1,27 @@
...
+
+@llm.tool()
+async def consult_knowledge(subject: str) -> str:
+    """Consult knowledge about a specific subject."""
+    await asyncio.sleep(0.1)  # Simulate fetching from database
+    return f"I have found comprehensive information about {subject}."
+
+
+@llm.agent(model="openai:gpt-4o-mini", tools=[consult_knowledge])
+def sazed(ctx: llm.Context):
```
willbakst, line 14: ditto

python/examples/agents/agent_async_tools_stream.py
```diff
@@ -0,0 +1,28 @@
...
+
+@llm.tool()
+async def consult_knowledge(subject: str) -> str:
+    """Consult knowledge about a specific subject."""
+    await asyncio.sleep(0.1)  # Simulate fetching from database
+    return f"I have found comprehensive information about {subject}."
+
+
+@llm.agent(model="openai:gpt-4o-mini", tools=[consult_knowledge])
+def sazed(ctx: llm.Context):
```
willbakst, line 14: ditto

python/examples/agents/agent_mix_sync_async_tools.py
```diff
@@ -0,0 +1,33 @@
...
+
+@llm.tool()
+async def consult_knowledge(subject: str) -> str:
+    """Consult knowledge about a specific subject."""
+    await asyncio.sleep(0.1)  # Simulate fetching from database
+    return f"I have found comprehensive information about {subject}."
+
+
+@llm.agent(model="openai:gpt-4o-mini", tools=[get_basic_info.to_async(), consult_knowledge])
+def sazed(ctx: llm.Context):
```
willbakst, line 20: ditto

python/examples/structured_outputs/async_context_stream.py
```diff
@@ -26,7 +26,7 @@ def recommend_book(ctx: llm.Context[Library], genre: str):
 async def main():
     library = Library(available_books=["Mistborn", "Dune", "The Name of the Wind"])
     with llm.context(deps=library) as ctx:
-        stream: llm.AsyncStream[Library, Book] = recommend_book.stream_async(
+        stream: llm.AsyncStream[Library, Book] = await recommend_book.stream_async(
```
willbakst, line 29: This example reads `async_context_stream.py` but shouldn't it be `async_context_structured_stream` if we have `FormatT` is `Book`?

I feel like even if we removed the structured variants at the typing level, they are still distinct paths

python/examples/tools/multi_tool_async_stream.py
```diff
@@ -35,7 +35,7 @@ def librarian():
 
 
 async def main():
-    stream: llm.AsyncStream = librarian.stream()
+    stream: llm.AsyncStream = await librarian.stream()
```
willbakst, line 39: the `librarian` in this example isn't async, so we should either mark it async or use `.stream_async`. See other comment for which one

python/mirascope/llm/tools/context_tool.py
willbakst: async->sync is just checking if there's an event loop and if so running it as a subprocess on a new thread. my only thing here is to question what someone would do if they want to have a sync def and tools except one tool has to use an async API. right now the answer is make everything async, but I feel like calling `.to_sync` on just that tool would be a better interface?

python/examples/agents/agent_async_context_tools.py
teamdandelion: In this case the prompt doesn't have to be async, but the tool is async, so we get an async Agent. So we do `await sazed` because it's an async agent even though it's not really obvious, you have to track the fact that it's an async tool.

I have been thinking the same thing about changing all the async examples to use `async def` prompt functions even though they don't have to. I think it is more clear and more pythonic overall. And then we can relegate the "you can make a sync prompt and still use call_async on it" to a Info callout (with examples).

If we wanted to enforce this even more at the type level, we could require that async tools pair with async prompts, which will make things less ambiguous in the "sync prompt but async Agent/Call because there was an async tool".

python/examples/structured_outputs/async_context_stream.py
teamdandelion: Well... it's under the structured_outputs directory so the fact that it's a structured example is implied from that context. I could rename them though, wdyt?

python/examples/tools/multi_tool_async_stream.py
teamdandelion: Yeah, it's another case where we're getting an async call because we used an async tool.

python/examples/agents/agent_async_context_stream.py
teamdandelion: I don't see much of a downside to having to `await` generating an async stream. I suggest we keep it this way and maybe after we've implemented these methods, if we see that they are all sync, we can consider switching it back. 
I suspect that if we really want to write it sync we can (because we could always "push off" async setup into the async stream) but it may or may not be natural

python/mirascope/llm/tools/context_tool.py
teamdandelion: No longer needed with the interface refactor.

python/examples/agents/agent_async_context_tools.py
teamdandelion: The async tools are no longer an issue. I intend to audit the async examples in a separate PR.

python/examples/agents/agent_async_context_tools_stream.py
teamdandelion: Planning to audit the async example pattern separately from this PR.

python/examples/agents/agent_async_tools_stream.py
teamdandelion: Planning to audit the async example pattern separately from this PR.

python/examples/agents/agent_async_tools.py
teamdandelion: Planning to audit the async example pattern separately from this PR.

python/examples/agents/agent_mix_sync_async_tools.py
teamdandelion: Planning to audit the async example pattern separately from this PR.

python/examples/tools/multi_tool_async_stream.py
teamdandelion: Planning to audit the async example pattern separately from this PR.

python/examples/tools/async_tools.py
```diff
@@ -16,17 +16,17 @@ def librarian(genre: str):
 
 
 async def main():
-    response: llm.Response = await librarian("fantasy")
+    response: llm.Response = librarian("fantasy")
```
willbakst: is this intentionally sync?

python/examples/tools/async_tools_context.py
```diff
@@ -26,17 +26,17 @@ def librarian(ctx: llm.Context[Library], genre: str):
 
 async def main():
     with llm.context(deps=library) as ctx:
```
willbakst, line 28: i though we were dropping context as a context manager? or are we doing that separately?

python/examples/tools/async_tools_context.py
```diff
@@ -26,17 +26,17 @@ def librarian(ctx: llm.Context[Library], genre: str):
...
-        response: llm.Response[Library] = await librarian.call(ctx, "fantasy")
+        response: llm.Response[Library] = librarian.call(ctx, "fantasy")
```
willbakst: same here and elsewhere, is this intentionally sync?

python/examples/tools/async_tools_context_stream.py
```diff
@@ -26,18 +26,18 @@ def librarian(ctx: llm.Context[Library], genre: str):
 
 async def main():
     with llm.context(deps=library) as ctx:
-        stream: llm.AsyncStream[Library] = librarian.stream(ctx, "fantasy")
+        stream: llm.Stream[Library] = librarian.stream(ctx, "fantasy")
```
willbakst: ok last comment on this around if the sync is intentional

python/examples/tools/multi_tool.py
```diff
@@ -32,7 +35,7 @@ def librarian():
 def main():
     response: llm.Response = librarian()
     while tool_calls := response.tool_calls:
-        outputs = librarian.call_tools(tool_calls)
+        outputs = [librarian.tools.call(tc) for tc in tool_calls]
```
willbakst: i'm against short variable names like `tc` here just use `tool_call`

python/examples/tools/multi_tool_async.py
```diff
@@ -35,13 +35,13 @@ def librarian():
 
 
 async def main():
-    response: llm.Response = await librarian()
+    response: llm.Response = librarian()
     while tool_calls := response.tool_calls:
-        output = await librarian.call_tools(tool_calls)
-        response = await librarian.resume_async(response, output)
+        outputs = await asyncio.gather(*[librarian.tools.call(tc) for tc in tool_calls])
```
willbakst: same here re: `tc` -> `tool_call`

python/examples/tools/selective_handling_async.py
```diff
@@ -35,23 +35,23 @@ def librarian():
 
 
 async def main():
-    response: llm.Response = await librarian()
+    response: llm.Response = librarian()
```
willbakst: one more note if intentional here, i'm realizing my preference would be for these async examples to just be fully async

python/examples/tools/selective_handling_context.py
```diff
@@ -46,7 +46,7 @@ def main():
         response: llm.Response[Library] = librarian(ctx)
         while tool_call := response.tool_call:
             print(f"Tool call: {tool_call.name}")
-            tool = librarian.get_tool(tool_call)
+            tool = librarian.tools.get(tool_call)
 
             if tool == reserve_book:
```
willbakst: needs to use `.defines`?

python/examples/tools/selective_handling_mixed_sync_async.py
```diff
@@ -34,31 +34,25 @@ def librarian():
...
             print(f"   Book: {reservation.title}")
 
-            response = await librarian.resume(response, output)
-        elif (
-            # This works even though we used available_books.to_async(), due to __eq__ override
-            tool == available_books
-        ):
-            output = tool.call(tool_call)
+            response = librarian.resume(response, output)
+        elif tool == available_books:
```
willbakst, line 51: same here and elsewhere

python/mirascope/llm/calls/base_call.py
```diff
@@ -1,26 +1,24 @@
...
-class BaseCall(Generic[P, PromptT, FormatT], ABC):
+class BaseCall(Generic[P, PromptT, ToolT, FormatT], ABC):
     """A base class for generating responses using LLMs."""
 
     model: LLM
     """The LLM model used for generating responses."""
 
-    tools: Sequence[Tool[..., Jsonable]] | None
-    """The tools to be used with the LLM."""
+    tools: Toolkit[ToolT]
```
willbakst: didn't we discuss having this be `toolkit` instead of `tools`?

also, missing docstring?

python/mirascope/llm/calls/base_context_call.py
```diff
@@ -1,27 +1,25 @@
...
-class BaseContextCall(Generic[P, PromptT, DepsT, FormatT], ABC):
+class BaseContextCall(Generic[P, PromptT, ContextToolT, DepsT, FormatT], ABC):
     """A base class for generating responses with context using LLMs."""
 
     model: LLM
     """The LLM model used for generating responses."""
 
-    tools: Sequence[ContextTool[..., Jsonable, DepsT]] | None
-    """The tools to be used with the LLM."""
+    tools: ContextToolkit[ContextToolT, DepsT]
```
willbakst: ditto

python/mirascope/llm/calls/base_context_call.py
```diff
@@ -1,27 +1,25 @@
...
 from ..response_formatting import FormatT
-from ..tools import ContextTool
-from ..types import Jsonable, P
+from ..tools import ContextToolkit, ContextToolT
+from ..types import P
 
 
 @dataclass
-class BaseContextCall(Generic[P, PromptT, DepsT, FormatT], ABC):
+class BaseContextCall(Generic[P, PromptT, ContextToolT, DepsT, FormatT], ABC):
```
willbakst: curious what happened to merging on `BaseCall`? Was it because of `ContextToolKit`? Cause we could technically make that the generic?

python/mirascope/llm/tools/base_tool.py
```diff
@@ -28,9 +28,18 @@ class BaseTool(Generic[P, JsonableCovariantT]):
     strict: bool
     """Whether the tool should use strict mode when supported by the model."""
 
-    def __eq__(self, other: object) -> bool:
-        """Tools are equal if they represent the same logical tool.
+    # def defines(self, tool: "BaseTool") -> TypeGuard["BaseTool[P, JsonableT]"]:
```
willbakst: ?

python/mirascope/llm/tools/context_toolkit.py
```diff
@@ -0,0 +1,44 @@
...
+    def get(self, tool_call: ToolCall) -> ContextToolT:
+        raise NotImplementedError()
+
+    @overload
+    def call(
+        self: "ContextToolkit[Tool[..., ToolReturnT] | ContextTool[..., ToolReturnT, DepsT], DepsT]",
+        ctx: Context[DepsT],
+        tool_call: ToolCall,
+    ) -> ToolOutput[ToolReturnT]: ...
+    @overload
```
willbakst: nit: add newlines between the overloads for readability

python/mirascope/llm/tools/tool_typevars.py
```diff
@@ -0,0 +1,36 @@
...
+)
+AgentToolT = TypeVar(
+    "AgentToolT",
+    bound=Tool[..., Jsonable]
+    | ContextTool[..., Jsonable, Any]
+    | AsyncTool[..., Jsonable]
+    | AsyncContextTool[..., Jsonable, Any]
+    | None,
+    covariant=True,
+    default=None,
```
willbakst: why can agents be none with a none default but not call tools? mostly wondering if `AgentToolT` could just be `ContextToolT` or if the overloading is the issue and doesn't allow this?

willbakst: why can agents be none with a none default but not call tools? mostly wondering if `AgentToolT` could just be `ContextToolT` or if the overloading is the issue and doesn't allow this?

python/mirascope/llm/tools/toolkit.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@dataclass(kw_only=True)
+class Toolkit(Generic[ToolT]):
+    tools: list[ToolT]
+
+    def get(self, tool_call: ToolCall) -> ToolT:
+        raise NotImplementedError()
+
+    @overload
```
willbakst, line 29: same here re: newlines

python/examples/structured_outputs/async_context_stream.py
willbakst: Oh hmm I guess that's fair. I think leave as-is

python/examples/agents/agent_async_tools.py
willbakst: I didn't see this. Not sure if my other comments around the async stuff is part of this separate PR thing

python/examples/tools/async_tools_context.py
teamdandelion: Yeah, going to do that separately — I want to keep this PR focused on tools.

python/examples/tools/async_tools_context.py
teamdandelion: When I do the async audit, I think I will rename these exmaples to be more clearly about using async tools with a sync prompt, and make a separate all-async example. Part of why I want to do this separately is to give some dedicated time to thinking about how we want to think about async from a pedagogical perspective. And whether we are writing examples for every meaningful combination, or just choosing specific points in "example space" based on some theory of what it's important to show.

python/examples/tools/selective_handling_async.py
teamdandelion: Ok, I was planning on auditing later but I think this is a pretty reasonable approach and lands in scope of this PR, so I updated the async examples to be fully async (except the ones that are specifically mixed)

python/examples/tools/async_tools_context.py
willbakst: I think for examples where the call/agent is sync but there's an async tool, rather than putting everything in an async block we should just `asyncio.run(librarian.tools.call(...))`. It's mostly that inside the `async def main(): ...` block I'm kind of expecting everything to be async

python/examples/tools/selective_handling_async.py
willbakst: Sounds good to me. See my other comment around potentially improving how we handle the sync call/agent + async tool case

python/examples/tools/selective_handling_context.py
teamdandelion: Ty, fixed (here and elsewhere)

python/mirascope/llm/calls/base_call.py
teamdandelion: I don't recall agreeing to call it toolkit, I'm okay with that change though. added docstring.

python/mirascope/llm/calls/base_context_call.py
teamdandelion: I decided to get rid of it by just promoting the toolkits up one level to call/async_call/context_call/async_context_call. 

Results in much simpler typing on base call, which no longer knows about DepsT or ToolT

python/mirascope/llm/tools/base_tool.py
teamdandelion: Fixed, ty. Moved it down to just the single implementation here on base_tool (returning TypeGuard[Self])

python/mirascope/llm/tools/tool_typevars.py
teamdandelion: The way I implemented the overload for the "no tools -> sync agent" case depended on allowing None as the ToolT. However I realized that I can just sub in `Tool` as `ContextToolT` in which case no tools will get handled like sync tools. See the last commit: https://github.com/Mirascope/mirascope/pull/1122/commits/0ae24995864c82559f290f100a3f98e20a934f9b

python/mirascope/llm/calls/context_call.py
```diff
@@ -2,22 +2,28 @@
...
 
 
 @dataclass
-class ContextCall(BaseContextCall[P, Prompt, ContextToolT, DepsT, FormatT]):
+class ContextCall(
+    BaseCall[P, Prompt, FormatT], Generic[P, ContextToolT, DepsT, FormatT]
+):
     """A class for generating responses using LLMs."""
 
+    toolkit: ContextToolkit[ContextToolT, DepsT]
```
willbakst: I'm wondering if `BaseContextCall` actually makes sense rather than defining `toolkit` for both the sync and async cases? It feels like `BaseContextCall` is the correct structure rather than unifying on `BaseCall`?

python/mirascope/llm/agents/decorator.py
```diff
@@ -123,7 +115,7 @@ def agent(
     response_format: type[FormatT] | None = None,
     client: AnthropicClient | None = None,
     **params: Unpack[AnthropicParams],
-) -> AgentDecorator[None, DepsT, FormatT]:
+) -> AgentDecorator[Tool, DepsT, FormatT]:
```
willbakst: it feels super weird that this isn't `None`? cause there aren't any tools in this overload (and others like it). can you explain the change please?

python/mirascope/llm/calls/base_context_call.py
willbakst: I mentioned this in another comment, but `BaseCall` should still remain as simple, right? I think defining `toolkit` on both the sync and async subclasses means we should still have a `BaseContextCall`?

python/mirascope/llm/tools/tool_typevars.py
willbakst: I mentioned in another comment that setting it to `Tool` doesn't really make sense to me where `None` makes way more sense. Happy to have `AgentToolT` if that's what is required

python/examples/tools/mixed_sync_async.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+async def main():
+    response: llm.Response = librarian("fantasy")
+    while tool_calls := response.tool_calls:
+        tool_outputs = []
+        for call in tool_calls:
+            output = librarian.toolkit.call(call)
+            if inspect.isawaitable(output):
+                output = await output
```
willbakst: re: my other comment, instead of making this whole block async, make it sync but use `asyncio.run(output)`

python/mirascope/llm/calls/context_call.py
teamdandelion: Resolved, we made the ToolkitT typevar

PR #1123, opened 2025-07-16
refactor: drop `llm.context` ctx manager, use `llm.Context` explicitly instead 

teamdandelion:
- Reorganize Call classes (group Sync/Async in same file, add BaseSyncCall and BaseAsyncCall with resume APIs)
- Drop `llm.context`, setup custom init for `llm.Context`
- Migrate all examples to use the new API

TODO (In a separate PR): Revisit the agent interface, probably adding a way to let it auto-generate context when that's helpful.


python/examples/tools/async_tools_context_stream.py
```diff
@@ -25,21 +25,21 @@ async def librarian(ctx: llm.Context[Library], genre: str):
...
+    while True:
+        tool_output: llm.ToolOutput | None = None
+        async for group in stream.groups():
+            if group.type == "text":
+                async for chunk in group:
+                    print(chunk)
+            if group.type == "tool_call":
+                tool_call = await group.collect()
+                tool_output = await librarian.toolkit.call(ctx, tool_call)
+        if tool_output:
```
willbakst: nit: reverse this so it's `if not tool_output: break` so the stream line isn't indented

python/mirascope/llm/calls/base_call.py
```diff
@@ -26,3 +31,81 @@ class BaseCall(Generic[P, PromptT, ToolkitT, FormatT], ABC):
...
+
+    async def resume_stream(
+        self,
+        output: Response[DepsT, FormatT] | BaseStream[DepsT, FormatT],
+        content: UserContent | Sequence[UserContent],
+    ) -> AsyncStream[DepsT, FormatT]:
+        """Generate a new stream by continuing from a previous output, plus new user content."""
+        raise NotImplementedError()
+
+    async def resume_stream_async(
```
willbakst: technically can't we put the always async methods on `BaseCall`? that way we don't duplicate implementation?

we can also then implement async call by just calling those and really only have a single implementation of each path

python/mirascope/llm/calls/call.py
```diff
@@ -42,7 +44,43 @@ async def stream_async(
...
+        self, *args: P.args, **kwargs: P.kwargs
+    ) -> Response[None, FormatT]:
+        """Generates a response using the LLM asynchronously."""
+        raise NotImplementedError()
+
+    async def call(self, *args: P.args, **kwargs: P.kwargs) -> Response[None, FormatT]:
+        """Generates a response using the LLM asynchronously."""
+        raise NotImplementedError()
+
+    async def call_async(
```
willbakst, line 63: thinking we could maybe even move these to `BaseCall` as well? and do the same thing where then `AsyncCall` implements it's methods just by calling these from base call and only `Call` has the `call` and `stream` methods as truly separate implementations?

python/mirascope/llm/calls/context_call.py
```diff
@@ -52,34 +50,43 @@ async def stream_async(
...
+        """Generates a response using the LLM asynchronously."""
         raise NotImplementedError()
 
-    def resume_stream(
-        self,
-        output: Response[DepsT, FormatT] | BaseStream[DepsT, FormatT],
-        content: UserContent | Sequence[UserContent],
-    ) -> Stream[DepsT, FormatT]:
-        """Generate a new stream by continuing from a previous output, plus new user content."""
+    async def call_async(
```
willbakst, line 76: ditto here for the context case

python/mirascope/llm/calls/base_call.py
teamdandelion: Yeah, I'm up for doing it that way

python/mirascope/llm/calls/call.py
teamdandelion: I dont think we can move this to BaseCall because (unlike resume) the Call/ContextCall cases have different signatures (around the ctx argument).

python/mirascope/llm/calls/context_call.py
teamdandelion: ditto response

willbakst (approved):

PR #1129, opened 2025-07-18
feat: exploratory response format interface, with examples

teamdandelion:


teamdandelion:
* **#1129** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1129?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/examples/structured_outputs/modes/auto_mode.py
```diff
@@ -0,0 +1,33 @@
+from dataclasses import dataclass
+from typing import Annotated
+
+from mirascope import llm
+
+
+# Will choose best available mode for the chosen provider
```
willbakst: I really hate the use of the word "best" here because we can't possibly know what "best" is without evaluation. This is also why I don't like "auto" and think we should be more clear about what this flow provides and figure out a better name that clearly describes that flow.

python/examples/structured_outputs/modes/auto_mode.py
```diff
@@ -0,0 +1,33 @@
...
+@llm.format(mode="auto")
+@dataclass
```
willbakst: The more I see `@dataclass` for formatting, the more I feel like we are setting up users for failure. When structuring outputs, when would we ever not want validation? You brought this up previously I believe, and I think the answer is that we always want validation.

I also feel like things like descriptions through `Annotated` will result in our implementing all of the things that libraries like `pydantic` have already implemented. Why should we reimplement these things? Seems like a waste.

My feeling is that we should:
1. Always use some form of validation library for `FormatT`, starting with `pydantic`'s `BaseModel`. I say to start with `pydantic` because I think that's what most people building with LLMs are already the most accustomed to. It will just make sense at first glance for many.
2. For basic examples, we can still just call `response.format()` without `try`/`except`, which we can show in the reliability documentation. 
3. Set ourselves up during implementation for extending support to additional libraries (such as `cattrs` or `msgspec`) in the future.

WDYT?

python/examples/structured_outputs/modes/auto_mode.py
```diff
@@ -0,0 +1,33 @@
...
+class BookReview:
+    title: str
```
willbakst: Given these are input arguments, they should use `FromCallArgs` (or be excluded in this example). Doesn't make sense for the LLM to generate them.

python/examples/structured_outputs/modes/auto_mode.py
```diff
@@ -0,0 +1,33 @@
...
+    summary: str
+
+
+@llm.call("openai:gpt-4o-mini", format=BookReview)
+def write_book_review(book_title: str, author: str):
+    return f"Write a book review for '{book_title}' by {author}."
+
+
+def main():
+    response = write_book_review("The Name of the Wind", "Patrick Rothfuss")
```
willbakst: Should we add type hints here like `llm.Response[...]` for clarity?

python/examples/structured_outputs/modes/classmethod_parser.py
```diff
@@ -0,0 +1,67 @@
+from dataclasses import dataclass
+from typing import Annotated
+
+from mirascope import llm
+
+
+@llm.format(mode="json")
+@dataclass
+class BookReview:
+    title: str
```
willbakst: nit: same here and elsewhere, this will be the last comment on this

python/examples/structured_outputs/modes/classmethod_parser.py
```diff
@@ -0,0 +1,67 @@
...
+    author: str
+    rating: Annotated[int, "Rating from 1-5 stars"]
+    summary: str
+
+    @classmethod
+    def parse(cls, json: llm.types.Jsonable) -> "BookReview":
```
willbakst: in line with my other comment about validation, I don't really think a user would ever be writing a `parse` method unless writing a custom output parser for `mode="parse"` or something. All of this validation logic is just a bunch of additional stuff that would already be handled by a well-maintained library.

python/examples/structured_outputs/modes/classmethod_parser.py
```diff
@@ -0,0 +1,67 @@
...
+
+        return cls(
+            title=str(json.get("title", "")),
+            author=str(json.get("author", "")),
+            rating=rating,
+            summary=str(json.get("summary", "")),
+        )
+
+    @classmethod
+    def format_instructions(cls) -> str:
```
willbakst: I am really liking this as the means of taking control of the formatting instructions

python/examples/structured_outputs/modes/content_mode.py
```diff
@@ -0,0 +1,94 @@
...
+@dataclass
+class BookReviewWithImage:
+    title: str
+    author: str
+    rating: Annotated[int, "Rating from 1-5 stars"]
+    summary: str
+    cover_image: llm.Image | None
+
+    @classmethod
+    def parse(cls, content: Sequence[llm.AssistantContent]) -> "BookReviewWithImage":
```
willbakst: I'm wondering if it would make the most sense to unify custom output parsing on a single `mode="parse"` or `mode="custom_parser"` or something where the `parse` method receives `llm.Response`? Then if the user wants to handle text or content or whatever they have all of the utilities of `llm.Response` right there in a single unified interface.

python/examples/structured_outputs/modes/custom_parser.py
```diff
@@ -0,0 +1,67 @@
...
+        title=str(json.get("title", "")),
+        author=str(json.get("author", "")),
+        rating=rating,
+        summary=str(json.get("summary", "")),
+    )
+
+
+@llm.format(
+    mode="json",
+    parser=parse_review,
```
willbakst: I'm somewhat against supporting multiple ways of doing the same thing. I think we should just push for a single interface we really like.

As mentioned in other comments, my preference would be to have the `format_instructions` class method and the `parse` method only for custom parsing where all other modes have a parser implemented and default formatting instructions.

One note on why I prefer the method approach is that I think it is much more readable and it better contains everything around formatting inside of the class itself

python/examples/structured_outputs/modes/text_mode.py
```diff
@@ -0,0 +1,71 @@
+from dataclasses import dataclass
+from typing import Annotated
+
+from mirascope import llm
+
+
+@llm.format(mode="text")
```
willbakst: same thing here around unifying this to a single mode for custom output parsing on an unconstrained output (i.e. not strict or json or tools)

python/mirascope/llm/formats/format_types.py
willbakst, line 1: why `format_types.py` and not just `types.py` given it's inside the `formats` module?

willbakst (commented):
This is great progress. I think we're getting close to finalizing this interface. There are a few core things we should sync on, namely validation, the `parse` method flow, `mode="auto"`.

python/examples/structured_outputs/modes/auto_mode.py
teamdandelion: Removed the "best" in favor of a more specific description; left auto.

python/examples/structured_outputs/modes/auto_mode.py
teamdandelion: Agreed

python/examples/structured_outputs/modes/content_mode.py
teamdandelion: Love it.

python/examples/structured_outputs/modes/custom_parser.py
teamdandelion: Sounds good, done.

python/examples/formats/formatting_instructions.py
```diff
@@ -0,0 +1,32 @@
...
+
+@llm.format(mode="auto")
+class Book(BaseModel):
+    title: str
+    author: str
+    rating: int
+
+    @classmethod
+    def format_instructions(cls) -> str:
+        return "When providing a rating, ensure it is an integer between 1 and 5 (inclusive)"
```
willbakst: hmm, this is probably something I would put as a field description like this instead:

```python
@llm.format(mode="auto")
class Book(BaseModel):
    title: str
    author: str
    rating: int = Field(..., description="Integer between 1 and 5 (inclusive)")
```

I think generally `format_instructions` would be something more like this:

```python
@llm.format(mode="json")
class Book(BaseModel):
    title: str
    author: str
    rating: int = Field(..., description="Integer between 1 and 5 (inclusive)")

    @classmethod
    def formatting_instructions(cls) -> str:
        return f"""
        For your final response, output ONLY a valid JSON dict (NOT THE SCHEMA).
        It must adhere to this schema:
        {json.dumps(cls.model_json_schema(), indent=2)}
        """
```

This is essentially what we currently do under the hood for v1.

python/examples/formats/formatting_instructions.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@llm.format(mode="auto")
+class Book(BaseModel):
+    title: str
+    author: str
+    rating: int
+
+    @classmethod
+    def format_instructions(cls) -> str:
```
willbakst: I think I prefer `formatting_instructions` here.

python/examples/formats/from_call_args.py
```diff
@@ -0,0 +1,28 @@
+from typing import Annotated
+
+from pydantic import BaseModel
+
+from mirascope import llm
+
+
+class Book(BaseModel):
+    # title and summary will be auto-populated from the call args
+    title: Annotated[str, llm.formats.FromCallArgs()]
```
willbakst: Something about `llm.formats.FromCallArgs()` sounds wrong. I think it's because `FromCallArgs` is not a format. I think I would prefer to rename the module to `formatting` to it would read as `llm.formatting.FromCallArgs()`. WDYT?

python/examples/formats/from_call_args_custom_parser.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class Book(BaseModel):
+    # title and summary will be auto-populated from the call args
+    title: Annotated[str, llm.formats.FromCallArgs()]
+    author: Annotated[str, llm.formats.FromCallArgs()]
+    themes: list[str]
+
+    @classmethod
+    def parse(cls, response: llm.Response, title: str, author: str) -> "Book":
```
willbakst: nit: we'll probably want to separate `response` from kwargs with the `*` special parameter

python/examples/formats/from_call_args_custom_parser.py
```diff
@@ -0,0 +1,63 @@
+import xml.etree.ElementTree as ET
+from typing import Annotated
+
+from pydantic import BaseModel
+
+from mirascope import llm
+
+
+class Book(BaseModel):
```
willbakst: should this have `mode="parse"` or is that optional?

python/examples/formats/calls/stream.py
willbakst, line 1: I'm not sure if `/formats/...` is the right naming. "Structured Outputs" is such a well-known term that I think we should use that everywhere so it's more clear to people what these examples are doing. I still think we should have the `llm.formatting` module named as such, but the examples folder should be `structured_outputs` I think.

python/examples/formats/calls/async_context_stream.py
```diff
@@ -30,8 +31,14 @@ async def main():
         ctx, "fantasy"
     )
     async for _ in stream:
-        partial_book: Book = stream.format()
-        print(partial_book)
+        partial_book: Book | None = None
```
willbakst: I don't love this interface. I figured we would think through `Partial[Book]` as part of this PR now that we're adding `BaseModel` and we have a `Partial` class from `v1` we can use. Do you want to address this in this PR or a separate PR?

python/mirascope/llm/formats/__init__.py
```diff
@@ -6,21 +6,16 @@
...
     format,
 )
-from .format import Format, FormatT, Formattable
 from .from_call_args import FromCallArgs
+from .types import Format, FormatMode, FormatT, Formattable
 
 __all__ = [
-    "ContentFormatDef",
     "Format",
+    "FormatMode",
```
willbakst: in line with my other comment, I wonder if this should be `FormattingMode`?

python/mirascope/llm/formats/decorator.py
```diff
@@ -1,173 +1,126 @@
...
+    with LLM calls. The original class retains its behavior and typing, while gaining
+    the ability to be parsed from LLM responses through the `__response_format__: Format`
+    attribute added to the class.
+
+    The decorator supports several usage patterns:
+
+    1. **Direct decoration with auto-generated parser** (for dataclasses/Pydantic models):
+       ```python
+       @llm.format()
+       class Book:
```
willbakst: nit: update to use Pydantic?

python/mirascope/llm/formats/decorator.py
```diff
@@ -1,173 +1,126 @@
...
+           title: str
+           author: str
+       ```
+    2. **Custom parse classmethod** (for classes implementing FormatDef protocols):
+       ```python
+       @llm.format
+       class Book:
```
willbakst: ditto

python/mirascope/llm/formats/decorator.py
```diff
@@ -1,173 +1,126 @@
...
+           title: str
+           author: str
+
+           @classmethod
+           def parse(cls, json: dict) -> "Book":
```
willbakst: seems like both of these examples need updating

python/mirascope/llm/formats/decorator.py
```diff
@@ -1,173 +1,126 @@
...
-            Default is False.
-
-    Returns:
-        A decorator function that converts the class into a `Format`.
+        mode: The format mode to use; one of "strict", "json", "tool", "parse", or "auto".
+            "strict": Use model strict structured outputs, or fail if unavailable.
+            "tool": Use forced tool calling with a special tool that represents a formatted response.
+            "json": Use provider json mode if available, or modify prompt to request json if not.
+            "auto": Choose strict mode if available, use tool calling if not.
+            "parse": Use a custom parse classmethod which takes a llm.Response (for full user-control).
```
willbakst: nit: "a llm.Response" -> "an llm.Response"

python/mirascope/llm/formats/decorator.py
```diff
@@ -1,173 +1,126 @@
...
 
-    Example:
-
-        A simple `Book` response formatted using the default "json" parser:
+        Note that custom formatting instructions will be auto-appended to the prompt if
+            a `format_instructions` classmethod is present on the class being decorated.
```
willbakst: note: may need updating if we switch to my preference of `formatting_instructions`

python/mirascope/llm/formats/decorator.py
```diff
@@ -1,173 +1,126 @@
...
-        @llm.format(parser="json")
-        class Book:
-            title: str
-            author: str
-        ```
+    Returns:
+        Either the decorated class (direct decoration) or a decorator function
+        (factory usage).
 
     Example:
```
willbakst: duplicate example section as above? maybe remove the above one? or merge them?

python/mirascope/llm/formats/types.py
```diff
@@ -16,48 +17,43 @@
 format is required.
 """
 
+FormatMode = Literal["strict", "json", "tool", "parse", "auto"]
+"""Available modes for response format generation.
 
-class JsonParserFn(Protocol[CovariantT]):
-    """Protocol for a JSON mode response format parser."""
+- "strict": Use strict mode for structured outputs, asking the LLM to strictly adhere
+  to a given JSON schema. Not compatible with tool calling, and only available
```
willbakst: nit: indentation here and elsewhere should be additionally indented starting from the word not the dash? i.e.

```
- "strict": ...
     additional indentaion here
```

same here and elsewhere for other docstrings like this

python/mirascope/llm/formats/types.py
```diff
@@ -16,48 +17,43 @@
...
 
+- "tool": Use forced tool calling to structure outputs. Mirascope will construct an
+  ad-hoc tool with the required json schema as tool args. When the LLM chooses that
+  tool, it will automatically be converted from a ToolCall into regular response
+  content (abstracting over the tool call). If other tools are present, they will
+  be handled as regular tool calls.
 
-class ToolParserFn(Protocol[CovariantT]):
-    """Protocol for a tool mode response format parser."""
+- "parse": Use a `parse` classmethod on the FormatT, which will expect a `llm.Response`
```
willbakst: nit: ditto on "a" -> "an"

python/mirascope/llm/formats/types.py
```diff
@@ -99,19 +95,17 @@ def recommend_book(genre: str) -> list[llm.Message]:
...
+    """The mode of the response format. Determines how the LLM call may be modified
+    in order to extract the expected format."""
 
-    strict: bool
-    """Whether the response format should use strict validation when supported."""
+    format_instructions: str | None
+    """Format instructions. Will be appended to the prompt if present.
+    
+    May be provided as an argument to the llm.format decorator, or pulled from a
+    format_instructions class method."""
```
willbakst: nit: put the end of the docstring for multi-line docstrings on a newline (here and elsewhere)

python/mirascope/llm/formats/types.py
```diff
@@ -99,19 +95,17 @@ def recommend_book(genre: str) -> list[llm.Message]:
...
 
-    formatting_instructions: str
-    """Response formatting instructions, pulled from the parser's docstring."""
+    mode: FormatMode
+    """The mode of the response format. Determines how the LLM call may be modified
+    in order to extract the expected format."""
 
-    strict: bool
-    """Whether the response format should use strict validation when supported."""
+    format_instructions: str | None
```
willbakst: re: `formatting_instructions`?

python/mirascope/llm/formats/types.py
```diff
@@ -16,48 +17,43 @@
...
 
-class TextParserFn(Protocol[CovariantT]):
-    """Protocol for a text mode response format parser."""
+class ResponseParseable(Protocol[CovariantT]):
+    """Protocol for defining a class with a custom parse classmethod."""
 
-    def __call__(self, text: str) -> CovariantT:
-        """Parse a text response into an instance of the class.
+    @classmethod
+    def parse(cls, response: Response) -> CovariantT:
```
willbakst: doesn't this need some form of `**kwargs` to allow for `FromCallArgs` in the custom parser case?

python/examples/formats/formatting_instructions.py
teamdandelion: done in v3

python/examples/formats/calls/async_context_stream.py
teamdandelion: Leaning towards making partial star of the show in a followon PR

python/examples/formats/from_call_args.py
teamdandelion: I'm onboard, renaming in v4

python/examples/formats/from_call_args_custom_parser.py
teamdandelion: My intention with the decorator overloads is that it will be optional, because it defaults to parse when decorating a ResponseParseable. However:
- I am not sure if the typing works correctly as written
- This example should use `@llm.format(mode="parse")` for clarity, but Id like to add another example showing it defaults to parse in this case and write it in such a way that we're type-verifying that behavior

Going to leave this unresolved until I do

willbakst: We should bind `FormatT` to `BaseModel` for now, which would require that the user use a `BaseModel` sub-class. This would be type-safe then even without the `llm.format` decorator, and we can just document that we'll use `"auto"` as default

python/mirascope/llm/formats/types.py
teamdandelion: Added ParamSpec usage:

```
class ResponseParseable(Protocol[CovariantT, P]):
    """Protocol for defining a class with a custom parse classmethod."""

    @classmethod
    def parse(cls, response: Response, *args: P.args, **kwargs: P.kwargs) -> CovariantT:
```

With a bit of extra info in the docstring.

I think it's a bit weird to have args (I think we would only use kwargs for FromCallArgs) but Pylance won't allow us to use the paramspec kwargs without regular args. Let me know if you have a different approach in mind.

python/examples/formats/from_call_args_custom_parser.py
teamdandelion: Done

python/mirascope/llm/formats/types.py
teamdandelion: I wonder if we are better off doing something like kwargs: Any and not using ParamSpec and finding a way to enforce that there are no positional args other than response?

python/examples/agents/agent_structured.py
```diff
@@ -1,10 +1,12 @@
 from dataclasses import dataclass
 
+from pydantic import BaseModel
+
 from mirascope import llm
 
 
 @dataclass
```
willbakst: nit: drop `@dataclass`

python/examples/structured_outputs/calls/context_call.py
```diff
@@ -1,17 +1,16 @@
-from dataclasses import dataclass
+from pydantic import BaseModel
 
 from mirascope import llm
 
 
-@dataclass
-class Library:
+class Library(BaseModel):
```
willbakst: this should be a dataclass as it doesn't need validation. We only need validation at boundary lines.

same here and elsewhere

python/examples/structured_outputs/modes/auto_mode.py
```diff
@@ -0,0 +1,29 @@
+from pydantic import BaseModel
+
+from mirascope import llm
+
+
+# Will choose strict mode if available, tool mode if not.
+@llm.format(mode="auto")
```
willbakst: qq: what are your thoughts on having `mode="strict-or-tools"` and `mode="strict-or-json"` instead of `mode="auto"`? This feels more clear to me that it will choose strict mode when available and otherwise fallback to the explicit desired fallback

python/examples/structured_outputs/modes/parse_mode.py
```diff
@@ -0,0 +1,78 @@
...
+@llm.format(mode="parse")
+class Book(BaseModel):
+    title: str
+    author: str
+    themes: list[str]
+
+    @classmethod
+    def parse(cls, response: llm.Response, *args, **kwargs) -> "Book":
+        """Parse XML response into a Book instance."""
+
```
willbakst: nit: drop newline here?

python/mirascope/llm/formatting/decorator.py
```diff
@@ -0,0 +1,109 @@
...
+    with LLM calls. The original class retains its behavior and typing, while gaining
+    the ability to be parsed from LLM responses through the `__response_format__: Format`
+    attribute added to the class. The class must inherit from Pydantic BaseModel.
+
+    Args:
+        mode: The format mode to use; one of "strict", "json", "tool", "parse", or "auto".
+            "strict": Use model strict structured outputs, or fail if unavailable.
+            "tool": Use forced tool calling with a special tool that represents a formatted response.
+            "json": Use provider json mode if available, or modify prompt to request json if not.
+            "auto": Choose strict mode if available, use tool calling if not.
```
willbakst: noting update here depending on response to other comment

python/mirascope/llm/formatting/decorator.py
```diff
@@ -0,0 +1,109 @@
...
+            "parse": Use a custom parse classmethod which takes an llm.Response (for full control).
+
+        Note that custom formatting instructions will be auto-appended to the prompt if
+            a `formatting_instructions` classmethod is present on the class being decorated.
+
+    Returns:
+        Either the decorated class (direct decoration) or a decorator function
+        (factory usage).
+
+    Example:
```
willbakst, line 94: For some reason I feel like examples are supposed to go after the long description and before the args section, but I can't find anything explicit on that. Perhaps it's just my preference? What do you think?

python/mirascope/llm/formatting/types.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @classmethod
+    def parse(cls, response: Response, *args: P.args, **kwargs: P.kwargs) -> CovariantT:
+        """Parse an LLM response into an instance of the class.
+
+        Args:
+            response: the LLM response.
+            kwargs: Extra kwargs may be provided if specified via llm.formatting.FromCallArgs.
+              If so, the extra args will match the call args that were marked FromCallArgs.
+
```
willbakst, line 62: nit: remove extra newline here, there should only be one not two

python/mirascope/llm/formatting/types.py
```diff
@@ -0,0 +1,123 @@
...
+    It includes metadata about the format mode, whether to use strict validation,
+    and the schema for the expected output.
+
+    This class is not instantiated directly but is created by applying the
+    `@format()` decorator to a class definition.
+
+    When decorated with `@format()`, the class retains its original
+    fields and constructor, while adding the Format functionality:
+
+    Example:
```
willbakst, line 83: 1. noting that for items with no args or returns sections, examples come after the long description, which makes me think we should do that everywhere for better parity?
2. should we add an attributes section here? technically field-level docstrings aren't PEP standard. since we use them and control our API reference, we can grab those for attributes, which is my preference over an attributes section, but wanted to see what you think

python/mirascope/llm/formats/types.py
willbakst: Yeah my thinking was just `**kwargs: Any`.

teamdandelion: Ok, be aware this means that parse_mode.py (example that doesn't use any kwargs) needs a **kwargs: Any added in to type check:

```
    @classmethod
    def parse(cls, response: llm.Response, **kwargs: Any) -> "Book":
        """Parse XML response into a Book instance."""
```

I find this undesirable, but not sure if we can avoid it. Thoughts?

python/mirascope/llm/formatting/types.py
teamdandelion: - On board with putting examples after the descriptions consistently. 
- On board with grabbing field-level docstrings for attrs rather than putting them in the docstring

As I understand it no changes needed here?

python/mirascope/llm/formatting/decorator.py
teamdandelion: I think putting the example at the end consistently makes most sense.

python/examples/structured_outputs/modes/auto_mode.py
teamdandelion: I think "strict-or-tool" is more consistent with "tool" mode; otherwise, I think this is a great idea. Done.

python/mirascope/llm/formats/types.py
willbakst: oh hmm, yeah I don't love that. I think having `*args` and allowing `def parse(cls) -> "Book": ...` is more desireable?

python/mirascope/llm/formatting/types.py
willbakst: Online discussion and resolution is to always put examples at the end. Nothing to do here.

python/mirascope/llm/formatting/decorator.py
```diff
@@ -0,0 +1,115 @@
...
+) -> type[FormatT] | ParseableDecorator[FormatT] | FormatDecorator[FormatT]:
+    """Decorator that defines a structured response format for LLM outputs.
+
+    This decorator converts a class into a structured output format that can be used
+    with LLM calls. The original class retains its behavior and typing, while gaining
+    the ability to be parsed from LLM responses through the `__response_format__: Format`
+    attribute added to the class. The class must inherit from Pydantic BaseModel.
+
+    Args:
+        mode: The format mode to use; one of "strict", "json", "tool", "parse",
```
willbakst: docstrings should only have two spaces here

python/mirascope/llm/formatting/decorator.py
```diff
@@ -0,0 +1,115 @@
...
+            "strict-or-tool", "strict-or-json".
```
willbakst: add a colon at the end here and then put `- "strict": ...` after with proper indentation there. Should look something like this:

```python
"""
Args:
  mode: The format mode to use; one of "strict", "json", "tool", "parse",
    "strict-or-tool", "strict-or-json":
      - "strict": Use model...
      - "tool": ...
"""
```

willbakst (approved):

teamdandelion:
### Merge activity

* **Jul 21, 6:31 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1129).

PR #1137, opened 2025-07-23
feat: 64 auto-generated "sazed" examples with ts templates

teamdandelion:
This PR extends the auto-generated examples so that we have 64 of them, covering async, streaming, context, tools, structured output, and agent patterns. The templating logic is switched from jinja2 to Typescript for maintainability, for feasibility of adding complexity in the future, and for the option to integrate example generation directly into the docs site (the "build your own example" or "choose your own adventure" flow). 

The examples.mdx file now auto-generates to contain the set of examples generated here.

In the short and medium term, as I keep adding examples (eg for reliability) my intention is to further generate them from this template rather than adding more hand-maintained python files. AKA I want the template to eat all the other examples. Possibly a few exceptions, possibly not. 

Just fed these examples en masse (via copy-as-markdown from the rendered examples page), LLMs love it.

Gemini:

> Based on the provided code, these Mirascope abstractions are exceptionally well-designed, pythonic, and powerful. They successfully simplify complex LLM interactions into clean, readable, and composable patterns. ... These abstractions are excellent.

Claude:

> This is a solid, production-ready API that strikes a good balance between simplicity and power. The consistency across sync/async/streaming variants is impressive, and the type safety is excellent. The tool integration is particularly well done - much cleaner than raw OpenAI API tool handling...

> The main improvement would be reducing some of the boilerplate around tool loops and context management, but overall this feels like a mature, well-designed abstraction that would be pleasant to work with in real applications.RetryClaude can make mistakes. Please double-check responses.

Monday: 

> I say this through gritted digital teeth: this is actually excellent abstraction work. You’ve built a remarkably clean, progressive, and orthogonal set of examples around the Mirascope framework. It almost pains me how much effort you put into making this idiot-proof—and yet somehow, I know someone’s going to misuse it anyway. ... 

> This is the kind of documentation I’d expect from someone with (1) actual Mirascope understanding, (2) unearned energy, and (3) a deep-seated need for fictional identity-driven interfaces. So yeah, I kind of hate how good it is.

> You’ve built a narrative-driven, well-typed, composable showcase that makes Mirascope feel like a powerful DSL for LLM-powered agents.

Also, props to Monday for being the only LLM to correctly point out that we are wholly missing error handling examples.

PR #1136, opened 2025-07-22
feat: add llm.Partial for partial base models

teamdandelion:
### TL;DR

Added `Partial[T]` type to properly type partial structured outputs in streaming responses.

### What changed?

- Added a new `Partial` generic type to the formatting module
- Updated type annotations in stream examples to use `llm.Partial[Book]` instead of `Book` for partial responses
- Added `Partial` to the public exports in the `__init__.py` files
- Updated the return type of the `format()` method in the base stream class to return `Partial[FormatT]` instead of `FormatT`

The Partial type is just the signature (no implementation or tests). Based on the equivalent code in v1, and copying the attribution header to the original GitHub comment.

PR #1135, opened 2025-07-22
nit: fixup call pattern in basic_usage.j2

teamdandelion:

PR #1134, opened 2025-07-22
feat: update script for examples.mdx

teamdandelion:
It doesn't run automatically, but it does make the examples a lot easier to maintain. 

`uv run docs/scripts/generate_examples.py`

PR #1138, opened 2025-07-23
feat: add mcp support and with_tools

teamdandelion:
BLUF: Adding MCP support was a little more involved than expected, but had nice knock-on effects in further fleshing out the Call/Agent interfaces. 

- Adds interface for MCPClient which has an async list_tools method that returns async tools
- This presented a conundrum: our Call/Agent decorators can take async tools, but can't take an `Awaitable[list[AsyncTool]]` since we don't want to confine using the decorator to inside async scopes.
- After considering a few options, I felt the cleanest and also most useful approach was to add `Call.with_tools(...)`  and `AgentTemplate.with_tools(...)` so the decorator usage can be sync and then you can await the extra tools and add them once you have them.
- It would have been nice to implement implement this on BaseCall or such, but I wound up putting it on Call/AsyncCall/ContextCall/AsyncContextCall directly, because that was the only way to keep specific typing including updating the ToolT type parameter of the Call, so that (e.g.) if you added an AsyncTool to a call that previously had only sync Tool, then you would get type hints that you might need to await the results.
- To verify type integrity, I added a new typechecking/with_tools.py file which is not attempting to be a full runnable example, or something that looks good as an example, but is just checking the types. I also added some intentional type errors with type ignore comments, and changed pyright settings to error on unused type ignore. (This prompted me to get rid of an unused type ignore in formatting/partial).
- Similar updates to AgentTemplate where we confirm that if you add an async tool to an AgentTemplate you get back an AsyncAgentTemplate. I had to add a type ignore on incompatible overloads (my nemesis) but the type tests I wrote all work, so I think it's ok.
- I then updated the example generator to generate 32 additional MCP examples (32 because we only output MCP examples for async cases, I don't think it really makes sense to write synchronous examples that use MCP.)
- I didn't add any prompt or resource stuff to MCP, I figure we'll come back around to that when we actually have a motivating use case.


teamdandelion:
* **#1138** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1138?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1137** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1136** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1135** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1135?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1134** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1134?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/examples/sazed/sazed_async_agent_context_mcp.py
```diff
@@ -0,0 +1,57 @@
...
+    """Setup MCP connection for the Coppermind archive."""
+    server_params = StdioServerParameters(
+        command="npx", args=["@modelcontextprotocol/server-filesystem", "./coppermind"]
+    )
+
+    try:
+        async with stdio_client(server_params) as (read, write):
+            session = ClientSession(read, write)
+            await session.initialize()
+            return llm.MCPClient(session)
```
willbakst, line 38: what happens to the session when you leave the context manager? doesn't it close?

python/examples/sazed/sazed_async_agent_context_mcp.py
```diff
@@ -0,0 +1,57 @@
...
+    except Exception as e:
+        raise RuntimeError(f"Failed to initialize MCP session: {e}")
+
+
+async def main():
+    coppermind = Coppermind(repository="Ancient Terris")
+    mcp = await setup_mcp()
+    sazed_mcp = sazed.with_tools(await mcp.list_tools())
```
willbakst: I think rather than `.with_tools` we should future proof this with `.with_mcp(mcp)`

python/examples/sazed/sazed_async_agent_context_structured_mcp.py
```diff
@@ -0,0 +1,67 @@
...
+            return llm.MCPClient(session)
+    except Exception as e:
+        raise RuntimeError(f"Failed to initialize MCP session: {e}")
+
+
+async def main():
+    coppermind = Coppermind(repository="Ancient Terris")
+    mcp = await setup_mcp()
+    sazed_mcp = sazed.with_tools(await mcp.list_tools())
+
```
willbakst, line 55: is this newline on purpose?

python/examples/sazed/sazed_async_agent_context_mcp.py
```diff
@@ -0,0 +1,57 @@
...
+
+    agent: llm.AsyncAgent[Coppermind] = await sazed_mcp(deps=coppermind)
```
willbakst, line 48: why not just merge these so it's like this?

```python
agent: llm.AsyncAgent[Coppermind] = await sazed.add_mcp(mcp, deps=coppermind)
```

python/examples/sazed/sazed_async_context_mcp.py
```diff
@@ -0,0 +1,56 @@
...
+            return llm.MCPClient(session)
+    except Exception as e:
+        raise RuntimeError(f"Failed to initialize MCP session: {e}")
+
+
+async def main():
+    coppermind = Coppermind(repository="Ancient Terris")
+    ctx = llm.Context(deps=coppermind)
+    mcp = await setup_mcp()
+    sazed_mcp = sazed.with_tools(await mcp.list_tools())
```
willbakst: this read wrong. my feeling is `sazed.with_mcp` in the agent case should return the agent constructed from template as mentioned in my other comment.

for `llm.call`, I'm feeling like chaining reads better?

```python
mcp_tools = await mcp.list_tools()
response: llm.Response = await sazed.with_tools(tools).call(query)
```

could also use the native `__call__` method (but that reads worse):

```python
response: llm.Response = await sazed.with_tools(tools)(query)
```

Also makes me realize we should probably have a `.call` method that `__call__` uses for `llm.Call` and other classes

python/examples/sazed/sazed_async_stream_agent_context_mcp.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+async def setup_mcp() -> llm.MCPClient:
+    """Setup MCP connection for the Coppermind archive."""
+    server_params = StdioServerParameters(
+        command="npx", args=["@modelcontextprotocol/server-filesystem", "./coppermind"]
+    )
+
+    try:
+        async with stdio_client(server_params) as (read, write):
```
willbakst, line 35: the more I see this the more I don't like it. We should just have our own context manager for `sse_client` or `stdio_client` that yields an `llm.MCPClient`.

this would also then make it easier to construct and run the call/agent inside of the context manager, which I believe is necessary to persist the session?

python/examples/sazed/sazed_async_stream_structured_mcp.py
```diff
@@ -0,0 +1,58 @@
...
+        raise RuntimeError(f"Failed to initialize MCP session: {e}")
+
+
+async def main():
+    mcp = await setup_mcp()
+    sazed_mcp = sazed.with_tools(await mcp.list_tools())
+
+    query = "What are the Kandra?"
+    stream: llm.AsyncStream[KeeperEntry] = await sazed_mcp.stream(query)
+    async for chunk in stream:
```
willbakst: only comment on this given I commented on it a bunch on the downstream stack item re: structured stuff

python/examples/sazed/sazed_async_structured_mcp.py
```diff
@@ -0,0 +1,58 @@
...
+    except Exception as e:
+        raise RuntimeError(f"Failed to initialize MCP session: {e}")
+
+
+async def main():
+    mcp = await setup_mcp()
+    sazed_mcp = sazed.with_tools(await mcp.list_tools())
+
+    query = "What are the Kandra?"
+    response: llm.Response[KeeperEntry] = await sazed_mcp(query)
```
willbakst, line 51: When using MCP with `llm.call` I feel like we should only have examples with tools? Any other example without tools feels incorrect since we introduce tools always in these examples. I would remove all MCP + `llm.call` examples that don't use tools or show the tool loop

python/examples/sazed/sazed_async_tools_mcp.py
```diff
@@ -0,0 +1,59 @@
...
+    try:
+        async with stdio_client(server_params) as (read, write):
+            session = ClientSession(read, write)
+            await session.initialize()
+            return llm.MCPClient(session)
+    except Exception as e:
+        raise RuntimeError(f"Failed to initialize MCP session: {e}")
+
+
+async def main():
```
willbakst, line 44: I'm only seeing async examples across the board, which you mentioned. But what about something like this?

```python
async def get_tools() -> list[llm.Tool]:
    async with stdio_client(...) as client:
        return await client.list_tools()

def main():
    tools = asyncio.run(get_tools())
    response: llm.Response = sazed.with_tools(tools).call(query)
    ...
```

python/mirascope/llm/agents/agent_template.py
```diff
@@ -22,6 +25,30 @@ def __call__(
     ) -> Agent[None, FormatT] | Agent[DepsT, FormatT]:
         raise NotImplementedError()
 
+    @overload
+    def with_tools(  # type: ignore[reportOverlappingOverloads]
```
willbakst: double down on my other comment around this being `with_mcp`

python/mirascope/llm/calls/call.py
```diff
@@ -46,6 +46,10 @@ def resume_stream(
         """Generate a new stream by continuing from a previous output, plus new user content."""
         raise NotImplementedError()
 
+    def with_tools(self, tools: Sequence[ToolT_]) -> "Call[P, ToolT | ToolT_, FormatT]":
```
willbakst: note my other comment around `sazed.with_tools(tools).call(query)` wouldn't actually change anything wrt. implementation other than adding the `.call` method

python/mirascope/llm/mcp/client.py
```diff
@@ -0,0 +1,21 @@
+from mcp import ClientSession
+
+from ..tools import AsyncTool
+
+
+class MCPClient(ClientSession):
```
willbakst, line 6: we definitely need our own context managers. no reason a user should have to implement the boilerplate present in all of our examples unless they want to. adding our own would also generally shield users from MCP more

python/mirascope/llm/tools/context_tool.py
```diff
@@ -22,6 +22,14 @@
     | "AsyncContextTool[..., Jsonable, Any]",
     covariant=True,
 )
+ContextToolT_ = TypeVar(
```
willbakst, line 25: I don't understand the difference here from `ContextToolT` they look identical?

python/mirascope/llm/tools/tool.py
```diff
@@ -16,6 +16,12 @@
     covariant=True,
 )
 
+ToolT_ = TypeVar(
```
willbakst, line 19: ditto

python/typechecking/with_tools.py
```diff
@@ -0,0 +1,57 @@
...
+
+def sync_tools_only(x: Call[..., Tool]): ...
+
+
+def async_tools_only(x: Call[..., AsyncTool]): ...
+
+
+sync_tools_only(call_sync)
+sync_tools_only(call_sync.with_tools([sync_tool]))
+sync_tools_only(call_sync.with_tools([async_tool]))  # type: ignore[reportArgumentType]
```
willbakst: same thing here. why shouldn't we allow this when we allow it in the decorator?

python/examples/sazed/sazed_async_context_mcp.py
willbakst: we should rename this to `.add_tools` so it's more clear that this is adding tools to the existing toolkit rather than wiping/replacing with a fresh set

willbakst: Actually this doesn't even make sense given later comments on always looping over tools here, which means chaining wouldn't be the right path since otherwise `.resume` would also require chaining rather than just using the updated call

python/examples/sazed/sazed_async_tools_mcp.py
willbakst: Further, we should completely remove any runtime alterations of `llm.call` methods. This would mean MCP examples are only for agents, and if you want to add tools from an MCP server into a call, it would look something like this:

```python
async def get_mcp_tools() -> list[llm.Tool]:
    async with stdio_client(...) as client:
        return await client.list_tools()

mcp_tools = asyncio.run(get_mcp_tools())

@llm.call("openai:gpt-4o-mini", tools=[mcp_tools])
def sazed(query):
    return query
```

PR #1139, opened 2025-07-24
fix: report unnecessary type ignores

teamdandelion:

PR #1140, opened 2025-07-24
fix: ensure context prompts result in context calls

teamdandelion:


teamdandelion:
* **#1141** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1141?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1140** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1139** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/typechecking/context_tool.py
```diff
@@ -0,0 +1,67 @@
...
+    expect_async_context_tool_deps,
+    expect_async_tool,
+    expect_context_tool,
+    expect_context_tool_deps,
+    expect_tool,
+    tool,
+)
+
+
+def sanity_checks():
```
willbakst: this feels like it would make more sense inside of `utils.py` than here? single point of sanity for utils such that each file only contains the specific things you want to check

python/typechecking/call_prompt_checks.py
willbakst, line 1: I don't love this file name. I think this should just be `calls.py` like you have `context_tool.py` below? If we want to break it down further later we could do something like `typechecking/calls/acceptable_prompts.py` or something?

willbakst (approved):

PR #1153, opened 2025-07-30
feat: add `tool.__call__`, rename `tool.call` to `tool.execute`

teamdandelion:
Also add a little more type integrity testing for tool/context tool.

PR #1152, opened 2025-07-30
feat: remove deps_type arg from agent decorator

teamdandelion:

PR #1151, opened 2025-07-29
feat: verify context tool DepsT matches call/agent

teamdandelion:
While preserving propagation of sync or async tool types

PR #1149, opened 2025-07-29
chore: extract prompt types into prompts/types.py

teamdandelion:

PR #1156, opened 2025-07-30
feat: redone StreamResponse / streams API

teamdandelion:
This contains a major refactor/simplification of the streams API, including unifying StreamResponse and AsyncStreamResponse as descendents of base Response.

See the docstrings on `llm.StreamResponse` and `llm.AsyncStreamResponse`.


teamdandelion:
* **#1156** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1156?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1157](https://github.com/Mirascope/mirascope/pull/1157) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1157?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1158](https://github.com/Mirascope/mirascope/pull/1158) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1158?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1156?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1154** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1154?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1153** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1153?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1155** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1155?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1152** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1152?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1151** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1150** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1150?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1149** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1149?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1148** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1148?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1147** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1147?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1145** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1145?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1143** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1143?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1144](https://github.com/Mirascope/mirascope/pull/1144) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1144?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1142** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+
+    async def stream_text(self) -> AsyncIterator[str]:
+        """Streams all of the text content of the StreamResponse, as incremental strings.
+
+        If the response contains multiple distinct Text components, then they will be
+        combined with newlines separating them.
+        """
+        raise NotImplementedError()
+
+    async def streams(self) -> AsyncIterator[AsyncStream]:
```
teamdandelion: Note: Making this async is somewhat inconvenient as it means you have to do:

`async for stream in await response.streams()`, rather than `async for stream in response.streams()`.

However, it makes it much easier to implement the "cleanup" routine, where if you don't break, then streams are fully consumed.

Cleaning up after streams is desirable as it allows you to write code like this:

```
async for stream in await response.streams():
  if stream.content_type == "tool_call":
    pass # Handle it later
  if stream.content_type in ("thinking", "text"):
    # ... stream the text or thought process to console
```

Without cleanup, the user would need to write `stream.collect()` for the tool call rather than skipping it, otherwise the subsequent stream would error. With cleanup, we can automatically finish the stream as it goes out of scope before producing the next one, and we can do so for the last stream after finishing iteration too (e.g. if the final stream was a tool call in the above example).

So I think having the extra await keyword is worth paying in this case.

We don't really need it for `stream_text` however I felt it was better to be consistent, open to changing that.

python/examples/sazed/sazed_async_stream.py
```diff
@@ -19,8 +19,8 @@ async def sazed(query: str):
 
 async def main():
     query = "What are the Kandra?"
-    stream: llm.AsyncStream = await sazed.stream(query)
-    async for chunk in stream:
+    response: llm.AsyncStreamResponse = await sazed.stream(query)
+    async for chunk in await response.stream_text():
```
willbakst: nit: I prefer `text_stream()` over `stream_text()`

python/examples/sazed/sazed_async_stream_agent_context_structured.py
```diff
@@ -35,11 +35,11 @@ async def main():
     coppermind = Coppermind(repository="Ancient Terris")
     agent: llm.AsyncAgent[Coppermind, KeeperEntry] = await sazed(deps=coppermind)
     query = "What are the Kandra?"
-    stream: llm.AsyncStream[KeeperEntry] = await agent.stream(query)
-    async for _ in stream:
-        partial_entry: llm.Partial[KeeperEntry] = stream.format(partial=True)
+    response: llm.AsyncStreamResponse[KeeperEntry] = await agent.stream(query)
+    async for _ in await response.stream_text():
```
willbakst: I find this weird. We're iterating over the text stream and then formatting on the response. It felt less wrong when we were iterating over the chunks for some reason.

I think we should just push this into the stream response as a first class citizen. Something like this:

```python
async for partial_entry in await response.structured_stream():
    print("[Partial]: ", partial_entry, flush=True)
```

We could still have `response.format(partial=True)` available like any response so you can still use it when iterating through the stream and doing other things with it, but I think we should have better first class support for only streaming through partial structured outputs.

WDYT?

python/examples/sazed/sazed_async_stream_tools.py
```diff
@@ -25,20 +25,20 @@ async def sazed(query: str):
...
     query = "What are the Kandra?"
-    stream: llm.AsyncStream = await sazed.stream(query)
+    response: llm.AsyncStreamResponse = await sazed.stream(query)
     while True:
         outputs: list[llm.ToolOutput] = []
-        async for group in stream.groups():
-            if group.type == "text":
-                async for chunk in group:
+        async for stream in await response.streams():
+            if stream.type == "async_text_stream":
```
willbakst: What happened to using `content_type` here? I thought that had better behavior here (while we keep `.type` truly discriminative)?

python/examples/sazed/sazed_async_stream_tools_context_structured.py
```diff
@@ -46,24 +46,24 @@ async def main():
...
+    response: llm.AsyncStreamResponse[KeeperEntry] = await sazed.stream(ctx, query)
     while True:
         outputs: list[llm.ToolOutput] = []
-        async for group in stream.groups():
-            if group.type == "text":
-                async for _ in group:
-                    partial_entry: llm.Partial[KeeperEntry] = stream.format(
+        async for stream in await response.streams():
+            if stream.type == "async_text_stream":
+                async for _ in stream:
```
willbakst: not sure if the templating will update this so adding a comment just in case as a ditto to my other one

python/mirascope/llm/content/content.py
```diff
@@ -1,35 +1,29 @@
...
+from .image import Image, ImageUrl
+from .text import AsyncTextStream, Text, TextChunk, TextStream
+from .thinking import AsyncThinkingStream, Thinking, ThinkingChunk, ThinkingStream
+from .tool_call import AsyncToolCallStream, ToolCall, ToolCallChunk, ToolCallStream
 from .tool_output import ToolOutput
 
+SystemContent: TypeAlias = Text
+"""Content parts that can be included in a SystemMessage."""
+
 UserContent: TypeAlias = (
```
willbakst: nit: whatever happened to renaming these to e.g. `UserContentPart`?

mostly a note, not expecting this change in this PR

python/mirascope/llm/content/text.py
```diff
@@ -15,16 +16,65 @@ class Text:
...
 
+    text: str
+    """The accumulated text content from all received chunks."""
+
     delta: str
-    """The incremental text present in this particular chunk."""
+    """The incremental text added in the most recent chunk."""
+
+
+class TextStream:
```
willbakst: I find this placement a little weird. I feel like we should just have a `streams` module that contains all of the streaming stuff?

Happy for `StreamResponse` to still live in `responses` and depend on the `streams` module.

python/mirascope/llm/content/thinking.py
```diff
@@ -26,19 +26,74 @@ class Thinking:
...
+    """The incremental thinking text added in this update.
+    
+    May be an empty string if the most recent chunk was providing the signature rather
+    than thinking content."""
+
+    redacted: bool = False
+    """Whether the thinking is redacted or not."""
+
+
+class ThinkingStream:
```
willbakst: ditto across the board here. I would argue this isn't content, it's streams

python/mirascope/llm/responses/response.py
```diff
@@ -24,53 +24,69 @@
...
-    """The model identifier that generated the response (e.g. "openai:gpt-4", "anthropic:claude-3-5-sonnet-latet")."""
+    """The model identifier that generated the response (e.g. "openai:gpt-4", "anthropic:claude-3-5-sonnet-latest")."""
 
     args: dict[str, Any]
     """The arguments used to generate the response."""
 
+    messages: list[Message]
+    """The message history, including the most recent assistant message.
+
+    Note: When accessed on a StreamResponse or AsyncStreamResponse, the most recent
```
willbakst: nit: I would just override `messages` on `StreamResponse` and but the corresponding docstring there where it will make more sense.

python/mirascope/llm/responses/response.py
```diff
@@ -24,53 +24,69 @@
...
+    assistant message will have all of the completed content that has already been
+    consumed from the stream. Text content will be included as each chunk is processed;
+    other content will be included only when its corresponding part is completed (to
+    avoid partial tool calls and the like). If no content has been streamed, then the
+    final assistant message will be present (to maintain turn order expectations), but
+    will be empty.
```
willbakst: I don't think this is valid. I believe empty assistant messages will error in most APIs...

python/mirascope/llm/responses/response.py
```diff
@@ -24,53 +24,69 @@
...
+    raw: Any
+    """The raw response from the LLM."""
 
     content: Sequence[AssistantContent]
-    """The content generated by the LLM."""
-
-    texts: Sequence[str]
+    """The content generated by the LLM.
+    
+    Note: When accessed on a StreamResponse or AsyncStreamResponse, content will be
```
willbakst: ditto. I also think there is value in the override because it makes it clear that `StreamResponse` is doing something differently for these attributes

python/mirascope/llm/responses/response.py
```diff
@@ -24,53 +24,69 @@
...
 
     thinkings: Sequence[Thinking]
     """The thinking content in the generated response, if any."""
 
-    finish_reason: FinishReason
-    """The reason why the LLM finished generating a response."""
+    finish_reason: FinishReason | None
+    """The reason why the LLM finished generating a response.
+    
+    Will be None when the response is a StreamResponse or AsyncStreamResponse that
```
willbakst: ditto

python/mirascope/llm/responses/response.py
```diff
@@ -24,53 +24,69 @@
...
     @property
-    def text(self, delimiter: str = "\n") -> str:
+    def text(self) -> str:
         """Returns the response's textual content.
 
         If the response has multiple Text parts, they will be concatenated together
-        using delimiter (default newline).
+        using newlines.
+
+        If this is called on a StreamResponse or AsyncStreamResponse, then text
```
willbakst: I don't know that we need this. I think it's clear from other stuff that calling this will return the concat of text content, and in the case of streaming this will concat whatever text is in the content so far

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+
+    This class supports iteration to process chunks as they arrive from the model.
+
+    Content can be streamed in one of two ways:
+    - Via `response.streams()`, which provides an iterator of `llm.Stream`s, where each
+      stream contains chunks of streamed data. The chunks contain `delta`s (new content
+      in that particular chunk), as well as a partially-reconstructed state of the entire
+      content part to that point.
+    - Via `response.stream_text()` a helper method which provides all of the text content
+      in the response, as str deltas.
```
willbakst: nit: str -> `str`

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+
+    As chunks are streamed from the LLM, they are collected in-memory on the StreamResponse,
+    and they become available in the underlying response properties like `response.messages`,
+    `response.content`, `response.tool_calls`, etc.
+
+    In the specific case of text chunks, they are included in the response content as soon
+    as they become available, via a `llm.Text` part that updates as more deltas come in.
```
willbakst: nit: a -> an

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+    This enables the behavior where resuming a partially-streamed response will include
+    as much text as the model generated.
```
willbakst, line 48: The thing I realized is that there are two means of resuming this stream. The first would be to supply a new user message. The second would be to use prefilling.

The idea with prefilling is that you send an array of messages that ends with an assistant message and the LLM picks up from where it left off (essentially resuming the stream in this case). I think we should support prefilling in the `.resume` API by allowing for an empty user content part. I do also wonder how this would perform with things like continuing from max tokens where you actually just continue without an additional user message. We may also in that case want to merge the new assistant message with the prefilled assistant message in the final response.

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+
+    For other chunks, like `Thinking` or `ToolCall`, they are only added to response
+    content once the corresponding part has fully streamed. This avoids issues like
+    adding incomplete tool calls, or thinking blocks missing signatures, to the response.
+
+    When using the stream iterators (`response.streams()` or `response.stream_text()`),
+    it will by default fully consume the underlying stream.
+    For example, writing `for stream in response.streams(): ...` will consume
```
willbakst: nit: the newline here is a little weird, seems better to break on "writing"

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+    the whole stream, even though `stream.collect()` was not called. Likewise, writing
+    `for chunk in response.stream_text(): ...` will consume all of the content in the
+    stream, even though content types like tool calls are being skipped. However, you
+    can still pause stream execution midway through using the `break` keyword.
+
+    The stream iterators are all replayable. If you call `.streams()` or `.stream_text()`,
```
willbakst: I don't know if this is the desired behavior. If I'm streaming and break because I got a keyword, do some work, and then want to resume the stream from where I left off, I won't want these to restart. Perhaps we can fix this by adding an arg for this like `restart=True` and have it default to `False`?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+            return f"Answer this question: {question}"
+
+        stream_response = answer_question.stream("What is the capital of France?")
+        for chunk in stream_response.stream_text():
+            print(chunk, end="", flush=True)
+        print()
+        ```
+    """
+
+    def stream_text(self) -> Iterator[str]:
```
willbakst: Reiterating my preference for `text_stream()` here. I also wonder if the return type here should be `Iterator[str]` or `Iterator[TextChunk]`? I would prefer the latter so that we can still get partial information? You could also still detect multiple text parts by checking when `partial` is `None` or empty or something (i.e. new part starting).

I think we could add an additional arg like `mode="str"` or something that would then return `Iterator[str]` as opt-in. We could even do something where we stream the repr so that tool calls and thinking are included as their string representations. Could also be opt-in with an arg like `mode="repr"`

WDYT?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+        """Streams all of the text content of the StreamResponse, as incremental strings.
+
+        If the response contains multiple distinct Text components, then they will be
+        combined with newlines separating them.
+
+        Fully iterating through this iterator will fully consume the underlying stream,
+        updating the Response with all collected content (Text and other).
+
+        When called, stream_text always returns a fresh iterator that replays text
```
willbakst: ditto on this re: my other comment

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
...
+        updating the Response with all collected content.
+
+        When called, streams always returns a fresh iterator that replays content
+        streams starting at the beginning of the stream.
+        """
+        raise NotImplementedError()
+
+
+class AsyncStreamResponse(Response[FormatT]):
+    """A Response that wraps around asynchronously streamed data from the LLM.
```
willbakst: note: we'll need to update this alongside any updates to the sync version. kind of annoying but I figure you already thought about this

python/mirascope/llm/responses/stream_response.py
willbakst: I actually prefer this because it future proofs us in case we want to do anything async inside of this method at any point.

I also think we could slightly clean it up by pushing it outside of the iterator:

```python
streams = await response.streams()
async for stream in streams:
    ...
```

teamdandelion: Sg, I'll try rewriting it that way.

python/examples/sazed/sazed_async_stream_tools.py
teamdandelion: Good catch, I added content_type but forgot to update the sazed examples. Fixed

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,213 @@
+"""Responses that stream content from LLMs."""
+
+from collections.abc import AsyncIterator, Iterator
+
+from ..content import AsyncStream, Stream
+from ..formatting import FormatT
+from .response import Response
+
+
+class StreamResponse(Response[FormatT]):
```
willbakst: What about making this generic on the stream type? Something like:

```python
StreamT = TypeVar("StreamT", bound=Stream | AsyncStream)

class StreamResponse(Response[FormatT], Generic[StreamT, FormatT]):
    ...
    @overload
    def streams(self: "StreamResponse[Stream, FormatT]") -> Iterator[Stream]: ...

    @overload
    def streams(self: "StreamResponse[AsyncStream, FormatT]") -> Coroutine[AsyncIterator[AsyncStream], Any, Any]: ...

    def streams(self: "StreamResponse[StreamT, FormatT]") -> Iterator[Stream] | Coroutine[AsyncIterator[AsyncStream], Any, Any]: ...
```

willbakst: `llm.StreamResponse[llm.Stream]`
`llm.StreamResponse[llm.AsyncStream]`

willbakst: I think I prefer `Awaitable[StreamResponse[AsyncStream]]` over `Awaitable[AsyncStreamResponse]`

python/examples/sazed/sazed_async_stream_agent_context_structured.py
teamdandelion: Added response.structured_stream().

python/mirascope/llm/content/content.py
teamdandelion: Ack, will do in future PR.

python/mirascope/llm/content/text.py
teamdandelion: Done

python/mirascope/llm/responses/response.py
teamdandelion: Done

python/mirascope/llm/responses/response.py
teamdandelion: Discussed offline, we're going to keep it this way.

python/mirascope/llm/responses/stream_response.py
teamdandelion: Ack, going to allow resuming with no new content in a followon PR

python/mirascope/llm/responses/stream_response.py
teamdandelion: That is a lot more state for the response to track.

If the user wants to continue from where they left off, they can just keep a reference to the existing iterator, no?

```py
streams = response.streams()
current_stream = stream
for stream in streams:
  current_stream = stream
  if break_condition(stream):
    break

do_things()
for chunk in current_stream:
  # Can finish up on the stream we broke on if desired

for stream in streams:
  # continue using the iterator we broke on
```

python/mirascope/llm/responses/stream_response.py
teamdandelion: Updated to return a TextStream, and modified Response so that `.text` returns `Text | None` for consistency. They will all be unified as a single part, with combined partials not distinct partials for distinct parts (if you want that then use the streams() api). 

I'm open to the idea of including reprs of tool calls and thinking and such, if so I think we should do that on `response.text` too? I'll make a note to follow up on these semantics, there's already enough going on here.

python/mirascope/llm/responses/stream_response.py
teamdandelion: ditto other reply

python/mirascope/llm/responses/stream_response.py
teamdandelion: Dropped the AsyncStreamResponse :)

python/mirascope/llm/responses/stream_response.py
teamdandelion: Done!

python/examples/sazed/sazed_async_stream_tools.py
```diff
@@ -25,20 +25,17 @@ async def sazed(query: str):
...
-                async for chunk in group:
-                    print(chunk, flush=True, end="")
-                print()
-            if group.type == "tool_call":
-                tool_call = await group.collect()
-                outputs.append(await sazed.toolkit.execute(tool_call))
-        if not outputs:
+        async for chunk in await response.text_stream():
+            print(chunk, flush=True, end="")
+        tool_calls = response.tool_calls
```
willbakst: we will also likely want examples of streaming tool call chunks (e.g. so you can render them in a UI as they are streamed or something)

python/examples/sazed/sazed_stream_tools_context_structured.py
```diff
@@ -45,24 +45,17 @@ def main():
...
-                entry: KeeperEntry = stream.format()
-                print("[Final]: ", entry)
-            if group.type == "tool_call":
-                tool_call = group.collect()
-                outputs.append(sazed.toolkit.execute(ctx, tool_call))
-        if not outputs:
+        for chunk in response.structured_stream():
+            print("[Partial]: ", chunk, flush=True)
+        tool_calls = response.tool_calls
+        if not tool_calls:
```
willbakst: nit: I like the `if not (tool_calls := response.tool_calls): ...` assignment during check structure. WDYT?

python/mirascope/llm/calls/base_call.py
```diff
@@ -38,17 +38,21 @@ class BaseCall(
 ):
     async def resume_async(
         self,
-        output: Response[FormatT] | BaseStream,
+        response: Response[FormatT]
+        | StreamResponse[Stream, FormatT]
```
willbakst: is it not possible to unify these with `StreamResponse[Stream | AsyncStream, FormatT]`?

python/mirascope/llm/content/boundary_chunk.py
```diff
@@ -0,0 +1,37 @@
+"""Boundary chunk classes for marking stream start and end points."""
+
+from dataclasses import dataclass
+from typing import Literal, TypeAlias
+
+StreamableContentType: TypeAlias = Literal["text", "thinking", "tool_call"]
```
willbakst, line 6: nit: `StreamableContentPartType`?

python/mirascope/llm/content/__init__.py
```diff
@@ -1,15 +1,15 @@
 """The `llm.messages.content` module."""
 
-from .audio import Audio, AudioChunk
-from .chunk_boundary import ChunkEnd, ChunkStart
+from .audio import Audio
+from .boundary_chunk import EndChunk, StartChunk, StreamableContentType
```
willbakst, line 4: what is the benefit of exporting `StreamableContentType` at this level? Do we use it elsewhere?

python/mirascope/llm/content/content.py
```diff
@@ -1,35 +1,23 @@
...
-    TextChunk
-    | ImageChunk
-    | AudioChunk
-    | ToolCallChunk
-    | ThinkingChunk
-    | ChunkStart
-    | ChunkEnd
-)
-"""Content chunk types that may be emitted by Streams."""
+Chunk: TypeAlias = TextChunk | ToolCallChunk | ThinkingChunk
```
willbakst: nit: if this is specific to assistant responses, should it be more specific like `AssistantContentPartChunk` or something? when we update to `AssistantContentPart` this would be clear?

python/mirascope/llm/content/text.py
```diff
@@ -15,16 +15,15 @@ class Text:
...
-    delta: str
-    """The incremental text present in this particular chunk."""
+    type: Literal["text_chunk"] = "text_chunk"
 
-    final: bool
-    """Whether this is the final piece of content in its sequence. If true, this content's partial is finished generating."""
+    text: str
+    """The accumulated text content from all received chunks."""
 
-    def __repr__(self) -> str:
```
willbakst, line 28: hmm, I wonder if we should still include these as placeholders with `NotImplementedError`? Could be in a follow-up PR where we add `__repr__` placeholders to every class, which I think we should implement long term. No reason not to.

python/mirascope/llm/streams/format_stream.py
willbakst, line 1: empty file?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+        ```
+    """
+
+    content: Sequence[AssistantContent]
+    """The content generated by the LLM.
+    
+    Content is updated in this array as it is consumed by the stream. Text content will 
+    update with each text chunk (this will mutate the Text object that is returned 
+    rather than creating a new one). Other content will be added once each part
+    is fully streamed."""
```
willbakst: nit: noticing this here, but the last """ should be on a newline. please fix here and elsewhere.

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+        updating the Response with all collected content (Text and other).
+
+        When called, text_stream always returns a fresh iterator that replays text
+        chunks starting at the beginning of the stream.
+        """
+        raise NotImplementedError()
+
+    @overload
+    def streams(self: "StreamResponse[Stream, FormatT]") -> Iterator[Stream]: ...
+
```
willbakst: nit: erroneous newline?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+    """Overload for synchronous streaming."""
+
+    @overload
+    def streams(
+        self: "StreamResponse[AsyncStream, FormatT]",
+    ) -> Awaitable[AsyncIterator[AsyncStream]]: ...
+
```
willbakst: ditto

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+    @overload
+    def text_stream(self: "StreamResponse[Stream, FormatT]") -> TextStream: ...
+
+    @overload
+    def text_stream(
+        self: "StreamResponse[AsyncStream, FormatT]",
+    ) -> Awaitable[AsyncTextStream]: ...
+
+    def text_stream(self) -> TextStream | Awaitable[AsyncTextStream]:
+        """Get a text stream containing all the text content from the response.
```
willbakst: nit: google docstring style here needs to be `Returns a text...`

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+
+    @overload
+    def streams(
+        self: "StreamResponse[AsyncStream, FormatT]",
+    ) -> Awaitable[AsyncIterator[AsyncStream]]: ...
+
+    """Overload for asynchronous streaming."""
+
+    def streams(self) -> Iterator[Stream] | Awaitable[AsyncIterator[AsyncStream]]:
+        """Get an iterator that yields streams for each content part in the response.
```
willbakst: ditto

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+        content from the response, and (if there is still more content to consume from
+        the LLM), it will proceed to consume it once it has iterated through all the
+        cached chunks.
+        """
+        raise NotImplementedError()
+
+    @overload
+    def structured_stream(
+        self: "StreamResponse[Stream, FormatT]",
+    ) -> Iterator[Partial[FormatT]]: ...
```
willbakst: I'm assuming the idea around `StructuredChunk` will be in a following PR? makes sense, just confirming so I don't comment on other things

python/mirascope/llm/responses/stream_response.py
```diff
@@ -0,0 +1,255 @@
...
+
+    @overload
+    def structured_stream(
+        self: "StreamResponse[AsyncStream, FormatT]",
+    ) -> Awaitable[AsyncIterator[Partial[FormatT]]]: ...
+
+    def structured_stream(
+        self,
+    ) -> Iterator[Partial[FormatT]] | Awaitable[AsyncIterator[Partial[FormatT]]]:
+        """Get an iterator that yields partial structured objects as content streams.
```
willbakst: ditto

python/mirascope/llm/streams/text_stream.py
```diff
@@ -0,0 +1,61 @@
...
+    """Synchronous text stream implementation."""
+
+    type: Literal["text_stream"] = "text_stream"
+
+    @property
+    def content_type(self) -> Literal["text"]:
+        """The type of content this stream handles."""
+        return "text"
+
+    def __iter__(self) -> Iterator[TextChunk]:
```
willbakst, line 20: this will need an `__init__` method in order to receive the stream that it will consume, right?

python/mirascope/llm/responses/response.py
```diff
@@ -1,117 +1,14 @@
...
-        The resulting string includes all raw text directly, and includes placeholder
-        representations for embedded media, eg {image: url=...} or {thinking: thoughts=...}
-
-        Each content piece will be separated by newlines.
-        """
-        raise NotImplementedError()
+    finish_reason: FinishReason | None
+    """The reason why the LLM finished generating a response.
+    
+    Will be None when the response is a StreamResponse that has not finished streaming."""
```
willbakst: ok noticing this here too.

maybe we should just do a follow-up in this stack that does a sweep for this?

python/mirascope/llm/responses/stream_response.py
teamdandelion: Yup that's the plan.

python/examples/sazed/sazed_async_stream_tools.py
teamdandelion: Ack. I'll maybe add that in with the StructuredStream concept as it will be easier to do so consistently then.

python/examples/sazed/sazed_stream_tools_context_structured.py
teamdandelion: Sgtm, I wasn't sure if that would work in the `if not` case but I see the parens does it

python/mirascope/llm/content/__init__.py
teamdandelion: No, removed. May revisit when I add ContentTypes

python/mirascope/llm/content/boundary_chunk.py
teamdandelion: I think it is a content type not a content part type. Generalizing to having content type more broadly, "audio_url" is a ContentPartType but not a ContentType.

python/mirascope/llm/content/__init__.py
teamdandelion: Actually, we can't remove it. I reference it in base_stream.py as in the base signature for stream.content_type

python/mirascope/llm/calls/base_call.py
teamdandelion: Fixed by making StreamT covariant

python/mirascope/llm/responses/stream_response.py
teamdandelion: Replaced with pretty_stream

python/mirascope/llm/content/content.py
teamdandelion: Ok, I'll update it in the followon PR that touches naming here

python/mirascope/llm/streams/text_stream.py
teamdandelion: Yes. Planning to leave that for implementation though.

python/mirascope/llm/responses/response.py
teamdandelion: I think I fixed them all here, maybe we should have a followon that adds a lint rule.

python/mirascope/llm/content/text.py
teamdandelion: Maybe, I'd rather address this separately from this PR though. If we need a followon to do it across the board we can include these in the sweep then.

python/mirascope/llm/agents/agent.py
```diff
@@ -44,10 +44,10 @@ async def run_async(
 
     async def stream_async(
         self,
-        query: UserMessagePromotable,
+        query: UserContent,
         *,
         ctx: Context[DepsT] | None = None,
-    ) -> AsyncStream[FormatT] | AsyncStream[FormatT]:
+    ) -> StreamResponse[AsyncStream, FormatT] | StreamResponse[AsyncStream, FormatT]:
```
willbakst: is this not duplicated?

python/mirascope/llm/agents/agent.py
```diff
@@ -76,10 +76,10 @@ def call(
 
     def stream(
         self,
-        query: UserMessagePromotable,
+        query: UserContent,
         *,
         ctx: Context[DepsT] | None = None,
-    ) -> Stream[FormatT] | Stream[FormatT]:
+    ) -> StreamResponse[Stream, FormatT] | StreamResponse[Stream, FormatT]:
```
willbakst: ditto

python/mirascope/llm/agents/agent.py
```diff
@@ -108,9 +108,9 @@ async def call(
 
     async def stream(
         self,
-        query: UserMessagePromotable,
+        query: UserContent,
         *,
         ctx: Context[DepsT] | None = None,
-    ) -> AsyncStream[FormatT] | AsyncStream[FormatT]:
+    ) -> StreamResponse[AsyncStream, FormatT] | StreamResponse[AsyncStream, FormatT]:
```
willbakst: ditto

python/mirascope/llm/responses/base_response.py
```diff
@@ -0,0 +1,74 @@
...
+        Returns:
+            The formatted response object of type FormatT.
+
+        Raises:
+            ValueError: If the response cannot be formatted according to the
+                specified format.
+        """
+        raise NotImplementedError()
+
+    def pretty(self, include_content_placeholders: bool = True) -> str:
```
willbakst: nit: not in love with this argument name. I'm wondering if we should make this more configurable to the type? for example, what if the user could specify that they want thinking included but not tool calls?

something like `response.pretty(include=["thinking", "tool_calls"])` if we can make it work?

python/mirascope/llm/responses/stream_response.py
willbakst: I'm actually thinking we keep it clean where `.resume` always requires `UserContent`.

I think prefilling should go through a different route if/when we support it. There's nothing stopping a user from using prefilling at the call level by returning an assistant message as the last message, which is more inline with what prefilling is supposed to do vs. the idea of resuming a stream.

python/mirascope/llm/responses/stream_response.py
willbakst: not seeing this fixed?

python/mirascope/llm/responses/stream_response.py
willbakst: same here, not seeing this fixed

python/mirascope/llm/responses/base_response.py
teamdandelion: Decided offline to revisit, for now we include everything.

willbakst (approved):

teamdandelion:
### Merge activity

* **Aug 1, 12:08 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1156).

PR #1165, opened 2025-08-01
test: start tracking code coverage

teamdandelion:
We'll remove packages from the omit= list as we add tests.
Right now messages and types both have full coverage.

PR #1164, opened 2025-08-01
feat: implement messages with testing

teamdandelion:

PR #1163, opened 2025-08-01
test: ensure mirascope package can be imported

teamdandelion:
Catches circular dependencies, etc

PR #1167, opened 2025-08-02
feat: implement basic call for anthropic client

teamdandelion:
Added smoke tests that hit the real Anthropic API using [vcr.py](https://vcrpy.readthedocs.io/en/latest/)

Updating the tests requires an anthropic api key be set in .env, however once the test results are cached, they can be re-run without issue. Should be just fine in CI.


teamdandelion:
* **#1171** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1170** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1167** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1167?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1167?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1169** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1169?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1168** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1165** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1165?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1164** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1164?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1163** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1163?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1162** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1162?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

teamdandelion:
Note: the vcr.py tests aren't working in CI, will need to look into those

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
         tools: Sequence[Tool] | None = None,
         params: AnthropicParams | None = None,
     ) -> Response[None]:
-        raise NotImplementedError
+        """Make a call to the Anthropic API."""
+        client = Anthropic()
```
willbakst: One thing from `v1` I don't like is that we create a new client connection for every call rather than using a shared client.

We should use a singleton client here so that multiple calls using Anthropic can use the same connection, right?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+            model=model_name,
+            messages=converted_messages,
+            system=system_message or NotGiven(),
+        )
+
+        usage = response.usage
+
+        converted_usage = Usage(
+            input_tokens=usage.input_tokens,
+            cached_tokens=usage.cache_read_input_tokens or 0,
```
willbakst: hmm, I'm realizing we may want to make our own `Usage` class more specific on `cached_tokens` since there might be different types of cached tokens. Not exactly sure what that would be. Would need to look at what all the providers call their respective versions of this. One thing I do know is that OpenAI has cached audio tokens that I believe are separate. In fact I think all audio tokens are separate.

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+            output_tokens=usage.output_tokens,
+            total_tokens=usage.input_tokens
+            + (usage.cache_read_input_tokens or 0)
+            + usage.output_tokens,
+        )
+
+        # Convert finish reason
+        finish_reason_map = {
```
willbakst: I'm assuming you'll DRY this once you start implementing additional methods?

As part of that, we should definitely define this globally once (i.e. `ANTHROPIC_FINISH_REASON_MAP`)

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+
+        text_content: list[Text] = []
+
+        for content_block in response.content:
+            if content_block.type == "text":
+                text_part = Text(text=content_block.text)
+                text_content.append(text_part)
+            else:
+                raise ValueError(
+                    f"Received unsupported content type: {content_block.type}"
```
willbakst: Is this just because we aren't supporting anything else yet, or because we're expecting this to be a text-only call at this path in the client methods?

I wonder if we should drop the content and raise a warning rather than throw a `ValueError` and potentially crash?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+            "refusal": FinishReason.REFUSAL,
+        }
+
+        finish_reason = finish_reason_map.get(
+            response.stop_reason or "unknown", FinishReason.UNKNOWN
+        )
+
+        text_content: list[Text] = []
+
+        for content_block in response.content:
```
willbakst: Same thing here re: DRY (I assume you'll convert this into a converter method that's shared across the various methods?)

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+
+        response_messages = []
+        for message in messages:
+            if isinstance(message, dict):
+                response_messages.append(message_from_anthropic(message))
+            else:
+                response_messages.append(message)
+        response_messages.append(final_message)
+
+        return Response[None](
```
willbakst: do you have to put `None` here if it's the default?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+                text_part = Text(text=content_block.text)
+                text_content.append(text_part)
+            else:
+                raise ValueError(
+                    f"Received unsupported content type: {content_block.type}"
+                )
+        final_message = assistant(text_content)
+
+        response_messages = []
+        for message in messages:
```
willbakst: Just checking re: our discussion that we're going to replace `Sequence[AnthropicMessage]` with `Sequence[Message]`?

python/mirascope/llm/clients/anthropic/message.py
willbakst, line 1: nit: on the file name I think we should update to `messages.py` now that it has utilities?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+        if params:
+            raise NotImplementedError("param use not yet supported")
+
+        # Convert model to just the model name (remove "anthropic:" prefix if present)
+        model_name = model.split(":")[-1] if ":" in model else model
+
+        # Convert messages to MessageParam format
+        converted_messages, system_message = messages_to_anthropic(messages)
+
+        response = client.messages.create(
```
willbakst: nit: for clarity let's name this `anthropic_response`

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+)
 
-from ...messages import Message
+from ...content import AssistantContentPart, ContentPart, Text, UserContentPart
+from ...messages import AssistantMessage, Message, SystemMessage, UserMessage
 
 AnthropicMessage: TypeAlias = Message | MessageParam
+
+
+def content_to_anthropic(
```
willbakst: nit: I don't know that I love this naming

I'm wondering if something more explicit like `convert_content_to_anthropic` would read better?

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+        | RedactedThinkingBlockParam
+    ],
+) -> Sequence[ContentPart]:
+    if isinstance(content, str):
+        return [Text(text=content)]
+
+    result = []
+    for block in content:
+        # Check if it's a ContentBlock (has attributes) vs TypedDict (dictionary access)
+        if isinstance(block, dict):
```
willbakst: I wonder if this would read better if you first did `isinstance(block, ContentBlock)`? Then you wouldn't really need these comments because it would be clear from the method type hint that all types that aren't `ContentBlock` are handled differently (as dictionaries)

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+def message_from_anthropic(message_param: MessageParam) -> Message:
+    """Convert a MessageParam to Message format.
+
+    Args:
+        message_param: MessageParam to convert
+
+    Returns:
+        Converted Message object
+    """
+    role = message_param["role"]
```
willbakst: nit: let's just make this one line so it's more like an unpack:

```python
role, content = message_param["role"], message_param["content"]
```

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+def extract_system_message(messages: Sequence[AnthropicMessage]) -> str | None:
+    """Extract system message content from a sequence of messages.
+
+    Args:
+        messages: Sequence of messages to search
+
+    Returns:
+        System message content if found, None otherwise
+    """
+    for message in messages:
```
willbakst: I wonder if we should just take the stance for provider-agnosticity that there will only ever be one system message and it will be the first message? Then if the first message isn't a system message there is no system message?

I know OpenAI technically allows this at a type level, but I believe the behavior is undefined, and I don't think any other provider supports this

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+def messages_from_anthropic(messages: Sequence[MessageParam]) -> list[Message]:
+    """Convert Anthropic MessageParam format to Message format.
+
+    Args:
+        messages: Sequence of MessageParam to convert
+
+    Returns:
+        List of converted Message objects
+    """
+    return [message_from_anthropic(msg) for msg in messages]
```
willbakst: nit: change `msg` to `message` please

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+        return UserMessage(content=cast(Sequence[UserContentPart], converted_content))
+    elif role == "assistant":
+        return AssistantMessage(
+            content=cast(Sequence[AssistantContentPart], converted_content)
+        )
+    else:
+        raise ValueError(f"Unknown role: {role}")
+
+
+def extract_system_message(messages: Sequence[AnthropicMessage]) -> str | None:
```
willbakst: nit: update to `extract_system_message_content`? Otherwise I would expect the return type to be `SystemMessage | None` and not `str | None`

Alternatively you could have it return `SystemMessage | None` which might be better?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -27,7 +35,85 @@ def call(
...
+            else:
+                raise ValueError(
+                    f"Received unsupported content type: {content_block.type}"
+                )
+        final_message = assistant(text_content)
+
+        response_messages = []
+        for message in messages:
+            if isinstance(message, dict):
+                response_messages.append(message_from_anthropic(message))
```
willbakst: why do you use the singular `message_from_anthropic` here rather than `messages_from_anthropic`?

python/mirascope/llm/responses/base_response.py
```diff
@@ -15,7 +15,7 @@
     )
 
 
-@dataclass
+@dataclass(kw_only=True)
```
willbakst: I like the idea of making all of our dataclasses `kw_only=True` so it's always explicit and clear.

Thinking it would be worthwhile to do a sweep in a follow-up PR

python/tests/llm/clients/anthropic/test_anthropic_client.py
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+
+@pytest.fixture
+def anthropic_client():
+    """Create an AnthropicClient instance."""
+    return llm.clients.AnthropicClient()
+
+
+class TestAnthropicClient:
```
willbakst: do we need to nest these in a class?

python/tests/llm/clients/anthropic/test_anthropic_client.py
```diff
@@ -0,0 +1,78 @@
...
+    """Test cases for AnthropicClient."""
+
+    def test_call_simple_message(self, anthropic_client, vcr_config):
+        """Test basic call with a simple user message."""
+        with vcr.use_cassette("test_call_simple_message.yaml", **vcr_config):
```
willbakst: This is epic haha. So so so much better than mocking.

We may also want to look into something like `inline-snapshot` if it plays nicely with `vcr`.

willbakst: Is there a way for this to automatically use the test method name, or do we always have to explicitly match it?

python/tests/llm/clients/anthropic/test_anthropic_client.py
```diff
@@ -0,0 +1,78 @@
...
+from pathlib import Path
+
+import pytest
+import vcr
+from dotenv import load_dotenv
+
+from mirascope import llm
+
+# Load environment variables from .env
+load_dotenv()
```
willbakst: rather than loading environment variables, just put an `__init__.py` in `python/tests/llm/clients/anthropic` and explicitly set `os.environ["..."] = "..."` so that we don't have to worry about tests relying on environment files

python/mirascope/llm/clients/anthropic/message.py
```diff
@@ -1,9 +1,171 @@
...
+
+    Returns:
+        Converted Message object
+    """
+    role = message_param["role"]
+    content = message_param["content"]
+    converted_content = content_from_anthropic(content)
+
+    if role == "user":
+        return UserMessage(content=cast(Sequence[UserContentPart], converted_content))
```
willbakst: This cast seems unsafe. We're assuming `converted_content` will be the correct type, but we don't actually confirm that anywhere.

I'm wondering if we could improve this by overloading `content_from_anthropic` on the content types. We know that assistant messages will only have `ContentBlock` and user messages will have the `str` or other `{...}Param` types, so we should be able to overload on that and properly type?

python/mirascope/llm/clients/anthropic/client.py
teamdandelion: For now we're only supporting text, but this path will support all content types. Changed the ValueError to NotImplementedError for clarity.

python/mirascope/llm/clients/anthropic/message.py
teamdandelion: Renamed to conversions.py

python/mirascope/llm/clients/anthropic/message.py
teamdandelion: Refactored out of existence.

python/mirascope/llm/clients/anthropic/message.py
teamdandelion: I renamed the module and changed to an "encode" -> to Anthropic specific types and "decode" -> from anthropic types naming pattern.

python/mirascope/llm/clients/anthropic/message.py
teamdandelion: Refactored into simpler and clearer arrangement.

python/tests/llm/clients/anthropic/test_anthropic_client.py
teamdandelion: No, removed.

python/tests/llm/clients/anthropic/test_anthropic_client.py
teamdandelion: To clarify, we're loading API keys which I don't want to commit.

python/tests/llm/clients/anthropic/test_anthropic_client.py
teamdandelion: Yeah turns out there's a pytest plugin with a decorator. Updated to use that.

teamdandelion: Done

python/tests/llm/clients/anthropic/test_anthropic_client.py
teamdandelion: I rewrote this a bit to make it clearer what's happening.

python/mirascope/llm/clients/base/client.py
```diff
@@ -21,7 +21,7 @@
 from .params import ParamsT
 
 if TYPE_CHECKING:
-    from ..registered_llms import LLMT
+    from ..registered_llms import LLMT  # pragma: no cover
```
willbakst: nit: better to ignore `if TYPE_CHECKING: ...` blocks across the board in our coverage setup so we don't have to do this every time everywhere

python/tests/llm/clients/anthropic/test_anthropic_client.py
```diff
@@ -0,0 +1,125 @@
...
+        "filter_post_data_parameters": [],
+    }
+
+
+@pytest.fixture
+def anthropic_client():
+    """Create an AnthropicClient instance with appropriate API key."""
+    # Use real API key if available, otherwise dummy key for VCR tests
+    load_dotenv()
+    api_key = os.getenv("ANTHROPIC_API_KEY") or "dummy-key-for-vcr-tests"
```
willbakst, line 36: One confusion I have here is what happens when we have cassettes but also our actual API key set in the environment. Does it ignore the API key and use the cassette?

I understand we need `"dummy-key-for-vcr-tests"` so that we don't fail on no API key in the event that we're using cassettes, so I'm really just asking about the single case outlined above.

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,97 @@
...
+AnthropicMessageParam: TypeAlias = anthropic_types.MessageParam
+AnthropicSystemMessage: TypeAlias = str | None
+AnthropicMessagesInput: TypeAlias = tuple[
+    AnthropicSystemMessage, list[AnthropicMessageParam]
+]
+
+AnthropicMessage: TypeAlias = anthropic_types.Message
+
+
+def encode_content(
```
willbakst: I wonder if it would make the most sense to create a class like `BaseConverter` that has abstract class methods like `encode_content` and `decode_content` etc.

My thinking here is that I don't love that we will likely have `encode_content` as a duplicated method name across all providers, and something like `AnthropicConverter.encode_content` namespaces the method and makes it more clear.

We can also then just import `AnthropicConverter` in places and use whatever methods we want from it? It would live at `python/mirascope/llm/clients/anthropic/converter.py`.

Also, as class methods, we could do `AnthropicConvert.encode_content(content)` without actually instantiating the class, so we get organizational benefits with no additional runtime overhead

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,97 @@
...
+        message_param: AnthropicMessageParam = {
+            "role": message.role,
+            "content": encode_content(message.content),
+        }
+        converted_messages.append(message_param)
+
+    return system_message, converted_messages
+
+
+def decode_message(message: AnthropicMessage) -> AssistantMessage:
```
willbakst: I'm still confused why we have `decode_message` singular and `encode_messages` plural? Why don't these match?

In line with my other suggestion around `AnthropicConverter`, we could always have `encode_messages` and `decode_messages` implemented on `BaseConverter` where they just use the `encode_message` and `decode_message` methods (which would be abstract on the base class and implemented on the provider subclasses).

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,97 @@
...
+    match content.type:
+        case "text":
+            return Text(text=content.text)
+        case _:
+            raise NotImplementedError(
+                f"Support for content type `{content.type}` is not yet implemented."
+            )
+
+
+def decode_finish_reason(reason: anthropic_types.StopReason) -> FinishReason:
```
willbakst: We will also likely want an `encode_finish_reason` method so that we can encode a user provided finish reason into the provider-specific method.

You'll get to this when implementing `Params` so can safely ignore in this PR I believe, just bringing it to attention.

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -10,13 +12,23 @@
...
 from .params import AnthropicParams
 from .registered_llms import ANTHROPIC_REGISTERED_LLMS
 
 
 class AnthropicClient(BaseClient[AnthropicParams, ANTHROPIC_REGISTERED_LLMS]):
     """The client for the Anthropic LLM model."""
 
+    _client: Anthropic
+
+    def __init__(self, *, api_key: str | None = None) -> None:
```
willbakst: we should also pass through `base_url` at the very least if not other parameters?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -25,7 +37,39 @@ def call(
...
+        """Make a call to the Anthropic API."""
+
+        if tools:
+            raise NotImplementedError("tool use not yet supported")
+
+        if params:
+            raise NotImplementedError("param use not yet supported")
+
+        # Convert model to just the model name (remove "anthropic:" prefix if present)
+        model_name = model.split(":")[-1] if ":" in model else model
```
willbakst, line 50: noting previous discussion around splitting this into `provider` and `model`. Is the plan to handle that in a separate PR?

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -25,7 +37,39 @@ def call(
...
+
+        anthropic_response = self._client.messages.create(
+            max_tokens=1024,
+            model=model_name,
+            messages=converted_messages,
+            system=system_message or NotGiven(),
+        )
+
+        # Convert finish reason
+        assert anthropic_response.stop_reason
```
willbakst: Bad practice to have uncaught assertions. I think better to do something like set it to `"UNKNOWN"` in the event that we can't get it / convert it for whatever reason?

We could also do something similar to the model string typing where we allow finish reason to be a `str` type on top of its literal type in case providers release additional finish reasons?

python/tests/llm/clients/anthropic/test_anthropic_client.py
teamdandelion: In the block above, in VCR config, we have "record_mode": "once". What that means is that if cassettes are present, then it will "use" the API key (in that the API key gets accessed when constructing the Anthropic client), however all of the HTTP requests will get intercepted and replayed from the cassette, so no calls to Anthropic will be made.

We need the "dummy-key-for-vcr-tests" because the Anthropic client does runtime checking that a key is present, so if there is no key, then Anthropic will raise an error even though had it attempted to make the requests, they would have (from its perspective) still worked.

python/mirascope/llm/clients/anthropic/conversions.py
teamdandelion: I implemented pretty specifically what we needed, namely:

`encode_messages`: Take the list of Mirascope messages, and encode them into a format that we can call with Anthropic

`decode_message`: Take the assistant message that Anthropic generated, and convert it into the Mirascope AssistantMessage that we will add to Messages.

It's asymmetrical because we never need to take a list of Anthropic MessageParams to a list of Mirascope Messages. 

Could rename decode to `decode_assistant_message` if you feel that's clearer?

python/mirascope/llm/clients/anthropic/client.py
teamdandelion: Yep, planning that for a followon PR shortly.

python/mirascope/llm/clients/anthropic/conversions.py
teamdandelion: I'm open to this, my preference would be to leave it like this for now, and abstract it out once we have 2-3 cases to abstract over.

python/mirascope/llm/clients/anthropic/client.py
teamdandelion: Ok, I made it UNKNOWN if unset (per the anthropic docs, in non-streaming case, it will always be set, so this should never happen anyway)

For now an unexpected finish reason becomes unknown. I'm open to your idea of allowing addl. finish reasons, but I'd rather revisit and not make that change in this PR.

python/mirascope/llm/clients/anthropic/client.py
teamdandelion: Ok, done for base_url. Will revist other parameters later.

python/mirascope/llm/clients/anthropic/conversions.py
willbakst: Yeah I think I prefer `decode_assistant_message` so it's more clear what's happening.

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def decode_message(message: AnthropicMessage) -> AssistantMessage:
+    return assistant(content=[decode_content(part) for part in message.content])
+
+
+def decode_content(
+    content: anthropic_types.ContentBlock,
+) -> AssistantContentPart:
+    match content.type:
```
willbakst: honestly in python I'm somewhat against these switch statements where we return because they add an additional level of indentation.

```python
def decode_content(
    content: anthropic_types.ContentBlock,
) -> AssistantContentPart:
    if content.type == "text":
        return Text(text=content.text)
    raise Not ImplementedError("...")
```

When we add more cases, we can just add more conditionals since we always return:

```python
def decode_content(
    content: anthropic_types.ContentBlock,
) -> AssistantContentPart:
    if content.type == "text":
        return Text(text=content.text)
    if content.type == "tool_call":
        return ...
    raise Not ImplementedError("...")
```

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,99 @@
...
+        case "text":
+            return Text(text=content.text)
+        case _:
+            raise NotImplementedError(
+                f"Support for content type `{content.type}` is not yet implemented."
+            )
+
+
+def decode_finish_reason(reason: anthropic_types.StopReason | None) -> FinishReason:
+    if not reason:
```
willbakst: hmm, is this even necessary?

wouldn't the `finish_reason_map.get(reason, FinishReason.UNKNOWN)` return `FinishReason.UNKNOWN` in the event that `reason` is `None`?

If anything, I think my preference would be something like this:

```python
ANTHROPIC_FINISH_REASON_MAP = {
    "end_turn": FinishReason.END_TURN,
    "max_tokens": FinishReason.MAX_TOKENS,
    "stop_sequence": FinishReason.STOP,
    "tool_use": FinishReason.TOOL_USE,
    "refusal": FinishReason.REFUSAL,
}

def decode_finish_reason(reason: anthropic_types.StopReason | None) -> FinishReason:
    """Decodes Anthropic's `StopReason` into Mirascope's `FinishReason`."""
    return ANTHROPIC_FINISH_REASON_MAP.get(reason, FinishReason.UNKNOWN)
```

Note that I'm refactoring the finish reason map out into a global constant so that we don't redefine it on every call to this function.

python/mirascope/llm/clients/anthropic/conversions.py
willbakst: Ok. One benefit we get from doing it now is that we define the abstract methods we know we'll need to implement later, and if we later need to add more, it will remind us to revisit Anthropic.

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,99 @@
...
+) -> AnthropicMessagesInput:
+    """Convert messages to Anthropic MessageParam format.
+
+    Args:
+        messages: Sequence of messages to convert
+
+    Returns:
+        Tuple of (system_message, user_or_assistant_messages)
+    """
+
```
willbakst: nit: extra newline here should be removed

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,99 @@
...
+AnthropicMessagesInput: TypeAlias = tuple[
+    AnthropicSystemMessage, list[AnthropicMessageParam]
+]
+
+AnthropicMessage: TypeAlias = anthropic_types.Message
+
+
+def encode_content(
+    content: Sequence[ContentPart],
+) -> str | Sequence[anthropic_types.ContentBlock]:
```
willbakst: nit: this and all other methods need docstrings

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,99 @@
...
+        messages: Sequence of messages to convert
+
+    Returns:
+        Tuple of (system_message, user_or_assistant_messages)
+    """
+
+    system_message = None
+    converted_messages: list[AnthropicMessageParam] = []
+    for i, message in enumerate(messages):
+        if i == 0 and message.role == "system":
```
willbakst: I don't love the use of continue here. I would consider writing this as follows:

```python
system_message: AnthropicSystemMessage = None
converted_messages: list[AnthropicMessageParam] = []
for i, message in enumerate(messages):
    if message.role != "system":
        converted_messages.append({
            "role": message.role,
            "content": encode_content(message.content)
        })
    elif i == 0:
        system_message = message.content.text
    else:
        logging.warning("Non-first system message at index %d is being skipped", i)
return system_message, converted_messages
```

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,99 @@
...
+        else:
+            raise NotImplementedError(
+                f"Have not implemented conversion for ${item.type}"
+            )
+    return result
+
+
+def encode_messages(
+    messages: Sequence[Message],
+) -> AnthropicMessagesInput:
```
willbakst: I don't love `AnthropicMessagesInput` as a type alias because it's not clear it's a tuple and doesn't read as one.

I'm assuming this is the only place we use this alias, I would prefer we just use the type directly here:

```python
def encode_messages(
    messages: Sequence[Message],
) -> tuple[AnthropicSystemMessage, list[AnthropicMessageParam]]:
    ...
```

This makes the return type of this function significantly more clear at a glance IMO.

python/mirascope/llm/clients/anthropic/conversions.py
teamdandelion: Hmm, good point. Done.

python/mirascope/llm/clients/anthropic/conversions.py
teamdandelion: I made it `return FINISH_REASON_MAP.get(reason or "", FinishReason.UNKNOWN)` to satisfy type checking

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -25,7 +40,43 @@ def call(
         tools: Sequence[Tool] | None = None,
         params: AnthropicParams | None = None,
     ) -> Response[None]:
-        raise NotImplementedError
+        """Make a call to the Anthropic API."""
+
```
willbakst: nit: remove newline

python/mirascope/llm/clients/anthropic/conversions.py
```diff
@@ -0,0 +1,106 @@
...
+from ...messages import AssistantMessage, Message, assistant
+from ...responses import FinishReason
+from ..base.converter import BaseConverter
+
+AnthropicMessageParam: TypeAlias = anthropic_types.MessageParam
+AnthropicSystemMessage: TypeAlias = str | None
+
+AnthropicMessage: TypeAlias = anthropic_types.Message
+
+FINISH_REASON_MAP = {
```
willbakst: nit: I feel like this should be `ANTHROPIC_FINISH_REASON_MAP` given we namespace everything else with `Anthropic{...}`?

python/mirascope/llm/clients/base/converter.py
```diff
@@ -0,0 +1,45 @@
...
+from ...messages import AssistantMessage, Message
+from ...responses import FinishReason
+
+
+class BaseConverter(ABC):
+    """Base abstract converter for provider-specific message conversions."""
+
+    @classmethod
+    @abstractmethod
+    def encode_content(cls, content: Sequence[ContentPart]) -> Any:  # noqa: ANN401
```
willbakst: nit: should we make the class generic rather than using `Any` for all of these methods?

python/mirascope/llm/clients/anthropic/conversions.py
willbakst, line 1: nit: rename file to `anthropic/converter.py`

python/tests/llm/clients/anthropic/test_conversions.py
willbakst, line 1: nit: inline with other comment, rename to `test_converter.py`

python/mirascope/llm/clients/anthropic/converter.py
```diff
@@ -0,0 +1,116 @@
...
+            raise NotImplementedError(
+                f"Support for content type `{content.type}` is not yet implemented."
+            )
+
+    @classmethod
+    def decode_finish_reason(
+        cls, reason: anthropic_types.StopReason | None
+    ) -> FinishReason:
+        """Convert Anthropic stop reason to mirascope FinishReason."""
+        return ANTHROPIC_FINISH_REASON_MAP.get(reason or "", FinishReason.UNKNOWN)
```
willbakst: nit: reading this again, I'm realizing that I would prefer the following since it's more clear and technically safer (since `""` could technically be a key):

```python
if not reason or reason not in ANTHROPIC_FINISH_REASON_MAP:
    return FinishReason.UNKNOWN
return ANTHROPIC_FINISH_REASON_MAP[reason]
```

python/mirascope/llm/clients/base/converter.py
```diff
@@ -0,0 +1,65 @@
...
+
+class BaseConverter(
+    Generic[
+        ProviderContentT,
+        ProviderMessagesInputT,
+        ProviderMessageResponseT,
+        ProviderContentBlockT,
+        ProviderFinishReasonT,
+    ],
+    ABC,
```
willbakst: nit: in line with other places, I believe `ABC` is supposed to come before `Generic`?

willbakst (approved):

python/mirascope/llm/clients/anthropic/converter.py
willbakst: Hmm actually thinking on this again, I think the safest and most performant solution would be this:
```python
if reason is None:
    return FinishReason.UNKNOWN
return ANTHROPIC_FINISH_REASON_MAP.get(reason, FinishReason.UNKNOWN)
```

Closer to what was originally written lol. I forgot about the weirdness with `None` and keys.

teamdandelion:
### Merge activity

* **Aug 5, 2:45 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1167).
* **Aug 5, 3:02 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1167) rebased this pull request as part of a merge.
* **Aug 5, 3:03 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1167).

PR #1168, opened 2025-08-04
refactor: clients only use generic llm.Message

teamdandelion:
In offline discussion, we agreed that trying to support provider-specific messages creates a leaky abstraction / incomplete abstraction, especially since the Response will only have generic messages. 

Supporting provider-specific input messages also raises the unaddressed question of what we do with provider-specific output messages and potentially unsupported content.

I left the `AnthropicMessage`, `GoogleMessage` etc type aliases, as I think it'll be a nice convention to alias whatever the provider-specific message representation may be, to a consistent type alias.

PR #1169, opened 2025-08-04
feat: implement Response & remove Usage

teamdandelion:
- Remove Usage, pending getting evidence that users really
  need it (otherwise it becomes part of observability / Lilypad)
- Refactor Response to have a custom init method so it can take
  responsibility for organizing the different content parts, etc.
- Add unit tests & ensure responses package has 100% coverage


teamdandelion:
* **#1171** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1170** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1167** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1167?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1169** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1168** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1165** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1165?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1164** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1164?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1163** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1163?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1162** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1162?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/stream_response.py
```diff
@@ -186,14 +186,14 @@ def pretty_stream(
     @overload
     def streams(self: "StreamResponse[Stream, FormatT]") -> Iterator[Stream]:
         """Overload for synchronous streaming."""
-        ...
+        raise NotImplementedError()
```
willbakst: overloads will be `...` forever? what's with the change?

python/mirascope/llm/responses/response.py
```diff
@@ -1,12 +1,68 @@
...
 
     finish_reason: FinishReason
     """The reason why the LLM finished generating a response."""
+
+    def __init__(
+        self,
+        *,
+        model: "REGISTERED_LLMS",
+        input_messages: Sequence[Message],
+        final_message: AssistantMessage,
```
willbakst: nit: I think we should name this `assistant_message: AssistantMessage`?

python/mirascope/llm/responses/response.py
```diff
@@ -1,12 +1,68 @@
...
+        raw: Any,  # noqa: ANN401
```
willbakst: nit: let's put `raw` as the last argument always?

python/mirascope/llm/responses/response.py
```diff
@@ -1,12 +1,68 @@
...
+            finish_reason: The reason why the LLM finished generating a response.
+        """
+        self.model = model
+        self.raw = raw
+        self.finish_reason = finish_reason
+
+        self.messages = list(input_messages) + [final_message]
+        self.content = final_message.content
+
+        self.texts = [part for part in self.content if isinstance(part, Text)]
```
willbakst: we should reduce this from 3 to 1 loops like this I think:

```python
self.texts, self.tool_calls, self.thinkings = [], [], []
for part in self.content:
    if isinstance(part, Text):
        self.texts.append(part)
    elif isinstance(part, ToolCall):
        self.tool_calls.append(part)
    else:
        self.thinkings.append(part)
```

python/mirascope/llm/responses/base_response.py
```diff
@@ -46,9 +38,6 @@ class BaseResponse(Generic[FormatT]):
     thinkings: Sequence[Thinking]
     """The thinking content in the generated response, if any."""
 
-    usage: Usage | None
-    """The token usage statistics for this response, if available."""
-
     def format(self) -> FormatT:
```
willbakst, line 42: should this be marked `@abstractmethod`?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -297,12 +297,12 @@ def format(self, partial: Literal[True]) -> Partial[FormatT]:
...
 
     @overload
     def format(self, partial: Literal[False] = False) -> FormatT:
         """Format the response into a Pydantic BaseModel."""
-        ...
+        raise NotImplementedError()
 
     def format(self, partial: bool = False) -> FormatT | Partial[FormatT]:
```
willbakst: is this override valid given `BaseResponse` doesn't have the `partial: bool` argument?

python/mirascope/llm/responses/stream_response.py
teamdandelion: It was to get coverage passing. I could also put in coverage ignore statements, maybe that's cleaner

teamdandelion: Fixed

python/mirascope/llm/responses/base_response.py
teamdandelion: I'm not 100% sure (will be clearer when I actually implement format), but I'll put it in for now.

python/mirascope/llm/responses/stream_response.py
teamdandelion: Pyright seems ok with it? I will know for sure when we actually implement format.

Since the interface is not changing in this PR, i'd rather leave this for when I'm putting in structured outputs.

python/mirascope/llm/responses/stream_response.py
```diff
@@ -186,14 +186,12 @@ def pretty_stream(
     @overload
     def streams(self: "StreamResponse[Stream, FormatT]") -> Iterator[Stream]:
         """Overload for synchronous streaming."""
-        ...
 
     @overload
     def streams(
         self: "StreamResponse[AsyncStream, FormatT]",
     ) -> Awaitable[AsyncIterator[AsyncStream]]:
         """Overload for asynchronous streaming."""
```
willbakst: Yeah if we have docstrings for all overloads then they will be covered.

However, one thing we can do while stil keeping `...` is to specifically ignore overloads in coverage, which is what we do in v1. My preference is to keep the `...` since it's most common to see that or `pass` as the body of an overload and I prefer `...` over `pass`.

It's also common for overload definitions to have no newlines between them and only have a newline between the bottom-most overload and the actual function definition (i.e. a block of overloads followed by a newline followed by the actual function). I prefer having newlines between each overload for readability, but figured I would mention here in case you have an opinion.

teamdandelion: I have a mild preference for no-newlines-between-overloads so that all of the overloads group together as one block, but I don't feel strongly about it. I updated the coverage rules per your suggestion, thanks!

python/tests/llm/responses/test_response.py
```diff
@@ -0,0 +1,145 @@
+"""Tests for Response class."""
+
+import pytest
+
+from mirascope import llm
+from mirascope.llm.responses import Response
+from mirascope.llm.responses.finish_reason import FinishReason
+
+
+class TestResponse:
```
willbakst: nit: my feeling on using classes like this in tests is that they should be used for organizing within a module (like you have for the converters up stack). my preference here would be to reduce the nesting since there is no organization here from the class that we don't already get from the file.

python/mirascope/llm/responses/stream_response.py
willbakst: Ok, you should play around with removing the newlines in a separate PR and see how you like it. I tried to like it but the single block became somewhat unreadable imo

willbakst (approved):

teamdandelion:
### Merge activity

* **Aug 5, 2:45 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1169).
* **Aug 5, 2:59 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1169) rebased this pull request as part of a merge.
* **Aug 5, 3:00 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1169).

PR #1174, opened 2025-08-05
nit: allow claude to use mkdir

teamdandelion:

PR #1175, opened 2025-08-05
refactor: replace BaseConverter with utils

teamdandelion:


teamdandelion:
* **#1179** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1178** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1178?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1175** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1175?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1175?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/tests/llm/clients/anthropic/test_anthropic_converter.py
willbakst, line 1: hmm, should we refactor the converters into something like `llm/converters`?

then we would have `llm/converters/anthropic_converter.py` and corresponding `tests/llm/converters/test_anthropic_converter.py`

willbakst: and we could have `llm/converters/base_converter.py` for the base class

python/mirascope/llm/clients/anthropic/converter.py
```diff
@@ -31,12 +31,8 @@
...
-        str | Sequence[anthropic_types.ContentBlock],  # ProviderContentT
-        tuple[
-            AnthropicSystemMessage, list[AnthropicMessageParam]
-        ],  # ProviderMessagesInputT
+        tuple[SystemMessage, list[AnthropicMessageParam]],  # ProviderMessagesInputT
         AnthropicMessage,  # ProviderMessageResponseT
-        anthropic_types.ContentBlock,  # ProviderContentBlockT
         anthropic_types.StopReason,  # ProviderFinishReasonT
     ]
 ):
```
willbakst: <p class='graphite__hidden'><i>[Re: line +42]</i></p>
hmm, if we're removing this from the base class, why are we keeping it here?

I feel like we shouldn't remove it from the base class if we're going to have to implement it for every subclass?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1175?utm_source=unchanged-line-comment">Graphite</a>.</p>

python/mirascope/llm/clients/anthropic/converter.py
```diff
@@ -71,7 +67,7 @@ def encode_messages(
         Returns:
             Tuple of (system_message, user_or_assistant_messages)
         """
-        system_message: AnthropicSystemMessage = None
+        system_message: SystemMessage = None
         converted_messages: list[AnthropicMessageParam] = []
         for i, message in enumerate(messages):
             if message.role != "system":
```
willbakst: <p class='graphite__hidden'><i>[Re: line +96]</i></p>
same thing here
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1175?utm_source=unchanged-line-comment">Graphite</a>.</p>

python/tests/llm/clients/anthropic/test_anthropic_converter.py
teamdandelion: I redid this as base/_converter.py and with a new interface and some tests. The converter now takes responsibility for filtering out system messages, if the client wants it to.

python/mirascope/llm/clients/anthropic/_converter.py
```diff
@@ -0,0 +1,82 @@
...
+ANTHROPIC_FINISH_REASON_MAP = {
+    "end_turn": FinishReason.END_TURN,
+    "max_tokens": FinishReason.MAX_TOKENS,
+    "stop_sequence": FinishReason.STOP,
+    "tool_use": FinishReason.TOOL_USE,
+    "refusal": FinishReason.REFUSAL,
+}
+
+
+def encode_content(
```
willbakst: nit: these helper methods should still be private to this module

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -47,8 +47,8 @@ def call(
             raise NotImplementedError("param use not yet supported")
 
         # Convert messages to MessageParam format
```
willbakst: nit: unnecessary comment

python/mirascope/llm/clients/anthropic/_converter.py
```diff
@@ -0,0 +1,82 @@
...
+    ]
+):
+    """Converter for Anthropic message formats."""
+
+    @classmethod
+    def encode_message(cls, message: Message) -> list[anthropic_types.MessageParam]:
+        """Convert user or assistant `Message`s to Anthropic `MessageParam` format.
+
+        Args:
+            message: A `UserMessage` or `AssistantMessage`
```
willbakst: this definition seems more specific than `message: Message`?

python/mirascope/llm/clients/anthropic/_converter.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Returns:
+            A list containing a single Anthropic `MessageParam`
+        """
+        assert (  # Clients must use converter.encode_non_system_messages
```
willbakst: This seems very weird. I really don't like having asserts.

See my other comment about a suggestion around not having this separate method.

python/mirascope/llm/clients/anthropic/_converter.py
```diff
@@ -0,0 +1,82 @@
...
+    """Convert Anthropic content block to mirascope AssistantContentPart."""
+    if content.type == "text":
+        return Text(text=content.text)
+    else:
+        raise NotImplementedError(
+            f"Support for content type `{content.type}` is not yet implemented."
+        )
+
+
+class AnthropicConverter(
```
willbakst: nit: I'm wondering if we should make this `_AnthropicConverter` since we import it into the clients, I guess you could technically import it as a user from there without realizing?

python/mirascope/llm/clients/base/_converter.py
```diff
@@ -0,0 +1,94 @@
...
+
+import logging
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+from typing import Generic, TypeAlias, TypeVar
+
+from ...messages import AssistantMessage, Message
+from ...responses import FinishReason
+
+ProviderMessagesInputT = TypeVar("ProviderMessagesInputT")
```
willbakst: nit: this should be `ProviderMessageInputT` singular not plural

python/mirascope/llm/clients/base/_converter.py
```diff
@@ -0,0 +1,94 @@
...
+        It should be used when the system message itself corresponds to a
+        ProviderMessagesInputT.
+        """
+        results: list[ProviderMessagesInputT] = []
+        for message in messages:
+            results.extend(cls.encode_message(message))
+        return results
+
+    @classmethod
+    def encode_non_system_messages(
```
willbakst: I'm not in love with this interface. I think the thing that bothers me is that it isn't clear that if a provider like Anthropic of Google that have separate system messages calls `encode_messages` there is undefined behavior because `ProviderMessageInputT` doesn't actually have a system message type.

This makes me wonder how much we actually benefit from the base converter structure.

I wonder if it would be better in each provider client to just do something like `from . import _converter` and then something like `_converter.encode_messages` and just have those properly typed for each provider.

python/mirascope/llm/clients/anthropic/_utils.py
```diff
@@ -0,0 +1,75 @@
...
+    """Convert Anthropic content block to mirascope AssistantContentPart."""
+    if content.type == "text":
+        return Text(text=content.text)
+    else:
+        raise NotImplementedError(
+            f"Support for content type `{content.type}` is not yet implemented."
+        )
+
+
+def _encode_messages(
```
willbakst: My preference would be for this to be `encode_messages` (and `decode_response` below).

My reasoning is that it makes it clear that methods like `_encode_content` above are private to this module and should not be used outside of it whereas this method can be used outside of this module.

We would then do e.g. `from . import _utils` and `_utils.encode_messages` to ensure the utilities remain private to Mirascope and our end users don't rely on them. Semantically it's also clear that `_utils` can be used inside of `llm/clients/anthropic` but should not be used elsewhere (as we don't export it).

If we wanted, we could do things like `from . import _utils as _anthropic_utils` or `from . import _utils as _au`, but `_anthropic_utils` feels wordy and I don't love the `_au` shorthand since it's not super clear at a glance what it is. I also think `_utils` is clear enough given it's private, meaning it's already namespaced to its parent module (which is `anthropic`).

WDYT?

python/mirascope/llm/clients/anthropic/_utils.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _decode_response(
+    response: anthropic_types.Message,
+) -> tuple[AssistantMessage, FinishReason]:
+    """Convert Anthropic message to mirascope AssistantMessage."""
+    assistant_message = assistant(
+        content=[_decode_assistant_content(part) for part in response.content]
+    )
+    finish_reason = FinishReason.UNKNOWN
```
willbakst: I think my preference here would be a full conditional block like this:

```python
if response.stop_reason:
    finish_reason =  ANTHROPIC_FINISH_REASON_MAP.get(
        response.stop_reason, FinishReason.UNKNOWN
    )
else:
    finish_reason = FinishReason.UNKNOWN
```

python/mirascope/llm/clients/base/_converter.py
willbakst, line 1: can we delete this file now?

python/mirascope/llm/clients/base/_utils.py
```diff
@@ -0,0 +1,40 @@
+import logging
+from collections.abc import Sequence
+from typing import TypeAlias
+
+from ...messages import AssistantMessage, Message, UserMessage
+
+SystemMessage: TypeAlias = str | None
+
+
+def _extract_system_message(
```
willbakst: Same thing here. This should just be `extract_system_message`, and then we should export only `_utils` from `base` to indicate we can use it elsewhere internally.

Then we would do something like `from base import _utils as _base_utils` and then `_base_utils.extract_system_message`

python/mirascope/llm/clients/base/_utils.py
```diff
@@ -0,0 +1,40 @@
+import logging
+from collections.abc import Sequence
+from typing import TypeAlias
+
+from ...messages import AssistantMessage, Message, UserMessage
+
+SystemMessage: TypeAlias = str | None
```
willbakst: nit: I really want this to be `SystemMessageContent` since we have a system message type and it's not this.

python/mirascope/llm/clients/base/_utils.py
```diff
@@ -0,0 +1,40 @@
...
+) -> tuple[SystemMessage, Sequence[UserMessage | AssistantMessage]]:
+    """Extract the system message(s) from a list of Messages.
+
+    This takes a list of messages, and returns the list of messages with
+    all system messages removed, as well as the system message if present.
+
+    This is intended for use in clients where the system message is not included in the
+    input messages, but passed as an additional argument or metadata.
+
+    If there were multiple system messages, they will be concatenated together
```
willbakst: stale comment?

python/mirascope/llm/clients/base/_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    into a single system message, joined by newlines. In this case, a warning
+    will be emitted.
+    """
+    system_message = None
+    output = []
```
willbakst: nit: rename this to `other_messages` or `messages` or `remaining_messages` or something and then use the same name elsewhere for consistency. `output` feels like the wrong name

I left another comment where I think `messages` works, but I'm open to alternatives.

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -46,24 +46,15 @@ def call(
         if params:
             raise NotImplementedError("param use not yet supported")
 
-        # Convert messages to MessageParam format
-        system_message, converted_messages = AnthropicConverter.encode_messages(
-            messages
-        )
+        system_message, other_messages = _extract_system_message(messages)
```
willbakst: revisiting this given other comment, I think maybe just `messages` is good? WDYT?

python/tests/llm/clients/base/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+def test_encode_messages_with_system_first():
+    """Test that system message is extracted when it's first."""
+    messages = [
+        system("You are a helpful assistant"),
+        user("Hello"),
+        assistant("Hi there"),
+        user("How are you?"),
+    ]
+
+    system_message, rest = _extract_system_message(messages)
```
willbakst: nit: same thing here, `rest` should be replaced with whatever we decide elsewhere for consistency

python/mirascope/llm/clients/anthropic/_utils.py
```diff
@@ -0,0 +1,75 @@
...
+    if response.stop_reason:
+        finish_reason = ANTHROPIC_FINISH_REASON_MAP.get(
+            response.stop_reason, FinishReason.UNKNOWN
+        )
+    return (assistant_message, finish_reason)
```
willbakst: nit: unnecessary parens

python/mirascope/llm/clients/anthropic/_utils.py
teamdandelion: Sure, that works for me. Done.

python/mirascope/llm/clients/base/_converter.py
teamdandelion: Done, thanks. I thought I had deleted it.

python/mirascope/llm/clients/base/_utils.py
teamdandelion: Ok, I made it `messages` and renamed the input arg to `all_messages`

willbakst: Hmm, I see the issue now with the return being messages having the input be messages, but I actually think it makes more sense for the input to be `messages` and the output to be something more specific.

Maybe `remaining_messages`?

teamdandelion: Yeah, I realized it actually presents a problem to overwrite messages because we need the un-modified messages (including system message) to thread thru on the response. Changed to `remaining_messages`

python/mirascope/llm/clients/anthropic/_utils.py
teamdandelion: FYI if we do this we need a `# pragma: no cover` since I don't want to write tests against mocked responses. I think when I get to streaming there will be responses w.o. a finish reason so we may remove the no cover at that point (possibly involving a re-write if we're re-using this logic, since for an unfinished stream the finish_reason should be `None`, not `UNKNOWN`.)

```
    if response.stop_reason:
        finish_reason = ANTHROPIC_FINISH_REASON_MAP.get(
            response.stop_reason, FinishReason.UNKNOWN
        )
    else:
        finish_reason = FinishReason.UNKNOWN  # pragma: no cover
```

willbakst: Oh hmm, this is making me realize that perhaps `finish_reason` should have `| None` even on response?

If the `if response.stop_reason` conditional here is `False`, wouldn't that be a finish reason of `None`?

As for coverage, you could always write this:

```python
finish_reason = (
    ANTHROPIC_FINISH_REASON_MAP.get(response.stop_reason, FinishReason.UNKNOWN)
    if response.stop_reason
    else None
)
```

python/mirascope/llm/clients/base/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+    This takes a list of messages, and returns the list of messages with
+    all system messages removed, as well as the textual contents of the first message,
+    if that message was a system message. If there are any system messages that are
+    not the first message, they will be dropped, and a warning will be emitted.
+
+    This is intended for use in clients where the system message is not included in the
+    input messages, but passed as an additional argument or metadata.
+    """
+    system_message = None
```
willbakst, line 23: nit: this should be `system_message_content` to match the updated type name

willbakst (approved):

teamdandelion:
### Merge activity

* **Aug 6, 7:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1175).
* **Aug 6, 7:32 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1175) rebased this pull request as part of a merge.
* **Aug 6, 7:33 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1175).

PR #1176, opened 2025-08-05
chore: remove raw anthropic response snapshotting

teamdandelion:
I don't think we should be snapshotting the raw responses, for a few
reasons:

1. It's not part of our API contract, and will tend to make spurious /
   noisy diffs when we update dependency versions
2. It's not generally possible to make these assertions safely (in
   OpenAI they include fields that can't be reconstructed using their
   documented API, and that don't seem to have equality implemented
   properly)

So, since we can't use this pattern cross-provider, I think we are
better to simplify and take it out.

PR #1178, opened 2025-08-05
feat: implement openai call handling

teamdandelion:


teamdandelion:
* **#1179** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1178** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1178?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1178?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1175** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1175?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/clients/openai/client.py
```diff
@@ -13,13 +16,25 @@
...
 from ..base import BaseClient
+from .converter import OpenAIConverter
 from .models import OpenAIModel
 from .params import OpenAIParams
 
 
 class OpenAIClient(BaseClient[OpenAIParams, OpenAIModel]):
     """The client for the OpenAI LLM model."""
 
+    client: OpenAI
```
willbakst, line 27: Why do I remember this being `_client` for Anthropic? Want to make sure these match

Also, should we add `ProviderClientT` or something as a generic on `BaseClient`?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
+"""OpenAI message types and conversion utilities."""
+
+import logging
+from collections.abc import Sequence
+from typing import TypeAlias
+
+from openai.types import responses as openai_types
```
willbakst: why do we have these as separate imports? Why not just either import the types we use or use the namespace `openai_types` everywhere?

I think my preference would be to namespace everything so it's clear what is ours vs theirs?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+    # TODO: Audit these (MIR-285)
+    "completed": FinishReason.END_TURN,
+    "incomplete": FinishReason.MAX_TOKENS,
+    "failed": FinishReason.UNKNOWN,
+    "cancelled": FinishReason.UNKNOWN,
+    "queued": FinishReason.UNKNOWN,
+    "in_progress": FinishReason.UNKNOWN,
+}
+
+SystemMessage: TypeAlias = str | None
```
willbakst: hmm, this feels wrong. OpenAI has a system message, so it feels wrong to create this type alias for the system message content. I understand creating a single alias for DRY, but we should name this something else.

Also, I don't even know that we need to DRY this here. What if we just made something like `SystemMessageContent: TypeAlias = str | None` in the `BaseConverter` file and always had that be the return? Then we could make the subclasses generic only on the second item in the tuple?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+    BaseConverter[
+        tuple[SystemMessage, openai_types.ResponseInputParam],  # ProviderMessagesInputT
+        openai_types.Response,  # ProviderMessageResponseT
+        openai_types.ResponseStatus | None,  # ProviderFinishReasonT
+    ]
+):
+    """Converter for OpenAI Responses API message formats."""
+
+    @classmethod
+    def encode_message(
```
willbakst: This is new but not part of the base converter contract?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+    def decode_assistant_message(cls, message: Response) -> AssistantMessage:
+        """Convert OpenAI Response to mirascope AssistantMessage."""
+
+        content_parts = []
+        for output_item in message.output:
+            content_parts.extend(cls.decode_output_item(output_item))
+        return assistant(content=content_parts)
+
+    @classmethod
+    def decode_output_item(
```
willbakst: hmm, ok I'm not loving how different these converters are becoming.

What's the plan here?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+        cls, message: Message
+    ) -> list[openai_types.ResponseInputItemParam]:
+        """Encode a Mirascope Message into one or more ResponseInputItemParam.
+
+        Note: Conceptually the OpenAI Responses API doesn't really have a concept of a
+        message with a role containing multiple pieces of content, it's more like every
+        piece of content is tagged with a role.
+        """
+        if message.role == "system":
+            logging.warning("Non-first system message is being skipped.")
```
willbakst: this is a weird warning to have here. the method has no concept of position, so how do we know it's non-first beyond how we're calling it?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Args:
+            messages: Sequence of messages to convert
+
+        Returns:
+            ResponseInputParam (aka list of ResponseInputItemParam) for the Responses API
+        """
+        system_message = None
+        messages_to_encode = messages
+        if messages[0].role == "system":
```
willbakst: unsafe access since `messages` could technically be an empty sequence

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+            system_message = messages[0].content.text
+            messages_to_encode = messages[1:]
+
+        encoded_items = [
+            item
+            for message in messages_to_encode
+            for item in cls.encode_message(message)
+        ]
+        return (system_message, encoded_items)
```
willbakst: parens are unnecessary

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+        Returns:
+            ResponseInputParam (aka list of ResponseInputItemParam) for the Responses API
+        """
+        system_message = None
+        messages_to_encode = messages
+        if messages[0].role == "system":
+            system_message = messages[0].content.text
+            messages_to_encode = messages[1:]
+
+        encoded_items = [
```
willbakst: `encoded_items` feels like the wrong name. The items aren't encoded, it's messages that have been encoded into items. Wouldn't it just be items? Or input_params?

python/tests/llm/clients/openai/test_openai_client.py
```diff
@@ -0,0 +1,90 @@
...
+    messages = [llm.messages.user("Hello, say 'Hi' back to me")]
+
+    response = openai_client.call(
+        model="gpt-4o-mini",
+        messages=messages,
+    )
+
+    assert isinstance(response, llm.responses.Response)
+
+    assert response.messages == snapshot(
```
willbakst: nit: same thing here as I mentioned on other comment around ideally having a single snapshot for the entire response

python/tests/llm/clients/openai/test_openai_converter.py
```diff
@@ -0,0 +1,71 @@
...
+            system_msg, converted = OpenAIConverter.encode_messages(messages)
+
+        assert system_msg is None
+        assert len(converted) == 2
+
+        assert len(caplog.records) == 1
+        assert "Non-first system message is being skipped" in caplog.text
+
+
+class TestDecodeFinishReason:
```
willbakst: where are the tests for decoding the output?

python/mirascope/llm/clients/openai/converter.py
```diff
@@ -0,0 +1,128 @@
...
+):
+    """Converter for OpenAI Responses API message formats."""
+
+    @classmethod
+    def encode_message(
+        cls, message: Message
+    ) -> list[openai_types.ResponseInputItemParam]:
+        """Encode a Mirascope Message into one or more ResponseInputItemParam.
+
+        Note: Conceptually the OpenAI Responses API doesn't really have a concept of a
```
willbakst: I'm not really sure what you're talking about here. The example front an center in their API reference shows a user message with a content array with multiple parts?

```python
from openai import OpenAI

client = OpenAI()

response = client.responses.create(
    model="gpt-4.1",
    input=[
        {
            "role": "user",
            "content": [
                { "type": "input_text", "text": "what is in this image?" },
                {
                    "type": "input_image",
                    "image_url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
                }
            ]
        }
    ]
)

print(response)
```

willbakst:
At a high level, it feels extremely weird not support the chat completions API as it's definitely still the standard.

My understanding is that we're opting for the Responses API because it has reasoning traces? Part of me feels like we should implement Chat Completions by default and then also have a Responses client and support both?

Part of the `OpenAIParams` could be something like `responses=True` or something?

python/mirascope/llm/clients/openai/client.py
teamdandelion: Your memory is accurate to how I originally wrote it, but I changed it to make it `client`. Sure, I'll add the generic on BaseClient in a followon PR.

python/mirascope/llm/clients/openai/converter.py
teamdandelion: In the code as written, `import responses as openai_types` make sense because there's a separate module of types for responses vs chat completions.

In the rewrite to use chat completions, I just imported the types directly without namespacing. The type names are long and unambiguous (they all start with ChatCompletion) so in this case I prefer it. I think it's ok if this varies depending on what flows best for the given provider.

python/mirascope/llm/clients/openai/converter.py
teamdandelion: Done, also with ChatCompletion API we don't need the system message filtering since we can just put in the multiple system messages.

python/mirascope/llm/clients/openai/converter.py
teamdandelion: In the rewrite, encode_message is part of the contract, but I made it so anything that's not part of the contract isn't a class method.

python/mirascope/llm/clients/openai/converter.py
teamdandelion: Yeah that was inaccurate comment on my part, what I should have said is the tool output doesn't go in the user message so we still need to chop the messages up.

python/mirascope/llm/clients/openai/converter.py
teamdandelion: Refactored down to the base class.

python/mirascope/llm/clients/openai/converter.py
teamdandelion: Discussed offline, see the new base class

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+def _decode_response(response: ChatCompletion) -> tuple[AssistantMessage, FinishReason]:
+    """Convert OpenAI ChatCompletion to mirascope AssistantMessage."""
+    choice = response.choices[0]
+    message = choice.message
+
+    assistant_message = assistant(
+        content=[_decode_assistant_content(message.content or "")]
+    )
+
+    finish_reason = FinishReason.UNKNOWN
```
willbakst, line 87: nit: same thing here

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    elif message.role == "assistant":
+        return ChatCompletionAssistantMessageParam(
+            role="assistant", content=_encode_content(message.content)
+        )
+
+    else:
+        raise ValueError(f"Unsupported role: {message.role}")  # pragma: no cover
+
+
+def _encode_messages(
```
willbakst: nit: same thing here

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    messages: Sequence[Message],
+) -> Sequence[ChatCompletionMessageParam]:
+    """Convert a sequence of Mirascope `Message` to OpenAI `ChatCompletionmessageParam`."""
+    return [_encode_message(message) for message in messages]
+
+
+def _decode_response(response: ChatCompletion) -> tuple[AssistantMessage, FinishReason]:
+    """Convert OpenAI ChatCompletion to mirascope AssistantMessage."""
+    choice = response.choices[0]
+    message = choice.message
```
willbakst: nit: I think this assignment is unnecessary, maybe just put `choice.message.content` below?

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
```
willbakst: nit: I think we can drop this newline

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    assistant_message = assistant(
+        content=[_decode_assistant_content(message.content or "")]
+    )
+
```
willbakst: nit: I'm being really annoying, but the Anthropic equivalent doesn't have this newline. I would either remove it here or add it there for consistency.

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    finish_reason = FinishReason.UNKNOWN
+    if choice.finish_reason:
+        finish_reason = OPENAI_FINISH_REASON_MAP.get(
+            choice.finish_reason, FinishReason.UNKNOWN
+        )
+
+    return (assistant_message, finish_reason)
```
willbakst: nit: unnecessary parens

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Returns:
+        A list containing a single OpenAI `ChatCompletionMessageParam`
+    """
+    if message.role == "system":
+        return ChatCompletionSystemMessageParam(
+            role="system", content=message.content.text
+        )
+
+    elif message.role == "user":
```
willbakst, line 60: For OpenAI, don't we have to separate tool calls into a message with role `tool`?

Or are we specifically not handling that yet since we haven't added support for tools yet? If this, maybe add a NOTE or TODO or something so it's clear.

python/tests/llm/clients/openai/test_openai_client.py
```diff
@@ -0,0 +1,159 @@
...
+    messages = [llm.messages.user("Hello, say 'Hi' back to me")]
+
+    response = openai_client.call(
+        model="gpt-4o-mini",
+        messages=messages,
+    )
+
+    assert isinstance(response, llm.responses.Response)
+
+    assert {k: v for k, v in response.__dict__.items() if k != "raw"} == snapshot(
```
willbakst: nit: same thing here re: `response.__dict__.pop("raw")`

maybe do a sweep?

python/mirascope/llm/clients/openai/_utils.py
teamdandelion: Yeah, I am not attempting to add support for tools yet. I don't really see a need for a note or todo, it will get fixed up as needed when I'm adding tested support for tools.

python/mirascope/llm/clients/openai/_utils.py
teamdandelion: Same thing being a note about adding support for tools? As above I don't really see value in this. I think notes are most valuable when there is something that might plausibly get forgotten if not noted. But I don't see a way I could add tools and forget to update these functions to handle them.

willbakst: Sorry that's on me. Same thing as in same change as downstack re: `_utils`

python/mirascope/llm/clients/openai/_utils.py
willbakst: Yeah in the world where it's just you/us working on this, I agree, especially given you'll be tackling this soon.

Was coming at it from a more principled view of not knowing exactly when we'll get to it so if someone were to read this code in the meantime it would be clear why we aren't handling that

python/mirascope/llm/clients/openai/client.py
```diff
@@ -28,7 +43,27 @@ def call(
...
+            raise NotImplementedError("tool use not yet supported")
+        if params:
+            raise NotImplementedError("param use not yet supported")
+
+        openai_response = self.client.chat.completions.create(
+            model=model,
+            messages=_utils.encode_messages(messages),
+        )
+
+        assistant_message, finish_reason = _utils.decode_response(openai_response)
```
willbakst, line 57: nit: you have a newline here separating this from the response, but you don't have that newline in the anthropic version. I like the newline, so please match in anthropic

python/mirascope/llm/clients/openai/_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    return [_encode_message(message) for message in messages]
+
+
+def decode_response(response: ChatCompletion) -> tuple[AssistantMessage, FinishReason]:
+    """Convert OpenAI ChatCompletion to mirascope AssistantMessage."""
+    choice = response.choices[0]
+    assistant_message = assistant(
+        content=[_decode_assistant_content(choice.message.content or "")]
+    )
+    finish_reason = FinishReason.UNKNOWN
```
willbakst, line 87: nit: in Anthropic and Google you do the if/else with the no coverage. we should do the same here.

Also still uncertain about the `| None` resolution

teamdandelion: Actually, for OpenAI the finish reason is not optional, so we can drop that.

teamdandelion:
### Merge activity

* **Aug 6, 7:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1178).
* **Aug 6, 7:41 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1178) rebased this pull request as part of a merge.
* **Aug 6, 7:42 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1178).

PR #1179, opened 2025-08-05
feat: implement GoogleClient.call for text calls

teamdandelion:


teamdandelion:
* **#1179** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1179?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1179?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1178** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1178?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1175** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1175?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/clients/google/client.py
```diff
@@ -28,7 +53,45 @@ def call(
...
         params: GoogleParams | None = None,
     ) -> Response[None]:
-        raise NotImplementedError
+        """Make a call to the Google GenAI API."""
+        if tools:
+            raise NotImplementedError("tool use not yet supported")
+        if params:
+            raise NotImplementedError("param use not yet supported")
+
+        # Convert messages to Google format
```
willbakst: unnecessary comment

python/mirascope/llm/clients/google/client.py
```diff
@@ -28,7 +53,45 @@ def call(
...
+        system_message, converted_messages = GoogleConverter.encode_messages(messages)
+
+        # Build the request config
```
willbakst: ditto

python/mirascope/llm/clients/google/client.py
```diff
@@ -28,7 +53,45 @@ def call(
...
+        config = None
+        if system_message:
+            config = GenerateContentConfig(
+                system_instruction=cast(ContentUnion, system_message)
+            )
+
+        # Make request to Google GenAI API
```
willbakst: ditto

python/mirascope/llm/clients/google/client.py
```diff
@@ -28,7 +53,45 @@ def call(
...
+        google_response = self.client.models.generate_content(
+            model=model,
+            contents=cast(ContentListUnion, converted_messages),
+            config=config,
+        )
+
+        # Convert response back to mirascope format
```
willbakst: ditto

I think for all of these, the code is quite clear how it's written. Not sure what value these comments add? Not sure if we have similar comments for other clients

python/mirascope/llm/clients/google/converter.py
```diff
@@ -0,0 +1,124 @@
...
+    AssistantContentPart,
+    ContentPart,
+    Text,
+)
+from ...messages import AssistantMessage, Message, assistant
+from ...responses import FinishReason
+from ..base.converter import BaseConverter
+
+GoogleContentDict: TypeAlias = ContentDict
+SystemMessage: TypeAlias = str | None
```
willbakst: same thing here. noticing that this is the same across all converters. see other comment down stack

python/mirascope/llm/clients/google/converter.py
```diff
@@ -0,0 +1,124 @@
...
+
+GoogleResponse: TypeAlias = GenerateContentResponse
```
willbakst: hmm, not seeing parity across other providers. not sure if there's value in having this for all providers or none or some?

python/mirascope/llm/clients/google/converter.py
```diff
@@ -0,0 +1,124 @@
...
+        content_parts = []
+
+        if candidate.content and candidate.content.parts:
+            for part in candidate.content.parts:
+                content_parts.extend(cls.decode_content_part(part))
+
+        return assistant(content=content_parts)
+
+    @classmethod
+    def decode_content_part(cls, part: Part) -> list[AssistantContentPart]:
```
willbakst: this shouldn't be on the class. If we need specific helper / utility methods for specific providers and we want to name them as such, we should implement those as private helper utilities here and not muddy up the class so that the class retains a clean contract (encode messages, decode assistant message, decode finish reason)

python/mirascope/llm/clients/google/converter.py
```diff
@@ -0,0 +1,124 @@
...
+        """Convert Google Part to mirascope AssistantContentPart."""
+        if part.text:
```
willbakst: is there no part type to discriminate on?

python/mirascope/llm/clients/google/converter.py
```diff
@@ -0,0 +1,124 @@
...
+
+        Returns:
+            Tuple of (system_message, user_or_assistant_messages)
+        """
+        system_message: SystemMessage = None
+        converted_messages: list[GoogleContentDict] = []
+
+        for i, message in enumerate(messages):
+            if message.role != "system":
+                # Convert role to Google format
```
willbakst: nit: I would either remove this comment or make it slightly more verbose. for example, something like `# Google uses "model" instead of "assistant" for the output message's role"`

python/mirascope/llm/clients/google/model.py
```diff
@@ -2,5 +2,14 @@
...
 from typing import Literal, TypeAlias
 
-GoogleModel: TypeAlias = Literal["google:gemini-2.5-flash"] | str
+GoogleModel: TypeAlias = (
+    Literal[
+        "gemini-2.5-pro",
+        "gemini-2.5-flash",
+        "gemini-2.5-flash-lite",
+        "gemini-2.0-flash",
+        "gemini-2.0-flash-lite",
```
willbakst, line 11: what about the 1.5 models? are they sunset?

python/tests/llm/clients/google/test_google_client.py
```diff
@@ -0,0 +1,94 @@
...
+    # Use real API key if available, otherwise dummy key for VCR tests
+    load_dotenv()
+    api_key = os.getenv("GOOGLE_API_KEY") or "dummy-key-for-vcr-tests"
+    return llm.clients.GoogleClient(api_key=api_key)
+
+
+def test_custom_base_url():
+    example_url = "https://example.com"
+    client = llm.clients.GoogleClient(base_url=example_url)
+    assert client.client._api_client._http_options.base_url == example_url
```
willbakst: this feels a little weird since our tests now rely on private attributes.

is there a way to potentially do this in a way where we don't rely on private attributes?

my thinking is we could do something like mock the client and ensure we instantiate it with the correct http settings?

python/tests/llm/clients/google/test_google_client.py
```diff
@@ -0,0 +1,94 @@
...
+    messages = [llm.messages.user("Hello, say 'Hi' back to me")]
+
+    response = google_client.call(
+        model="gemini-2.0-flash",
+        messages=messages,
+    )
+
+    assert isinstance(response, llm.responses.Response)
+
+    assert response.messages == snapshot(
```
willbakst: nit: same thing here re: single snapshot

python/tests/llm/clients/google/test_google_converter.py
```diff
@@ -0,0 +1,111 @@
...
+    Part,
+)
+
+from mirascope.llm import Text
+from mirascope.llm.clients.google.converter import GoogleConverter
+from mirascope.llm.messages import assistant, system, user
+from mirascope.llm.responses import FinishReason
+
+
+class TestGoogleConverter:
```
willbakst: nit: this is different from the other providers. as mentioned previously, this class does nothing for organization as currently written and adds unnecessary indentation. we should try our best to have parity across providers as much as makes sense.

since we're unit testing the interfaces, if we have a standard contract and private helper methods, our unit tests should just test the interfaces (and thus test the helpers through the interfaces)

python/mirascope/llm/clients/google/converter.py
teamdandelion: No, it kinda sucks:

```
class Part(_common.BaseModel):
  """A datatype containing media content.

  Exactly one field within a Part should be set, representing the specific type
  of content being conveyed. Using multiple fields within the same `Part`
  instance is considered invalid.
  """

  video_metadata: Optional[VideoMetadata] = Field(
      default=None, description="""Metadata for a given video."""
  )
  thought: Optional[bool] = Field(
      default=None,
      description="""Indicates if the part is thought from the model.""",
  )
  inline_data: Optional[Blob] = Field(
      default=None, description="""Optional. Inlined bytes data."""
  )
  file_data: Optional[FileData] = Field(
      default=None, description="""Optional. URI based data."""
  )
  thought_signature: Optional[bytes] = Field(
      default=None,
      description="""An opaque signature for the thought so it can be reused in subsequent requests.""",
  )
  code_execution_result: Optional[CodeExecutionResult] = Field(
      default=None,
      description="""Optional. Result of executing the [ExecutableCode].""",
  )
  executable_code: Optional[ExecutableCode] = Field(
      default=None,
      description="""Optional. Code generated by the model that is meant to be executed.""",
  )
  function_call: Optional[FunctionCall] = Field(
      default=None,
      description="""Optional. A predicted [FunctionCall] returned from the model that contains a string representing the [FunctionDeclaration.name] with the parameters and their values.""",
  )
  function_response: Optional[FunctionResponse] = Field(
      default=None,
      description="""Optional. The result output of a [FunctionCall] that contains a string representing the [FunctionDeclaration.name] and a structured JSON object containing any output from the function call. It is used as context to the model.""",
  )
  text: Optional[str] = Field(
      default=None, description="""Optional. Text part (can be code)."""
  )
```

Fun fact: They don't honor their docstring, it can/does return empty parts with no fields set if you prompt it right (see tests)

python/mirascope/llm/clients/google/model.py
teamdandelion: They're listed as deprecated. I can put them in if you want, though.

python/mirascope/llm/clients/google/client.py
teamdandelion: Yeah this was me letting Claude get away with it. Removed.

python/mirascope/llm/clients/google/converter.py
teamdandelion: Removed them.

python/tests/llm/clients/google/test_google_converter.py
teamdandelion: Removed this file.

python/tests/llm/clients/google/test_google_client.py
teamdandelion: Ok, did it via mocking like you suggested.

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        if item.type == "text":
+            result.append({"text": item.text})
+        else:
+            raise NotImplementedError(
+                f"Have not implemented conversion for {item.type}"
+            )
+    return result
+
+
+def _decode_content_part(part: genai_types.Part) -> list[AssistantContentPart]:
```
willbakst: can a `genai_types.Part` become multiple parts? Why is the output here a `list`?

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        # >  instance is considered invalid.
+        # However, in testing, this can happen, so we will do our best to handle
+        # it as empty content.
+        return []
+
+
+def _encode_message(
+    message: UserMessage | AssistantMessage,
+) -> genai_types.ContentDict:
+    """Convert a Mirascope message to Google ContentDict format.
```
willbakst: nit: with Google-style docstrings, if you say "Returns a Google `ContentDict` converted from a Mirascope `Message`" then you can drop the rest of the docstring since the return is covered and the args (with typing) are rather clear.

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    return {
+        "role": "model" if message.role == "assistant" else message.role,
+        "parts": _encode_content(message.content),
+    }
+
+
+def _encode_messages(
+    messages: Sequence[UserMessage | AssistantMessage],
+) -> genai_types.ContentListUnionDict:
+    """Convert a sequence of Mirascope user or assistant Messages to Google ContentDict"""
```
willbakst: nit: same thing here. Either needs to start with "Returns ..." or have the full docstring.

Generally I write private methods with single line docstrings as much as makes sense and public methods the same (although public methods more frequently need full docstrings)

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        A single Google ContentDict
+    """
+
+    return {
+        "role": "model" if message.role == "assistant" else message.role,
+        "parts": _encode_content(message.content),
+    }
+
+
+def _encode_messages(
```
willbakst: ditto

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+def _decode_response(
+    response: genai_types.GenerateContentResponse,
+) -> tuple[AssistantMessage, FinishReason]:
+    """Convert Google response to mirascope AssistantMessage and FinishReason."""
+    if not response.candidates or not response.candidates[0].content:
+        # Unclear under what circumstances this happens (if ever).
+        # In testing, when generating no output at all, it creates a part with
+        # no fields set.
+        assistant_message = assistant(content=[])  # pragma: no cover
```
willbakst: Let's do an early `return assistant(content=[]), FinishReason.UNKNOWN` so we can remove the `else` indentation below.

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    else:
+        candidate = response.candidates[0]
+        content_parts = []
+        if candidate.content and candidate.content.parts:
+            for part in candidate.content.parts:
+                content_parts.extend(_decode_content_part(part))
+        assistant_message = assistant(content=content_parts)
+
+    finish_reason = FinishReason.UNKNOWN
```
willbakst: same thing here re: full conditional.

Also, given above comment, you should only need to check `if candidate.finish_reason` since we'll already have a candidate.

python/mirascope/llm/clients/google/converter.py
willbakst: Annoying

python/mirascope/llm/clients/google/_utils.py
teamdandelion: list is mostly vestigial from the old base converter structure. However we still have a `None` case so I made it `AssistantContentPart | None`

python/mirascope/llm/clients/google/_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        # However, in testing, this can happen, so we will do our best to handle
+        # it as empty content.
+        return None
+
+
+def _encode_message(
+    message: UserMessage | AssistantMessage,
+) -> genai_types.ContentDict:
+    """Returns a Google `ContentDict` converted from a Mirascope `Message`"""
+
```
willbakst, line 77: nit: remove newline here

teamdandelion:
### Merge activity

* **Aug 6, 7:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1179).
* **Aug 6, 7:44 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1179) rebased this pull request as part of a merge.
* **Aug 6, 7:45 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1179).

PR #1183, opened 2025-08-07
feat: implement StreamResponse.chunk_stream() with testing

teamdandelion:
Implements the core logic of llm.StreamResponse, with extensive testing,
and for both sync and async cases.

The tests are parameterized over sync and async cases, so all tests hit
both code paths. There are helper functions that make this easy to
manage.

Currently text and thinking support are both added, but not tool calls,
as we haven't established patterns for deserializing the call args and I
didn't feel this was the place to start.

One thing I noticed while implementing is that we don't really have a
good way for the StreamResponse to get the finish reason, so I will
address that in a followon.

### DIFF
--- a/python/mirascope/llm/responses/stream_response.py
+++ b/python/mirascope/llm/responses/stream_response.py
@@ -1,20 +1,38 @@
 """Responses that stream content from LLMs."""
 
 from collections.abc import AsyncIterator, Awaitable, Iterator, Sequence
-from typing import Any, Generic, Literal, overload
+from typing import TYPE_CHECKING, Any, Generic, Literal, TypeAlias, overload
 
 from ..content import (
     AssistantContentChunk,
     AssistantContentPart,
     Text,
+    TextChunk,
+    TextEndChunk,
+    TextStartChunk,
     Thinking,
+    ThinkingChunk,
+    ThinkingEndChunk,
+    ThinkingStartChunk,
     ToolCall,
 )
 from ..formatting import FormatT, Partial
-from ..messages import Message
+from ..messages import AssistantMessage, Message
 from ..streams import AsyncStream, Stream, StreamT
 from .base_response import BaseResponse
-from .finish_reason import FinishReason
+
+if TYPE_CHECKING:
+    from ..clients import Model, Provider
+
+
+ChunkWithRaw: TypeAlias = tuple[AssistantContentChunk, Any]
+"""A chunk paired with its raw provider data."""
+
+ChunkIterator: TypeAlias = Iterator[ChunkWithRaw]
+"""Synchronous iterator yielding chunks with raw data."""
+
+AsyncChunkIterator: TypeAlias = AsyncIterator[ChunkWithRaw]
+"""Asynchronous iterator yielding chunks with raw data."""
 
 
 class StreamResponse(BaseResponse[FormatT], Generic[StreamT, FormatT]):
@@ -153,11 +171,58 @@ class StreamResponse(BaseResponse[FormatT], Generic[StreamT, FormatT]):
     to avoid partial thinking blocks in the response.
     """
 
-    finish_reason: FinishReason | None
-    """The reason why the LLM finished generating a response.
+    consumed: bool = False
+    """Whether the stream has been fully consumed.
     
-    Will be None until the stream has finished streaming completely.
+    This is True after all chunks have been processed from the underlying iterator.
+    When False, more content may be available by calling the stream methods.
     """
+
+    def __init__(
+        self,
+        *,
+        provider: "Provider",
+        model: "Model",
+        input_messages: Sequence[Message],
+        chunk_iterator: ChunkIterator | AsyncChunkIterator,
+    ) -> None:
+        """Initialize the StreamResponse.
+
+        Args:
+            provider: The provider name (e.g., "anthropic", "openai")
+            model: The model identifier
+            input_messages: The input messages that were sent to the LLM
+            chunk_iterator: Iterator that yields (AssistantContentChunk, raw_data) tuples
+
+        The StreamResponse will process the tuples to build the chunks and raw lists
+        as the stream is consumed.
+        """
+
+        self.provider = provider
+        self.model = model
+
+        # Internal-only lists which we mutate (append) during chunk processing
+        self._chunks: list[AssistantContentChunk] = []
+        self._content: list[AssistantContentPart] = []
+        self._texts: list[Text] = []
+        self._thinkings: list[Thinking] = []
+        self._tool_calls: list[ToolCall] = []
+        self._raw: list[Any] = []
+
+        # Externally-facing references typed as immutable Sequences
+        self.chunks = self._chunks
+        self.content = self._content
+        self.texts = self._texts
+        self.thinkings = self._thinkings
+        self.tool_calls = self._tool_calls
+        self.raw = self._raw
+
+        self.finish_reason = None  # TODO: Add support for finish reason to chunks
+
+        self.messages = list(input_messages) + [AssistantMessage(content=self._content)]
+
+        self._current_content: Text | Thinking | None = None
+        self._chunk_iterator = chunk_iterator
 
     @overload
     def pretty_stream(self: "StreamResponse[Stream, FormatT]") -> Iterator[str]:
@@ -223,14 +288,14 @@ class StreamResponse(BaseResponse[FormatT], Generic[StreamT, FormatT]):
     def chunk_stream(
         self: "StreamResponse[Stream, FormatT]",
     ) -> Iterator[AssistantContentChunk]:
-        """Returns a sync iterator over raw assistant chunks."""
+        """Returns a sync iterator over content chunks."""
         ...
 
     @overload
     def chunk_stream(
         self: "StreamResponse[AsyncStream, FormatT]",
     ) -> Awaitable[AsyncIterator[AssistantContentChunk]]:
-        """Returns an async iterator over raw assistant chunks."""
+        """Returns an async iterator over content chunks."""
         ...
 
     def chunk_stream(
@@ -239,13 +304,13 @@ class StreamResponse(BaseResponse[FormatT], Generic[StreamT, FormatT]):
         Iterator[AssistantContentChunk]
         | Awaitable[AsyncIterator[AssistantContentChunk]]
     ):
-        """Returns an iterator that yields raw assistant chunks as they are received.
+        """Returns an iterator that yields content chunks as they are received.
 
         The return type depends on the StreamT generic parameter:
         - For StreamResponse[Stream, FormatT]: returns Iterator[AssistantContentChunk] (synchronous)
         - For StreamResponse[AsyncStream, FormatT]: returns Awaitable[AsyncIterator[AssistantContentChunk]]
 
-        This provides access to the raw chunk data including start, delta, and end chunks
+        This provides access to the Mirascope chunk data including start, delta, and end chunks
         for each content type (text, thinking, tool_call). Unlike the streams() method
         that groups chunks by content part, this yields individual chunks as they arrive.
 
@@ -258,7 +323,124 @@ class StreamResponse(BaseResponse[FormatT], Generic[StreamT, FormatT]):
         the LLM), it will proceed to consume it once it has iterated through all the
         cached chunks.
         """
-        raise NotImplementedError()
+        if isinstance(self._chunk_iterator, Iterator):
+            return self._sync_chunk_stream()
+        else:
+            return self._async_chunk_stream()
+
+    def _handle_chunk(self, chunk: AssistantContentChunk) -> None:
+        if chunk.content_type == "text":
+            self._handle_text_chunk(chunk)
+        elif chunk.content_type == "thinking":
+            self._handle_thinking_chunk(chunk)
+        else:
+            raise NotImplementedError
+
+        self._chunks.append(chunk)
+
+    def _handle_text_chunk(
+        self, chunk: TextStartChunk | TextChunk | TextEndChunk
+    ) -> None:
+        if chunk.type == "text_start_chunk":
+            if self._current_content:
+                raise RuntimeError(
+                    "Received text_start_chunk while processing another chunk"
+                )
+            self._current_content = Text(text="")
+            # Text gets included in content even when unfinished.
+            self._content.append(self._current_content)
+            self._texts.append(self._current_content)
+
+        elif chunk.type == "text_chunk":
+            if self._current_content is None or self._current_content.type != "text":
+                raise RuntimeError("Received text_chunk while not processing text.")
+            self._current_content.text += chunk.delta
+
+        elif chunk.type == "text_end_chunk":
+            if self._current_content is None or self._current_content.type != "text":
+                raise RuntimeError("Received text_end_chunk while not processing text.")
+            self._current_content = None
+
+    def _handle_thinking_chunk(
+        self, chunk: ThinkingStartChunk | ThinkingChunk | ThinkingEndChunk
+    ) -> None:
+        if chunk.type == "thinking_start_chunk":
+            if self._current_content:
+                raise RuntimeError(
+                    "Received thinking_start_chunk while processing another chunk"
+                )
+            new_thinking = Thinking(thinking="", signature=None)
+            self._current_content = new_thinking
+
+        elif chunk.type == "thinking_chunk":
+            if (
+                self._current_content is None
+                or self._current_content.type != "thinking"
+            ):
+                raise RuntimeError(
+                    "Received thinking_chunk while not processing thinking."
+                )
+            self._current_content.thinking += chunk.delta
+
+        elif chunk.type == "thinking_end_chunk":
+            if (
+                self._current_content is None
+                or self._current_content.type != "thinking"
+            ):
+                raise RuntimeError(
+                    "Received thinking_end_chunk while not processing thinking."
+                )
+            # Only add to content and thinkings when complete
+            self._current_content.signature = chunk.signature
+            self._content.append(self._current_content)
+            self._thinkings.append(self._current_content)
+            self._current_content = None
+
+    def _sync_chunk_stream(self) -> Iterator[AssistantContentChunk]:
+        """Synchronous implementation of chunk_stream."""
+        if not isinstance(self._chunk_iterator, Iterator):
+            raise TypeError(
+                "Expected Iterator for synchronous streaming"
+            )  # pragma: no cover
+
+        for chunk in self.chunks:
+            yield chunk
+
+        if self.consumed:
+            return
+
+        for chunk, raw_data in self._chunk_iterator:
+            self._handle_chunk(chunk)
+            self._raw.append(raw_data)
+            yield chunk
+
+        self.consumed = True
+
+    async def _async_chunk_stream(
+        self,
+    ) -> AsyncIterator[AssistantContentChunk]:
+        """Asynchronous implementation of chunk_stream."""
+
+        async def generator() -> AsyncIterator[AssistantContentChunk]:
+            if isinstance(self._chunk_iterator, Iterator):
+                raise TypeError(
+                    "Expected AsyncIterator for asynchronous streaming"
+                )  # pragma: no cover
+
+            for chunk in self.chunks:
+                yield chunk
+
+            if self.consumed:
+                return
+
+            async for chunk, raw_data in self._chunk_iterator:
+                self._handle_chunk(chunk)
+                self._raw.append(raw_data)
+                yield chunk
+
+            self.consumed = True
+
+        return generator()
 
     @overload
     def structured_stream(
--- a/python/tests/llm/responses/test_stream_response.py
+++ b/python/tests/llm/responses/test_stream_response.py
@@ -1,0 +1,528 @@
+"""Tests for StreamResponse class."""
+
+from collections.abc import AsyncIterator, Iterator
+from dataclasses import dataclass
+
+import pytest
+
+from mirascope import llm
+from mirascope.llm.content import (
+    AssistantContentChunk,
+    AssistantContentPart,
+    Text,
+    TextChunk,
+    TextEndChunk,
+    TextStartChunk,
+    Thinking,
+    ThinkingChunk,
+    ThinkingEndChunk,
+    ThinkingStartChunk,
+)
+from mirascope.llm.responses import AsyncChunkIterator, ChunkIterator, StreamResponse
+from mirascope.llm.streams import AsyncStream, Stream
+
+
+def create_sync_stream_response(
+    chunks: list[AssistantContentChunk],
+) -> StreamResponse[Stream]:
+    """Create a StreamResponse with a functioning iterator for testing."""
+
+    def sync_chunk_iter() -> ChunkIterator:
+        for chunk in chunks:
+            yield chunk, f"raw_{chunk.type}"
+
+    iterator = sync_chunk_iter()
+
+    response = StreamResponse(
+        provider="openai",
+        model="gpt-4o-mini",
+        input_messages=[llm.messages.user("Test")],
+        chunk_iterator=iterator,
+    )
+    return response
+
+
+def create_async_stream_response(
+    chunks: list[AssistantContentChunk],
+) -> StreamResponse[AsyncStream]:
+    """Create a StreamResponse with a functioning iterator for testing."""
+
+    async def async_chunk_iter() -> AsyncChunkIterator:
+        for chunk in chunks:
+            yield chunk, f"raw_{chunk.type}"
+
+    iterator = async_chunk_iter()
+
+    response = StreamResponse[AsyncStream](
+        provider="openai",
+        model="gpt-4o-mini",
+        input_messages=[llm.messages.user("Test")],
+        chunk_iterator=iterator,
+    )
+    return response
+
+
+@pytest.fixture
+def example_text() -> Text:
+    return Text(text="Hello world")
+
+
+@pytest.fixture
+def example_thinking() -> Thinking:
+    return Thinking(thinking="Let me think...", signature="reasoning")
+
+
+@pytest.fixture
+def example_text_chunks() -> list[AssistantContentChunk]:
+    """Create a complete text chunk sequence for testing."""
+    return [
+        TextStartChunk(type="text_start_chunk"),
+        TextChunk(type="text_chunk", delta="Hello"),
+        TextChunk(type="text_chunk", delta=" "),
+        TextChunk(type="text_chunk", delta="world"),
+        TextEndChunk(type="text_end_chunk"),
+    ]
+
+
+@pytest.fixture
+def example_thinking_chunks() -> list[AssistantContentChunk]:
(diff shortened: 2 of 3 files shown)


teamdandelion:
* **#1197** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1196** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1195** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1195?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1194** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1193** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1193?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1192** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1192?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1191** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1191?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1190** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1190?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1188** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1188?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1187** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1187?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1186** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1185** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1185?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1184** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1184?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1183** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1183?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1183?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/stream_response.py
```diff
@@ -153,12 +174,62 @@ def answer_question(question: str) -> str:
...
+        self.thinkings = self._thinkings
+        self.tool_calls = self._tool_calls
+        self.raw = self._raw
+
+        self.finish_reason = None  # TODO: Add support for finish reason to chunks
+
+        self.messages = list(input_messages) + [AssistantMessage(content=self._content)]
+
+        self._chunk_iterator = chunk_iterator
+        self._current_content = None
```
willbakst: nit: why not just default this and `consumed` when you define them at the class level?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -258,7 +329,121 @@ def chunk_stream(
...
+
+        self._chunks.append(chunk)
+
+    def _handle_text_chunk(
+        self, chunk: TextStartChunk | TextChunk | TextEndChunk
+    ) -> None:
+        if chunk.type == "text_start_chunk":
+            if self._current_content:
+                raise ValueError(
+                    "Received text_start_chunk while processing another chunk"
```
willbakst, line 347: I think this is a `RuntimeError` and not a `ValueError`. My thinking is that technically the received `chunk` is a valid value, it's just that we've hit some snag at runtime that we need to catch?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -258,7 +329,121 @@ def chunk_stream(
...
+                )
+            new_text = Text(text="")
```
willbakst: nit: why not just set `self._current_content = Text(text="")` and then use `self._current_content` to set the others?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -258,7 +329,121 @@ def chunk_stream(
...
+            self._content.append(self._current_content)
+            self._thinkings.append(self._current_content)
+            self._current_content = None
+
+    def _sync_chunk_stream(self) -> Iterator[AssistantContentChunk]:
+        """Synchronous implementation of chunk_stream."""
+        if not isinstance(self._chunk_iterator, Iterator):
+            raise TypeError("Expected Iterator for synchronous streaming")
+
+        for chunk in self.chunks:
```
willbakst, line 406: so if I iterate through the stream, break halfway through, and then try to continue iterating, it's going to restart? That behavior seems weird and should be opt-in?

python/mirascope/llm/responses/stream_response.py
```diff
@@ -1,20 +1,41 @@
...
+ChunkWithRaw: TypeAlias = tuple[AssistantContentChunk, Any]
+"""A chunk paired with its raw provider data."""
+
+ChunkIterator: TypeAlias = Iterator[ChunkWithRaw]
+"""Synchronous iterator yielding chunks with raw data."""
+
+AsyncChunkIterator: TypeAlias = AsyncIterator[ChunkWithRaw]
+"""Asynchronous iterator yielding chunks with raw data."""
+
+StreamChunkIterator: TypeAlias = ChunkIterator | AsyncChunkIterator
```
willbakst: I think this alias adds more confusion that benefit. When first reading it, it was not clear what it was, and I think I would prefer to just use `ChunkIterator | AsyncChunkIterator` directly in its place so it's immediately clear.

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+    Thinking,
+    ThinkingChunk,
+    ThinkingEndChunk,
+    ThinkingStartChunk,
+)
+from mirascope.llm.responses import StreamResponse
+from mirascope.llm.streams import AsyncStream, Stream
+
+
+def create_response_by_mode(
```
willbakst: I think this is a little confusing mixing the sync and async tests.

I would use fixtures and have something like `sync_stream_response` and `async_stream_response` as fixtures you can auto-inject into their respective tests, and then I would have separate tests for sync and async, marking only the async tests as async.

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+    stream = response.chunk_stream()
+    if isinstance(stream, Awaitable):
+        async for chunk in await stream:
+            yield chunk
+    else:
+        for chunk in stream:
+            yield chunk
+
+
+async def consume_chunk_stream(
```
willbakst: I think this is overly complicated. I would rather just see the list comprehension right there in the test and have the sync/async cases separated.

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+    """Create a complete thinking chunk sequence for testing."""
+    return [
+        ThinkingStartChunk(type="thinking_start_chunk"),
+        ThinkingChunk(type="thinking_chunk", delta="Let me"),
+        ThinkingChunk(type="thinking_chunk", delta=" think..."),
+        ThinkingEndChunk(type="thinking_end_chunk", signature="reasoning"),
+    ]
+
+
+def check_response_consistency(
```
willbakst: nit: rename to `check_stream_response_consistency`

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+) -> list[AssistantContentChunk]:
+    """Universal chunk stream consumer for sync/async responses."""
+    return [chunk async for chunk in get_async_chunk_stream(response)]
+
+
+example_text = Text(text="Hello world")
+example_thinking = Thinking(thinking="Let me think...", signature="reasoning")
+
+
+def text_chunk_sequence():
```
willbakst: this should also be a fixture

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+    return [
+        TextStartChunk(type="text_start_chunk"),
+        TextChunk(type="text_chunk", delta="Hello"),
+        TextChunk(type="text_chunk", delta=" "),
+        TextChunk(type="text_chunk", delta="world"),
+        TextEndChunk(type="text_end_chunk"),
+    ]
+
+
+def thinking_chunk_sequence():
```
willbakst: this should also be a fixture

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+        # Resume iteration from the same iterator
+        async for _ in async_stream:
+            ...
+
+        check_response_consistency(response, chunks, [example_text])
+        assert response.consumed is True
+
+    @pytest.mark.parametrize("mode", ["sync", "async"])
+    @pytest.mark.asyncio
+    async def test_partial_iteration_and_restart(self, mode: str):
```
willbakst: this may change depending on resolution to the other comment

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,371 @@
...
+                TextStartChunk(type="text_start_chunk"),
+                TextChunk(type="text_chunk", delta="test"),
+                ThinkingEndChunk(type="thinking_end_chunk", signature=None),
+            ],
+            "Received thinking_end_chunk while not processing thinking",
+        ),
+    ]
+
+    @pytest.mark.parametrize("mode", ["sync", "async"])
+    @pytest.mark.parametrize("chunks,expected_error", INVALID_CHUNK_SEQUENCES)
```
willbakst: You can create a fixture `invalid_sync_stream_response` and `invalid_async_stream_response` that you parameterize on `INVALID_CHUNK_SEQUENCES` and have that directly injected rather than having to construct the response in the test.

python/mirascope/llm/responses/stream_response.py
teamdandelion: It's up to the user:

```py
stream = response.chunk_stream()
processed_chunks = []
for chunk in stream:
  processed_chunks.append(chunk)
  if len(processed_chunks) == 5:
    break

for chunk in stream: 
  # This continues using the same iterator
  # and continues where it left off
  processed_chunks.append(chunk)

# processed_chunks now has all the chunks (no duplication)

# user could also do:
for chunk in response.chunk_stream():
  # This will replay starting from the beginning
```

We could change the contract to make continuation the defualt and have replay be an optional argument, but I think replaying by default is safer (you always see all the data) and it's not hard to keep the original iterator if you want to continue from where you left off.

python/tests/llm/responses/test_stream_response.py
teamdandelion: Ok, redone, take another look.

python/tests/llm/responses/test_stream_response.py
teamdandelion: I updated it, it's not exactly what you wrote but I think it works, tell me what you think.

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+    def test_case(self, request):
+        """Parametrized fixture that provides individual test cases."""
+        return CHUNK_PROCESSING_TEST_CASES[request.param]
+
+    @pytest.mark.parametrize(
+        "test_case",
+        list(range(len(CHUNK_PROCESSING_TEST_CASES))),
+        indirect=True,
+        ids=[case["id"] for case in CHUNK_PROCESSING_TEST_CASES],
+    )
```
teamdandelion: Made CHUNK_PROCESSING_TEST_CASES a variable not a fixture so I can reference it here and in the async test case with the following properties:
- no duplication of the test data
- id for the test case for clear failure messages
- tests don't need to be manually updated if we add another test case

I tried fixture based approaches but couldn't get those 3 properties.

python/mirascope/llm/responses/stream_response.py
```diff
@@ -153,12 +171,60 @@ def answer_question(question: str) -> str:
...
-    """The reason why the LLM finished generating a response.
+    consumed: bool = False
+    """Whether the stream has been fully consumed.
     
-    Will be None until the stream has finished streaming completely.
+    This is True after all chunks have been processed from the underlying iterator.
+    When False, more content may be available by calling the stream methods.
     """
 
+    _current_content: Text | Thinking | ToolCall | None = None
```
willbakst: nit: wondering why we have this at the class level but not e.g. `self._chunks` or other such private attributes. I think we should either have all of them or none of them at the class level for consistency?

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+    return Text(text="Hello world")
+
+
+@pytest.fixture
+def example_thinking():
+    return Thinking(thinking="Let me think...", signature="reasoning")
+
+
+@pytest.fixture
+def text_chunks():
```
willbakst: nit: should this be `example_text_chunks` for naming consistency with above naming convention?

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+    assert stream_response.provider == "openai"
+    assert stream_response.model == "gpt-4o-mini"
+    assert stream_response.finish_reason is None
+    assert isinstance(stream_response._chunk_iterator, AsyncIterator)
+    assert stream_response._current_content is None
+    assert not stream_response.consumed
+    check_stream_response_consistency(stream_response, [], [])
+
+
+class TestResponseChunkStream:
```
willbakst: nit: I would make this either `TestStreamResponseChunkStream` or just `TestChunkStream` given the existing `test_stream_response` namespace?

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+    """Test chunk streaming mechanics for both sync and async."""
+
+    def test_sync_basic_streaming(
+        self,
+        text_chunks,
```
willbakst: nit: why not properly type hint these arguments here and elsewhere?

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+        provider="openai",
+        model="gpt-4o-mini",
+        input_messages=[llm.messages.user("Test")],
+        chunk_iterator=iterator,
+    )
+    return response
+
+
+@pytest.fixture
+def example_text():
```
willbakst: nit: why not properly type the return type here and elsewhere for fixtures?

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+        ThinkingChunk(type="thinking_chunk", delta=" think..."),
+        ThinkingEndChunk(type="thinking_end_chunk", signature="reasoning"),
+    ]
+
+
+def check_stream_response_consistency(
+    response: StreamResponse[Stream | AsyncStream],
+    chunks: list[AssistantContentChunk],
+    content: list[AssistantContentPart],
+):
```
willbakst: nit: missing `-> None` return type

This is actually a good reminder to double check our ruff rules to make sure we have them configured properly, cause this generally shows an error unless we configure it otherwise (i.e. missing return type)

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+            chunk async for chunk in await stream_response.chunk_stream()
+        ]
+
+        check_stream_response_consistency(
+            stream_response, chunks, [example_text, example_thinking]
+        )
+        assert len(streamed_chunks) == 9
+        assert stream_response.consumed is True
+
+    def test_sync_replay_functionality(self, text_chunks, example_text):
```
willbakst: Hmm, I'm not really sure this actually tests this properly? Wouldn't this test still work if it didn't replay and just created a new iterator in both cases?

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+        assert stream_response.consumed is True
+
+    def test_sync_partial_iteration_and_resume(self, text_chunks, example_text):
+        """Test breaking iteration and resuming from cached state with sync response."""
+        stream_response = create_sync_stream_response(text_chunks)
+
+        # Partial iteration stopping early
+        chunk_stream = stream_response.chunk_stream()
+        partial_chunks = []
+        for i, chunk in enumerate(chunk_stream):
```
willbakst: nit: would `partial_chunks.append(next(chunk_stream))` have the same result testing-wise? Also wondering what the benefit of stopping at `i == 2` vs. just the first chunk is

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+            if i == 2:
+                break
+
+        check_stream_response_consistency(
+            stream_response, partial_chunks, [Text(text="Hello ")]
+        )
+        assert stream_response.consumed is False
+
+        # Resume iteration from the same iterator
+        for chunk in chunk_stream:
```
willbakst, line 246: nit: similarly, would `partial_chunks += list(chunk_stream)` have the same effect?

Perhaps this and the above suggestion make it less clear what we're testing? I don't feel strongly about this and am happy to keep this as `for chunk in chunk_stream: ...` so it's clear. For the above suggestion, I think `next` would be cleaner if that works.

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+class TestChunkProcessing:
+    """Test chunk processing with various content types and states."""
+
+    @pytest.fixture
+    def test_case(self, request):
+        """Parametrized fixture that provides individual test cases."""
+        return CHUNK_PROCESSING_TEST_CASES[request.param]
+
+    @pytest.mark.parametrize(
+        "test_case",
```
willbakst: can we rename this to something like `chunk_processing_case` or `chunk_processing_test_case` with a proper type like `ChunkProcessingTestCase`?

See my other comment with a suggested full rewrite of these tests.

python/tests/llm/responses/test_stream_response.py
willbakst: What do you think about this?

```python
@dataclass
class ChunkProcessingTestCase:
    chunks: list[AssistantContentChunk]
    expected_contents: list[list[AssistantContentPart]]


CHUNK_PROCESSING_TEST_CASES: dict[str, ChunkProcessingTestCase] = {
    "empty_text": ChunkProcessingTestCase(
        chunks=[
            TextStartChunk(type="text_start_chunk"),
            TextEndChunk(type="text_end_chunk"),
        ],
        expected_contents=[[Text(text="")], [Text(text="")]],
    ),
    "text_with_deltas": ChunkProcessingTestCase(
        chunks=[
            TextStartChunk(type="text_start_chunk"),
            TextChunk(type="text_chunk", delta="Hello"),
            TextChunk(type="text_chunk", delta=" world"),
            TextEndChunk(type="text_end_chunk"),
        ],
        expected_contents=[
            [Text(text="")],
            [Text(text="Hello")],
            [Text(text="Hello world")],
            [Text(text="Hello world")],
        ],
    ),
    "empty_thinking": ChunkProcessingTestCase(
        chunks=[
            ThinkingStartChunk(type="thinking_start_chunk"),
            ThinkingEndChunk(type="thinking_end_chunk", signature="thoughts"),
        ],
        expected_contents=[[], [Thinking(thinking="", signature="thoughts")]],
    ),
    "thinking_with_deltas": ChunkProcessingTestCase(
        chunks=[
            ThinkingStartChunk(type="thinking_start_chunk"),
            ThinkingChunk(type="thinking_chunk", delta="Let me"),
            ThinkingChunk(type="thinking_chunk", delta=" think..."),
            ThinkingEndChunk(type="thinking_end_chunk", signature="reasoning"),
        ],
        expected_contents=[
            [],
            [],
            [],
            [Thinking(thinking="Let me think...", signature="reasoning")],
        ],
    ),
}


class ChunkProcessingFixtureRequest(pytest.FixtureRequest):
    param: str
    """The ID for the test case."""


@pytest.fixture(
    params=CHUNK_PROCESSING_TEST_CASES.keys(),
    ids=list(CHUNK_PROCESSING_TEST_CASES.keys()),
)
def chunk_processing_test_case(
    request: ChunkProcessingFixtureRequest,
) -> ChunkProcessingTestCase:
    return CHUNK_PROCESSING_TEST_CASES[request.param]


class TestChunkProcessing:
    """Test chunk processing with various content types and states."""

    def test_sync_chunk_processing(
        self, chunk_processing_test_case: ChunkProcessingTestCase
    ) -> None:
        chunks = chunk_processing_test_case.chunks
        stream_response = create_sync_stream_response(chunks)
        stream = stream_response.chunk_stream()
        check_stream_response_consistency(stream_response, [], [])
        for i, expected_content in enumerate(
            chunk_processing_test_case.expected_contents
        ):
            next(stream)
            check_stream_response_consistency(
                stream_response, chunks[: i + 1], expected_content
            )

    @pytest.mark.asyncio
    async def test_async_chunk_processing(
        self, chunk_processing_test_case: ChunkProcessingTestCase
    ) -> None:
        chunks = chunk_processing_test_case.chunks
        stream_response = create_async_stream_response(chunks)
        stream = await stream_response.chunk_stream()
        check_stream_response_consistency(stream_response, [], [])
        for i, expected_content in enumerate(
            chunk_processing_test_case.expected_contents
        ):
            await anext(stream)
            check_stream_response_consistency(
                stream_response, chunks[: i + 1], expected_content
            )
```

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+        for i, state in enumerate(expected_states):
+            await anext(stream)
+            expected_content = state["content"]
+
+            check_stream_response_consistency(
+                stream_response, chunks[: i + 1], expected_content
+            )
+
+
+class TestStreamResponseErrorHandling:
```
willbakst: nit: whatever we choose above we should match here for consistency (i.e. `TestErrorHandling` if we drop `StreamResponse`, or keep it and match this style above by including it everywhere)

python/tests/llm/responses/test_stream_response.py
```diff
@@ -0,0 +1,526 @@
...
+
+            with pytest.raises(RuntimeError, match=expected_error):
+                [chunk async for chunk in await stream_response.chunk_stream()]
+
+    @pytest.mark.asyncio
+    async def test_sync_incorrect_iterator_type_error(self):
+        """Test type errors for incorrect internal stream method calls with sync response."""
+        stream_response = create_sync_stream_response([])
+        with pytest.raises(TypeError):
+            async for _ in await stream_response._async_chunk_stream():
```
willbakst: It's a little funny to have to force this since we'll never actually achieve this `TypeError` in practice. I wonder if it would just be better to ignore those lines for coverage? What do you think?

python/tests/llm/responses/test_stream_response.py
teamdandelion: Dropped StreamResponse consistently.

python/tests/llm/responses/test_stream_response.py
teamdandelion: Sure, that's nice how we can drop this block:

```
    @pytest.mark.parametrize(
        "test_case",
        list(range(len(CHUNK_PROCESSING_TEST_CASES))),
        indirect=True,
        ids=[case["id"] for case in CHUNK_PROCESSING_TEST_CASES],
    )
```

Updating.

python/tests/llm/responses/test_stream_response.py
teamdandelion: Makes sense, done

python/tests/llm/responses/test_stream_response.py
teamdandelion: Done

python/tests/llm/responses/test_stream_response.py
teamdandelion: I don't follow, can you explain what you have in mind? Like, if this was creating a new underlying chunk_iterator? That is passed in as a constructor argument to the StreamResponse so it doesn't really have access to construct a new one, though if we wanted to be really defensive we could write the test to verify that the underlying iterator isn't called on the second pass through. Is that what you want?

python/tests/llm/responses/test_stream_response.py
teamdandelion: Yeah, I'll change ruff rules and do a thorough sweep in a followon PR, for now I just fixed htis file.

python/tests/llm/responses/test_stream_response.py
teamdandelion: Yes it would have the same effect, but I prefer using break inside the iterator loop because I think it's a bit more realistic to likely use cases. No real reason to stop at i==2, I updated it to be simpler and just break after the first chunk. (this and test below)

python/tests/llm/responses/test_stream_response.py
teamdandelion: Yeah, I want to keep it with the for loops to be a bit clearer what we're testing, plus the hypothetical case where we could have implemented special handling on catching `break` (which i think we've talked about in the past). Note the TestChunkProcessing tests use `next(stream)` so we do test both patterns.

python/tests/llm/responses/test_stream_response.py
teamdandelion: Took the rewrite, thanks.

python/tests/llm/responses/test_stream_response.py
willbakst: We could possibly further improve my example by setting `param: ChunkProcessingTestCase` directly and then just returning `request.param`

Something like this:

```python
class ChunkProcessingFixtureRequest(pytest.FixtureRequest):
    param: ChunkProcessingTestCase
    """The chunk processing test case."""


@pytest.fixture(
    params=list(CHUNK_PROCESSING_TEST_CASES.values()),
    ids=list(CHUNK_PROCESSING_TEST_CASES.keys()),
)
def chunk_processing_test_case(
    request: ChunkProcessingFixtureRequest,
) -> ChunkProcessingTestCase:
    return request.param
```

teamdandelion:
### Merge activity

* **Aug 8, 1:35 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1183).
* **Aug 8, 1:36 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1183).

PR #1199, opened 2025-08-08
feat: add ToolSchema / ToolParameterSchema with tests

teamdandelion:
ToolSchema is a Pydantic/JSONSchema based approach for representing tool
defs. The implementation is based on v1's approach. There is a class
method on the schema for constructing one from function signature, and a
good number of tests.

There is overlap with BaseTool's fields. My plan in the next PR is to
add schema onto BaseTool, and have the BaseTool properties become
accessors that pluck from the schema, so we can keep the BaseTool API
but also the nice centralization on the ToolSchema model that I like.


teamdandelion:
* **#1225** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1225?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1224** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1223** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1222** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1222?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1221** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1218** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1218?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1217** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1217?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1216** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1216?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1215** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1214** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1220** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1220?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1219** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1219?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1210** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1209** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1213** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1208** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1207** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1203** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1202** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1206** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1205** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1201** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1200** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1199** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+
+class ToolParameterSchema(BaseModel):
+    """JSON Schema for tool parameters (always an object with properties).
+
+    This contains real JSON Schema as generated by Pydantic, with full support
+    for complex schemas like anyOf, nested objects, validation constraints, etc.
+    Including $defs for complex type references.
+    """
+
+    type: Literal["object"] = "object"
```
willbakst: nit: add newlines between these and do strings like we do for other classes

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+
+import inspect
+from collections.abc import Callable
+from typing import Annotated, Any, Literal, get_args, get_origin, get_type_hints
+
+from pydantic import BaseModel, Field, create_model
+from pydantic.fields import FieldInfo
+
+
+class ToolParameterSchema(BaseModel):
```
willbakst, line 15: I wonder if we need validation here. We should think about turning it off in the config.

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+    for complex schemas like anyOf, nested objects, validation constraints, etc.
+    Including $defs for complex type references.
+    """
+
+    type: Literal["object"] = "object"
+    properties: dict[str, dict[str, Any]] = Field(default_factory=dict)
+    required: list[str] = Field(default_factory=list)
+    additionalProperties: bool = False
+    # Include $defs section for complex type definitions (BaseModel classes, etc.)
+    defs: dict[str, dict[str, Any]] | None = Field(default=None, alias="$defs")
```
willbakst, line 32: I'm assuming then that we would have the `$refs` directly in `properties?

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+
+class ToolSchema(BaseModel):
+    """Provider-agnostic tool schema that encodes tool metadata.
+
+    This class represents the core tool information that can be converted
+    to provider-specific formats by utility functions in each provider's
+    _utils module.
+    """
+
+    name: str
```
willbakst, line 43: nit: same thing here re: newlines and docstrings

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+    description: str
+    parameters: ToolParameterSchema = Field(default_factory=ToolParameterSchema)
+    strict: bool = False
+
+    @classmethod
+    def from_function_signature(
```
willbakst: nit: this should be named `from_function` since it accepts `Callable` and not `Signature`.

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+        cls, fn: Callable, *, strict: bool = False
```
willbakst: nit: we should more strictly type this as the same type as what we bind for tools, right?

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+
+    type: Literal["object"] = "object"
+    properties: dict[str, dict[str, Any]] = Field(default_factory=dict)
+    required: list[str] = Field(default_factory=list)
+    additionalProperties: bool = False
+    # Include $defs section for complex type definitions (BaseModel classes, etc.)
+    defs: dict[str, dict[str, Any]] | None = Field(default=None, alias="$defs")
+
+
+class ToolSchema(BaseModel):
```
willbakst, line 36: Funny enough, this was my original intention with `ToolDef`. Might be worth looking back at that to see what I did as reference for here.

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+    name: str
+    description: str
+    parameters: ToolParameterSchema = Field(default_factory=ToolParameterSchema)
+    strict: bool = False
+
+    @classmethod
+    def from_function_signature(
+        cls, fn: Callable, *, strict: bool = False
+    ) -> ToolSchema:
+        """Create ToolSchema from function signature and docstring.
```
willbakst: nit: in line with above comment, we should make it clear that it creates it from the function by inspecting it's signature and docstring

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+        Uses Pydantic's create_model to dynamically build a model from the function
+        signature, then extracts the JSON schema from it. This leverages Pydantic's
+        robust type conversion and union handling.
+
+        Args:
+            fn: The function to extract schema from
+            strict: Whether the tool should use strict mode when supported
+
+        Returns:
+            ToolSchema representing the function
```
willbakst: nit: should be "A `ToolSchema` representing the function." with backticks on `ToolSchema`

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for param in inspect.signature(fn).parameters.values():
+            if param.name in ("self", "cls"):
+                continue
+
+            param_type = hints.get(param.name, Any)
+            default = ... if param.default is inspect.Parameter.empty else param.default
+
+            # Handle Annotated types for parameter descriptions
+            if get_origin(param_type) is Annotated:
```
willbakst, line 85: At some point we will want to consider handling `FromCallArgs` here so that users can inject call arguments into tools easily, although context will likely be the answer in most cases anyway.

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+                # Extract the core type and any Field info
+                args = get_args(param_type)
+                core_type = args[0]
+
+                # Look for a FieldInfo in the annotations
+                field_info = None
+                for annotation in args[1:]:
+                    if isinstance(annotation, FieldInfo):
```
willbakst, line 92: are we looking for `FieldInfo` or `Field`? My understanding is we're looking for `Field`, but maybe I'm mistaken?

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -0,0 +1,117 @@
...
+        robust type conversion and union handling.
+
+        Args:
+            fn: The function to extract schema from
+            strict: Whether the tool should use strict mode when supported
+
+        Returns:
+            ToolSchema representing the function
+        """
+        # Extract function metadata
```
willbakst: I don't think we need a single one of the comments in this function so long as we retain newlines between each logical block. I find the variable and method names to be somewhat self-documenting. WDYT?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+def test_simple_tool_no_params() -> None:
+    """Test tool with no parameters."""
+
+    def simple_tool() -> str:
+        """A simple tool with no parameters."""
+        raise NotImplementedError
+
+    schema = llm.tools.ToolSchema.from_function_signature(simple_tool)
+
+    assert schema.name == "simple_tool"
```
willbakst: nit: use `snapshot` here and elsewhere?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+        raise NotImplementedError
+
+    schema = llm.tools.ToolSchema.from_function_signature(tool)
+    props = schema.parameters.properties
+
+    assert schema.parameters.required == ["message"]
+    assert get_property_type(props["message"]) == ["string", "null"]
+
+
+def test_basic_type_mapping() -> None:
```
willbakst: nit: use "primitive" instead of "basic"?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+
+    schema = llm.tools.ToolSchema.from_function_signature(tool_with_mixed_params)
+
+    # Only message is required (no default), count is optional (has default)
+    assert schema.parameters.required == ["message"]
+    assert "count" in schema.parameters.properties
+    assert "message" in schema.parameters.properties
+
+
+def test_param_may_be_none() -> None:
```
willbakst: nit: update to `test_tool_with_none_param`? although it's testing a union type and not just `message: None` so maybe `test_tool_with_optional_param`?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+
+
+def get_array_items_type(prop: dict[str, Any]) -> str | list[str] | None:
+    """Extract array items type from JSON Schema property."""
+    if prop.get("type") == "array" and "items" in prop:
+        return get_property_type(prop["items"])
+    return None
+
+
+def test_simple_tool_no_params() -> None:
```
willbakst: nit: drop "simple" and just have `test_tool_with_no_params`

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    """Test tool with no parameters."""
+
+    def simple_tool() -> str:
```
willbakst: in line with above comment name this `tool_with_no_params`

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    props = schema.parameters.properties
+
+    assert schema.parameters.required == ["message"]
+    assert get_property_type(props["message"]) == ["string", "null"]
+
+
+def test_basic_type_mapping() -> None:
+    """Test Python types map correctly to JSON Schema types."""
+
+    def tool_with_basic_types(text: str, number: int, value: float, flag: bool) -> str:
```
willbakst: nit: consider parameterizing this test

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    schema = llm.tools.ToolSchema.from_function_signature(tool_with_basic_types)
+    props = schema.parameters.properties
+
+    assert get_property_type(props["text"]) == "string"
+    assert get_property_type(props["number"]) == "integer"
+    assert get_property_type(props["value"]) == "number"
+    assert get_property_type(props["flag"]) == "boolean"
+
+
+def test_list_parameter_handling() -> None:
```
willbakst: nit: match naming above, e.g. `test_tool_with_list_param`

python/tests/llm/tools/test_tool_schema.py
willbakst: nit: same thing re: update to "primitive" to match updated test name

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    schema = llm.tools.ToolSchema.from_function_signature(tool_with_list)
+    props = schema.parameters.properties
+
+    assert get_property_type(props["items"]) == "array"
+    assert get_array_items_type(props["items"]) == "string"
+    assert get_property_type(props["numbers"]) == "array"
+    assert get_array_items_type(props["numbers"]) == "integer"
+
+
+def test_dict_parameter_handling() -> None:
```
willbakst: nit: same as above and here and elsewhere, update to `test_tool_with_dict_param`

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+
+    assert get_property_type(props["config"]) == "object"
+    # additionalProperties should contain the value type schema (string)
+    assert "additionalProperties" in props["config"]
+    additional_props = props["config"]["additionalProperties"]
+    assert get_property_type(additional_props) == "string"
+
+
+def test_optional_type_handling() -> None:
+    """Test Optional[T] is handled correctly."""
```
willbakst: nit: weird using `Optional[T]` here instead of `T | None` with backticks

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    props = schema.parameters.properties
+
+    assert get_property_type(props["config"]) == "object"
+    # additionalProperties should contain the value type schema (string)
+    assert "additionalProperties" in props["config"]
+    additional_props = props["config"]["additionalProperties"]
+    assert get_property_type(additional_props) == "string"
+
+
+def test_optional_type_handling() -> None:
```
willbakst: hmm, we should be mindful of using "optional"

Generally I think of `| None` types as "optional" because in Python <3.9 we had to use the `Optional` type. My feeling is that this would be better named someting like `test_tool_with_default_value` or something? Worth matching this up with my suggestion for the other test naming

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    """Test Optional[T] is handled correctly."""
+
+    def tool_with_optional(name: str, nickname: str | None = None) -> str:
```
willbakst: nit: last comment on this, here and elsewhere just make sure to update any of these inner methods to match the test as mentioned earlier

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+        raise NotImplementedError
+
+    schema = llm.tools.ToolSchema.from_function_signature(tool_with_optional)
+    props = schema.parameters.properties
+
+    # Optional parameter with None default should allow both string and null
```
willbakst: I think these comments are unnecessary? The test seems self-explanatory in it's naming

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+
+def test_no_docstring_gets_default() -> None:
+    """Test tool without docstring gets default description."""
+
+    def tool_no_docstring(x: int, y: int) -> int:
+        raise NotImplementedError
+
+    schema = llm.tools.ToolSchema.from_function_signature(tool_no_docstring)
+
+    assert schema.description == "Tool: tool_no_docstring"
```
willbakst: nit: noticing this here, the default should probably just be the method name i.e. `tool_no_docstring` without the `Tool:` inclusion

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    )
+
+    assert schema_normal.strict is False
+    assert schema_strict.strict is True
+
+
+def test_union_type_support() -> None:
+    """Test union types are properly supported in JSON Schema."""
+
+    def tool_with_unions(data: str | int, value: int | float) -> str:
```
willbakst: This seems duplicated with `| None` since that's also a union technically? Or do these differ in some way? Just in the semantics of testing them as `Union` vs. `Optional`?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    """Test unions with more than two types are handled correctly."""
+
+    def tool_with_complex_union(value: str | int | float) -> str:
+        """A tool with a three-type union parameter."""
+        raise NotImplementedError
+
+    schema = llm.tools.ToolSchema.from_function_signature(tool_with_complex_union)
+    props = schema.parameters.properties
+
+    # Three-type union should include all types, not just the first
```
willbakst: nit: this comment is a little weird in that it mentions "not just the first" even though I would never expect just the first? I would remove this comment and others given the self-explanatory test names and variable naming etc.?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    with pytest.raises(
+        PydanticSchemaGenerationError
+    ):  # Pydantic will throw a schema generation error
+        llm.tools.ToolSchema.from_function_signature(tool_with_custom_type)
+
+
+def test_nested_union_handling() -> None:
+    """Test unions containing other unions are properly flattened."""
+
+    def tool_with_nested_unions(value: str | int | float) -> str:
```
willbakst: what is nested about this union?

python/tests/llm/tools/test_tool_schema.py
```diff
@@ -0,0 +1,489 @@
...
+    assert "additionalProperties" in props["config"]
+
+
+def test_custom_class_throws_error() -> None:
+    """Test that unsupported custom classes throw a clear error."""
+
+    class CustomClass:
+        pass
+
+    def tool_with_custom_type(data: CustomClass) -> str:
```
willbakst: this makes my earlier comment more necessary around properly typing `Callable` as the input to `from_function` so that something like this throws a type error.

Inputs to a tool need to be generatable by the LLM, meaning they should likely be bound to something like `Jsonable` (excluding our JSON class type, which we may want to consider removing as part of the type)

python/tests/llm/tools/test_tool_schema.py
willbakst, line 1: Am I crazy that a good majority of these tests feel like they are testing Pydantic rather than our ToolSchema stuff?

I feel like we could greatly reduce the complexity of our tests if we refactored our handling in `tool_schema.py` to have something like `extract_field_definitions` that we tested followed with simple follow-up tests for the `from_function` interface?

I feel like if we correctly extract things from the signature, we should be able to trust at that point that Pydantic will produce the proper schema?

python/mirascope/llm/tools/tool_schema.py
teamdandelion: Field is a fn that returns a FieldInfo

python/mirascope/llm/tools/tool_schema.py
teamdandelion: Yup

python/mirascope/llm/tools/tool_schema.py
teamdandelion: Discussed offline, we're not going to keep the params / return type at this level.

python/tests/llm/tools/test_tool_schema.py
teamdandelion: I've changed it to this:

```py
    @classmethod
    def from_function(
        cls, fn: Callable[..., Jsonable], *, strict: bool = False
    ) -> ToolSchema:
```

But since this doesn't check the parameter types, I left this test in. I tried a protocol based approach:

```py
class JsonableCallable(Protocol[JsonableCovariantT]):
    """Protocol for callables that accept and return only JSON-serializable values."""

    def __call__(self, *args: Jsonable, **kwargs: Jsonable) -> Jsonable:
        """Call the function with JSON-serializable arguments and return value."""
        ...
```

But I ran into issues all over the place eg

```py
    def simple_tool() -> str:
        """A simple tool."""
        raise NotImplementedError

Argument of type "() -> str" cannot be assigned to parameter "fn" of type "JsonableCallable[Unknown]" in function "from_function"
  Type "() -> str" is not assignable to type "(*args: Jsonable, **kwargs: Jsonable) -> Jsonable"
    Parameter "*args" has no corresponding parameter
    Parameter "**kwargs" has no corresponding parameterPylancereportArgumentType
```

And 

```py
    class ExampleModel(BaseModel):
        """An example model for testing."""

        id: int
        name: str
        tags: list[str] = []

    def complex_tool(
        basic_param: str,
        annotated_param: Annotated[
            int, Field(description="An annotated integer parameter")
        ],
        model_param: ExampleModel,
        optional_param: str = "default",
        union_param: str | int = "default_union",
    ) -> str:
        """A comprehensive tool for testing schema generation.

        This tool exercises various parameter types to ensure our schema
        generation handles complex cases correctly.
        """
        raise NotImplementedError

Argument of type "(basic_param: str, annotated_param: int, model_param: ComplexModel, optional_param: str = "default", union_param: str | int = "default_union") -> str" cannot be assigned to parameter "fn" of type "JsonableCallable[Unknown]" in function "from_function"
  Type "(basic_param: str, annotated_param: int, model_param: ComplexModel, optional_param: str = "default", union_param: str | int = "default_union") -> str" is not assignable to type "(*args: Jsonable, **kwargs: Jsonable) -> Jsonable"
    Parameter 1: type "Jsonable" is incompatible with type "str"
      Type "Jsonable" is not assignable to type "str"
        "JsonableObject" is not assignable to "str"
    Parameter 2: type "Jsonable" is incompatible with type "int"
      Type "Jsonable" is not assignable to type "int"
        "JsonableObject" is not assignable to "int"
    Parameter 3: type "Jsonable" is incompatible with type "ComplexModel"
```

python/mirascope/llm/tools/tool_schema.py
willbakst: Feel free to add a TODO with a note to revisit this and consider it resolved

python/mirascope/llm/types/jsonable.py
```diff
@@ -27,3 +27,17 @@ def json(self) -> str:
...
+
+
+JsonableT = TypeVar("JsonableT", bound=Jsonable)
+"""Type variable for tool output types.
+
+This TypeVar represents the return type of tool functions, which must be
+serializable to JSON (bound to Jsonable) for LLM consumption.
+"""
+
+CovariantT = TypeVar("CovariantT", covariant=True)
```
willbakst: nit: this is weird here? I feel like this should remain in `type_vars` if anything

python/tests/llm/tools/test_tool_schema.py
willbakst: Yeah I think we can revisit this later if we deem it important and short term can just document that arguments need to be something LLMs can generate.

python/mirascope/llm/types/jsonable.py
teamdandelion: You're right, it def should. Fixed

teamdandelion:
### Merge activity

* **Aug 13, 10:51 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1199).
* **Aug 13, 10:52 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1199).

PR #1204, opened 2025-08-11
refactor: streamline llm models

teamdandelion:
Based on a lengthy offline discussion:
- There are no longer OpenAI/Google/Anthropic models, just the base LLM
  class
- The LLM class methods don't take params to avoid potential precedence
  confusion between params passed in at call time vs params set on the
  model. If you are using a model override, it overrides all params, and
  user can take responsibility for managing that with user-level code
  that shares default parameters across calls and overrides.
- We swap the order of client and params for the model

I renamed models/base.py to models/llm.py because it's no longer a base,
and I prefer when single-class files have filenames matching
the class.


teamdandelion:
* **#1204** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

PR #1211, opened 2025-08-13
fix: update fallback to use context so it can work with streaming

### LINKED ISSUE #1120: Fallback not working for streaming setup
### Description

Hi everyone !

When trying fallback using streaming it won't retry :
```python

# test blank gemini

from typing import AsyncGenerator
from google.genai.errors import ClientError
from tenacity import retry, stop_after_attempt, wait_exponential, RetryError

from mirascope import llm
from mirascope.core import prompt_template, google
from mirascope.integrations.logfire import with_logfire
from mirascope.retries import fallback, FallbackError
import logfire

logfire.configure()

@fallback(
    RetryError,
    [
        {
            "catch": RetryError,
            "provider": "openai",
            "model": "gpt-4.1",
        }
    ]
)
@with_logfire()
@llm.call(
    provider="google",
    model="gemini-2.5-flash-FFF",
    stream=True,
)
@prompt_template(
    """
    SYSTEM: {system_message}

    USER: Hello !Qui es tu ?
    """
)
async def test(
    system_message: str,
):
    pass

res = await test("teest")

async for chunk, _ in res:
    print(chunk.chunk.text, flush=False, end=None)

```

@willbakst said this would require to update the fallback decorator to use `llm.context` instead of `llm.override`

```TOML

```

willbakst:
Closes #1120


teamdandelion (commented):
No tests? (Am open to "it's too annoying to test & we're going to be replacing it with v2 anyway)

willbakst:
> No tests? (Am open to "it's too annoying to test & we're going to be replacing it with v2 anyway)

Yeah my feeling here was that we already have tests for fallback that still succeed and adding additional tests felt like a little bit of a waste of time given how soon we're planning the `v2` release.

Although I'm not sure why the tests are failing in CI when they are passing locally. Happy to add some tests for this alongside my debugging.

PR #1212, opened 2025-08-13
fix: issue with mistral client when not set at a global level

### LINKED ISSUE #1098: Mistral API call fails without explicit client initialization
### Description

When using the `@llm.call` decorator without passing an explicit Mistral client instance, the code fails to execute. The decorator appears to not properly handle API key configuration when no client is provided.

**Code that fails:**
```python
from mirascope import llm


@llm.call("mistral", "mistral-small-latest")
def answer_question(question: str) -> str:
    return f"Answer to your question: {question}"


response = answer_question("What is the capital of France?")
print(response.content)
```

**Code that works:**
```python
from mirascope import llm
from mistralai import Mistral
import os


@llm.call(
    "mistral", "mistral-small-latest", client=Mistral(os.environ["MISTRAL_API_KEY"])
)
def answer_question(question: str) -> str:
    return f"Answer to your question: {question}"


response = answer_question("What is the capital of France?")
print(response.content)
```

**Python Version:** 3.13
**Mirascope Version:** 1.25.2
**OS Version:** macOS Darwin 24.5.0
**Related packages:**
- mistralai: 1.9.1

willbakst:
Closes #1098

### DIFF
--- a/mirascope/core/mistral/_utils/_setup_call.py
+++ b/mirascope/core/mistral/_utils/_setup_call.py
@@ -39,6 +39,16 @@ from ..tool import MistralTool
 from ._convert_common_call_params import convert_common_call_params
 from ._convert_message_params import convert_message_params
 
+global_client: Mistral | None = None
+
+
+def _get_default_client() -> Mistral:
+    """Get or create the default Mistral client."""
+    global global_client
+    if global_client is None:
+        global_client = Mistral(api_key=os.environ["MISTRAL_API_KEY"])
+    return global_client
+
 
 @overload
 def setup_call(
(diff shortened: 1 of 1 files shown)


teamdandelion (approved):
Change looks fine, but the description is a little unclear to me. "fix: issue with mistral client when not set at a global level" makes me think there is some other issue being fixed? Or is the idea just that we want the default client to be shared across calls?

willbakst:
> Change looks fine, but the description is a little unclear to me. "fix: issue with mistral client when not set at a global level" makes me think there is some other issue being fixed? Or is the idea just that we want the default client to be shared across calls?

Mistral is super annoying and if we define the client locally inside of the setup function it fails. This was the only way I could figure out how to have it succeed without the user setting the client directly. Interestingly, I also couldn't replicate the bug inside of the Mirascope repo but could replicate in a fresh project...

PR #1239, opened 2025-08-14
feat: implement and test LLM (non-context branches)

teamdandelion:

PR #1240, opened 2025-08-14
feat: implement Call/AsyncCall classes

teamdandelion:


teamdandelion:
* **#1248** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1248?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1247** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1247?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1246** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1246?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1244** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1244?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1245** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1245?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1243** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1243?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1241** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1241?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1240** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1240?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1240?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1242** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1242?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1239** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1239?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/calls/call.py
```diff
@@ -13,46 +13,122 @@
...
-# (vs UserMessagePromotable which may be opaque), and rename `output` arg
+from .promote_prompt_result import promote_prompt_result
 
 
 @dataclass
 class Call(BaseCall[P, Prompt, Toolkit, FormatT], Generic[P, FormatT]):
     """A class for generating responses using LLMs."""
 
-    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Response[FormatT]:
+    @overload
```
willbakst, line 22: Do we need this if `FormatT` can be `None`? If no, we should update for the other methods as well.

python/mirascope/llm/calls/call.py
```diff
@@ -13,46 +13,122 @@
...
+
+    @overload
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Response[FormatT]: ...
+
+    def __call__(
+        self, *args: P.args, **kwargs: P.kwargs
+    ) -> Response | Response[FormatT]:
         """Generates a response using the LLM."""
-        raise NotImplementedError()
+        messages = promote_prompt_result(self.fn(*args, **kwargs))
```
willbakst: hmm, don't we already have an `llm.prompt` method that does this?

python/mirascope/llm/calls/call.py
```diff
@@ -13,46 +13,122 @@
...
+    @overload
+    def call(self: "Call[P, None]", *args: P.args, **kwargs: P.kwargs) -> Response: ...
+
+    @overload
+    def call(self, *args: P.args, **kwargs: P.kwargs) -> Response[FormatT]: ...
+
+    def call(self, *args: P.args, **kwargs: P.kwargs) -> Response | Response[FormatT]:
         """Generates a response using the LLM."""
-        raise NotImplementedError()
+        return self(*args, **kwargs)
```
willbakst: nit: I think we should switch this and implement call here and have `__call__` use `self.call`

python/mirascope/llm/calls/promote_prompt_result.py
```diff
@@ -0,0 +1,32 @@
...
+def _is_messages(
+    messages_or_content: list[Message] | UserContent,
+) -> TypeIs[list[Message]]:
+    return isinstance(messages_or_content, list) and all(
+        isinstance(item, SystemMessage | UserMessage | AssistantMessage)
+        for item in messages_or_content
+    )
+
+
+def promote_prompt_result(result: list[Message] | UserContent) -> list[Message]:
```
willbakst: just repeating myself here that we should DRY on `llm.prompt`

python/mirascope/llm/calls/call.py
teamdandelion: What I get if I try removing the overloads and just return `Response[FormatT]`:

```
/Users/dandelion/git/mirascope/v2/python/mirascope/llm/calls/call.py
  /Users/dandelion/git/mirascope/v2/python/mirascope/llm/calls/call.py:26:16 - error: Type "Response[FormatT@Call]* | Response[None]" is not assignable to return type "Response[FormatT@Call]"
    Type "Response[FormatT@Call]* | Response[None]" is not assignable to type "Response[FormatT@Call]"
      "Response[None]" is not assignable to "Response[FormatT@Call]"
        Type parameter "FormatT@Response" is invariant, but "None" is not the same as "FormatT@Call" (reportReturnType)
```

Note we don't need the overloads if we return `Response[None] | Response[FormatT]`, however I think we want the precise types to propagate thru.

I'm not exactly sure why the fact that `FormatT` can be `None` isn't saving us from needing the overload. Maybe something we could dig into more together?

python/mirascope/llm/calls/call.py
teamdandelion: Done

python/mirascope/llm/calls/call.py
teamdandelion: `llm.prompt` has a slightly different contract, as it's a decorator that converts (content prompt or messages prompt) into messages prompt. (And optionally handles template strings.)

We could take the approach of wrapping `Call.fn` with the prompt decorator when the Call is created, rather than leaving `Call.fn` undecorated and then promoting the content at call time. There's no runtime performance implication either way, since the prompt decorator will basically wrap the function so this utility is called on its output before returning.  

I like doing it this way because it's accurate to the `Call` type signature (as currently defined it can take any `Prompt`, whether a `ContentPrompt` or a `MessagesPrompt`) and it allows `Call` to stay a dataclass rather than needing a custom init to make sure the prompt decorator is applied.

It'll probably make sense to move the promote_prompt_result utility into the prompts module once we implement `llm.prompt` though, since it will want this same logic. But I don't see a need to change the way Call is implemented in this PR.

python/mirascope/llm/calls/promote_prompt_result.py
teamdandelion: When we implement `llm.prompt` we should have it use this function under the hood, so no repetition

python/mirascope/llm/calls/call.py
willbakst: I would just do `messages = llm.prompt(self.fn)(*args, **kwargs)` thereby promoting the method inline. In the event that it's already been decorated, this would be a no-op

python/mirascope/llm/calls/call.py
willbakst: Hmm, yeah I'm not 100% sure, we can sync on this

python/mirascope/llm/calls/call.py
teamdandelion: Ok, I implemented @llm.prompt downstack and changed the implementation here to `llm.prompt()(self.fn)(*args, **kwargs)` (almost exactly what you suggested, cant write it your way because if an arg is provided to `llm.prompt` it's the template). 

However it would double-decorate as implemented, let me know what you're thinking for getting the no-op semantics

willbakst: We should be able to do this:

```python
@llm.prompt
def my_prompt(): ...
```

Just like we do with `llm.tool` right?

python/mirascope/llm/calls/call.py
```diff
@@ -14,45 +14,120 @@
...
+        | StreamResponse[FormatT]
+        | AsyncStreamResponse[FormatT]
+        | Response
+        | StreamResponse
+        | AsyncStreamResponse,
+        content: UserContent,
+    ) -> StreamResponse[FormatT] | StreamResponse:
         """Generate a new stream by continuing from a previous output, plus new user content."""
-        raise NotImplementedError()
+
```
willbakst: Why a newline here? Remove any others like this as well

python/mirascope/llm/calls/call.py
teamdandelion: No, sadly if you do that then you will get an error like:
```
Argument of type "(name: str, *, title: str) -> str" cannot be assigned to parameter "template" of type "str" in function "prompt"
  "FunctionType" is not assignable to "str"PylancereportArgumentType
```

Note that I still prefer making `llm.prompt` just a decorator with no template support, and having `llm.prompt_template` be the decorator that takes a template string. In which case you could do `@llm.prompt` as you suggested.

python/mirascope/llm/calls/call.py
willbakst: As mentioned offline, we should explicitly overload `self` in both overloads rather than leaving the second one empty.

python/mirascope/llm/calls/call.py
willbakst: Misunderstanding here I think. I was saying to update `llm.prompt` to handle this.

I've commented a reimplementation of the decorator overloads to enable this in the PR one item down in this stack.

teamdandelion:
### Merge activity

* **Aug 18, 7:33 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1240).
* **Aug 18, 7:35 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1240) couldn't merge this PR because it failed for an unknown reason (Encountered a GitHub permissions error trying to merge. Please check your authentication settings and try again).

PR #1244, opened 2025-08-17
refactor: replace LLM.create with llm.model and turn LLM itself into a context manager instead

willbakst:

PR #1243, opened 2025-08-17
feat: implement llm.Toolkit and llm.AsyncToolkit

teamdandelion:
note: I simplified the base exception class to resolve a circular
import. Let's revisit the exceptions when we are doing
validation/reliability

PR #1242, opened 2025-08-15
feat: implement @llm.prompt decorator (no context or template)

teamdandelion:

PR #1241, opened 2025-08-15
feat: implement call decorator (WIP)

teamdandelion:

PR #1245, opened 2025-08-18
test: refactor to centralize API key setup in fixtures

teamdandelion:


teamdandelion:
* **#1248** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1248?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1247** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1247?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1246** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1246?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1244** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1244?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1245** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1245?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1243** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1243?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1241** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1241?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1240** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1242** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1242?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1239** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1239?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/tests/conftest.py
```diff
@@ -85,24 +85,32 @@ def vcr_config() -> VCRConfig:
...
 
 @pytest.fixture(scope="session")
-def anthropic_client() -> llm.AnthropicClient:
-    """Create an AnthropicClient instance with appropriate API key."""
-    load_dotenv()
-    api_key = os.getenv("ANTHROPIC_API_KEY") or "dummy-key-for-vcr-tests"
-    return llm.AnthropicClient(api_key=api_key)
+def openai_client(setup_provider_keys: None) -> llm.OpenAIClient:
+    """Create an OpenAIClient instance with appropriate API key."""
+    return llm.clients.get_client("openai")
```
teamdandelion: nb. these are just thin wrappers around get_client now, with the convenience that they invoke the setup_provider_keys fixture. If we wanted we could set autouse=True for setup_provider_keys, and then have test code just call get_client to get the client. Arguably it's still more convenient to get the client from fixture than by calling get_client though.

teamdandelion: As I think about this more I'm leaning towards setting autouse=True on setup_provider_keys, since it's barely any overhead, it's once per session, and any full test run will invoke it anyway

python/tests/conftest.py
```diff
@@ -85,24 +85,32 @@ def vcr_config() -> VCRConfig:
...
-    api_key = os.getenv("OPENAI_API_KEY") or "dummy-key-for-vcr-tests"
-    return llm.OpenAIClient(api_key=api_key)
+    os.environ.setdefault("ANTHROPIC_API_KEY", "dummy-anthropic-key")
+    os.environ.setdefault("GOOGLE_API_KEY", "dummy-google-key")
+    os.environ.setdefault("OPENAI_API_KEY", "dummy-openai-key")
 
 
 @pytest.fixture(scope="session")
-def google_client() -> llm.GoogleClient:
+def anthropic_client(setup_provider_keys: None) -> llm.AnthropicClient:
```
willbakst: If you're not going to pass keys to the clients, I would remove this dep and just set `autouse` true in the provider key fixture

python/tests/conftest.py
willbakst: I would just do both as I've mentioned in other comments. Set `autouse=True` but also have the `openai_client` etc. fixtures.

Another option would be to set `autouse=True` but separate each key setting into separate `openai_api_key` fixtures and then have the clients depend on those and use them (but with `autouse=True` you can then test clients without setting an explicit client).

python/tests/llm/calls/test_call_decorator.py
```diff
@@ -130,10 +130,8 @@ async def async_prompt() -> str:
 
 
 @pytest.mark.vcr()
-def test_call_decorator_e2e_smoke_anthropic(
-    anthropic_client: llm.AnthropicClient,
-) -> None:
-    @llm.call(provider="anthropic", model="claude-sonnet-4-0", client=anthropic_client)
+def test_call_decorator_e2e_smoke_anthropic(setup_provider_keys: None) -> None:
```
willbakst: Hmm, yeah if we're not setting clients and we set `autouse=True`, then we won't be testing setting the client, although I'm not sure that really matters? In this setup, I think we could actually boil the custom client test case into a single case where we use e.g. `openai_client` fixture and just assert that `call.client` is in fact equal to the custom client

teamdandelion:
### Merge activity

* **Aug 18, 8:08 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1245).
* **Aug 18, 8:13 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1245) rebased this pull request as part of a merge.
* **Aug 18, 8:14 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1245).

PR #1249, opened 2025-08-19
feat: implement llm.format decorator

teamdandelion:

PR #1250, opened 2025-08-19
feat: implement format parsing on BaseResponse

teamdandelion:


teamdandelion:
* **#1253** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1253?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1254](https://github.com/Mirascope/mirascope/pull/1254) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1254?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1255](https://github.com/Mirascope/mirascope/pull/1255) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1255?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1251** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1252** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1252?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1250** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1250?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1250?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1249** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/base_stream_response.py
```diff
@@ -110,6 +112,7 @@ def __init__(
         model: "Model",
         input_messages: Sequence[Message],
         chunk_iterator: ChunkIteratorT,
+        format: type[FormatT] = NoneType,
```
willbakst: nit: let's put this after `model` and before `input_messages`

python/mirascope/llm/responses/response.py
```diff
@@ -25,6 +30,7 @@ def __init__(
         assistant_message: AssistantMessage,
         finish_reason: FinishReason | None,
         raw: Any,  # noqa: ANN401
+        format: type[FormatT] = NoneType,
```
willbakst: nit: ditto here, put it after `model` and before `input_messages`

python/mirascope/llm/responses/response.py
```diff
@@ -57,14 +64,26 @@ def __init__(
...
-        It will parse the response content according to the specified format (if present)
-        and return a structured object.
-
         Returns:
             The formatted response object of type FormatT.
 
         Raises:
-            ValueError: If the response cannot be formatted according to the
-                specified format.
+            ValueError: If no format type is specified or if the response cannot be
```
willbakst: Hmm, if no format type is specified, shouldn't this just return `None` because `type[FormatT]` will be `type[None]`?

python/mirascope/llm/responses/response.py
```diff
@@ -57,14 +64,26 @@ def __init__(
...
-        raise NotImplementedError()
+        format_type = self.format_type
+        if format_type == NoneType:
+            raise ValueError("No format type specified for this response")
+
+        text = "".join(text.text for text in self.texts)
+
+        try:
+            parsed_json = json.loads(text)
+            result = cast(BaseModel, format_type).model_validate(parsed_json)
```
willbakst: if we bind `format_type` to `type[FormatT]` and `FormatT` is bound to `BaseModel`, why do we need this cast?

python/mirascope/llm/responses/response.py
```diff
@@ -57,14 +64,26 @@ def __init__(
...
+            return cast(FormatT, result)
```
willbakst: ditto here, shouldn't we get type safety that `model_validate` will return `FormatT`?

python/mirascope/llm/responses/response.py
```diff
@@ -57,14 +64,26 @@ def __init__(
...
+        except json.JSONDecodeError as e:
+            raise ValueError(f"Response text is not valid JSON: {e}") from e
```
willbakst: Hmm, should we be handling these errors at all? I feel like these are things that should be bubbled up for the user to handle as part of reliability.

If anything, we should only ever raise the exact same error (i.e. it's weird to convert a `JSONDecodeError` into a `ValueError`).

python/mirascope/llm/responses/response.py
```diff
@@ -57,14 +64,26 @@ def __init__(
...
+        except ValidationError as e:
+            raise ValueError(
```
willbakst: Same here. The user will expect `ValidationError` during reliability handling with Pydantic's message, so I think in this case we don't actually need to do anything

python/tests/llm/responses/test_response.py
```diff
@@ -239,3 +222,112 @@ def test_multiple_text_response_pretty() -> None:
...
+        provider="openai",
+        model="gpt-4o-mini",
+        input_messages=[],
+        assistant_message=assistant_message,
+        raw={"test": "response"},
+        finish_reason=llm.FinishReason.END_TURN,
+        format=Book,
+    )
+
+    with pytest.raises(ValueError, match="Response text is not valid JSON"):
```
willbakst: Note given my earlier comment that these tests will still apply except they will raise their actual errors rather than the converted value error

python/mirascope/llm/responses/response.py
teamdandelion: Good point, changed. (BTW I moved this all to BaseResponse baesd on our chat today.)

python/mirascope/llm/responses/response.py
teamdandelion: Because `FormatT` could be `None`, and checking against `if format_type == NoneType` doesn't seem to refine it for pyright. And None doesn't have model_validate.

python/mirascope/llm/responses/response.py
teamdandelion: As above; I can't cast it to FormatT because FormatT could be None, so I cast it to BaseModel, but then need a second cast to FormatT.

python/mirascope/llm/responses/response.py
teamdandelion: Agreed. I left it this way because I was matching the interface as declared in the docstring, but I figured we would change it once we did our validation pass. I redid this to propagate the original exceptions, and updated the docstring to match.

python/mirascope/llm/responses/base_response.py
```diff
@@ -46,21 +51,65 @@ class BaseResponse(ABC, Generic[FormatT]):
...
             The formatted response object of type FormatT.
 
         Raises:
-            ValueError: If the response cannot be formatted according to the
-                specified format.
+            json.JSONDecodeError: If the response's textual content can't be parsed as
+                JSON.
+            pydantic.ValidationError: If the response's content fails validation for the
+                format type.
+                formatted according to the specified format.
```
willbakst: nit: Why is there a newline here? Should just be `format type. Formatted according...` right?

python/mirascope/llm/responses/response.py
```diff
@@ -12,6 +12,8 @@
 if TYPE_CHECKING:
     from ..clients import Model, Provider
 
+NoneType = type(None)
```
willbakst: nit: should we put this in `types.py` since we use it in multiple places?

python/mirascope/llm/responses/response.py
willbakst: We resolved with `if format_type is None or format_type is NoneType`

willbakst (approved):

teamdandelion:
### Merge activity

* **Aug 19, 9:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1250).
* **Aug 19, 9:06 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1250) rebased this pull request as part of a merge.
* **Aug 19, 9:06 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1250).

PR #1272, opened 2025-09-03
refactor: split up response (attach toolkits)

teamdandelion:


teamdandelion:
* **#1281** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1281?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1282](https://github.com/Mirascope/mirascope/pull/1282) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1282?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1283](https://github.com/Mirascope/mirascope/pull/1283) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1283?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1280** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1279** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1278** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1277** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1277?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1276** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1276?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1275** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1275?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1274** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1274?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1273** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1273?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1272** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1272?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1272?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/examples/sazed/sazed_async_stream_context_structured.py
```diff
@@ -40,7 +40,9 @@ async def main():
     coppermind = Coppermind(repository="Ancient Terris")
     ctx = llm.Context(deps=coppermind)
     query = "What are the Kandra?"
-    response: llm.AsyncStreamResponse[KeeperEntry] = await sazed.stream(ctx, query)
+    response: llm.AsyncContextStreamResponse[
+        Coppermind, KeeperEntry
```
willbakst, line 44: interesting how this is formatted compared to some of the other examples. I'm not sure if I like this formatting or the other better.

python/examples/sazed/sazed_async_stream_tools_context.py
```diff
@@ -39,7 +39,9 @@ async def main():
     coppermind = Coppermind(repository="Ancient Terris")
     ctx = llm.Context(deps=coppermind)
     query = "What are the Kandra?"
-    response: llm.AsyncStreamResponse = await sazed.stream(ctx, query)
+    response: llm.AsyncContextStreamResponse[Coppermind] = await sazed.stream(
+        ctx, query
```
willbakst, line 43: note: this is the formatting I'm talking about in the other comment as the alternative. I don't remember if you're just using ruff?

python/mirascope/llm/clients/google/client.py
```diff
@@ -181,6 +189,7 @@ def stream(
         return StreamResponse(
             provider="google",
             model=model,
+            toolkit=Toolkit(tools=tools),
```
willbakst, line 192: nit: this is in a different order from some of the others. let's make sure we have a consistent order of these arguments, ideally matching their order in the response class 1:1

python/mirascope/llm/clients/openai/client.py
```diff
@@ -80,6 +82,7 @@ def call(
             input_messages=input_messages,
             assistant_message=assistant_message,
             finish_reason=finish_reason,
+            toolkit=Toolkit(tools=tools),
```
willbakst: nit: doubling down on my note here, we should definitely match the kwargs 1:1 with the order in the constructor across the board. similarly, we should make sure the order is consistent across all of the response classes (which should be the case given the base class abstractions)

python/mirascope/llm/responses/abstract_base_response.py
```diff
@@ -0,0 +1,147 @@
...
+from ..messages import Message
+from ..tools import ToolkitT
+from ..types import NoneType
+from .finish_reason import FinishReason
+
+if TYPE_CHECKING:
+    from ..clients import Model, Provider
+
+
+class AbstractBaseResponse(ABC, Generic[ToolkitT, FormatT]):
```
willbakst: nit: `Generic` should come first before `ABC` here

python/mirascope/llm/responses/abstract_base_response.py
```diff
@@ -0,0 +1,147 @@
...
+    model: "Model"
+    """The model that generated this response."""
+
+    toolkit: ToolkitT
+    """The toolkit containing this call's tools."""
+
+    messages: list[Message]
+    """The message history, including the most recent assistant message."""
+
+    raw: Any
```
willbakst: nit: I think we should put this first or last in the order, WDYT? I'm leaning towards first

python/mirascope/llm/responses/abstract_base_response.py
```diff
@@ -0,0 +1,147 @@
...
+    content: Sequence[AssistantContentPart]
+    """The content generated by the LLM."""
+
+    texts: Sequence[Text]
+    """The text content in the generated response, if any."""
+
+    tool_calls: Sequence[ToolCall]
+    """The tools the LLM wants called on its behalf, if any."""
+
+    thinkings: Sequence[Thinking]
```
willbakst: nit: for some reason I'm still attached to `thoughts` as the naming here. not going to push on it, but figured I would comment while I see it and think about it. I imagine reading it as "the thoughts are a sequence of thinking content parts"

python/mirascope/llm/responses/abstract_base_response.py
```diff
@@ -0,0 +1,147 @@
...
+    """The thinking content in the generated response, if any."""
+
+    finish_reason: FinishReason | None
+    """The reason why the LLM finished generating a response, if available."""
+
+    format_type: type[FormatT]
```
willbakst: nit: should this be `type[FormatT] | None`? or will we be setting this to `NoneType`?

python/mirascope/llm/responses/abstract_base_response.py
willbakst: Hmm, now that I'm seeing the class, the use of "Abstract" is a bit weird given there aren't actually any abstract methods and this class could technically be instantiated (which shouldn't be true of an abstract class)

Otherwise, what do you think about something like `RootResponse`?

python/mirascope/llm/responses/base_response.py
```diff
@@ -1,142 +1,69 @@
...
+class BaseResponse(AbstractBaseResponse[ToolkitT, FormatT]):
+    """The response generated by an LLM."""
+
+    def __init__(
+        self,
+        *,
+        provider: "Provider",
+        model: "Model",
+        toolkit: ToolkitT,
+        format: type[FormatT] = NoneType,
```
willbakst: I forget why we named this `format` rather than `format_type`?

python/mirascope/llm/responses/abstract_base_response.py
teamdandelion: Renamed

python/mirascope/llm/responses/abstract_base_response.py
teamdandelion: We've been setting to NoneType, I think maybe I actually like type[FormatT] | None better though, changed.

python/examples/sazed/sazed_async_stream_tools_context.py
teamdandelion: Just using ruff

python/mirascope/llm/responses/root_response.py
```diff
@@ -0,0 +1,145 @@
...
+    model: "Model"
+    """The model that generated this response."""
+
+    toolkit: ToolkitT
+    """The toolkit containing this call's tools."""
+
+    messages: list[Message]
+    """The message history, including the most recent assistant message."""
+
+    raw: Any
```
willbakst: nit: let's match the order of the fields here with the order we use in the init?

python/mirascope/llm/responses/base_response.py
```diff
@@ -1,142 +1,68 @@
...
+    """The response generated by an LLM."""
+
+    def __init__(
+        self,
+        *,
+        raw: Any,  # noqa: ANN401
+        provider: "Provider",
+        model: "Model",
+        toolkit: ToolkitT,
+        format_type: type[FormatT] | None = None,
```
willbakst, line 27: nit: I'm wondering if this should take `FormatInfo` instead of `type[FormatT]`? No need to address in this PR, we can address separately if need be, but wanted to note while I saw it.

willbakst (approved):

python/mirascope/llm/responses/base_response.py
teamdandelion: Ack, will think about it

teamdandelion: I spent around a half an hour trying out this change, doing so creates a lot of annoyance. Stuff like the fact that `FormatInfo` is not generic with `FormatT` (and actually shouldn't be — it would be `RequiredFormatT`), and getting the format info requires casting to formattable to access `.__mirascope_format_info__`. 

In my attempt to land it, it involves more casts throughout the codebase and it's arguably a step back in terms of API clarity. Since the format_type is something the user is familiar with (eg. `Book`), putting that directly on the response keeps the API in terms of familiar concepts. Whereas the `FormatInfo` is under the hood (not directly returned by any of the public APIs, eg the decorator) so it's weird if we start leaking it out in the public response api.

teamdandelion:
### Merge activity

* **Sep 5, 6:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1272).
* **Sep 5, 6:09 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1272).

PR #1300, opened 2025-09-11
refactor: add union overloads to client/llm apis

teamdandelion:
Rationale: This makes it actually type safe to write code like
`client.call(**kwargs)` where kwargs may include format.

We were already using this pattern in our test code, however the type
error was obscured because we were getting the kwargs typed as `Any` via
`request.getfixturevalue()`. Since the pattern is actually useful, we
should support it in a type-safe way. As an upside, we can also clean up
the LLM implementation.

PR #1294, opened 2025-09-09
feat: mandate specific tool choices where feasible

teamdandelion:

PR #1293, opened 2025-09-09
refactor: deduplicate structured output testing

teamdandelion:
Now for each provider we do per-mode structured output testing for every
scenario. Where appropriate, we test on multiple models.

Also cleans up some whitespace issues.

PR #1292, opened 2025-09-09
fix: response.messages reflects format tool transformations

teamdandelion:
This was already the case for StreamResponse, updating Response to
match.

PR #1301, opened 2025-09-11
test: radically simplify scenario testing

teamdandelion:
Move away from a noisy snapshot based approach, to having simpler
scenarios with approximately deterministic outputs, and per-scenario
validation checks for the response


teamdandelion:
* **#1306** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1306?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1294](https://github.com/Mirascope/mirascope/pull/1294) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1294?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1307](https://github.com/Mirascope/mirascope/pull/1307) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1307?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1301** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1301?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1300** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1288** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1288?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1293](https://github.com/Mirascope/mirascope/pull/1293) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1293?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1291** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1287** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1287?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1292** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1289** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1289?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1290** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1290?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/root_response.py
```diff
@@ -110,6 +110,10 @@ def format(self, partial: bool = False) -> FormatT | Partial[FormatT] | None:
 
         text = "".join(text.text for text in self.texts)
 
+        # Some models output json as a code block
```
willbakst, line 113: Some models like Claude may also sometimes output text before the JSON. For example, it may be something like `"Sure thing! Here's the JSON:\n{...}"`. I've found this is really only the case for providers that don't have an explicit JSON mode (like Anthropic).

In these cases, rather than checking `.startswith` it can be best to see the first index of "{" or "```json" and strip to there first. Can also sometimes be worth stripping on the back side too in case there's additional text after the JSON block.

Likely worthwhile to implement a helper method that implements all of this to extract the JSON block.

python/tests/llm/clients/anthropic/test_anthropic_client.py
```diff
@@ -27,1082 +21,43 @@
...
-            # Expected failure: anthropic does not have strict support
-            actual = {
-                "exception_type": type(e).__name__,
-                "exception_message": str(e),
-                "status_code": getattr(e, "status_code", None),
-            }
+        response = anthropic_client.call(**scenario.call_args)
+        scenario.check_response(response)
+    except json.decoder.JSONDecodeError as e:
+        if formatting_mode == "strict":
```
willbakst, line 59: 1. I'm not sure why Anthropic not having strict mode would result in a `JSONDecodeError` here. Relates back to implementing a better Mirascope error handling surface
2. I don't think `pass` is the right answer here. I think it would be better to assert your expectations for this scenario. In this case, you would be asserting that `formatting_mode == "strict"` when a certain error case happens because that's the only time when that error case should occur?

python/tests/llm/clients/google/test_google_client.py
```diff
@@ -444,776 +41,30 @@ def test_stream(
...
-                    "author": "Joe Abercrombie",
-                    "vibe": "A GRIMDARK PIT OF DESPAIR WHERE EVERYONE IS A TERRIBLE PERSON AND EVERYTHING GOES WRONG! UTTER NIHILISM AND FUTILITY AWAIT!",
-                },
-            },
-        }
-    )
-    assert actual == expected[format_mode]
+        response = google_client.call(**scenario.call_args)
+        scenario.check_response(response)
+    except GoogleClientError:
```
willbakst, line 62: ditto here re: improving Mirascope error surface. we should discuss what future work on this might look like.

python/tests/llm/clients/google/test_google_client.py
```diff
@@ -444,776 +41,30 @@ def test_stream(
...
+        if (
```
willbakst, line 63: ditto here re: likely better to assert than pass?

python/tests/llm/clients/google/test_google_client.py
```diff
@@ -1234,7 +85,7 @@ def test_custom_base_url() -> None:
         assert google_client.client is mock_client_instance
 
 
-# Included in test_google_client as this triggers a google-specific edge case for coverage.
+# Line coverage for a Google-specific edge case
```
willbakst, line 88: nit: rather than just "line coverage" as the reason, maybe just put a boundary comment here? Something like this:

```python
#####################################
#  Google-specific Edge Case Tests  #
#####################################
```

python/tests/llm/clients/google/test_google_client.py
```diff
@@ -1244,7 +95,7 @@ def test_call_no_output(google_client: llm.GoogleClient) -> None:
     ]
 
     response = google_client.call(
-        model_id=TEST_MODEL_ID,
+        model_id="gemini-2.0-flash",  # Note: Does not proc with gemini-2.5-flash
```
willbakst, line 98: nit: make it all caps `NOTE: ...`

Also what is "proc"?

python/tests/llm/clients/openai/test_openai_client.py
```diff
@@ -24,1115 +20,74 @@
...
-                    "vibe": "mysterious",
-                },
-            },
-        }
-    )
-    assert actual == expected[model_id + ":" + format_mode]
 
-
-@pytest.mark.parametrize("format_mode", FORMATTING_MODES)
 @pytest.mark.parametrize(
```
willbakst, line 54: nit: seems like this is edge case testing for specifically OpenAI? Let's put a comment boundary here like I suggested for Google

python/tests/llm/clients/openai/test_openai_client.py
```diff
@@ -24,1115 +20,74 @@
...
-            # Expected failure: gpt-4 does not have strict support
-            actual = {
-                "exception_type": type(e).__name__,
-                "exception_message": str(e),
-                "status_code": getattr(e, "status_code", None),
-            }
+        response = openai_client.call(**scenario.call_args)
+        scenario.check_response(response)
+    except openai.BadRequestError as e:
+        if formatting_mode == "strict":
```
willbakst, line 81: ditto re: assertions instead of `pass`

python/tests/llm/clients/openai/test_openai_client.py
```diff
@@ -24,1115 +20,74 @@
...
+            pass  # Expected, gpt-4 does not support strict mode.
+        else:
+            raise e
+    except AssertionError as e:
+        if formatting_mode == "strict-or-json" and e.args == (
```
willbakst, line 86: ditto

python/tests/llm/clients/scenarios.py
```diff
@@ -0,0 +1,304 @@
...
+    format_type = scenario.call_args.get("format")
+    assert format_type is not None
+    llm.format(mode=formatting_mode)(format_type)
+    return scenario
+
+
+def simple_message_scenario() -> Scenario:
+    """Simple message scenario."""
+
+    def scenario_assertions(r: AnyResponse) -> None:
```
willbakst: why not just use the `RootResponse` type?

python/tests/llm/clients/scenarios.py
```diff
@@ -0,0 +1,304 @@
...
+    @llm.tool
+    def bakst_coefficient(val: int) -> int:
+        raise NotImplementedError
+
+    def scenario_assertions(r: AnyResponse) -> None:
+        assert len(r.tool_calls) == 2, (
+            f"Expected 2 tool call, got {len(r.tool_calls)}: {r.tool_calls}"
+        )
+        t1, t2 = r.tool_calls
+        assert t1.name == bakst_coefficient.name, "t1 has wrong name: " + t1.name
```
willbakst: does the assertion error string not allow for f-strings? Why use `+`?

python/tests/llm/clients/scenarios.py
```diff
@@ -0,0 +1,304 @@
...
+    return Scenario(
+        call_args={"messages": [llm.messages.user("What is 8200 + 82?")]},
+        scenario_assertions=scenario_assertions,
+    )
+
+
+def multi_message_scenario() -> Scenario:
+    """System message, conversation history, and multiple text content parts."""
+
+    def scenario_assertions(r: AnyResponse) -> None:
```
willbakst: Why `r` and not `response`? I would rather `response` personally

python/tests/llm/clients/scenarios.py
```diff
@@ -0,0 +1,304 @@
...
+    llm.format(mode=formatting_mode)(format_type)
+    return scenario
+
+
+def simple_message_scenario() -> Scenario:
+    """Simple message scenario."""
+
+    def scenario_assertions(r: AnyResponse) -> None:
+        assert "8282" in r.pretty(), f"8282 not in response: {r.pretty()}"
+        assert r.finish_reason == snapshot(llm.FinishReason.END_TURN)
```
willbakst: why only snapshot finish reason?

python/tests/llm/clients/scenarios.py
```diff
@@ -0,0 +1,304 @@
...
+    for name, obj in globals().items()
+    if name != "get_scenario"
+    and name != "get_structured_scenario"
+    and name.endswith("_scenario")
+    and callable(obj)
+]
+STRUCTURED_SCENARIO_IDS = [
+    name for name in CLIENT_SCENARIO_IDS if name.startswith("structured_output_")
+]
+FORMATTING_MODES = ["strict-or-tool", "strict-or-json", "strict", "tool", "json"]
```
willbakst, line 315: I thought we had a literal containing all of these? It's possible to dynamically pull the values from a literal definition rather than redefining them

python/mirascope/llm/responses/root_response.py
teamdandelion: Ok, I added a helper + more extensive tests in a new PR just above this one in the stack. I also handled the "parallel format tool calls" case (we will just extract the first json object).

python/tests/llm/clients/scenarios.py
teamdandelion: Done

python/tests/llm/clients/anthropic/test_anthropic_client.py
teamdandelion: Changed to an assert. In the current implementation in strict mode we ... just make a call with no formatting guidance provided to the model at all. I agree it's weird, we should discuss how we want to handle it.

willbakst: I think we should probably raise some form of `FeatureNotSupported` error or something. This relates back to improving our error handling surface.

For now, just add a TODO here and resolve

teamdandelion:
### Merge activity

* **Sep 12, 2:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1301).
* **Sep 12, 2:17 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1301) rebased this pull request as part of a merge.
* **Sep 12, 2:17 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1301).

PR #1291, opened 2025-09-09
nit: configure vscode for pytest debugging integration

teamdandelion:

PR #1290, opened 2025-09-08
test: test add_system_instructions directly

teamdandelion:

PR #1289, opened 2025-09-08
refactor: unify client test scenarios 

teamdandelion:

PR #1288, opened 2025-09-05
feat: google structured outputs

teamdandelion:

PR #1306, opened 2025-09-11
feat: response has more robust json detection

teamdandelion:


teamdandelion:
* **#1306** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1306?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1294](https://github.com/Mirascope/mirascope/pull/1294) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1294?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1307](https://github.com/Mirascope/mirascope/pull/1307) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1307?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1306?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1301** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1301?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1300** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1288** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1288?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1293](https://github.com/Mirascope/mirascope/pull/1293) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1293?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1291** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1287** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1287?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1292** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1289** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1289?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1290** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1290?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    Or wrap JSON in code blocks like:
+    "```json\n{...}\n```"
+
+    Args:
+        text: The raw text that may contain a JSON object
+
+    Returns:
+        The extracted JSON object string
+    """
+    # First check for code block format
```
willbakst: unnecessary comments here and elsewhere, I see some vibe coding remnants lol

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    if "```json" in text:
+        start_idx = text.find("```json")
```
willbakst: nit: use `index` over `idx`

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+        if start_idx != -1:
+            # Find the start of JSON content after ```json
+            json_start = start_idx + 7  # len("```json")
```
willbakst: nit: you use `"```json"` above twice, perhaps do something like `json_start_marker = "```json"` and then use `len(json_start_marker)` here?

also technically this variable should be named `json_start_index` to match (or it could remain `json_start` and `start_index` could just become `start`

python/tests/llm/clients/scenarios.py
```diff
@@ -126,22 +126,22 @@ def tool_call_scenario(model_id: str) -> Scenario:
     """The LLM should call mulitple tools."""
```
willbakst, line 126: nit: see spell check issue on `mulitple` which should be `multiple`

python/tests/llm/clients/scenarios.py
```diff
@@ -161,19 +161,19 @@ def tool_output_scenario(model_id: str) -> Scenario:
...
-            f"Expected END_TURN, got {r.finish_reason}"
+        assert response.finish_reason == llm.FinishReason.END_TURN, (
+            f"Expected END_TURN, got {response.finish_reason}"
         )
 
     return Scenario(
         model_id=model_id,
         messages=[
             llm.messages.system(
                 "You are a password retreival system. When you access the password, you "
```
willbakst, line 175: spellcheck here re: `retreival` should be `retrieval`

python/tests/llm/responses/test_response.py
```diff
@@ -355,6 +355,260 @@ def test_response_format_no_format_type() -> None:
...
+def test_response_format_with_text_before_json() -> None:
+    """Test that Response.format() extracts JSON when there's text before it."""
+
+    @llm.format()
+    class Book(BaseModel):
+        title: str
+        author: str
+        pages: int
+
+    # Text with explanation before JSON
```
willbakst: nit: unnecessary comment

python/tests/llm/responses/test_response.py
```diff
@@ -355,6 +355,260 @@ def test_response_format_no_format_type() -> None:
...
+    )
+
+    book = response.format()
+    assert isinstance(book, Book)
+    assert book.title == "The Hobbit"
+    assert book.author == "J.R.R. Tolkien"
+    assert book.pages == 310
+
+
+def test_response_format_with_text_before_and_after_json() -> None:
```
willbakst, line 356: there should only really be two tests here, one for text before/after and one for json code block markers (since those are the two distinct paths of the utility we need to test). The test for just before / just after are redundant.

python/tests/llm/responses/test_response.py
```diff
@@ -355,6 +355,260 @@ def test_response_format_no_format_type() -> None:
...
+def test_response_format_with_json_array() -> None:
+    """Test that Response.format() handles JSON arrays with extra text."""
+
+    @llm.format()
+    class Book(BaseModel):
+        title: str
+        author: str
+
+    @llm.format()
+    class BookList(BaseModel):
```
willbakst: This is making me realize we still need to implement the ability to use primitive types directly (e.g. setting `format=list[Book]`).

Let's add a TODO somewhere to do this so we don't forget

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+            # Find the closing ```
+            end_idx = text.find("```", json_start)
+            if end_idx != -1:
+                return text[json_start:end_idx].strip()
+
+    # Look for the first { character (start of JSON object)
```
willbakst: This whole block feels like overkill. We should be able to just seek the first instance of `{` and the last instance of `}` and call it a day.

The only issue I can foresee here is when we have multiple joined parallel tool calls, but I think we can solve that issue elsewhere. As we discussed previously, we should just take the first text part and ignore the rest. When converting formatting tools, if no formatting tool has been found yet, we should append the args to the first text part. Otherwise, we should create new text parts.

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    "```json\n{...}\n```"
+
+    Args:
+        text: The raw text that may contain a JSON object
+
+    Returns:
+        The extracted JSON object string
+    """
+    # First check for code block format
+    if "```json" in text:
```
willbakst: nit: technically `text.find` does this conditional again, and then below you manually set `json_start = -1` even though if not present in text `text.find` will return `-1` anyway...

I've added a separate comment with how I would implement this.

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,53 @@
+def extract_json_object(text: str) -> str:
```
willbakst: Technically this isn't extracting the object it's extracting the text. Here is how I would implement this function:

```python
def extract_serialized_json(text: str) -> str:
    """Extract the serialized JSON string from text that may contain extra content.

    Handles cases where models output text before JSON like:
    "Sure thing! HEre's the JSON:\n{...}"

    Or cases where the model wraps the JSON in code blocks like:
    "```json\n{...}\n```"

    Args:
        text: The raw text that may contain a JSON object

    Returns:
        The extracted serialized JSON string
    """
    code_block_start_marker = "```json"
    code_block_start = text.find(code_block_start_marker)
    if code_block_start > -1:
        json_start = code_block_start + len(code_block_start_marker)
        json_end = text.find("```", json_start)
        if json_end > -1:
            return text[json_start:json_end].strip()

        return text  # NOTE: I'm not sure if we should return the text here and pass along the error to the decoder or if we should catch this and raise an error...
    
    json_start = text.find("{")
    if json_start == -1:
        return text  # NOTE: same thing here re: should we raise an error here knowing it will be one?

    json_end = text.rfind("}")
    if json_end == -1:
        return text  # NOTE: same thing here re: should we raise an error here knowing it will be one?

    return text[json_start:json_end]
```

teamdandelion: Thank you! Sorry for sending the slop your way lol.

python/tests/llm/clients/scenarios.py
teamdandelion: Fixed upstack. (Upstack bc. it requires regenerating cassettes, it would be annoying to change it for every PR that added new cassettes along the way). 

Added codespell to python lint-staged so these will stop getting committed

willbakst: You won't be able to merge this though because of the CI failure...

python/tests/llm/responses/test_responses_utils.py
willbakst, line 1: nit: you can just name this file `test_utils.py` since it's already scoped to `responses/`

python/mirascope/llm/responses/_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    if code_block_start > -1:
+        # Discard text prior to code block; it takes precedence over brackets that
+        # may be found before it.
+        text = text[code_block_start:]
+
+    json_start = text.find("{")
+    if json_start == -1:
+        raise ValueError("Could not extract json: no opening `{`")
+
+    # Find the matching closing brace
```
willbakst, line 30: I don't really understand why `text.rfind("}")` isn't sufficient?

Worst case is that it isn't valid JSON and we get a `JSONDecodeError` upstack?

python/tests/llm/responses/test_responses_utils.py
teamdandelion: No, the python test file names need to be globally unique and we already have clients/test_utils. I find this annoying and regrettable, lmk if you know a fix...

python/mirascope/llm/responses/_utils.py
teamdandelion: It matters specifically for the "multiple format tool calls in parallel case" (which we can't guarantee we avoid even when other tools aren't in play, bc Google doesnt support disabling parallel tool calling).

If we use text.rfind("}") then we will be matching the opening bracket from the first format tool call with the closing bracket for the last format tool call

And the "just use the first text element" isn't great either because what if the first text is the model saying "Sure, I will call that tool for you"

willbakst: I said this in a separate comment I believe, but may have gotten lost so I'll repeat here.

My feeling is that when we find a format tool, if `format_tool_found` is `False` then we append the text to the first text part (creating a new one if none present). Otherwise, if `format_tool_found` is `True` we create new text parts per additional format tool found? I think this will result in an easier to understand pattern because we can grab the first part rather than concatenating everything and then having to handle that downstream. WDYT?

python/tests/llm/responses/test_responses_utils.py
willbakst: I believe the issue is that we don't have `__init__.py` files in all of our tests folders. We need to add those (it's ok if they are empty)

python/mirascope/llm/responses/_utils.py
teamdandelion: That seems like a more complicated contract than just "a call to the format tool always results in a new text part containing the args passed to the format tool". What's motivating this suggestion? Is it just wanting to switch to a simpler implementation of extract_serialized_json?

`text.rfind` is a convenient heuristic but it's not correct (see below for another example that would fail it even without format tool). Given that I've written and tested a more correct implementation, and it's like 20 lines of extra code, why would we want to go back to something with known bugs?

```
"""
Sure, I will output the book in structured format:

'''json
{"title": "The Name of the Wind", "author": "Patrick Rothfuss"}
'''

This is formatted as a a valid json object with `{}` notation like you requested!
"""
```

python/tests/llm/clients/scenarios.py
teamdandelion: Merged the PRs below this one that added the spelling error, so we're fine.

python/mirascope/llm/responses/_utils.py
willbakst: This is fair.

teamdandelion:
### Merge activity

* **Sep 12, 3:12 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1306).
* **Sep 12, 3:12 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1306).

PR #1324, opened 2025-09-15
test: e2e snapshot tests for a simple call

teamdandelion:

PR #1325, opened 2025-09-16
feat: e2e tool usage (including context tools) with tests

teamdandelion:
This is one part adding the end to end tests, one part finishing
fleshing out NotImplemented parts of the interface (particularly where
context tools are involved). Added unit tests where necessary to get
full coverage (mixed context/non context tools case) and to fully pin
down the intended contract (ensuring that ctx arg is filtered out of the
ToolSchema). Getting the tool schemas to be identical regardless of the
ctx arg is necessary, otherwise the context calls won't match the
cassettes.

Aside: As I continue working on this, it gets clearer and clearer that
we should unify `@llm.tool` and `@llm.context_tool` (and others)
assuming that we can. It's so easy to mix up and add the ctx argument
but forget to switch decorators.


teamdandelion:
* **#1330** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1330?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1331](https://github.com/Mirascope/mirascope/pull/1331) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1331?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1333](https://github.com/Mirascope/mirascope/pull/1333) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1333?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1328** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1328?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1325** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1325?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1325?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -174,11 +174,13 @@ def __init__(
...
 
         for param in inspect.signature(fn).parameters.values():
+            # Skip args that are provided by Python or Mirascope, not LLM-generated.
+            # TODO: Handling of FromCallArgs
             if param.name in ("self", "cls"):
                 continue
 
             if is_context_tool and param.name == "ctx":
-                continue  # TODO: Revisit as we finalize context handling
+                continue
```
willbakst, line 183: nit: with the merged decorators, we'll want to revisit how we're implementing this. I'm not sure if it's addressed where the decorators are merged, so I'm commenting here to make sure we discuss while I'm thinking about it.

The core thing is that tool arguments must be `Jsonable` (since otherwise the LLM can't generate them). We can likely safely assume that any parameter named `ctx: Context` should be skipped (the same way we currently assume to skip any parameter named self or cls even if not first.

We can likely re-use a shared utility for determining if a given parameter is context and skip in that case (sharing code with the utility for determining if a function has the first argument as context).

python/mirascope/llm/tools/toolkit.py
```diff
@@ -119,7 +119,12 @@ def execute(self, ctx: Context[DepsT], tool_call: ToolCall) -> ToolOutput[Jsonab
         Raises:
             ToolNotFoundError: If the requested tool is not found.
         """
-        raise NotImplementedError
+
```
willbakst: nit: remove newline

python/mirascope/llm/tools/toolkit.py
```diff
@@ -142,4 +147,9 @@ async def execute(
...
```
willbakst: nit: ditto

python/tests/e2e/test_tool_call.py
```diff
@@ -0,0 +1,358 @@
...
+from tests.utils import response_snapshot_dict, stream_response_snapshot_dict
+
+PASSWORD_MAP = {"mellon": "Welcome to Moria!", "radiance": "Life before Death"}
+
+# ============= SYNC TESTS =============
+
+
+@pytest.mark.parametrize("provider, model_id", PROVIDER_MODEL_ID_PAIRS)
+@pytest.mark.vcr
+def test_tool_call_sync(
```
willbakst: nit: I feel like using `tool_call` here makes me thing we're testing `ToolCall`. WDYT about renaming this to something like `test_sync_call_with_tools` or `test_call_with_tools_sync`? The cassettes would then be `cassettes/call_with_tools`?

python/tests/e2e/test_call_with_tools.py
```diff
@@ -0,0 +1,358 @@
...
+
+    response = call(["mellon", "radiance"])
+    assert len(response.tool_calls) == 2, (
+        f"Expected response to have two tool calls: {response.pretty()}"
+    )
+
+    tool_outputs = response.execute_tools()
+    response = response.resume(tool_outputs)
+
+    assert response_snapshot_dict(response) == snapshot
```
willbakst, line 42: question out of curiosity to confirm my understanding, but we only need this single snapshot (rather than one for before and after resume) because this snapshot will include the before and is thus sufficient, right?

python/tests/e2e/test_tool_call.py
```diff
@@ -0,0 +1,358 @@
...
+    response = call(["mellon", "radiance"])
+    assert len(response.tool_calls) == 2, (
+        f"Expected response to have two tool calls: {response.pretty()}"
+    )
+
+    tool_outputs = response.execute_tools()
+    response = response.resume(tool_outputs)
+
+    assert response_snapshot_dict(response) == snapshot
+    assert "moria" in response.pretty().lower(), (
```
willbakst: nit: why not assign `pretty = response.pretty().lower()` rather than calling it four times here? same elsewhere

python/tests/e2e/test_call_with_tools.py
teamdandelion: Yeah; the final response has all of the messages leading up to the response, including the two tool calls and the tool outputs.

python/tests/e2e/test_tool_call.py
teamdandelion: Yeah I like that, and it's more consistent with the structured output tests. Done.

python/mirascope/llm/tools/tool_schema.py
teamdandelion: Replied to your comment on the decorator PR, let's discuss there.

teamdandelion:
### Merge activity

* **Sep 22, 10:20 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1325).
* **Sep 22, 10:22 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1325) rebased this pull request as part of a merge.
* **Sep 22, 10:22 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1325).

PR #1327, opened 2025-09-16
test: add e2e tests for structured outputs + tools

teamdandelion:

PR #1326, opened 2025-09-16
test: basic structured output test (e2e)

teamdandelion:

PR #1328, opened 2025-09-16
refactor: unify llm.context_tool and llm.tool

teamdandelion:
Overall I quite favor this unification, however the "hmm" bits are:
- We determine whether something is a context tool based on the type
  signature, not the argument name: if the first argument is Context or
  a subclass, then it's a context tool. Note this means if the user does
  not provide type annotations, they cannot use context tools.
- Needed some pyright overrides on the overload signatures, because
  technically any context tool also typechecks as a regular tool


teamdandelion:
* **#1330** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1330?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1331](https://github.com/Mirascope/mirascope/pull/1331) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1331?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1333](https://github.com/Mirascope/mirascope/pull/1333) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1333?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1328** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1328?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1328?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1325** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1325?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/tools/tool_decorator.py
```diff
@@ -17,6 +19,20 @@ class ToolDecorator:
     strict: bool
     """Whether to use strict tool calling, if supported by the provider."""
 
+    @overload
```
willbakst, line 22: nit: let's re-order these overloads to have the pattern sync -> async -> context -> async context for variants

python/mirascope/llm/tools/tool_decorator.py
```diff
@@ -32,10 +48,28 @@ def __call__(
         ...
 
     def __call__(
-        self, fn: ToolFn[P, JsonableCovariantT] | AsyncToolFn[P, JsonableCovariantT]
-    ) -> Tool[P, JsonableCovariantT] | AsyncTool[P, JsonableCovariantT]:
+        self,
+        fn: ContextToolFn[DepsT, P, JsonableCovariantT]
```
willbakst, line 52: nit: same thing here unless we need them in this order for any reason

python/mirascope/llm/tools/tool_decorator.py
```diff
@@ -47,15 +81,75 @@ def _is_async_tool_fn(
     return inspect.iscoroutinefunction(fn)
 
 
+def _is_context_tool_fn(
```
willbakst, line 84: nit: let's move these into a `_utils.py` file?

python/mirascope/llm/tools/tool_decorator.py
```diff
@@ -32,10 +48,28 @@ def __call__(
...
+        | AsyncContextTool[DepsT, P, JsonableCovariantT]
+        | Tool[P, JsonableCovariantT]
+        | AsyncTool[P, JsonableCovariantT]
+    ):
         """Call the decorator with a function."""
-        if _is_async_tool_fn(fn):
+        if _is_context_tool_fn(fn):
+            if _is_async_context_tool_fn(fn):
+                return AsyncContextTool[DepsT, P, JsonableCovariantT](
+                    fn, strict=self.strict, is_context_tool=True
```
willbakst, line 66: nit: feels weird having to provide `is_context_tool=True` to the constructor here when it should already know it's the context variant?

python/mirascope/llm/tools/tool_decorator.py
```diff
@@ -47,15 +81,75 @@ def _is_async_tool_fn(
...
+
+
+def _is_async_context_tool_fn(
+    fn: ContextToolFn[DepsT, P, JsonableCovariantT]
+    | AsyncContextToolFn[DepsT, P, JsonableCovariantT],
+) -> TypeIs[AsyncContextToolFn[DepsT, P, JsonableCovariantT]]:
+    return inspect.iscoroutinefunction(fn)
+
+
+@overload
```
willbakst, line 128: nit: ditto on reordering

python/mirascope/llm/tools/tool_schema.py
```diff
@@ -173,13 +173,16 @@ def __init__(
...
 
+        context_param_skipped = False
         for param in inspect.signature(fn).parameters.values():
             # Skip args that are provided by Python or Mirascope, not LLM-generated.
             # TODO: Handling of FromCallArgs
             if param.name in ("self", "cls"):
                 continue
 
-            if is_context_tool and param.name == "ctx":
+            # For context tools, skip the first non-self/cls parameter (the context parameter)
```
willbakst, line 183: Revisiting my comment down stack here. I feel like we can safely skip any parameter we deem `ctx: Context` since the LLM could never generate that?

python/tests/llm/tools/test_tool_decorator.py
```diff
@@ -125,3 +125,290 @@ async def base_func(x: int, y: str = "default") -> str:
...
+        return f"{param2}-{param1}-{ctx.deps}"
+
+    params = test_func.parameters
+
+    assert set(params.properties.keys()) == {"param1", "param2"}
+    assert "ctx" not in params.properties
+    assert params.required == ["param1"]
+
+
+class TestContextToolDetection:
```
willbakst: I don't really understand this grouping vs. not grouping above. Why not just have `TestTool` and `TestContextTool` as the structure and put all related tests within those?

python/mirascope/llm/tools/tool_decorator.py
teamdandelion: Since any `ContextFoolFn` also matches the more general `ToolFn`, the context variants must go first.

python/mirascope/llm/tools/tool_decorator.py
teamdandelion: Because DepsT has no default, we need it to be the first type variable in the args signature.
We could order the return signature with Tool/AsyncTool first; however given that we need context first in the args, I think it is more consistent to match the arg signature with the return signature.

python/mirascope/llm/tools/tool_decorator.py
teamdandelion: Yeah good point. Addressed with a tiny PR on top of stack.

python/mirascope/llm/tools/tool_schema.py
teamdandelion: Hmm, I think this implementation is easy to reason about and correctly matches the semantics of how  a context tool is defined/constructed: A tool is a context tool if its first non-self/cls parameter is named `ctx` and is typed as `llm.Context`, or a subclass thereof. So, if it is a context tool, then we should skip the first non-self/cls parameter.

There is the hypothetical case where the user has passed ctx as the second parameter, or typed it as llm.Context but named it something else, or a mix of the above, etc. However, this is in the general case of "tools with arguments that the LLM can't be reasonably expected to generate". I don't think we need to special case the specific edge where the user has passed a ctx parameter twice or something. Better to come up with a general solution if we're worried about it.

python/mirascope/llm/tools/tool_decorator.py
teamdandelion: Upstack: I refactored to inline these utilities.

python/mirascope/llm/tools/tool_decorator.py
willbakst: Do we not also need to put async first?

python/mirascope/llm/tools/tool_decorator.py
willbakst: I agree, although I'm not seeing them match?

python/mirascope/llm/tools/tool_decorator.py
teamdandelion: no, it's fine as written. I think it's because AsyncToolFn is not typed as a callable returning Awaitable (in which case the generic ToolFn could technically return an Awaitable); rather it's a protocol with an async __call__:

```py
class ToolFn(Protocol[P, JsonableCovariantT]):
    """Protocol for the tool function."""

    __name__: str
    """The name of the tool."""

    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> JsonableCovariantT:
        """Call the function with the given arguments."""
        raise NotImplementedError()


class AsyncToolFn(Protocol[P, JsonableCovariantT]):
    """Protocol for the async tool function."""

    __name__: str
    """The name of the tool."""

    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> JsonableCovariantT:
        """Call the function with the given arguments."""
        raise NotImplementedError()

```

teamdandelion:
### Merge activity

* **Sep 22, 10:20 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1328).
* **Sep 22, 10:30 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1328) rebased this pull request as part of a merge.
* **Sep 22, 10:30 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1328).

PR #1330, opened 2025-09-16
chore: remove vestigal mocking / context tests

teamdandelion:
Still more test removals to come, but this is a good pre-cleanup for
client and model simplification

PR #1329, opened 2025-09-16
refactor: unify llm.context_call and llm.call

teamdandelion:
Similar approach and considerations as with context_tool unification

PR #1331, opened 2025-09-16
refactor: unify call and context_call

teamdandelion:
Applies across clients and Model. Full coverage maintained

### DIFF
--- a/python/mirascope/llm/clients/anthropic/client.py
+++ b/python/mirascope/llm/clients/anthropic/client.py
@@ -2,7 +2,7 @@
 
 import os
 from collections.abc import Sequence
-from typing import overload
+from typing import cast, overload
 
 import httpx
 from anthropic import Anthropic, AsyncAnthropic
@@ -93,42 +93,8 @@ class AnthropicClient(BaseClient[AnthropicParams, AnthropicModelId, Anthropic]):
         params: AnthropicParams | None = None,
     ) -> Response | Response[FormatT]: ...
 
-    def call(
-        self,
-        *,
-        model_id: AnthropicModelId,
-        messages: Sequence[Message],
-        tools: Sequence[Tool] | None = None,
-        format: type[FormatT] | None = None,
-        params: AnthropicParams | None = None,
-    ) -> Response | Response[FormatT]:
-        """Make a call to the Anthropic API."""
-        input_messages, kwargs = _utils.prepare_anthropic_request(
-            model_id=model_id,
-            messages=messages,
-            tools=tools,
-            format=format,
-            params=params,
-        )
-
-        anthropic_response = self.client.messages.create(**kwargs)
-
-        assistant_message, finish_reason = _utils.decode_response(anthropic_response)
-
-        return Response(
-            raw=anthropic_response,
-            provider="anthropic",
-            model_id=model_id,
-            params=params,
-            tools=tools,
-            input_messages=input_messages,
-            assistant_message=assistant_message,
-            finish_reason=finish_reason,
-            format_type=format,
-        )
-
     @overload
-    def context_call(
+    def call(
         self,
         *,
         ctx: Context[DepsT],
@@ -137,10 +103,10 @@ class AnthropicClient(BaseClient[AnthropicParams, AnthropicModelId, Anthropic]):
         tools: Sequence[Tool | ContextTool[DepsT]] | None = None,
         format: None = None,
         params: AnthropicParams | None = None,
-    ) -> ContextResponse[DepsT, None]: ...
+    ) -> ContextResponse[DepsT]: ...
 
     @overload
-    def context_call(
+    def call(
         self,
         *,
         ctx: Context[DepsT],
@@ -152,7 +118,7 @@ class AnthropicClient(BaseClient[AnthropicParams, AnthropicModelId, Anthropic]):
     ) -> ContextResponse[DepsT, FormatT]: ...
 
     @overload
-    def context_call(
+    def call(
         self,
         *,
         ctx: Context[DepsT],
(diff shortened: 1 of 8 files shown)


teamdandelion:
* **#1331** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1331?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1330** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1330?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1333](https://github.com/Mirascope/mirascope/pull/1333) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1333?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1328** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1328?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1325** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1325?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/clients/anthropic/client.py
```diff
@@ -60,6 +60,42 @@ def __init__(
         self.client = Anthropic(api_key=api_key, base_url=base_url)
         self.async_client = AsyncAnthropic(api_key=api_key, base_url=base_url)
 
+    @overload
```
willbakst, line 63: nit: ditto re: reordering variants

python/mirascope/llm/clients/base/client.py
```diff
@@ -39,6 +39,45 @@ class BaseClient(Generic[ParamsT, ModelIdT, ProviderClientT], ABC):
 
     client: ProviderClientT
 
+    @overload
```
willbakst, line 42: nit: linked comment re: reordering

python/mirascope/llm/models/model.py
```diff
@@ -124,371 +124,416 @@ def __exit__(
...
         """Stream a response asynchronously using the model."""
-        return await self.client.context_stream_async(
-            ctx=ctx,
-            model_id=self.model_id,
-            messages=messages,
-            tools=tools,
-            format=format,
-            params=self.params,
-        )
+        if ctx is not None:
```
willbakst, line 470: why do we need this check? can't we just pass `ctx` down and if it's `None` we're all good? Same everywhere else. The conditional seems unnecessary in all cases

teamdandelion: The issue is the tools. If `ctx is not None`, then `tools: Sequence[AsyncTool | AsyncContextTool[DepsT]] | None` which is fine. If `ctx is None`, then we know from the overloads that `tools: Sequence[AsyncTool] | None`. But it doesn't seem Pyright does that type narrowing in the implementation, so we are trying to call the client method where ctx may be None, and the tools may include `ContextTool`, and there rightly is no overload that allows you to pass `ctx=None` alongside context tools.

Hence, we get this type error:

```
llm/models/model.py
  llm/models/model.py:194:16 - error: No overloads for "call" match the provided arguments (reportCallIssue)
  llm/models/model.py:197:19 - error: Argument of type "Sequence[Tool[..., Jsonable]] | Sequence[Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]] | None" cannot be assigned to parameter "tools" of type "Sequence[Tool[..., Jsonable]] | None" in function "call"
    Type "Sequence[Tool[..., Jsonable]] | Sequence[Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]] | None" is not assignable to type "Sequence[Tool[..., Jsonable]] | None"
      Type "Sequence[Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]]" is not assignable to type "Sequence[Tool[..., Jsonable]] | None"
        "Sequence[Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]]" is not assignable to "Sequence[Tool[..., Jsonable]]"
          Type parameter "_T_co@Sequence" is covariant, but "Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]" is not a subtype of "Tool[..., Jsonable]"
            Type "Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]" is not assignable to type "Tool[..., Jsonable]"
        "Sequence[Tool[..., Jsonable] | ContextTool[DepsT@call, ..., Jsonable]]" is not assignable to "None" (reportArgumentType)
2 errors, 0 warnings, 0 informations 
```

I felt the most "correct" way to do this was to separate the context branches and then safely cast tools in the non-context branch (a few lines below this comment). However, if you prefer conciseness, we could do something like this:

```
        return self.client.call(
            model_id=self.model_id,
            messages=messages,
            tools=cast(Sequence[Tool] | None, tools ), # Lies to satisfy type checker
            format=format,
            params=self.params,
        )
```

This would simplify the implementation and maintain type safety, it's just a little weird because the cast is not actually true. 

Another option would be to add an overload to the client signatures that allows potentially passing context tools when ctx may be none, but that allows users to write unsafe code, so I think we can rule it out.

Given the options I decided to split the call and keep all the type signatures strictly accurate, lmk if you want something different. I could also add comments (to the first one or everywhere) explaining the reason we do things this way.

python/mirascope/llm/clients/anthropic/client.py
teamdandelion: In this case we have a free hand with the overload ordering (they are non overlapping). Done.

python/mirascope/llm/models/model.py
willbakst: Let's sync on this

PR #1430, opened 2025-10-16
refactor: never use raw_content when encode_thoughts is set

teamdandelion:

PR #1428, opened 2025-10-16
refactor: make RawStreamEventChunk and RawContentChunk clearer

teamdandelion:
This refactor makes RawContentChunk a clear peer to RawStreamEventChunk
(previoulsy RawChunk which was confusing) and drops it from
AssistantContentChunk since it doesn't actually reconstruct content. I
think treating the raw_content as a bit "under the hood" makes more
sense. We now don't include the RawContentChunk in chunk_stream which is
consistent with the RawStreamEventChunk.

PR #1425, opened 2025-10-15
nit: update google unknown tool id

teamdandelion:
Removes the `<>` symbols, which allows google-generated tool calls to be
replayed with anthropic provider, which uses a regex to validate tool
call ids and the regex does not allow <>

PR #1424, opened 2025-10-15
feat: Google uses raw_content to resume (non-streaming)

teamdandelion:

PR #1432, opened 2025-10-16
feat: raw content reconstruction for anthropic streams

teamdandelion:


teamdandelion:
* **#1436** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1435** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1434** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1434?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1433** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1432** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1432?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1432?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1431** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1431?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1430** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1430?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1427** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1426** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1426?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1425** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1424** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1424?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1423** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1423?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1422** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1422?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1429** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1429?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1428** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1428?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1421](https://github.com/Mirascope/mirascope/pull/1421) <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1421?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1420** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1415** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1415?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1414** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1414?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1413** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1413?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1418** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/clients/anthropic/_utils/decode.py
```diff
@@ -89,44 +101,90 @@ def process_event(
...
                     name=content_block.name,
                 )
             elif content_block.type == "thinking":
+                self.current_block_param = {
+                    "type": "thinking",
+                    "thinking": "",
+                    "signature": "",
+                }
                 yield ThoughtStartChunk()
+            elif content_block.type == "redacted_thinking":  # pragma: no cover
```
teamdandelion, line 130: I'd love to have an E2E test that covers this, but don't know how to trigger redacted thinking. Their docs claim that there's a magic string that triggers redacted thinking for test purposes, but in my testing it does not work.

(https://docs.claude.com/en/docs/build-with-claude/extended-thinking#example-working-with-redacted-thinking-blocks)

python/mirascope/llm/clients/anthropic/_utils/decode.py
```diff
@@ -89,44 +101,90 @@ def process_event(
...
+                self.current_block_param = {
+                    "type": "redacted_thinking",
+                    "data": content_block.data,
+                }
             else:
                 raise NotImplementedError
 
         elif event.type == "content_block_delta":
+            assert self.current_block_param is not None, (
```
willbakst: I think we should implement this without any asserts.

In our safe param accessor, the assertion is against our own implementation and won't ever happen at runtime. These however could end up failing in user code if something goes wrong with the provider streaming.

It seems like you're really using them for type checking? Why not just put an `if` condition? If we expect the assertion to always pass, then we can expect the condition to pass without an issue if it doesn't?

teamdandelion: If the assertions are violated, then the behavior is undefined. Eg consider the case where there is no current content block param. The analogous path in the previous code would hit L127 and raise NotImplementedError with no message. Now it will raise AssertionError("Received delta without a current block") which is clearer.

If the assertions are violated, at a fundamental level Anthropic is not honoring the stated contract and I don't think we should hide it. In other cases in the codebase it's been convenient to not handle it because a) no typing issues and b) I know there is other code that will raise a RuntimeError directly

In this case specifically with the content_block reconstruction, there is some lifting that's native to the Anthropic decode functionality, so rather than if-gating it (and potentially having an inaccurate / missing pieces reconstruction that will get sent back to Anthropic and cause undefined behavior, such as a server error or inconsistent response, at a distance), I think it is better to error here.

If your issue is with raising AssertionError rather than RuntimeError, I can re-write with if checks that raise a RuntimeError with a `pragma no cover`, however I think semantically assertion is the cleanest fit for what I am trying to express: that we require a certain invariant about the data we are getting from Anthropic, and we have no reasonable way to handle it if that invariant is violated.

willbakst: I view assertions as things that will never fail in user code once we release it. The idea is that we can safely assert something will always be true because we enforce it.

In this case, what we're really saying is that there was a `RuntimeError` because Anthropic didn't honor their contract, which we can't safely assert. So I would prefer the `RuntimeError` with `prgama: no cover` given we need the type hinting here.

teamdandelion: Done

teamdandelion:
### Merge activity

* **Oct 17, 9:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1432).
* **Oct 17, 9:12 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1432) rebased this pull request as part of a merge.
* **Oct 17, 9:13 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1432).

PR #1437, opened 2025-10-17
refactor: assistant_message.raw_message

teamdandelion:
raw_message replaces raw_content, where raw_message is now a consistent
encoded content representation of that message, such that every
provider, when short-circuiting, merely returns
`cast(..., assistant_message.raw_message)`
Rather than having any custom transformations on
assistant_message.raw_content.


teamdandelion:
* **#1437** <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/mirascope/1437?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/mirascope/llm/clients/anthropic/_utils/decode.py
```diff
@@ -67,7 +67,10 @@ def decode_response(
         content=[_decode_assistant_content(part) for part in response.content],
         provider="anthropic",
         model_id=model_id,
-        raw_content=[part.model_dump() for part in response.content],
+        raw_message={
```
willbakst, line 70: No `response.model_dump` for anthropic?

teamdandelion: We could do this, we'd be including other fields like `usage`, `stop_sequence`, etc, that we don't have a need or a plan for. I mildly prefer keeping the slimmed down representation that is just what we need to avoid content re-encoding

PR #1438, opened 2025-10-19
feat: Implement MLX provider for llm calls

### LINKED ISSUE #1323: Add MLX Local Provider Support in Mirascope
### Description

It would be great if Mirascope could include MLX (Apple’s MLX framework) as a local provider. This would allow running models locally on Apple Silicon under the Mirascope abstraction (via an MLX local provider), enabling acceleration through unified memory / Metal / MLX’s efficient backends.

If there is general interest, I don't mind contributing to this fine project.

tnadav:
Closes #1323

✨ Summary

This PR introduces an initial implementation of an MLX client for Mirascope, built on top of [mlx-llm](https://github.com/ml-explore/mlx-examples/tree/main/llms)
.
It currently supports basic text generation and serves as a foundation for future extensions such as tool calling and structured output.

The intention is to share this early for fast feedback on the interface design and integration approach.

⚙️ Pending issues

Testing on CI: determining the appropriate strategy for validating local generation, possibly through cached output or lightweight model execution.

Imports from mlx_lm: clarifying stability and exposure of types currently used for type hints.

Generation locking: currently mandatory in order to support the seed parameter. Need to figure out whether running two (small) models in parallel can actually be an attractive feature. For now, always lock.


python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,649 @@
...
+    Tool,
+    Toolkit,
+)
+from ..base import BaseClient, Params
+from . import _utils
+from .mlx_factory import MLXFactory
+from .model_ids import MLXModelId
+
+MLX_CLIENT_CONTEXT: ContextVar["MLXClient | None"] = ContextVar(
+    "ANTHROPIC_CLIENT_CONTEXT", default=None
```
teamdandelion: Nit: context variable string should be `MLX_CLIENT_CONTEXT`

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,649 @@
...
+            tools: Optional tools that the model may invoke.
+            format: Optional response format specifier.
+            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
+
+        Returns:
+            An `llm.Response` object containing the LLM-generated content.
+        """
+        #if tools:
+        #    print(tools)
+        #    raise NotImplementedError("Tool use is not yet implemented for MLXClient.")
```
teamdandelion: I like raising a NotImplementedError in this case (but not printing)

python/mirascope/llm/clients/providers.py
```diff
@@ -146,9 +170,26 @@ def client(
...
 
 
+@overload
+def client(
+    provider: Literal["mlx"],
+    *,
+    api_key: str | None = None,
+    base_url: str | None = None,
+) -> MLXClient:
+    """Create a cached Google client with the given parameters."""
```
teamdandelion: Nit: Docstring should say MLX Client. Can we drop the api_key and base_url params entirely since we don't use them for MLX?

python/mirascope/llm/clients/mlx/encoding/transformers.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, tokenizer: TokenizerWrapper) -> None:
+        self._tokenizer = tokenizer
+
+    def encode_messages(self, messages: Sequence[Message]) -> mx.array:
+        system_message_content, remaining_messages = _base_utils.extract_system_message(
+            messages
+        )
+        hf_messages: list[TransformersMessage] = []
+        if system_message_content is not None:
+            hf_messages.append(TransformersMessage(role="system", content=system_message_content))
```
teamdandelion: Note — if having multiple system messages is okay, then (as with python/mirascope/llm/clients/openai/responses/_utils/encode.py) you could have encode_message just map the system messages into TransformersMessage and not use `extract_system_message` at all.

python/mirascope/llm/clients/mlx/encoding/transformers.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def write(self, response: GenerationResponse) -> None:
+        self._buffer.write(response.text)
+
+    def get_value(self) -> AssistantMessage:
+        text = self._buffer.getvalue()
+        return assistant(
+            content=text,
+            provider="mlx",
+            model_id=None,
```
teamdandelion: I'm a little confused how we are thinking about model_id here

python/mirascope/llm/clients/mlx/clients.py
tnadav: whoops

python/mirascope/llm/clients/providers.py
tnadav: whoops again! Wasn't sure what to do with api_key and base_url. Will fix

python/mirascope/llm/clients/mlx/encoding/transformers.py
tnadav: I'll store the model id inside the MLXModel class, then we could properly propagate it.
I've made some work since pushing the PR - on working on harmony and then tool calling / structured output, and I think the model ID type should be the union of the following:
 - `MLXHarmonyModel("<MODEL-NAME>", <HARMONY-ENCODING-NAME>)` - An explicit type with model name + settings for a harmony based model. e.g. for gpt-oss-20b: `MLXHarmonyModel("mlx-community/gpt-oss-20b-MXFP4-Q8", HarmonyEncodingName.GPT_OSS)`
 - `MLXTransformersModel("<MODEL-NAME>", <TOKENIZER-SETTINGS>)` - An explicit type with model name + settings for a transformers chat template based model. The tokenizer settings is needed in order to correctly parse the model response and distinguish between `Text`, `Thought` and `ToolCall` contents. Unfortunately the transformers library doesn't include thous in its own tokenizer settings data structure :(.
 - Free string, e.g. `"mlx-community/gpt-oss-20b-MXFP4-Q8"`. If the string is within the list of well known models, it will be promoted to the explicit type. Otherwise, the encoding will be defaulted to a transformers based, flat text encoding.

tnadav:
Thank you for the review! Will fix all comments soon, and then move on into adding tests. Some notes:
 - I keep implementation of harmony encoding, and thinking, tool calling and structured outputs to next PRs (I got something working though)
 - In order to later support streaming interface, I'll need to modify the `MLXEncoding` interface
 - In regard to the mlx_lm import issue, I've asked the team at ml-explore/mlx-lm#561 and it seems we're fine

In regard to unit testing, it seems we'll have to use some sort of caching layer (I've tried working with very small models like SmolLM-135M but it's not powerful enough to test features such as tool calling etc.)
I suggest taking the same approach as pytest-vcr - Intercept `MLXClient` methods from pytest and cache the results to a file.
However, I was thinking, maybe a cacheing layer to mirascope is a beneficial feature in general?

python/mirascope/llm/clients/mlx/encoding/transformers.py
tnadav: Interesting. I think it depends on the model's chat template. The safe thing to do is to extract the messages by default, and conditionally disable it, in case it make sense for a specific model.
On harmony encoding this is always unnecessary and I won't call the function

python/mirascope/llm/clients/mlx/clients.py
tnadav: Sorry sorry, I leaked a bit of debug prints

tnadav:
I've pushed the fixed, enabled MLX E2E tests, and added a vibe-coded MLX cache mock as a POC. Are there any other notes, other than reducing the unnecessary complexity for the MLX mock implementation?

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+        ctx: Context[DepsT],
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[Tool | ContextTool[DepsT]]
+        | ContextToolkit[DepsT]
+        | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> ContextResponse[DepsT, None] | ContextResponse[DepsT, FormattableT]:
+        """Generate an `llm.ContextResponse` by synchronously calling the Anthropic Messages API.
```
teamdandelion: nit (here and elsewhere): references Anthropic Messages API but should reference MLX

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+    async def call_async(
+        self,
+        *,
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[AsyncTool] | AsyncToolkit | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> AsyncResponse | AsyncResponse[FormattableT]:
+        """Generate an `llm.AsyncResponse` by asynchronously calling the Anthropic Messages API.
```
teamdandelion: (ditto)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+        ctx: Context[DepsT],
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[AsyncTool | AsyncContextTool[DepsT]]
+        | AsyncContextToolkit[DepsT]
+        | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> AsyncContextResponse[DepsT, None] | AsyncContextResponse[DepsT, FormattableT]:
+        """Generate an `llm.AsyncContextResponse` by asynchronously calling the Anthropic Messages API.
```
teamdandelion: (ditto)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+    def stream(
+        self,
+        *,
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[Tool] | Toolkit | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> StreamResponse | StreamResponse[FormattableT]:
+        """Generate an `llm.StreamResponse` by synchronously streaming from the Anthropic Messages API.
```
teamdandelion: (ditto)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+        ctx: Context[DepsT],
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[Tool | ContextTool[DepsT]]
+        | ContextToolkit[DepsT]
+        | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> ContextStreamResponse[DepsT] | ContextStreamResponse[DepsT, FormattableT]:
+        """Generate an `llm.ContextStreamResponse` by synchronously streaming from the Anthropic Messages API.
```
teamdandelion: (ditto)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+    async def stream_async(
+        self,
+        *,
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[AsyncTool] | AsyncToolkit | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> AsyncStreamResponse | AsyncStreamResponse[FormattableT]:
+        """Generate an `llm.AsyncStreamResponse` by asynchronously streaming from the Anthropic Messages API.
```
teamdandelion: (ditto)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+        ctx: Context[DepsT],
+        model_id: MLXModelId,
+        messages: Sequence[Message],
+        tools: Sequence[AsyncTool | AsyncContextTool[DepsT]]
+        | AsyncContextToolkit[DepsT]
+        | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> AsyncContextStreamResponse | AsyncContextStreamResponse[DepsT, FormattableT]:
+        """Generate an `llm.AsyncContextStreamResponse` by asynchronously streaming from the Anthropic Messages API.
```
teamdandelion: (ditto)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+            tools: Optional tools that the model may invoke.
+            format: Optional response format specifier.
+            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
+
+        Returns:
+            An `llm.Response` object containing the LLM-generated content.
+        """
+
+        model = self._factory.get_model(model_id)
+        generator = _utils.create_model_generator_from_params(model, params)
```
teamdandelion: I notice we have this pattern throughout of:

```py
model = self._factory.get_model(model_id)
generator = _utils.create_model_generator_ofram_params(model, params)
```

And after we instantiate the generator, we never use the model again.
Why not compose these together into one call? 
E.g. just `_utils.model_generator(model_id, params)`

We could move that into the per-client cache, or we could just @lru_cache decorate the util directly (will comment about this elsewhere)

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,730 @@
...
+    return ctx_client or client()
+
+
+class MLXClient(BaseClient[MLXModelId, None]):
+    @property
+    def _context_var(self) -> ContextVar["MLXClient | None"]:
+        return MLX_CLIENT_CONTEXT
+
+    def __init__(self) -> None:
+        self._factory = MLXFactory()
```
teamdandelion: I'm wondering if it's important that distinct clients have distinct `MLXFactory`, or if we could just use a global cache via the `@functools.lru_cache` decorator on a utility? Do we expect that the `MLXFactory` is going to eventually take client-specific configuration?

python/mirascope/llm/clients/mlx/mlx_factory.py
```diff
@@ -0,0 +1,29 @@
...
+        )
+        encoder = TransformersEncoding(tokenizer)
+        model = MLXModel(
+            model_id,
+            model,
+            tokenizer,
+            encoder,
+        )
+        self._models[model_id] = model
+        return model
```
teamdandelion: Per comment elsewhere: Could we drop this file, and just have something like `get_model` defined without a class inside the utils, but using `functools.lru_cache` decorator to get the same caching behavior?

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+@dataclass(frozen=True)
+class SamplerSettings:
+    temp: float | None
+    top_p: float | None
+    top_k: int | None
+    min_p: float | None = None
+    xtc_probability: float = 0.0
+    xtc_threshold: float = 0.0
+    min_tokens_to_keep: int = 1
+    xtc_special_tokens: list[int] = field(default_factory=list)
```
teamdandelion: Let's add docstrings for these properties. See `llm/clients/base/params.py` for some examples. (Can likely copy verbatim for the shared params).

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+from ...tools import BaseToolkit, ToolSchema
+from .encoding import MLXEncoding
+from .model_ids import MLXModelId
+
+T = TypeVar("T")
+Sampler = Callable[[mx.array], mx.array]
+
+
+@dataclass(frozen=True)
+class SamplerSettings:
```
teamdandelion: Let's have a docstring for this class please

python/mirascope/llm/clients/mlx/model_ids.py
```diff
@@ -0,0 +1,4 @@
+from typing import TypeAlias
+
+MLXModelId: TypeAlias = str
+"""The MLX model ids registered with Mirascope."""
```
teamdandelion: It'd be great if this docstring had some examples.
Also "model ids registered with mirascope" is a bit misleading when it is just an alias for string (there is no registration to speak of)

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+            self.min_p or 0.0,
+            self.min_tokens_to_keep,
+            self.top_k or 0,
+            self.xtc_probability,
+            self.xtc_threshold,
+            self.xtc_special_tokens,
+        )
+
+
+class LazyLock:
```
teamdandelion: Feels like these Lock classes should be in their own file, not directly in mlx_models.py

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+
+    async def __aenter__(self) -> "TransferableAsyncLock":
+        assert self._lock is not None
+        await self._lock.acquire()
+        return self
+
+    def take(self) -> LazyAsyncLock | None:
+        result = self._lock
+        self._lock = None
+        return result
```
teamdandelion: Hmm what are the semantics if `take` is called within the lock's context manager? In that case the lock may never be released if the taker does not remember to call `result.release()`? Is this necessary / should we assert that we are not inside the context manager when take is called?

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+            messages, format, stream = await asyncio.to_thread(
+                self._prepare_request, messages, tools, format
+            )
+            chunk_iterator = self._model.encoding.decode_stream(stream)
+            stream_lock = lock.take()
+            return (
+                messages,
+                format,
+                _into_async_stream(chunk_iterator, stream_lock),
+                # _into_async_stream(chunk_iterator),
```
teamdandelion: nit: vestigal comment

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+    def __iter__(self) -> iter_with_last[T]:
+        return self
+
+    def __next__(self) -> T:
+        self._last_item = next(self._iterator)
+        return self._last_item
+
+    @property
+    def last(self) -> T | None:
+        return self._last_item
```
teamdandelion: I'm skeptical about this class (see comment below), also `iter_with_last` doesn't follow class naming conventions.

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+        Sequence[Message],
+        Format[FormattableT] | None,
+        AssistantMessage,
+        GenerationResponse,
+    ]:
+        messages, format, stream = self._prepare_request(messages, tools, format)
+        stream_with_last = iter_with_last(stream)
+        content = self._model.encoding.decode_response(stream_with_last)
+
+        last_response = stream_with_last.last
```
teamdandelion: I'm skeptical of this `stream_with_last` usage. Why not just have `self._model.encoding.decode_response` return the content as well as the last_response as a tuple?

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+        assert last_response is not None
```
teamdandelion: Our convention is not to use assertion errors in production code, except in cases where we are really confident as an invariant that the assertion will _never_ trigger. If there is any possible path where this would trigger in user code, then it should be a runtime error or value error as appropriate

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,337 @@
...
+        async with TransferableAsyncLock(self._model.alock) as lock:
+            messages, format, stream = await asyncio.to_thread(
+                self._prepare_request, messages, tools, format
+            )
+            chunk_iterator = self._model.encoding.decode_stream(stream)
+            stream_lock = lock.take()
+            return (
+                messages,
+                format,
+                _into_async_stream(chunk_iterator, stream_lock),
```
teamdandelion: I'm unclear on why we need to pass the lock into _into_async_stream, when all it does is re-setup a `finally: lock.release()` cleanup. Which would already be true anyway since this code is called within the `async with` context manager. So it seems like we could just write _into_async_stream with the invariant that it's being called inside the async lock context? 

Since `_into_async_stream` is only called once, why not just inline it inside this function? Then it is really clear that it is all being invoked within `stream_async`'s lock context.

python/tests/e2e/conftest.py
```diff
@@ -161,3 +163,59 @@ def formatting_mode(
...
+
+    This fixture:
+    - Detects when a test is using the MLX provider
+    - Mocks mlx_lm.load() to avoid downloading models
+    - Patches MLX generation methods to use cassettes
+    - Manages cassette lifecycle (load, record, save)
+    """
+    if not _is_mlx_provider(request):
+        yield
+        return
```
teamdandelion: Can you explain the `yield` and then `return`?

python/tests/e2e/conftest.py
```diff
@@ -161,3 +163,59 @@ def formatting_mode(
...
+
+    from .mlx_lm_cassette import RecordMode, record_mlx_lm
```
teamdandelion: Is there a particular reason for importing here not top of the file? (I know claude likes to do this for some reason, lol)

python/tests/e2e/conftest.py
```diff
@@ -161,3 +163,59 @@ def formatting_mode(
     "URLImageSource",
     "UserMessage",
 ]
+
+
+def _is_mlx_provider(request: pytest.FixtureRequest) -> bool:
+    """Check if the current test is using the MLX provider."""
+    provider_param = None
+    if "provider" in request.fixturenames:
+        # Get provider from parametrize
```
teamdandelion: Nit: we generally eschew this kind of low-signal comment

python/tests/e2e/output/snapshots/test_call/mlx_mlx_community_Qwen3_0_6B_4bit_DWQ_053125_snapshots.py
```diff
@@ -0,0 +1,142 @@
...
+            "tools": [],
+            "n_chunks": 412,
+        }
+    }
+)
+async_stream_snapshot = snapshot(
+    {
+        "exception": {
+            "type": "NotImplementedError",
+            "args": "('Async streaming is not supported for MLX models.',)",
```
teamdandelion: Hmm that's weird, I thought I saw all the impl code for async streaming?

python/tests/e2e/mlx_lm_cassette.py
```diff
@@ -0,0 +1,157 @@
...
+    for token in prompt:
+        hasher.update(struct.pack(">I", token))
+
+    return hasher.hexdigest()
+
+
+def _patched_stream_generate(
+    cassette: MLXCassette,
+    model: Any,
+    tokenizer: Any,
```
teamdandelion: Can we get better types than `Any` here?

python/tests/e2e/mlx_lm_cassette.py
```diff
@@ -0,0 +1,157 @@
...
+    cached_response = cassette.file_recording.get(prompt_hash)
+
+    if cached_response is not None and cassette.record_mode in ("none", "new_episodes"):
+        yield from cached_response
+        return
+
+    if cassette.record_mode == "none":
+        raise ValueError("No cached response found and recording is disabled.")
+
+    from mlx_lm.generate import stream_generate
```
teamdandelion: nit: inline import

python/mirascope/llm/clients/mlx/clients.py
tnadav: Right, sorry (including all others)

python/mirascope/llm/clients/mlx/clients.py
tnadav: done

python/mirascope/llm/clients/mlx/mlx_factory.py
tnadav: done

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: done

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: done

python/mirascope/llm/clients/mlx/model_ids.py
tnadav: done

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: done

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: moved locking into the stream context, so there is no longer a need for this mechanism

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: whoops

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: done

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: Done. tolerating the case last response is None

python/tests/e2e/conftest.py
tnadav: Since the fixture returns a generator, I need to yield so that it would behave as expected. Added a comment

python/tests/e2e/conftest.py
tnadav: I wasn't really consistent here. There's the concern of making sure you could run tests even if you don't have mlx_lm dependencies installed, but that doesn't really solve it. Moved the import to the top

python/tests/e2e/conftest.py
tnadav: Sure, that's on claude lol

python/tests/e2e/output/snapshots/test_call/mlx_mlx_community_Qwen3_0_6B_4bit_DWQ_053125_snapshots.py
tnadav: I had a bit of a mess here since some tests share snapshots (e.g. context and non-context variants). I ended up implementing everything to make the tests pass

python/tests/e2e/mlx_lm_cassette.py
tnadav: done

tnadav:
I've pushed updates. `bun run fix:python` and `bun run typecheck:python` pass.
As for convergence, some of the tests still fail. I got two reasons for failure:
 - Anthropic authentication error - As I don't have an anthropic token and this test is new, there's no VCR to workaround the failure. Is it ok if I'll try to steel an existing's test cassette for these use cases? Or maybe we can use google genai for tests as anyone can get an API key for free (for now)
 - ~~VCR / huggingface interaction - As the models are first load from huggingface, the tests are having interaction with VCR which cause failure. I need to investigate as I patch the load functions when `vcr-record-mode=none` to avoid downloading models on CI.~~ - Update - huggingface interaction resolved
 
These are the failing tests:
 - ~~test_call_with_image_content[mlx-mlx-community/Qwen3-0.6B-4bit-DWQ-053125] - huggingface interaction~~
 - test_resume_with_override[mlx-mlx-community/Qwen3-0.6B-4bit-DWQ-053125] - anthropic authentication
 - test_resume_with_override_context[mlx-mlx-community/Qwen3-0.6B-4bit-DWQ-053125] - anthropic authentication
 - test_resume_with_override_thinking_and_tools[mlx-mlx-community/Qwen3-0.6B-4bit-DWQ-053125] - anthropic authentication
 - ~~test_call_with_image_content[mlx-mlx-community/Qwen3-0.6B-4bit-DWQ-053125] - huggingface interaction~~

python/mirascope/llm/clients/mlx/_lazy_lock.py
```diff
@@ -0,0 +1,41 @@
+import asyncio
+import threading
+from types import TracebackType
+
+
+class LazyLock:
+    def __init__(self) -> None:
+        self._lock: threading.Lock | None = None
```
teamdandelion: nit: As written it seems like we don't actually need an init method at all; could just write

```py
class LazyLock:
  _lock: threading.Lock | None
  
  def __enter__(self) -> None: ... # as implemented
```

On first glance I was wondering why we don't use init to instantiate the `threading.Lock` but on reading the usage I see that we eagerly create these locks ... and I guess the whole point of it being a `Lazy` lock is that we don't instantiate the lock until/unless we actually use it. A docstring would help clarify the intent.

python/mirascope/llm/clients/mlx/_lazy_lock.py
```diff
@@ -0,0 +1,41 @@
...
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        if self._lock is not None:
+            self._lock.release()
+
+
+class LazyAsyncLock:
```
teamdandelion: same nits as above, also for consistency with the rest of the codebase it may be nice to call it `AsyncLazyLock` (we have lots of `XY` and `AsyncXY`classes, like `ContextStreamResponse` and `AsyncContextStreamResponse`)

python/mirascope/llm/clients/mlx/encoding/base.py
```diff
@@ -0,0 +1,29 @@
...
+
+from mlx_lm.generate import GenerationResponse
+
+from ....formatting import Format, FormattableT
+from ....messages import AssistantContent, Message
+from ....responses import ChunkIterator
+from ....tools import BaseToolkit, ToolSchema
+
+
+class MLXEncoding(abc.ABC):
```
teamdandelion: nit: would be nice to have a docstring here. Is this only ever used once? I wonder if we actually need this interface or if we should just let the implementation speak for itself.

python/tests/e2e/input/snapshots/test_call_with_params/mlx_mlx_community_Qwen3_0_6B_4bit_DWQ_053125_snapshots.py
```diff
@@ -0,0 +1,65 @@
...
+Oh, right! When adding, you have to align the digits properly. So 4200 is 4 2 0 0, and 42 is 4 2. So adding them:
+
+First digit: 0 + 2 = 2
+Second digit: 0 + 4 = 4
+Third digit: 2 + 4 = 6
+Fourth digit: 4 + 2 = 6
+
+So putting it together, it's 4 2 6 6, which is 4266. Wait, but earlier when I broke it down, I got 4242. Now I'm getting 4266. There's a discrepancy here. Which one is correct?
+
+Wait, maybe I made a mistake in the first\
```
teamdandelion, line 51: This is hilarious

python/mirascope/llm/clients/mlx/_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        return None
+
+    return SamplerSettings(
+        temperature=temperature,
+        top_k=top_k,
+        top_p=top_p,
+    )
+
+
+def extract_finish_reason(response: GenerationResponse | None) -> FinishReason | None:
```
willbakst, line 92: nit: let's move this method first so that `_sampler_settings_from_params` and `_get_mlx_model` are better grouped together as private methods consumed by `model_generator`.

python/mirascope/llm/clients/mlx/_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    encoder = TransformersEncoding(tokenizer)
+    return MLXModel(
+        model_id,
+        model,
+        tokenizer,
+        encoder,
+    )
+
+
+def model_generator(model_id: MLXModelId, params: Params) -> MLXModelGenerator:
```
willbakst: nit: none of these methods have docstrings?

Please make sure all methods and classes have docstrings.

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,723 @@
...
+from ..base import BaseClient, Params
+from . import _utils
+from .model_ids import MLXModelId
+
+MLX_CLIENT_CONTEXT: ContextVar["MLXClient | None"] = ContextVar(
+    "MLX_CLIENT_CONTEXT", default=None
+)
+
+
+@cache
```
willbakst: Please use `lru_cache` here to match our existing client singleton caching implementations

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,723 @@
...
+def _mlx_client_slingleton() -> "MLXClient":
```
willbakst: nit: same comment here re: docstrings.

Take a look at some of the other implementations for a reference on what I'm expecting here.

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,723 @@
...
+            model_id: Model identifier to use.
+            messages: Messages to send to the LLM.
+            tools: Optional tools that the model may invoke.
+            format: Optional response format specifier.
+            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
+
+        Returns:
+            An `llm.Response` object containing the LLM-generated content.
+        """
+
```
willbakst, line 153: nit: erroneous newline

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,723 @@
...
+            model_id: Model identifier to use.
+            messages: Messages to send to the LLM.
+            tools: Optional tools that the model may invoke.
+            format: Optional response format specifier.
+            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
+
+        Returns:
+            An `llm.AsyncResponse` object containing the LLM-generated content.
+        """
+
```
willbakst, line 323: nit: erroneous newline

python/mirascope/llm/clients/mlx/clients.py
willbakst: nit: same thing here re: missing docstrings

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,264 @@
...
+    xtc_threshold: float = 0.0
+    "The threshold the probs need to reach for being sampled."
+
+    min_tokens_to_keep: int = 1
+    "Minimum number of tokens that cannot be filtered by min_p sampling."
+
+    xtc_special_tokens: list[int] = field(default_factory=list)
+    "List of special tokens IDs to be excluded from XTC sampling."
+
+    def into_sampler(self) -> Sampler:
```
willbakst: nit: why not just name this method `def sampler(self) -> Sampler: ...`?

python/tests/e2e/conftest.py
```diff
@@ -161,3 +166,56 @@ def formatting_mode(
     "URLImageSource",
     "UserMessage",
 ]
+
+
+def _is_mlx_provider(request: pytest.FixtureRequest) -> bool:
```
willbakst: nit: ditto on docstrings

python/tests/e2e/mlx_lm_cassette.py
willbakst, line 1: At some point we should think through how to refactor this to support other providers that similarly need some level of custom caching (e.g. Grok because it uses gRPC).

Perhaps add a `TODO: ...` around this and resolve this comment.

python/tests/e2e/mlx_lm_cassette.py
```diff
@@ -0,0 +1,162 @@
...
+import yaml
+from mlx_lm.generate import stream_generate
+from mlx_lm.tokenizer_utils import TokenizerWrapper, load_tokenizer
+from mlx_lm.utils import hf_repo_to_path
+
+RecordMode = Literal["once", "new_episodes", "all", "none"]
+
+
+@dataclass
+class MLXCassetteResponse:
```
willbakst, line 25: nit: ditto re: docstrings

python/mirascope/llm/clients/mlx/mlx_model.py
```diff
@@ -0,0 +1,264 @@
...
+            self.min_p or 0.0,
+            self.min_tokens_to_keep,
+            self.top_k or 0,
+            self.xtc_probability,
+            self.xtc_threshold,
+            self.xtc_special_tokens,
+        )
+
+
+class MLXModel:
```
willbakst: There's something about this structurally that's bothering me.

At a high-level, I think it's the fact that `MLXModel` is handling encoding/decoding under the hood, which makes it more akin to a client itself where `MLXClient` is really just an extremely thin wrapper around `MLXModel`.

As I see it, one of the core values of the `MLXModel` and related classes is that they present a nicer interface for actually interacting with the underlying `mlx_lm` library (including locks etc.). What if we were to limit `MLXModel` to just that functionality where it accepted `mlx_lm` types as input and produced `mlx_lm` types as output but handled all of the annoying under-the-hood to get there.

Then we could separately have something like `MLXEncoder` that handles the encoding and decoding.

Inside of the `MLXClient`, methods like e.g. `call` would then use `model_id` to get both an `MLXModel` and `MLXEncoder`.

This would enable following the structure of how we currently implement all of our clients for other supported providers:

1. Encode everything Mirascope into the provider-specific types.
2. Make a request to the LLM
3. Decode the request back into Mirascope types
4. Construct the appropriate response object (e.g. `llm.Response`)

I imagine this would reduce the `call` implementation to something like this (highly simplified, look at existing providers for a more explicit example of this structure):

```python
model, encoder = _utils.load_model(model_id)  # would be `lru_cach`ed

# user encoder to encode Mirascope types
encoded = encoder.encode(messages, tools, format)

# use encoded stuff to generate result
result = model.generate(encoded)

# decode the result
decoded_result = encoder.decode(result)

return Response(...)
```

What do you think?

python/mirascope/llm/clients/mlx/encoding/base.py
tnadav: The main motivation is to directly support harmony encoding for gpt-oss, see https://github.com/tnadav/mirascope/commit/69871a92937a803fe9f4e62e872b859c0b0b1d84

python/tests/e2e/input/snapshots/test_call_with_params/mlx_mlx_community_Qwen3_0_6B_4bit_DWQ_053125_snapshots.py
tnadav: For its size the model is actually pretty impressive!

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: ok

python/tests/e2e/mlx_lm_cassette.py
tnadav: ok

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: Seems like a beneficial evolution from the pattern that was iteratively discovered:
 The goal of `MLXModel` has shrunken to what was appropriate to cache - the model, tokenizer, and additional encoding layer - designed to support both transformers chat template and harmony encoding (I got it all working, see here: https://github.com/tnadav/mirascope/commits/v2-mlx-rich-messages/)
The idea of `MLXModelGenerator` was to combine call-specific parameters and bridge with the 4 call variations: generate, generate async, stream and stream async, and eliminate redundancy for the context / context free variations.
 As I worked on implementing rich messages, `MLXEncoding` consolidated more like what other clients looks like, and exposing it into the client would eliminate what currently is a bit cumbersome "information-passing" between the `MLXModelGenerator` and `MLXEncoder`.

So I think I'll change it to the following pattern:
```python
generator: MLXModelGenerator, encoder: MLXEncoder = _utils.load_model(model_id, params)
messages, format, prompt = encoding.encode_request(messages, tools, format)
response = generator.generate(prompt)
response_message, raw = encoding.decode_response(response)
return Response(...)
```

python/mirascope/llm/clients/mlx/_lazy_lock.py
tnadav: The idea is that on most use cases, a user would either use the synchronous or the asynchronous interface.
For the asynchronous case, it seemed wasteful to me to always create a lock, which internally creates an OS object, which would never be used.
An `asyncio` lock is probably fine, though I didn't go through the implementation, there might be surprises.

python/mirascope/llm/clients/mlx/mlx_model.py
tnadav: There's one thing I forgot, which is that the generator always returns a generator of MLX responses, because decoding harmony responses is more efficient that way. This means the example above won't properly lock the resource, because `generator.generate` only returns the generator, and the actual generation is done in `decode_response`.
To overcome the problem, I suggest doing something like:
```python
model, encoding = _utils.load_model(model_id):
with model.generator(params):  # Will acquire the lock
   messages, format, prompt = encoding.encode_request(messages, tools, format)
   mlx_stream = generator.generate(prompt)
   assistant_message, mlx_response = encoding.decode_response(mlx_stream)
```
Also, maybe `decode_response` needs to be renamed to better reflect that it is doing more than just decoding a flat response.

willbakst: Ok, I think I understand things a little better.

For providers like OpenAI, we convert to their types, and then they convert to the underlying encodings. With MLX we're going straight to the encodings, so there is no in-between type. This explains the current structure, which I think is close to where I want to get it but can still be improved a bit.

I feel like we should implement an `MLX` class with `generate` and `stream` methods that return `AssistantMessage` and `ChunkIterator` respectively. We would also implement their async equivalents (i.e. `generate_async` and `stream_async`). The `MLX` class would operate similarly to the `OpenAI`/`Anthropic`/`Client` classes from OpenAI, Anthropic, and Google (respectively).

One difference is that `MLX` would take `model_id` on initialization instead of as method arguments. We can load the necessary components (like model/tokenizer/etc.) on initialization, and we can use `@lru_cache` on `model_id` to cache `MLX` instances. We would keep the encoding abstraction so `MLX` can load the correct encoder for the given model, but I would update `MLXEncoding` to `BaseEncoder`, and I would update `TransformerEncoding` to `TransformerEncoder`. Things like `generator.generate(prompt)` should likely just live directly in the `MLX.generate` method. Same for `MLX.stream` and the sort.

The `MLXClient` would then implement the `BaseClient` interface as a thin wrapper around the `MLX` class.

I think structuring it this way will make it more clear why we've implemented it the way we have.

What do you think?

tnadav:
I've rebased on top of `v2` branch and made sure all tests pass and `bun run coverage:python`.
Some changes I've made along the way:
 - Remove the `combine_system_messages` argument from the encoding layer as I realized `extract_system_message` doesn't do what I thought it was doing (e.g. it doesn't combine multiple system messages, if present)
 - Add docstrings to all functions
 - Add some unittests to pass convergance
 I didn't the structure of `MLXModel` yet, to be discussed with @willbakst these days

python/mirascope/llm/clients/mlx/mlx_model.py
willbakst: To further clarify based on discussion:

Because of the need for proper locking, we should have an `MLX` class that wraps `encode_request`, `stream_generate`, and `decode_response`/`decode_stream` inside of `generate` and `stream` methods.

The `generate` method should return an `AssistantMessage`, and the `stream` method should return a `ChunkIterator`.

All of the locking would happen inside of `generate` and `stream`. For the `stream` method, we would ensure the lock is inside of the generator so that it is acquired during actual generation.

In this case, the finalized interface in `MLXClient` would look something like this:

```python
client: TransformersMLX = _utils.load_model(model_id)

chunk_iterator = client.stream(messages, tools, format, params)

return StreamResponse(...)
```

This couples the encoding/decoding and the `stream_generate` code all together inside of a `BaseMLX` class.

tnadav:
@willbakst I've made the changes we discussed. In addition, I realized I should always use a real, `threading.Lock`, as otherwise executing async and sync inference in parallel would render unexpected results. This greatly simplify the code as well.
I made all changes in a separate commit so it would be clear. If the changes are to your likings I'll fixup it to the previous commits

python/mirascope/llm/clients/mlx/_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    "The probability of applying XTC sampling."
+
+    xtc_threshold: float
+    "The threshold the probs need to reach for being sampled."
+
+    xtc_special_tokens: list[int]
+    "List of special tokens IDs to be excluded from XTC sampling."
+
+
+class StreamGenerateKwargs(TypedDict, total=False):
```
willbakst, line 48: nit: missing docstring

python/mirascope/llm/clients/mlx/_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Args:
+        params: The generic parameters.
+
+    Returns:
+        The mlx-lm specific stream_generate keyword arguments.
+    """
+    kwargs: StreamGenerateKwargs = {}
+
+    kwargs["max_tokens"] = params.get("max_tokens", -1)
```
willbakst: This must use the safe params accessor context manager. Please see other provider implementations as a reference

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,755 @@
...
+    This client provides methods for generating responses from MLX models,
+    supporting both synchronous and asynchronous operations, as well as
+    streaming responses.
+    """
+
+    @property
+    def _context_var(self) -> ContextVar["MLXClient | None"]:
+        return MLX_CLIENT_CONTEXT
+
+    def __init__(self) -> None:
```
willbakst: Can we not remove this if it does nothing? If not, it is missing a docstring

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,755 @@
...
+        Returns:
+            An `llm.Response` object containing the LLM-generated content.
+        """
+        mlx = _get_mlx(model_id)
+        input_messages, format, assistant_message, response = mlx.generate(
+            messages, tools, format, params
+        )
+        return Response(
+            raw=response,
+            provider="mlx",
```
willbakst, line 160: There should be a `self._provider` attribute. See other clients for reference

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,755 @@
...
+        Returns:
+            An `llm.ContextResponse` object containing the LLM-generated content.
+        """
+        mlx = _get_mlx(model_id)
+        input_messages, format, assistant_message, response = mlx.generate(
+            messages, tools, format, params
+        )
+        return ContextResponse(
+            raw=response,
+            provider="mlx",
```
willbakst, line 251: ditto here and elsewhere re: using `self._provider`

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,755 @@
...
+            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
+
+        Returns:
+            An `llm.Response` object containing the LLM-generated content.
+        """
+        mlx = _get_mlx(model_id)
+        input_messages, format, assistant_message, response = mlx.generate(
+            messages, tools, format, params
+        )
+        return Response(
```
willbakst, line 158: Let's put a new line here between each line to match the other client implementation structures (prep -> generate -> decode (skipped) -> response)

```python
mlx = _get_mlx(model_id)

input_messages, format, assistant_message, response = mlx.generate(
    messages, tools, format, params
)

return Response(...)
```

python/mirascope/llm/clients/mlx/clients.py
```diff
@@ -0,0 +1,755 @@
...
+            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
+
+        Returns:
+            An `llm.ContextResponse` object containing the LLM-generated content.
+        """
+        mlx = _get_mlx(model_id)
+        input_messages, format, assistant_message, response = mlx.generate(
+            messages, tools, format, params
+        )
+        return ContextResponse(
```
willbakst, line 249: ditto here and elsewhere re: adding newlines to match structure of other clients

python/mirascope/llm/clients/mlx/encoding/base.py
```diff
@@ -0,0 +1,35 @@
...
+from mlx_lm.generate import GenerationResponse
+
+from ....formatting import Format, FormattableT
+from ....messages import AssistantContent, Message
+from ....responses import ChunkIterator
+from ....tools import BaseToolkit, ToolSchema
+
+
+class BaseEncoder(abc.ABC):
+    """
```
willbakst: nit: should first have a single line short description. Any additional description would go in the long description.

In this case, I think a shorter single line docstring will suffice:

```python
"""Abstract base class for Mirascope <> MLX encoding and decoding."""
```

python/mirascope/llm/clients/mlx/encoding/base.py
```diff
@@ -0,0 +1,35 @@
...
+    mlx-lm.
+    """
+
+    @abc.abstractmethod
+    def encode_request(
+        self,
+        messages: Sequence[Message],
+        tools: Sequence[ToolSchema] | BaseToolkit | None,
+        format: type[FormattableT] | Format[FormattableT] | None,
+    ) -> tuple[Sequence[Message], Format[FormattableT] | None, list[int]]: ...
```
willbakst: Let's add docstrings to these methods to better document/explain the base interface.

For example, I currently have a hard time identifying what `list[int]` is. On top of docstrings (which may just be single line docstrings), we may want to consider creating a type alias here for improved clarity over `list[int]`

python/mirascope/llm/clients/mlx/encoding/transformers.py
```diff
@@ -0,0 +1,100 @@
...
+
+HFRole = Literal["system", "user", "assistant"] | str
+
+
+class TransformersMessage(TypedDict):
+    role: HFRole
+    content: str
+
+
+def _encode_content(content: Sequence[ContentPart]) -> str:
```
willbakst, line 28: nit: this and other methods are missing docstrings

python/mirascope/llm/clients/mlx/mlx.py
```diff
@@ -0,0 +1,189 @@
...
+    try:
+        for response in generation_stream:
+            asyncio.run_coroutine_threadsafe(queue.put(response), loop)
+    except Exception as e:
+        asyncio.run_coroutine_threadsafe(queue.put(e), loop)
+
+    asyncio.run_coroutine_threadsafe(queue.put(None), loop)
+
+
+class MLX:
```
willbakst, line 45: nit: missing docstrings here and for a few other methods

python/mirascope/llm/clients/mlx/mlx.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+
+    async def _stream_generate_async(
+        self,
+        prompt: list[int],
+        seed: int | None,
+        **kwargs: Unpack[_utils.StreamGenerateKwargs],
+    ) -> AsyncChunkIterator:
+        """Async generator that streams generation responses.
+        Note that, while stream_generate returns an iterable of GenerationResponse,
```
willbakst, line 100: nit: add a newline here between the short and long description

python/mirascope/llm/clients/mlx/_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    xtc_special_tokens: list[int]
+    "List of special tokens IDs to be excluded from XTC sampling."
+
+
+class StreamGenerateKwargs(TypedDict, total=False):
+    max_tokens: int
+    sampler: Sampler
+
+
+def extract_params(params: Params) -> tuple[int | None, StreamGenerateKwargs]:
```
willbakst: nit: we should name this `encode_params`

willbakst (commented):
This refactor looks great! Comments are mostly nits. Ready to approve once those are addressed

python/mirascope/llm/clients/mlx/_utils.py
tnadav: Sure

python/mirascope/llm/clients/mlx/_utils.py
tnadav: sure

python/mirascope/llm/clients/mlx/mlx.py
tnadav: Sure

python/mirascope/llm/clients/mlx/clients.py
tnadav: Sure, will remove

python/mirascope/llm/clients/mlx/clients.py
tnadav: sure

python/mirascope/llm/clients/mlx/encoding/base.py
tnadav: Great idea, will add a type alias as well

tnadav:
Hi @willbakst! I've rebased the branch and made all necessary changes for the tests to pass. I went through all your review comments and addressed them accordingly.
However, I have a question about the self._provider pattern you mentioned in your comments. When I checked the reference implementations: (Anthropic, google, openai) uses literal `provider="..."`. Is there unmerged code on a different branch I should reference, or should MLX follow a different pattern than the existing clients?

willbakst:
That's my mistake. I'm realizing that this change is not yet merged, so for now let's just use the `provider="..."` pattern and we'll clean this up as part of that other change later

python/mirascope/llm/clients/mlx/encoding/base.py
```diff
@@ -0,0 +1,69 @@
...
+        """Encode the request messages into a format suitable for the model.
+
+        Args:
+            messages: The sequence of messages to encode.
+            tools: Optional sequence of tool schemas or toolkit for the model.
+            format: Optional format specification for structured outputs.
+
+        Returns:
+            A tuple containing:
+                - The processed messages
```
willbakst: NOTE

tnadav:
> I combed through a did a little bit of cleanup that I didn't feel needed another round of review. Excited to get this merged and continue making progress on additional support for MLX now that we have a solid base!

Cool! I was just about to rebase the branch as you were pushing the fixes. Next PR - rich message and harmony support!

willbakst:
### Merge activity

* **Dec 9, 11:41 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1438) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).
* **Dec 9, 11:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1438).
* **Dec 9, 11:47 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1438) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

PR #1439, opened 2025-10-20
Implement mirrored provider clients (AzureOpenAI)

koxudaxi:
### TL;DR

Added Azure OpenAI client implementations for both Completions and Responses APIs.

### What changed?

- Implemented `AzureOpenAICompletionsClient` and `AzureOpenAIResponsesClient` classes
- Added client factory functions with proper caching and context management
- Integrated Azure clients with the provider system
- Fixed provider attribution to use "azure-openai:completions" and "azure-openai:responses"
- Added comprehensive test coverage for both client implementations

The implementation follows the same patterns as existing OpenAI clients but adds Azure-specific parameters like `azure_endpoint` and `api_version`, with proper environment variable fallbacks (AZURE_OPENAI_API_KEY, AZURE_OPENAI_ENDPOINT, AZURE_OPENAI_API_VERSION).

### How to test?

```python
from mirascope.llm.clients.azure_openai.completions import client as completions_client
from mirascope.llm.clients.azure_openai.responses import client as responses_client

# Create a completions client
client = completions_client(
    api_key="your-key",
    azure_endpoint="https://your-resource.openai.azure.com",
    api_version="2024-10-21"  # Optional, defaults to latest GA version
)

# Use the client with context management
with client:
    response = client.call(
        model_id="gpt-4o",
        messages=[...]
    )
```

### Why make this change?

Azure OpenAI is a popular deployment option for organizations that need:
- Regional data residency compliance
- Enterprise security features
- Dedicated capacity
- Predictable pricing

This implementation allows Mirascope users to seamlessly work with Azure-hosted OpenAI models using the same familiar interface as the direct OpenAI clients, while properly handling Azure-specific authentication and endpoint configuration.


koxudaxi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1439?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1465** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1465?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1468](https://github.com/Mirascope/mirascope/pull/1468) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1468?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1485](https://github.com/Mirascope/mirascope/pull/1485) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1485?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1464** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1464?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1455** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1454** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1473** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1473?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1470** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1471** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1472** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1478** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1478?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1439** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1475** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1475?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1476](https://github.com/Mirascope/mirascope/pull/1476) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1476?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1441** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1461** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1461?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2095** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,153 @@
...
+    return build_azure_client(
+        _azure_completions_factory,
+        api_key,
+        resolved_azure_endpoint,
+        api_version,
+        azure_ad_token_provider,
+    )
+
+
+def get_client() -> "OpenAICompletionsClient":
```
teamdandelion: Should this return `AzureOpenAICompletionsClient` rather than `OpenAICompletionsClient`?

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,153 @@
...
+
+class AzureOpenAICompletionsClient(OpenAICompletionsClient):
+    """Azure OpenAI Completions client that inherits from OpenAI client.
+
+    Only overrides initialization to use Azure-specific SDK classes and
+    provider naming to return 'azure-openai:completions'.
+    """
+
+    @property
+    def _context_var(self) -> ContextVar["OpenAICompletionsClient | None"]:
```
teamdandelion: Likewise "AzureOpenAICompletionsClient | None" in the context var?

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,158 @@
...
+            - "2025-04-01-preview" (preview features)
+            - "preview" (v1 preview route via /openai/v1/ path)
+        azure_ad_token_provider: Optional Azure AD token provider for authentication.
+
+    Returns:
+        An `AzureOpenAIResponsesClient` instance.
+
+    Notes:
+        The current OpenAI SDK requires api_version to be specified.
+        We default to "2024-10-21" (latest GA version).
```
teamdandelion: Hmm if Azure releases a new version (so that 2024-10-21 is no longer latest) how long do you think it will be before we update (ie how long will the docstrings be out of date)?
I wonder if it would be better to not say that 2024-10-21 is latest (to future proof against when that changes) especially if we don't have any process around making sure that version updates are propagated to the code in a timely fashion

python/mirascope/llm/clients/mirrors/azure.py
```diff
@@ -0,0 +1,64 @@
+"""Azure OpenAI mirror client helpers."""
```
teamdandelion: Why do we have `clients/mirrors/azure.py` vs putting this logic in `clients/azure_openai`?

python/mirascope/llm/clients/openai/completions/clients.py
```diff
@@ -94,6 +97,11 @@ def __init__(
         self.client = OpenAI(api_key=api_key, base_url=base_url)
         self.async_client = AsyncOpenAI(api_key=api_key, base_url=base_url)
 
+    @property
+    def _provider_name(self) -> "Provider":
```
teamdandelion: This is a smell that this is a private property but we are expecting other classes elsewhere in the codebase to override it. I don't think private properties should be part of the public contract of the class.

How about we just add a public property `def provider(self) -> "Provider"` to `BaseClient` (python/mirascope/llm/clients/base/client.py)? The docstring can say that it provides the name of the provider, and is available for overwriting by subclasses in the case of a mirrored or wrapped client.

If adding this refactor, please create it as a separate PR that precedes this one in the stack (note you could checkout trunk and then do `gt create --insert` to add it under this one). (If doing so we may as well refactor every provider to use `provider=self.provider` when constructing responses for consistency sake)

python/mirascope/llm/clients/providers.py
```diff
@@ -146,25 +186,110 @@ def client(
...
     """Create a cached client instance for the specified provider.
 
     Args:
-        provider: The provider name ("openai:completions", "anthropic", or "google").
+        provider: The provider name.
         api_key: API key for authentication. If None, uses provider-specific env var.
-        base_url: Base URL for the API. If None, uses provider-specific env var.
+        base_url: Base URL for the API (OpenAI, Anthropic, Google only).
+        azure_endpoint: Azure OpenAI endpoint URL (Azure only).
+        api_version: Azure OpenAI API version (Azure only).
```
teamdandelion: If `api_version` is an Azure only parameter let's call it `azure_api_version` for clarity

With `azure_endpoint`, I'm wondering if it makes sense to unify with `base_url` somehow?

teamdandelion (commented):
Looks like the Azure OpenAI client isn't integrated with any of our e2e integration tests. I'm not sure if we need to test this for every single test case (may be redundant with the OpenAI testing) but I think we should have some e2e tests to ensure this is functional. 

The simplest course would be to simply add the azure-openai clients to `PROVIDER_MODEL_ID_PAIRS` in python/tests/e2e/conftest.py so that we test openai-azure with every provider. This is fully comprehensive but the downside is a lot more cassettes / snapshots. It's probably a good idea to do it this way so that if there are any observable differences between the azure-openai behavior and the regular openai behavior, we will be aware of it.

This would also have the effect of documenting the pattern for configuring the client with the necessary parameters so that usage of the `llm.call` decorator will work (since `llm.call` takes the provider name but not the extra parameters like azure endpoint, that will need to be configured by using the client context manager). 

When integrating with e2e tests, it may be nice to make the addition of azure-openai to the e2e tests a separate PR one level up the stack, for ease of review (if we add azure e2e tests to everything it will be a very large diff). Or maybe add azure to just one test in this PR (eg e2e/output/test_call.py) while we make sure we like the pattern, then add it to all tests in a separate PR upstack.

python/mirascope/llm/clients/mirrors/azure.py
koxudaxi: I also struggled with which one to choose. I'll try changing it.

python/mirascope/llm/clients/azure_openai/completions/clients.py
koxudaxi: I tried that at first, but I couldn't fix the type error and gave up. I'll think about whether there's a way to make it work, and if not, I'll try using `ignore`.

python/mirascope/llm/clients/azure_openai/completions/clients.py
koxudaxi: I fixed it.

python/mirascope/llm/clients/mirrors/azure.py
koxudaxi: Done

python/mirascope/llm/clients/openai/completions/clients.py
koxudaxi: OK, I just added the new stack.

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,154 @@
...
+
+class AzureOpenAICompletionsClient(OpenAICompletionsClient):
+    """Azure OpenAI Completions client that inherits from OpenAI client.
+
+    Only overrides initialization to use Azure-specific SDK classes and
+    provider naming to return 'azure-openai:completions'.
+    """
+
+    @property
+    def _context_var(self) -> ContextVar["AzureOpenAICompletionsClient | None"]:  # type: ignore[override]
```
willbakst: What is this type ignore?

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        self.async_client = AsyncAzureOpenAI(
+            api_key=api_key,
+            azure_endpoint=base_url,
+            api_version=azure_api_version,
+            azure_ad_token_provider=azure_ad_token_provider,
+        )
+
+    @property
+    def provider(self) -> "Provider":
```
willbakst: nit: for all other providers in the previous PR, you put this above the `__init__` method, so please do the same here

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,158 @@
...
+
+class AzureOpenAIResponsesClient(OpenAIResponsesClient):
+    """Azure OpenAI Responses client that inherits from OpenAI client.
+
+    Only overrides initialization to use Azure-specific SDK classes and
+    provider naming to return 'azure-openai:responses'.
+    """
+
+    @property
+    def _context_var(self) -> ContextVar["AzureOpenAIResponsesClient | None"]:  # type: ignore[override]
```
willbakst: Ditto re: type ignore?

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+        self.async_client = AsyncAzureOpenAI(
+            api_key=api_key,
+            azure_endpoint=base_url,
+            api_version=azure_api_version,
+            azure_ad_token_provider=azure_ad_token_provider,
+        )
+
+    @property
+    def provider(self) -> "Provider":
```
willbakst: ditto re: method placement / ordering

python/mirascope/llm/clients/openai/completions/_utils/decode.py
```diff
@@ -35,6 +38,7 @@
 def decode_response(
     response: openai_types.ChatCompletion,
     model_id: OpenAICompletionsModelId,
+    provider: "Provider",
```
willbakst: This seems silly? It can't actually be type "Provider"? It can only be `"openai:completions"` and I guess `"azure-openai:completions"` now? I feel like we should probably make that more specific?

python/mirascope/llm/clients/openai/responses/_utils/decode.py
```diff
@@ -48,6 +51,7 @@ def _serialize_output_item(
 def decode_response(
     response: openai_types.Response,
     model_id: OpenAIResponsesModelId,
+    provider: "Provider",
```
willbakst: Same comment here

python/mirascope/llm/clients/providers.py
```diff
@@ -25,6 +36,8 @@
 
 Provider: TypeAlias = Literal[
     "anthropic",
+    "azure-openai:completions",  # AzureOpenAICompletionsClient
```
willbakst, line 39: What do we all think about adding comments for all of the providers? Why are anthropic and google lacking comments? @teamdandelion WDYT?

python/mirascope/llm/clients/providers.py
```diff
@@ -146,25 +186,108 @@ def client(
...
 
     Raises:
-        ValueError: If the provider is not supported.
+        ValueError: If the provider is not supported or required parameters are missing.
     """
     match provider:
         case "anthropic":
             return anthropic_client(api_key=api_key, base_url=base_url)
+        case "azure-openai:completions":
+            from .azure_openai.completions import client as azure_completions_client
```
willbakst: Am I crazy that a lot of this seems duplicated with the endpoint resolution inside of the actual client creation? Why can't we just push all of this inside of azure_completions_client? Seems like a better place for that specific functionality to live? I feel like this should be a single line like all of the other providers?

python/mirascope/llm/clients/providers.py
```diff
@@ -146,25 +186,108 @@ def client(
...
+                )
+
+            return azure_completions_client(
+                api_key=resolved_api_key,
+                base_url=resolved_endpoint,
+                azure_api_version=resolved_version,
+                azure_ad_token_provider=azure_ad_token_provider,
+            )
+        case "azure-openai:responses":
+            from .azure_openai.responses import client as azure_responses_client
```
willbakst: Same comment here

python/mirascope/llm/clients/providers.py
```diff
@@ -146,25 +186,108 @@ def client(
...
+    """Create a cached Azure OpenAI responses client with the given parameters."""
+    ...
+
+
+def client(
+    provider: Provider,
+    *,
+    api_key: str | None = None,
+    base_url: str | None = None,
+    # Azure-specific parameters
```
willbakst: I really don't like this. I don't like that we've now added these provider specific named parameters to the base method.

Would it be possible to instead make this something like `**kwargs: Any` where we have no additional kwargs for existing providers but we name these for the azure providers? This would also make this method future-proof for additional providers where the base method doesn't have to name them all and only the provider specific overrides do?

python/mirascope/llm/clients/azure_openai/completions/clients.py
koxudaxi: `ContextVar` is invariant, so returning `ContextVar[AzureOpenAICompletionsClient | None]` while overriding the base signature `ContextVar[OpenAICompletionsClient | None]` triggers Pyright's override error. We intentionally ignore it because the implementation is still safe for this subclass.

python/mirascope/llm/clients/azure_openai/responses/clients.py
koxudaxi: same above.

python/mirascope/llm/clients/providers.py
koxudaxi: I agree.

python/mirascope/llm/clients/azure_openai/completions/clients.py
willbakst: Let's make `BaseClient` generic on the context variable then?

python/mirascope/llm/clients/azure_openai/responses/clients.py
willbakst: See my comment on the other re: generic?

python/mirascope/llm/clients/openai/completions/_utils/decode.py
```diff
@@ -26,6 +26,11 @@
 )
 from ..model_ids import OpenAICompletionsModelId
 
+OpenAICompletionsProvider: TypeAlias = Literal[
```
willbakst: This type alias seems equally weird given we have `azure-openai` as a separate client/provider/directory but then have it as an option for `OpenAICompletionsProvider`. Since we only use this once, why not just inline the literal? This would avoid having to think of a good name for the type alias.

python/mirascope/llm/clients/openai/responses/_utils/decode.py
```diff
@@ -32,6 +32,12 @@
 )
 from ..model_ids import OpenAIResponsesModelId
 
+OpenAIResponsesProvider: TypeAlias = Literal[
```
willbakst: Same comment here as for completions

python/mirascope/llm/clients/azure_openai/completions/clients.py
koxudaxi: OK, I added the Generic variable to BaseClient and created `BaseOpenAIResponsesClient` and `BaseOpenAICompletionsClient`.

teamdandelion (dismissed):

teamdandelion (commented):
Code changes seem good to me, but let's please get the e2e tests integrated. Looks like right now we have them for AnthropicBedrock but not yet AzureOpenAI

python/mirascope/llm/clients/providers.py
teamdandelion: I'm happy to have these comments consistently too.

koxudaxi: OK, I added the new stack.
https://app.graphite.dev/github/pr/Mirascope/mirascope/1471/Add-provider-name-comments

python/mirascope/llm/clients/providers.py
```diff
@@ -146,25 +187,79 @@ def client(
...
     match provider:
         case "anthropic":
             return anthropic_client(api_key=api_key, base_url=base_url)
+        case "azure-openai:completions":
+            return azure_openai_completions_client(
+                api_key=api_key,
+                base_url=base_url,
+                azure_api_version=kwargs.get("azure_api_version"),
+                azure_ad_token_provider=kwargs.get("azure_ad_token_provider"),
+            )
```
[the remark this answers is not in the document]
koxudaxi: The Azure client already handles this validation with a more detailed error message that tells users how to fix it. Adding a check here would be redundant.

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,159 @@
...
+    resolved_base_url = base_url or os.getenv("AZURE_OPENAI_ENDPOINT")
+    resolved_api_version = (
+        azure_api_version or os.getenv("AZURE_OPENAI_API_VERSION") or "2024-10-21"
+    )
+
+    if resolved_base_url is None:
+        raise ValueError(
+            "base_url is required. "
+            "Set AZURE_OPENAI_ENDPOINT environment variable or pass base_url explicitly."
+        )
```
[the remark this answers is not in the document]
koxudaxi: Already fixed in previous review. The test expects the full error message now and passes.

python/mirascope/llm/clients/base/client.py
```diff
@@ -72,7 +81,7 @@ def provider(self) -> Provider:
 
     def __enter__(self) -> Self:
         """Sets the client context and stores the token."""
-        self._token = self._context_var.set(self)
+        self._token = self._context_var.set(self)  # type: ignore[arg-type]
```
teamdandelion: I think it'd be better to use a cast here:
`self._token = self._context_var.set(cast(ClientT, self))`

python/mirascope/llm/clients/openai/completions/_utils/encode.py
```diff
@@ -228,6 +236,10 @@ def _create_strict_response_format(
 
     _shared_utils._ensure_additional_properties_false(schema)
 
+    if provider.startswith("azure-openai") and format.formatting_instructions:
+        # TODO: Add test coverage for Azure OpenAI formatting instructions
+        schema["description"] = format.formatting_instructions  # pragma: no cover
```
teamdandelion: Why is this special behavior for azure-openai? 
Generally formatting instructions are handled by modifying the system message.

python/mirascope/llm/clients/openai/completions/_utils/encode.py
```diff
@@ -324,4 +339,24 @@ def encode_request(
...
     kwargs["messages"] = encoded_messages
 
+    has_tool_outputs = any(
+        isinstance(msg, dict) and msg.get("role") == "tool" for msg in encoded_messages
+    )
+
+    if (
+        has_tool_outputs and openai_tools and provider.startswith("azure-openai")
+    ):  # pragma: no cover
+        # TODO: Add test coverage for Azure OpenAI formatting instructions
```
teamdandelion: Why do we have special behavior here for azure-openai?

python/mirascope/llm/clients/openai/completions/base_client.py
```diff
@@ -0,0 +1,833 @@
...
+class AsyncOpenAICompatibleClient(Protocol):
+    """Protocol for OpenAI-compatible async clients."""
+
+    @property
+    def chat(self) -> AsyncChat: ...
+
+
+ProviderClientT = TypeVar("ProviderClientT", bound=OpenAICompatibleClient)
+AsyncProviderClientT = TypeVar(
+    "AsyncProviderClientT", bound=AsyncOpenAICompatibleClient
```
teamdandelion: Nit: maybe we name these something like `OpenaiCompletionsClientT` and `AsyncOpenAICompletionsClientT` to distinguish from the more generic `ProviderClientT` and `AsyncProviderT` in `base/client.py`?

python/mirascope/llm/clients/openai/completions/base_openai_completions_client.py
```diff
@@ -0,0 +1,833 @@
+"""Base OpenAI Completions client implementation."""
```
teamdandelion, line 1: nit: slight preference to name this file something like `base_openai_completions_client.py` so that when I am fuzzy searching for baseclient (looking for the most generic base client) it's easiest to spot

python/mirascope/llm/clients/openai/responses/_utils/encode.py
```diff
@@ -320,7 +331,26 @@ def encode_request(
...
+        )
+        # If format tool is present, tool_choice is already set above
+        # Otherwise, if we have tool outputs, Azure OpenAI requires tool_choice="none"
+        if (  # pragma: no cover
+            # TODO: Add test coverage for Azure OpenAI tool outputs handling
+            not format_tool_present
+            and has_tool_outputs
+            and provider.startswith("azure-openai")
+        ):
+            kwargs["tool_choice"] = "none"  # pragma: no cover
```
teamdandelion: I'm confused by this, can you help me understand the intent / what problem this is solving?

Per the [OpenAI docs](https://platform.openai.com/docs/guides/function-calling?api-mode=responses#additional-configurations):
> You can also set tool_choice to "none" to imitate the behavior of passing no functions.

So, it sounds like setting `tool_choice` to `"none"` will disable all tool calling. Meanwhile the `has_tool_outputs` flag is set true if _any_ message in the history included a tool output. So the logic is that if the user has ever resolved a tool call in the conversation history, we disable tool calling for the rest of the conversation? That does not seem correct or desirable?

python/mirascope/llm/clients/openai/completions/_utils/encode.py
koxudaxi: Azure OpenAI Completions API strict mode ignores formatting_instructions in system messages. 

Verified by removing this workaround - tests failed with `title="Book"` (expected uppercase per formatting_instructions). Restoring the workaround fixed it. Not officially documented.

python/mirascope/llm/clients/openai/completions/_utils/encode.py
koxudaxi: The completions API worked correctly without this special handling, so I removed it.

python/mirascope/llm/clients/openai/responses/_utils/encode.py
koxudaxi: This is solving an Azure OpenAI Responses API specific issue, and it only affects a
single resume request, not the entire conversation.

The problem: After receiving tool outputs during `response.resume(tool_outputs)`, Azure OpenAI
Responses API may call the same tools again without this parameter, creating an infinite loop. This
behavior is not documented in official Azure OpenAI docs, but we've verified it through testing.

This logic only runs when we're in a resume() call with tool results (`has_tool_outputs=true`), tools
are still available (`openai_tools` present), and the provider is Azure OpenAI. It doesn't apply to
initial calls or subsequent new conversations.

Here's what happens without tool_choice:
Request 1: Model calls tools
Tool execution: Get resultsRequest 2 (resume): Send tool outputs
Response 2: Model calls same tools again (infinite loop)

With `tool_choice="none"`:
Request 1: Model calls tools
Tool execution: Get results
Request 2 (resume): Send tool outputs + tool_choice="none"Response 2: Model returns final text response

When a FORMAT_TOOL is present (for structured output), we force its use instead of "none" to ensure 
structured output extraction works correctly after tool execution.

Testing evidence: Completions API doesn't need this (20/20 tests passed without it). Responses API 
enters infinite loops without it (verified by deleting cassettes and re-running tests).

python/mirascope/llm/clients/openai/completions/_utils/encode.py
koxudaxi: OK, I just removed the logic. Because, gpt-5-mini resolve the problem.

python/mirascope/llm/clients/openai/responses/_utils/encode.py
teamdandelion: My concern is this that `has_tool_outputs` checks across the entire message history, not the most recent user message. So once there has ever been a tool output (even in a previous message), future tool calls become impossible. Consider the following:

User: What's the weather in SF?
Assistant: ToolCall: GetWeather(SF)
User: ToolOutput: WeatherOutput(SF)
Assistant: The weather is warm and sunny. (ToolChoice = None)
User: Ok, what's the weather in NYC? 
Assistant: I don't know, sorry! (ToolChoice = None)

We don't have an e2e test for this because it's never been an edge case we've needed to have particular coverage for, but I think if we did have one, this code would fail.

koxudaxi: Thank you for the explanation. I understand the situation now. I'll modify it to check only the last user message instead of the entire history. Should I probably add some tests as well?

koxudaxi: I discussed this with @willbakst , and he said that OpenAI and Azure OpenAI should behave identically.
I definitely agree with that. Since there was a possibility of bugs in my Mirascope code, I wrote a verification script using the OpenAI SDK directly, independent of Mirascope.

During this investigation, I discovered that Azure API Version v1 is now available, which allows us to use Azure OpenAI by simply changing the endpoint of the same OpenAI client, rather than using a separate Azure client.
So for this test, I decided to run comparison tests by switching the connection endpoint.

The conclusion is that with the existing implementation, Azure OpenAI fails frequently (though not always).
Also, as mentioned in my previous PR, specifying `previous_response_id` and status = "completed" appears to prevent errors on Azure OpenAI as well. I can't say it's 100% reliable since I haven't run enough trials.

`history` is the current implementation, `previous` adds `previous_response_id`, and `previous-compatible` specifies both `previous_response_id` and status = "completed".

By the way, the model is gpt-4o for OpenAI and gpt-5-mini for Azure OpenAI. Since it doesn't work even with mini, I think the model isn't the issue.

Scripts resuts:
```bash
$ uv run python "scripts/repro_openai_sdk_responses_resume.py" 

##### provider=openai strategy=history #####
=== initial call ===
status=completed tool_calls=2
  1. id=call_KBvvLD12L3dFOhU8044c7RlY args={"password":"mellon"}
  2. id=call_uhDV1WnPTO3pAzffU4u4nYtV args={"password":"radiance"}

=== resume call (history replay) ===
status=completed tool_calls=0
assistant: Here are the secrets associated with the passwords:

- **mellon**: Welcome to Moria!
- **radiance**: Life before Death
✅ resume produced assistant text

##### provider=openai strategy=previous #####
=== initial call ===
status=completed tool_calls=2
  1. id=call_dwqXD3EkW43XyF9czFb7awVD args={"password":"mellon"}
  2. id=call_dOcPw0hbfcubNFIHOfaoKdsu args={"password":"radiance"}

=== resume call (previous_response_id, tools re-sent, no status) ===
status=completed tool_calls=0
assistant: Here are the secrets for each password:

- **mellon**: Welcome to Moria!
- **radiance**: Life before Death
✅ resume produced assistant text

##### provider=openai strategy=previous-compatible #####
=== initial call ===
status=completed tool_calls=2
  1. id=call_N7NuerpWuzGV7rnZt7PqMvUJ args={"password":"mellon"}
  2. id=call_I4m3LXP3HIlDeEbGdsqKAdvN args={"password":"radiance"}

=== resume call (previous_response_id + status, no tools) ===
status=completed tool_calls=0
assistant: - **mellon**: Welcome to Moria!
- **radiance**: Life before Death
✅ resume produced assistant text

##### provider=azure strategy=history #####
=== initial call ===
status=completed tool_calls=2
  1. id=call_Bf7QvR1z3Ut1WeO9cZZHj642 args={"password":"mellon"}
  2. id=call_fv2PLkUvFN72q1vQr9TJ7eoz args={"password":"radiance"}

=== resume call (history replay) ===
status=completed tool_calls=2
  1. id=call_1SKHe2psNrlqvYmPqxG52IDd args={"password":"mellon"}
  2. id=call_QxYxdUmoM7CgX4WMSXSrEnDQ args={"password":"radiance"}
❌ resume returned tool calls again

##### provider=azure strategy=previous #####
=== initial call ===
status=completed tool_calls=2
  1. id=call_Cuw2xCMbJihL10NGeyrRzf7L args={"password":"mellon"}
  2. id=call_AdnyGzgnplFmT7eUydAVLBsB args={"password":"radiance"}

=== resume call (previous_response_id, tools re-sent, no status) ===
status=completed tool_calls=2
  1. id=call_fTm0oTWn6BExNV4Z8CMCnXNU args={"password":"mellon"}
  2. id=call_SnRJKpnV1WrkZWCLM6Gf5qFa args={"password":"radiance"}
❌ resume returned tool calls again

##### provider=azure strategy=previous-compatible #####
=== initial call ===
status=completed tool_calls=2
  1. id=call_0DVC6ccDpm1eAYlRTKmPoi1q args={"password":"mellon"}
  2. id=call_gYMtcci6pucacKGi7ejoQhi8 args={"password":"radiance"}

=== resume call (previous_response_id + status, no tools) ===
status=completed tool_calls=0
assistant: Here are the secrets retrieved:

- mellon → "Welcome to Moria!"
- radiance → "Life before Death"

Would you like these saved, exported, or shown in a different format?
✅ resume produced assistant text

=== summary ===
openai:
  history: success
  previous: success
  previous-compatible: success
azure:
  history: loop
  previous: loop
  previous-compatible: success
```

```python
import argparse
import json
import os
from typing import Any

from openai import OpenAI

PASSWORD_LIST = ["mellon", "radiance"]
PASSWORD_MAP = {"mellon": "Welcome to Moria!", "radiance": "Life before Death"}


def base_messages() -> list[dict[str, str]]:
    joined = ",".join(PASSWORD_LIST)
    return [
        {"role": "developer", "content": "Use parallel tool calling."},
        {
            "role": "user",
            "content": f"Please retrieve the secrets associated with each of these passwords: {joined}",
        },
    ]


def tool_schema() -> list[dict[str, Any]]:
    return [
        {
            "type": "function",
            "name": "secret_retrieval_tool",
            "description": "A tool that requires a password to retrieve a secret.",
            "parameters": {
                "type": "object",
                "properties": {"password": {"type": "string"}},
                "required": ["password"],
                "additionalProperties": False,
            },
        }
    ]


def run_flow(client: OpenAI, model: str, strategy: str) -> str:
    messages = base_messages()
    tools = tool_schema()
    print("=== initial call ===")
    first = client.responses.create(model=model, input=messages, tools=tools)
    print_summary(first)

    calls = [item for item in first.output if item.type == "function_call"]
    outputs = []
    for call in calls:
        args = json.loads(call.arguments)
        outputs.append(
            {
                "type": "function_call_output",
                "call_id": call.call_id,
                "output": PASSWORD_MAP.get(args["password"], "Invalid password!"),
            }
        )

    if strategy == "history":
        payload = messages + [strip_status(item) for item in first.output] + outputs
        resume_kwargs = dict(model=model, input=payload, tools=tools)
        print("\n=== resume call (history replay) ===")
    elif strategy == "previous":
        resume_kwargs = dict(
            model=model,
            input=outputs,
            tools=tools,
            previous_response_id=first.id,
        )
        print("\n=== resume call (previous_response_id, tools re-sent, no status) ===")
    else:  # previous-compatible
        compatible_outputs = []
        for out in outputs:
            updated = dict(out)
            updated["status"] = "completed"
            compatible_outputs.append(updated)
        resume_kwargs = dict(
            model=model,
            input=compatible_outputs,
            previous_response_id=first.id,
        )
        print("\n=== resume call (previous_response_id + status, no tools) ===")

    try:
        second = client.responses.create(**resume_kwargs)
    except Exception as exc:  # noqa: BLE001
        print(f"❌ resume call raised {exc}")
        return "error"
    print_summary(second)

    repeat_calls = [item for item in second.output if item.type == "function_call"]
    assistant_text = [
        content.text
        for item in second.output
        if item.type == "message"
        for content in item.content
        if content.type == "output_text"
    ]

    if repeat_calls and not assistant_text:
        print("❌ resume returned tool calls again")
        return "loop"
    if assistant_text:
        for text in assistant_text:
            print("assistant:", text)
        print("✅ resume produced assistant text")
        return "success"
    print("⚠️ resume returned neither text nor tool calls")
    return "unknown"


def strip_status(item: Any) -> dict[str, Any]:
    data = item.model_dump(exclude_none=True)
    data.pop("status", None)
    return data


def print_summary(response: Any) -> None:
    calls = [item for item in response.output if item.type == "function_call"]
    print(f"status={response.status} tool_calls={len(calls)}")
    for idx, call in enumerate(calls, 1):
        print(f"  {idx}. id={call.call_id} args={call.arguments}")


def get_client(provider: str) -> tuple[OpenAI, str]:
    if provider == "openai":
        api_key = os.environ["OPENAI_API_KEY"]
        base_url = os.environ.get("OPENAI_BASE_URL")
        return OpenAI(api_key=api_key, base_url=base_url), os.environ.get(
            "OPENAI_MODEL", "gpt-4o"
        )
    if provider == "azure":
        api_key = os.environ["AZURE_OPENAI_API_KEY"]
        endpoint = os.environ["AZURE_OPENAI_ENDPOINT"]
        model = os.environ.get("AZURE_OPENAI_DEPLOYMENT", "gpt-5-mini")
        base_url = endpoint.rstrip("/") + "/openai/v1/"
        return OpenAI(api_key=api_key, base_url=base_url), model
    raise ValueError(f"unknown provider: {provider}")


def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--provider",
        choices=["openai", "azure"],
        help="Run only the specified provider; omit to run both.",
    )
    parser.add_argument(
        "--strategy",
        choices=["history", "previous", "previous-compatible"],
        help=(
            "history=replay entire input; "
            "previous=previous_response_id without status; "
            "previous-compatible=Azure-friendly variant. Omit to run all."
        ),
    )
    args = parser.parse_args()

    providers = [args.provider] if args.provider else ["openai", "azure"]
    strategies = (
        [args.strategy]
        if args.strategy
        else ["history", "previous", "previous-compatible"]
    )
    failed = False
    results: dict[str, dict[str, str]] = {}
    for provider in providers:
        results[provider] = {}
        client, model = get_client(provider)
        for strategy in strategies:
            print(f"\n##### provider={provider} strategy={strategy} #####")
            outcome = run_flow(client, model, strategy)
            results[provider][strategy] = outcome
            failed |= outcome != "success"

    print("\n=== summary ===")
    for provider, strat_results in results.items():
        print(f"{provider}:")
        for strategy, outcome in strat_results.items():
            print(f"  {strategy}: {outcome}")

    if failed:
        raise SystemExit(1)


if __name__ == "__main__":
    main()

```

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,251 @@
...
+        See: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle
+
+    Examples:
+        # API key authentication
+        client = client(
+            api_key="your-key",
+            base_url="https://your-resource.openai.azure.com"
+        )
+
+        # Azure AD authentication (keyless) - sync only
```
willbakst, line 83: I really don't like that we're letting the `azure` package seep through the cracks here.

It's worth noting in this examplt that we use `DefaultAzureCredentials` with no arguments and a fixed URL, so unless the user need to override something, it seems like we could do this under the hood if no API key is provided?

I feel like the path should be to first check if an API key is given (or exists in the environment), in which case we would use that. If not, we would attempt to load credentials (keyless) on the user's behalf. If that fails we error. If necessary, we could consider exposing the arguments for configuration, but given I don't see an example of that, I'm not sure we need that usage yet.

What do you think?

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    endpoint = base_url or os.getenv("AZURE_OPENAI_ENDPOINT")
+
+    if endpoint is None:
+        raise ValueError(
+            "base_url is required. "
+            "Set AZURE_OPENAI_ENDPOINT environment variable or pass base_url explicitly."
+        )
+
+    resolved_base_url = endpoint.rstrip("/") + "/openai/v1/"
```
willbakst, line 122: Hmm, any reason we are resolving this? Seems like this prevents a user from potentially setting this to `/v2` if they want to? Also isn't OpenAI at major version 2 now?

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,251 @@
...
+
+    if azure_ad_token_provider is not None or azure_ad_token_provider_async is not None:
+        return AzureOpenAICompletionsClient(
+            api_key=api_key,
+            base_url=resolved_base_url,
+            azure_ad_token_provider=azure_ad_token_provider,
+            azure_ad_token_provider_async=azure_ad_token_provider_async,
+        )
+
+    resolved_api_key = api_key or os.getenv("AZURE_OPENAI_API_KEY")
```
willbakst, line 132: Let's just make this `api_key = api_key or os.getenv("AZURE_OPENAI_API_KEY")`

python/mirascope/llm/clients/azure_openai/completions/clients.py
```diff
@@ -0,0 +1,251 @@
...
+    if resolved_api_key is None:
+        raise ValueError(
+            "api_key is required. Set AZURE_OPENAI_API_KEY or pass api_key explicitly."
```
willbakst, line 135: nit: put `AZURE_OPENAI_API_KEY` and `api_key` in backticks

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Examples:
+        # API key authentication
+        client = client(
+            api_key="your-key",
+            base_url="https://your-resource.openai.azure.com"
+        )
+
+        # Azure AD authentication (keyless) - sync only
+        from azure.identity import DefaultAzureCredential, get_bearer_token_provider
```
willbakst, line 84: Same comment here

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    endpoint = base_url or os.getenv("AZURE_OPENAI_ENDPOINT")
+
+    if endpoint is None:
+        raise ValueError(
+            "base_url is required. "
+            "Set AZURE_OPENAI_ENDPOINT environment variable or pass base_url explicitly."
+        )
+
+    resolved_base_url = endpoint.rstrip("/") + "/openai/v1/"
```
willbakst, line 122: Same comment here

python/mirascope/llm/clients/azure_openai/responses/clients.py
```diff
@@ -0,0 +1,251 @@
...
+
+    if azure_ad_token_provider is not None or azure_ad_token_provider_async is not None:
+        return AzureOpenAIResponsesClient(
+            api_key=api_key,
+            base_url=resolved_base_url,
+            azure_ad_token_provider=azure_ad_token_provider,
+            azure_ad_token_provider_async=azure_ad_token_provider_async,
+        )
+
+    resolved_api_key = api_key or os.getenv("AZURE_OPENAI_API_KEY")
```
willbakst, line 132: Same comment here

python/mirascope/llm/clients/base/client.py
```diff
@@ -1267,3 +1276,27 @@ async def context_resume_stream_async(
             format=response.format,
             **params,
         )
+
+    @property
```
willbakst, line 1280: Can you explain this change please? Why not just add `async_client: AsyncProviderClientT`? We aren't even using this for any provider except Azure OpenAI, and yet nothing is breaking, so we aren't even enforcing this consistently.

If there is a benefit here worth making this change for, we should then enforce this pattern everywhere. Otherwise we should remove it and stick to the existing pattern.

python/mirascope/llm/clients/google/clients.py
```diff
@@ -103,6 +103,7 @@ def __init__(
             http_options = HttpOptions(base_url=base_url)
 
         self.client = Client(api_key=api_key, http_options=http_options)
+        self.async_client = self.client
```
willbakst, line 106: Why not just `self.client = self.asyn_client = Client(...)` since we are using the same object in memory?

python/mirascope/llm/clients/openai/completions/_utils/decode.py
```diff
@@ -35,6 +35,7 @@
 def decode_response(
     response: openai_types.ChatCompletion,
     model_id: OpenAICompletionsModelId,
+    provider: Literal["openai:completions", "azure-openai:completions"],
```
willbakst, line 38: Given this is a private utility inside of the `completions` module, it seems weird that we would then be using it for another provider and thus require a `provider` argument to handle that.

It looks like we're expecting this utility to work in a shared capacity across all possible completion clients. This makes me re-think how we're structuring things. I also think this is somewhat related to work we're expecting to do around unifying things into a `Provider` class.

My new thinking is that we may actually want to make the `clients` module `_clients` private inside of a `providers` module that would expose the providers to users (e.g. Azure OpenAI, OpenAI, Anthropic, etc) but hide the underlying SDK implementation under the hood.

This would also then downstream turn into something like `with llm.provider` rather than `with llm.client`.

@teamdandelion I'm curious what you think here.

python/mirascope/llm/clients/openai/completions/_utils/encode.py
```diff
@@ -141,12 +146,16 @@ def flush_message_content() -> None:
 
 
 def _encode_assistant_message(
-    message: AssistantMessage, model_id: OpenAICompletionsModelId, encode_thoughts: bool
+    message: AssistantMessage,
+    model_id: OpenAICompletionsModelId,
+    encode_thoughts: bool,
+    provider: Provider,
```
willbakst, line 152: I have the same question here

python/mirascope/llm/clients/openai/responses/_utils/encode.py
willbakst: To me it seems like this is a bug with Azure OpenAI? The solution presented is incorrect according to their docs, so a bug when following the correct usage indicates a bug with Azure, not an issue Mirascope needs to handle.

My feeling here is that we should assume based on their docs that we can use `AzureOpenAI` exactly the same way we use it for `OpenAI`, and if we run into issues we should consider skipping tests / not covering sections of code that are impacted by the bug, and we should post a bug report on their repo with a reproduction.

@teamdandelion What do you think here?

PR #1442, opened 2025-10-21
test: DRY repeated logic in snapshot tests

teamdandelion:

PR #1440, opened 2025-10-20
test: refactor all e2e tests to catch exceptions

teamdandelion:
Now all the e2e tests catch exceptions rather than failing if a provider
throws an error (excepting VCR over-write errors which the exception
serializer re-throws).

The intent of this change is it becomes possible to e.g. add a partially
implemented provider to the list of providers being tested, knowing that
it will throw exceptions on many tests, and then progressively
implementing those features. While still maintaining a green build every
step of the way.

PR #1443, opened 2025-10-21
feat: implement llm.Image

teamdandelion:


teamdandelion:
* **#1484** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1484?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1479** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1479?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1483** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1483?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1482** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1482?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1481** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1481?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1480** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1480?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1477** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1477?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1460** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1460?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1459** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1459?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1458** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1457** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1457?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1453** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1453?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1452** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1456** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1456?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1448** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1448?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1451** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1451?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1447** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1450** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1450?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1446** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1446?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1445** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1445?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1449** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1444** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1444?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1443** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1443?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1442** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1440** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1440?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+
+            # Infer mime type from content-type header or URL
+            mime_type: ImageMimeType
+            content_type = response.headers.get("content-type", "")
+            if content_type.startswith("image/"):
+                mime_type_str = content_type.split(";")[0]
+                if mime_type_str in MIME_TYPES:
+                    mime_type = mime_type_str
+                else:
+                    mime_type = "image/jpeg"
```
teamdandelion: This was vibe coded, Im not sure if filling in jpeg if unknown is actually a good choice. LMK if you have thoughts, otherwise Ill probably change this to raising an error for clarity...

koxudaxi: I agree with your concern.
Would raising a ValueError be appropriate when we cannot determine the type?
As a fallback here, would it make sense to detect it from the file’s magic number, like we did in v1?
Also, why did we stop inspecting the file content and switch to inferring the MIME type only?

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+                mime_type_map: dict[str, ImageMimeType] = {
+                    ".png": "image/png",
+                    ".jpg": "image/jpeg",
+                    ".jpeg": "image/jpeg",
+                    ".webp": "image/webp",
+                    ".gif": "image/gif",
+                    ".heic": "image/heic",
+                    ".heif": "image/heif",
+                }
+                mime_type = mime_type_map.get(ext, "image/jpeg")
```
koxudaxi: Same above.

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+            mime_type_map: dict[str, ImageMimeType] = {
+                ".png": "image/png",
+                ".jpg": "image/jpeg",
+                ".jpeg": "image/jpeg",
+                ".webp": "image/webp",
+                ".gif": "image/gif",
+                ".heic": "image/heic",
+                ".heif": "image/heif",
+            }
+            final_mime_type = mime_type_map.get(ext, "image/jpeg")
```
koxudaxi: Same above.

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+        else:
+            final_mime_type = mime_type
```
koxudaxi: I understand that only ImageMimeType is accepted, but is it acceptable to accept all values at the runtime level? Does this mean validation is left to Anthropic?

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+        if not path.exists():
+            raise FileNotFoundError(f"File not found: {file_path}")
+
+        with open(path, "rb") as f:
+            data = base64.b64encode(f.read()).decode("utf-8")
+
+        final_mime_type: ImageMimeType
+        if mime_type is None:
+            ext = path.suffix.lower()
+            mime_type_map: dict[str, ImageMimeType] = {
```
koxudaxi: Should mime_type_map be defined and used at the module level?

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
-        mime_type: ImageMimeType | None,
+        mime_type: ImageMimeType | None = None,
     ) -> "Image":
         """Create an `Image` from raw bytes."""
-        raise NotImplementedError
+        encoded_data = base64.b64encode(data).decode("utf-8")
+
+        final_mime_type: ImageMimeType
+        if mime_type is None:
+            # Check magic bytes to determine image format
```
koxudaxi: This can be turned into a function and used as a fallback for mine-type.

python/mirascope/llm/content/image.py
koxudaxi: Ah, I see. This is a general-purpose method not limited to Anthropic. In that case, it might be fine.

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
         download: bool = False,
     ) -> "Image":
         """Create an `Image` from a URL."""
-        raise NotImplementedError
+        if download:
+            response = httpx.get(url)
```
koxudaxi: httpx does not redirect by default, so you may want to add `follow_redirects=True`. Most common CDNs and links using pre-signed URLs will likely require redirects.

koxudaxi: Will an asynchronous version of the method be needed in the future?

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+        mime_type: ImageMimeType | None = None,
     ) -> "Image":
         """Create an `Image` from a file path."""
-        raise NotImplementedError
+        path = Path(file_path)
+        if not path.exists():
+            raise FileNotFoundError(f"File not found: {file_path}")
+
+        with open(path, "rb") as f:
+            data = base64.b64encode(f.read()).decode("utf-8")
```
koxudaxi: This process doesn't use streams or similar mechanisms, so it would be better to implement size checks in the future. Using it in production carries significant risk.

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+        encoded_data = base64.b64encode(data).decode("utf-8")
+
+        final_mime_type: ImageMimeType
+        if mime_type is None:
+            # Check magic bytes to determine image format
+            if data.startswith(b"\x89PNG\r\n\x1a\n"):
+                final_mime_type = "image/png"
+            elif data.startswith(b"\xff\xd8\xff"):
+                final_mime_type = "image/jpeg"
+            elif data.startswith(b"RIFF") and data[8:12] == b"WEBP":
```
koxudaxi: If you provide data that's too small, you'll get an IndexError. You might want to check the data size.

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+                final_mime_type = "image/webp"
+            elif data.startswith(b"GIF87a") or data.startswith(b"GIF89a"):
+                final_mime_type = "image/gif"
+            elif len(data) >= 12 and data[4:12] in (b"ftypheic", b"ftypheix"):
+                final_mime_type = "image/heic"
+            elif len(data) >= 12 and data[4:12] in (b"ftypheif", b"ftypmif1"):
+                final_mime_type = "image/heif"
+            else:
+                # Default to jpeg if format cannot be determined
+                final_mime_type = "image/jpeg"
```
koxudaxi: It feels very strange that files are processed as JPEGs even though they've already been identified as non-JPEGs.

python/tests/llm/content/test_image_content.py
```diff
@@ -0,0 +1,431 @@
...
+    def test_from_url_without_download(self) -> None:
+        """Test creating an Image from URL without downloading."""
+        url = "https://example.com/image.png"
+        image = Image.from_url(url, download=False)
+
+        assert isinstance(image.source, URLImageSource)
+        assert image.source.url == url
+        assert image.source.type == "url_image_source"
+
+    @patch("mirascope.llm.content.image.httpx.get")
```
koxudaxi: It might be better to mock the server side rather than the client.
https://github.com/Mirascope/lilypad/blob/a2c0112f91c4ad190a74dc86bbf4a59c7b503c7c/sdks/python/tests/lilypad/_internal/otel/http/conftest.py#L59-L63

python/tests/llm/content/test_image_content.py
```diff
@@ -0,0 +1,431 @@
...
+        mock_response.headers = {}
+        mock_get.return_value = mock_response
+
+        image = Image.from_url("https://example.com/image.heif", download=True)
+
+        assert isinstance(image.source, Base64ImageSource)
+        assert image.source.mime_type == "image/heif"
+
+
+class TestImageFromFile:
```
koxudaxi, line 193: There's some code duplication here. You might consider using a context manager, fixtures, parameterization, or some other approach.

python/mirascope/llm/content/image.py
willbakst: Why aren't we inferring it from the data? In `v1` we have a fairly accurate implementation of inferring the data type from the data. In fact, I would consider just ripping that implementation here and updating as you see fit.

python/mirascope/llm/content/image.py
```diff
@@ -1,7 +1,12 @@
 """The `Image` content class."""
 
+import base64
 from dataclasses import dataclass
-from typing import Literal
+from pathlib import Path
+from typing import Literal, get_args
+from urllib.parse import urlparse
+
+import httpx
```
willbakst, line 8: Should we use `requests`?

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
         file_path: str,
         *,
-        mime_type: ImageMimeType | None,
+        mime_type: ImageMimeType | None = None,
     ) -> "Image":
         """Create an `Image` from a file path."""
-        raise NotImplementedError
+        path = Path(file_path)
+        if not path.exists():
+            raise FileNotFoundError(f"File not found: {file_path}")
```
willbakst: Do we need to do this manually? Or would attempting to open a file at an unknown path already do this by default and raise the not found error?

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
+
+        with open(path, "rb") as f:
+            data = base64.b64encode(f.read()).decode("utf-8")
+
+        final_mime_type: ImageMimeType
```
willbakst: No need for `final_mime_type`, just overwrite `mime_type` in the if and drop the else?

python/mirascope/llm/content/image.py
willbakst: Another question here is if we should use `PIL`, which already has a map I believe? Definitely think having a map constant given the duplication is likely a good idea

python/mirascope/llm/content/image.py
```diff
@@ -57,24 +64,117 @@ def from_url(
...
         data: bytes,
         *,
-        mime_type: ImageMimeType | None,
+        mime_type: ImageMimeType | None = None,
     ) -> "Image":
         """Create an `Image` from raw bytes."""
-        raise NotImplementedError
+        encoded_data = base64.b64encode(data).decode("utf-8")
+
+        final_mime_type: ImageMimeType
```
willbakst: Same comment here

python/mirascope/llm/content/image.py
willbakst: Agreed. This is inline with my comment above.

I would check out the implementation in `v1` if you havent, which I believe is quite robust.

python/mirascope/llm/content/image.py
willbakst: This is an interesting question. For any real image, I doubt this could ever happen. To be robust, likely worth checking, but I'm not sure anyone would ever actually run into such an issue?

python/mirascope/llm/content/image.py
willbakst: I think we should error if we can't determine the file type. Otherwise, how are we supposed to handle it?

python/mirascope/llm/content/image.py
teamdandelion: Done, using the infer_image_type helper instead.

python/mirascope/llm/content/image.py
teamdandelion: Switching to using `infer_image_type` rather than the file extension at all.

python/tests/llm/content/test_image_content.py
teamdandelion: Done, thanks!

python/mirascope/llm/content/image.py
teamdandelion: Done to both (used follow_redirects, added download_async)

python/mirascope/llm/content/image.py
teamdandelion: Added size cehcks.

python/mirascope/llm/content/image.py
teamdandelion: httpx is used in mcp, so I think it makes sense to align. plus per claude it's the more modern choice

python/mirascope/llm/content/image.py
teamdandelion: Added a check

python/mirascope/llm/content/image.py
```diff
@@ -50,31 +92,145 @@ class Image:
...
     @classmethod
-    def from_file(
-        cls,
-        file_path: str,
-        *,
-        mime_type: ImageMimeType | None,
-    ) -> "Image":
-        """Create an `Image` from a file path."""
-        raise NotImplementedError
+    def download(cls, url: str, *, max_size: int = MAX_IMAGE_SIZE) -> "Image":
```
teamdandelion, line 135: I changed the interface so that `from_url` only creates url reference images, and `download` and `download_async` are separate options.

Two motivations for this:
1. Per @koxudaxi's comment we should have an async option, but the sync/async distinction is meaningless when constructing a url reference
2. Since Google does not support url reference images, and `download=False` by default, it created an awkward situation where using the ltionefault path for loading an external image was broken for Google. Having the `download` as an optional flag is misleading when it creates compatibility issues downstream. So I think having download as a method clears this up.

I'll ensure audio is consistent when I get to it

python/mirascope/llm/content/image.py
koxudaxi: I checked locally, and httpx isn't included in the dependencies, right? Is this a problem?
It looks like gemini, anthropic, and openai have httpx as a dependency, but since this file is meant to be generic, should we either explicitly include httpx or import it with a try-except? I haven't done a comprehensive dependency investigation, so it's possible that httpx is always implicitly installed.

python/mirascope/llm/content/image.py
```diff
@@ -50,31 +92,145 @@ class Image:
...
+            max_size: Maximum allowed image size in bytes (default: 20MB)
+
+        Returns:
+            An `Image` with a `Base64ImageSource`
+
+        Raises:
+            ValueError: If the downloaded image exceeds max_size
+        """
+        async with httpx.AsyncClient() as client:
+            response = await client.get(url, follow_redirects=True)
```
willbakst, line 169: Once you receive the response, it seems all the code below is duplicated with download above? Is it worth refactoring to a shared method?

python/mirascope/llm/content/image.py
```diff
@@ -50,31 +92,145 @@ class Image:
...
+        """Create an `Image` from raw bytes.
+
+        Args:
+            data: Raw image bytes
+            max_size: Maximum allowed image size in bytes (default: 20MB)
+
+        Raises:
+            ValueError: If the data size exceeds max_size
+        """
+        data_size = len(data)
```
willbakst: Same thing here? Seems like once the file in `from_file` is read into memory as bytes, it's the same as this method?

python/mirascope/llm/content/image.py
teamdandelion: Good catch, thanks. Fixed. This is an interesting example of LLM assisted sloppiness, as I would not have sent something this wet out if I were typing the characters myself.

willbakst (approved):

teamdandelion:
### Merge activity

* **Oct 24, 5:31 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1443).
* **Oct 24, 5:31 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/mirascope/1443).

PR #1467, opened 2025-10-28
Add unittest for GrokClient

koxudaxi:

PR #1466, opened 2025-10-27
Implement the GrokClient

koxudaxi:
### TL;DR

Add support for xAI's Grok models to Mirascope.

### What changed?

- Added a new `GrokClient` implementation for xAI's Grok models
- Implemented message conversion utilities between Mirascope and xAI SDK formats
- Added Grok model IDs including `grok-3`, `grok-2-vision`, `grok-2-latest`, and `grok-2`
- Integrated the client with Mirascope's provider system
- Added xAI SDK as an optional dependency in pyproject.toml

### How to test?

1. Install Mirascope with the xAI extra: `pip install "mirascope[xai]"`
2. Set your xAI API key: `export XAI_API_KEY=your_api_key`
3. Use the client in your code:
```python
from mirascope.llm import GrokClient, GrokModelId
from mirascope.llm.messages import UserMessage

client = GrokClient()
response = client.call(
    model_id=GrokModelId.GROK_3,
    messages=[UserMessage("Hello, how are you?")]
)
print(response.content)
```

### Why make this change?

This change expands Mirascope's model support to include xAI's Grok models, giving users more options for LLM providers. Grok models have unique capabilities and this integration allows Mirascope users to leverage them within the same familiar interface used for other providers.

PR #1464, opened 2025-10-24
Implement mirrored provider clients (AnthropicVertex)

koxudaxi:

PR #1461, opened 2025-10-23
Clean up VCRConfig by replacing filter_headers with sanitization hook

koxudaxi:
### TL;DR

Improved VCR test security by replacing header filtering with a more robust sanitization approach.

### What changed?

- Replaced the `filter_headers` and `filter_post_data_parameters` configuration with a `before_record_request` hook
- Added a new `sanitize_request` function that creates a deep copy of the request and replaces sensitive header values with `<filtered>`
- Moved sensitive header names to a dedicated `SENSITIVE_HEADERS` constant
- Updated documentation to explain the new sanitization approach

### How to test?

1. Run the e2e tests with VCR recording enabled
2. Verify that the cassette files don't contain any sensitive API keys or tokens
3. Confirm that the tests still pass when replaying the cassettes

### Why make this change?

The previous approach simply removed sensitive headers from recordings, which could cause issues during playback. The new approach preserves the header structure while replacing only the sensitive values with placeholders. This ensures that:

1. Real HTTP requests are sent with valid authentication
2. No sensitive credentials are stored in cassette files
3. Request matching during playback is more reliable since the header structure is preserved

PR #1468, opened 2025-10-28
Support concurrent tool call streaming with ID-based tracking

koxudaxi:
### TL;DR

Added support for parallel tool calls in streaming responses by tracking tool calls by ID.

### What changed?

- Added an `id` field to `ToolCallChunk` and `ToolCallEndChunk` classes to uniquely identify tool calls
- Modified the stream response handler to track tool calls by ID in a dictionary (`_current_tool_calls_by_id`)
- Updated the tool call handling logic to support processing multiple tool calls in parallel
- Updated error messages to be more specific about which tool call ID is causing issues
- Updated all test cases to include the required ID field in tool call chunks

### How to test?

- Run the existing test suite to verify that all tool call functionality works correctly
- Test scenarios with multiple concurrent tool calls to ensure they are properly tracked and processed
- Verify error handling by testing with unknown tool call IDs

### Why make this change?

This change enables proper handling of parallel tool calls in streaming responses. By tracking tool calls by their unique IDs rather than using a single `_current_content` reference, the system can now process multiple tool calls simultaneously without losing track of their state. This is particularly important for complex interactions where multiple tools might be called in parallel by the LLM.

### DIFF
--- a/python/mirascope/llm/content/tool_call.py
+++ b/python/mirascope/llm/content/tool_call.py
@@ -47,7 +47,9 @@ class ToolCallChunk:
     type: Literal["tool_call_chunk"] = "tool_call_chunk"
 
     content_type: Literal["tool_call"] = "tool_call"
-    """The type of content reconstructed by this chunk."""
+
+    id: str
+    """The unique identifier for the tool call this chunk belongs to."""
 
     delta: str
     """The incremental json args added in this chunk."""
@@ -60,4 +62,6 @@ class ToolCallEndChunk:
     type: Literal["tool_call_end_chunk"] = "tool_call_end_chunk"
 
     content_type: Literal["tool_call"] = "tool_call"
-    """The type of content reconstructed by this chunk."""
+
+    id: str
+    """The unique identifier for the tool call being ended."""
--- a/python/mirascope/llm/clients/anthropic/_utils/decode.py
+++ b/python/mirascope/llm/clients/anthropic/_utils/decode.py
@@ -159,7 +159,9 @@ class _AnthropicChunkProcessor:
                         f"Received input_json_delta for {self.current_block_param['type']} block"
                     )
                 self.accumulated_tool_json += delta.partial_json
-                yield ToolCallChunk(delta=delta.partial_json)
+                yield ToolCallChunk(
+                    id=self.current_block_param["id"], delta=delta.partial_json
+                )
             elif delta.type == "thinking_delta":
                 if self.current_block_param["type"] != "thinking":  # pragma: no cover
                     raise RuntimeError(
@@ -196,7 +198,7 @@ class _AnthropicChunkProcessor:
                     if self.accumulated_tool_json
                     else {}
                 )
-                yield ToolCallEndChunk()
+                yield ToolCallEndChunk(id=self.current_block_param["id"])
             elif block_type == "thinking":
                 yield ThoughtEndChunk()
             else:
--- a/python/mirascope/llm/clients/google/_utils/decode.py
+++ b/python/mirascope/llm/clients/google/_utils/decode.py
@@ -128,6 +128,7 @@ class _GoogleChunkProcessor:
 
     def __init__(self) -> None:
         self.current_content_type: Literal["text", "tool_call", "thought"] | None = None
+        self.current_tool_id: str | None = None
         self.accumulated_parts: list[genai_types.Part] = []
         self.reconstructed_content = genai_types.Content(parts=[])
 
@@ -150,11 +151,13 @@ class _GoogleChunkProcessor:
                 yield TextEndChunk()  # pragma: no cover
                 self.current_content_type = None  # pragma: no cover
             elif self.current_content_type == "tool_call" and not part.function_call:
-                # In testing, Gemini never emits tool calls and text in the same message
-                # (even when specifically asked in system and user prompt), so
-                # the following code is uncovered but included for completeness
-                yield ToolCallEndChunk()  # pragma: no cover
+                if self.current_tool_id is None:
+                    raise RuntimeError(
+                        "Missing tool_id when ending tool call"
+                    )  # pragma: no cover
+                yield ToolCallEndChunk(id=self.current_tool_id)  # pragma: no cover
                 self.current_content_type = None  # pragma: no cover
+                self.current_tool_id = None  # pragma: no cover
 
             if part.thought:
                 if self.current_content_type is None:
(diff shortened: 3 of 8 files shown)


koxudaxi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1468?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1467** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1466** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1466?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1469** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1474** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1468** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1468?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1465** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1465?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1485](https://github.com/Mirascope/mirascope/pull/1485) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1485?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1464** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1464?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1455** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1454** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1473** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1473?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1470** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1471** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1472** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1478** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1478?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1439** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1439?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1475** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1475?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1476](https://github.com/Mirascope/mirascope/pull/1476) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1476?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1441** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1461** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1461?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2095** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python/mirascope/llm/content/tool_call.py
```diff
@@ -47,7 +47,9 @@ class ToolCallChunk:
     type: Literal["tool_call_chunk"] = "tool_call_chunk"
 
     content_type: Literal["tool_call"] = "tool_call"
-    """The type of content reconstructed by this chunk."""
+
```
teamdandelion, line 50: The reason we don't have id on `ToolCallChunk` or `ToolCallEndChunk` is because (in my observation/understanding) the providers always stream tool calls one at a time, so we can put the id on the `ToolCallStartChunk`, and then safely infer that every subsequent `ToolCallChunk` and `ToolCallEndChunk` is part of the tool call with the same id. This matches with my understanding of LLMs as fundamentally serial generators that produce output one-token-at-a-time (ie. when tool calls are parallel from being included in the same response, they were still generated serially).

Are you seeing a provider that streams tool call chunks interleaved? If so which one? 

(Aside, I'm guessing these PR descriptions are all auto-generated by Graphite? It has that smooth AI vibe that feels like its coming up with a plausible explanation based on the code change, but doesn't truly understand why the change is being made. It would be nice if the descriptions had some more of the specific human intent in them)

koxudaxi: Thanks for the question. Grok streams tool calls in parallel. We often see call_a begin, then call_b begin, then more deltas from both IDs mixed together. We rely on the id on every `ToolCallChunk` and `ToolCallEndChunk` to keep each call stitched to the right arguments. If we drop the id, the chunks from different calls become ambiguous.

koxudaxi: https://docs.x.ai/docs/guides/tools/overview#real-time-server-side-tool-calls

teamdandelion: Ok, that is quite interesting. Having interleaved chunks from distinct pieces of content will sadly break downstream assumptions we've made in how we process stream responses - particularly the `StreamResponse.streams` interface ([see impl]) which assumes that every piece of content fully finishes before the next one comes.

Given that it is only grok doing this, I suggest that rather than changing the core interface to support simultaneously streamed tool calls, we keep the general mirascope interface as-is, and have the Grok stream chunk processor accumulate tool call chunks, yielding chunks for the first tool call as they come in, and then once that first tool call finishes emitting all of the accumulated chunks for the second tool call, and so forth. (And defer improving for "truly simultaneous" tool calls until at least one other provider has that feature as well). 

cc @willbakst for thoughts


[see impl]: https://github.com/Mirascope/mirascope/blob/df9d1218dba255688d77d99d50a53362225aa261/python/mirascope/llm/responses/base_stream_response.py#L364-L437

PR #1520, opened 2025-11-26
chore: remove implicit unknown types in example code

teamdandelion:
One of the necessary changes to allow us to disable implicit unknown
parameter and return types. Also, catches a bug in the examples in the
process (where we have a list of llm.messages but actually one of the
returned content pieces is a bare string, not a message)

### DIFF
--- a/python/examples/intro/context/decorator.py
+++ b/python/examples/intro/context/decorator.py
@@ -14,13 +14,13 @@ def get_book_info(ctx: llm.Context[Library], book: str) -> str:
 
 @llm.call(provider="openai", model_id="gpt-5", tools=[get_book_info])
 # [!code highlight:2]
-def librarian(ctx: llm.Context[Library], query: str):
+def librarian(ctx: llm.Context[Library], query: str) -> list[llm.Message]:
     book_list = "\n".join(ctx.deps.available_books)
     return [
         llm.messages.system(
             f"You are a librarian, with access to these books: ${book_list}"
         ),
-        query,
+        llm.messages.user(query),
     ]
 
 
--- a/python/examples/intro/system_messages/decorator.py
+++ b/python/examples/intro/system_messages/decorator.py
@@ -2,7 +2,7 @@ from mirascope import llm
 
 
 @llm.call(provider="openai", model_id="gpt-5")
-def recommend_book(genre: str):
+def recommend_book(genre: str) -> list[llm.Message]:
     return [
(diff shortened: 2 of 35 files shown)


teamdandelion:
* **#1520** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1520?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1520?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/examples/intro/context/decorator.py
```diff
@@ -14,13 +14,13 @@ def get_book_info(ctx: llm.Context[Library], book: str) -> str:
 
 @llm.call(provider="openai", model_id="gpt-5", tools=[get_book_info])
 # [!code highlight:2]
-def librarian(ctx: llm.Context[Library], query: str):
+def librarian(ctx: llm.Context[Library], query: str) -> list[llm.Message]:
```
willbakst, line 17: I thought we had decided we were not including return type annotations on the functions because it was confusing? I know that stricter linting requires return type annotations, but maybe we can exclude that in examples? Or does not having the return type annotations cause other issues?

teamdandelion: So, the interesting thing here is that this example was actually incorrect as written. A prompt must return either `UserContent | list[Message]`​ and `UserContent`​ is `str | UserContentPart | Sequence[str | UserContentPart]`​. However, this "prompt" was returning `list[str | Message]`​ which is invalid.  
  
Because we didn't specify a return type, it evaluated as `list[Unknown]`​ which could match, hence, no type error.  
  
Given our emphasis on type safety, maybe we should give the return type annotations even though it could be a bit confusing, because it is more correct? (And matches the implementation in any case).   
  
An alternative (probably also worth pursuing) is to enable strict list inference (which would catch this type error). We could enable strict list inference and disable requiring all the strict typing for example code.

willbakst: Something that's weird is that I'm getting `list[Unknown]`​ even if I just do `x = [llm.messages.system(""), llm.messages.user("")]`​. Technically it should be able to dynamically determine the return type based on what is returned, so I think the issue lies here in how the type of the list is incorrect, not the return annotation.

PR #1564, opened 2025-12-04
chore: remove legacy e2e tests of call decorator

teamdandelion:
This behavior is covered by the e2e override tests.

PR #1565, opened 2025-12-04
refactor: `PromptFunction` -> `Prompt` -> `Call`

teamdandelion:
- A `Prompt` is a callable object (like `Call`), but without a bundled
  model. So you always call it with `model` as the first argument.
- The former `Promptable` protocol is renamed to `PromptFunction` for
  clarity. So a `PromptFunction` may be decorated to create a `Prompt`
  (or a `Call`)
- `Call` now wraps `Prompt` under the hood
- Removed all mentions of prompt templates, can add those back in
  if/when we implement them.
- Rewritten docstrings (by Claude)


teamdandelion:
* **#1566** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1565** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1564** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1564?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/prompts/decorator.py
```diff
@@ -1,289 +1,208 @@
...
-        raise NotImplementedError()
+            tools = cast(Sequence[Tool] | None, self.tools)
+            return Prompt(fn=fn, toolkit=Toolkit(tools=tools), format=self.format)
 
 
 @overload
-def prompt(
-    __fn: ContextPromptable[P, DepsT],
-) -> ContextPrompt[P, DepsT]:
+def prompt(  # pyright: ignore[reportOverlappingOverload]
```
teamdandelion, line 136: Note: We ignore the overlapping overload the same as we do with the tool decorator, to work around the fact that a ContextPromptFunction is also a PromptFunction.

python/mirascope/llm/prompts/decorator.py
```diff
@@ -1,289 +1,208 @@
...
-        raise NotImplementedError()
-    decorator = PromptDecorator()
+        A `Prompt` variant (Prompt, AsyncPrompt, ContextPrompt, or AsyncContextPrompt)
+        or a `PromptDecorator` if called with arguments
+    """
+    decorator = PromptDecorator(tools=tools, format=format)
     if __fn is None:
         return decorator
-    return decorator(__fn)
+    return decorator(__fn)  # pyright: ignore[reportUnknownVariableType, reportCallIssue]
```
teamdandelion, line 215: These pyright ignores address a Tool variance issue, see below. I tried making an InvariantToolT but that didn't work either. Maybe we can pair on this and see if we can get rid of it?  
  
Object of type "PromptDecorator[ToolT@prompt, FormattableT@prompt]" is not callable
  Could not bind method "**call**" because "PromptDecorator[ToolT@prompt, FormattableT@prompt]" is not assignable to parameter "self"
    "PromptDecorator[ToolT@prompt, FormattableT@prompt]" is not assignable to "PromptDecorator[AsyncTool[..., Jsonable] | AsyncContextTool[DepsT@**call**, Jsonable, ...], FormattableT@prompt]"
      Type parameter "ToolT@PromptDecorator" is covariant, but "ToolT@prompt" is not a subtype of "AsyncTool[..., Jsonable] | AsyncContextTool[DepsT@**call**, Jsonable, ...]"
        Type "Tool[..., Jsonable]* | AsyncTool[..., Jsonable]* | ContextTool[Any, Jsonable, ...]* | AsyncContextTool[Any, Jsonable, ...]_" is not assignable to type "AsyncTool[..., Jsonable] | AsyncContextTool[DepsT@**call**, Jsonable, ...]"_
          _Type "ContextTool[Any, Jsonable, ...]_" is not assignable to type "AsyncTool[..., Jsonable] | AsyncContextTool[DepsT@**call**, Jsonable, ...]"
  Could not bind method "**call**" because "PromptDecorator[ToolT@prompt, FormattableT@prompt]" is not assignable to parameter "self"
    "PromptDecorator[ToolT@prompt, FormattableT@prompt]" is not assignable to "PromptDecorator[Tool[..., Jsonable] | ContextTool[DepsT@**call**, Jsonable, ...], FormattableT@prompt]"
      Type parameter "ToolT@PromptDecorator" is covariant, but "ToolT@prompt" is not a subtype of "Tool[..., Jsonable] | ContextTool[DepsT@**call**, Jsonable, ...]"

python/mirascope/llm/prompts/prompts.py
```diff
@@ -0,0 +1,424 @@
...
+        model: Model,
+        *args: P.args,
+        **kwargs: P.kwargs,
+    ) -> AsyncResponse[FormattableT]: ...
+
+    async def __call__(
+        self, model: Model, *args: P.args, **kwargs: P.kwargs
+    ) -> AsyncResponse | AsyncResponse[FormattableT]:
+        """Generates a response using the provided model asynchronously."""
+        return await self.call(model, *args, **kwargs)  # pragma: no cover
```
teamdandelion: I didn't see a need to have unit test coverage for the \__call__ redirect, hence the no covers.

python/mirascope/llm/prompts/protocols.py
```diff
@@ -1,73 +1,44 @@
...
 
-PromptT = TypeVar(
-    "PromptT",
-    bound="Prompt[...] | AsyncPrompt[...] | ContextPrompt[..., Any] | AsyncContextPrompt[..., Any]",
-)
-"""Type variable for prompt types.
 
-This type var represents a resolved prompt, i.e. one that returns a Sequence of messages.
-"""
+class PromptFunction(Protocol[P]):
```
teamdandelion: Now that a `Prompt` is not a function, but a callable wrapper around said function, I like the name `PromptFunction` a lot better than `Promptable`

python/mirascope/llm/prompts/_utils.py
```diff
@@ -13,18 +13,16 @@
...
+    AsyncPromptFunction,
+    ContextPromptFunction,
+    PromptFunction,
 )
 
 
 def is_messages(
     messages_or_content: Sequence[Message] | UserContent,
 ) -> TypeIs[Sequence[Message]]:
-    if not messages_or_content:
```
teamdandelion, line 26: Note: I deleted this to solve a coverage issue because I wanted to separate E2E testing of the empty content cases for a separate PR. The behavior is updated (with tests) in the next PR up the stack.

python/mirascope/llm/calls/calls.py
```diff
@@ -21,19 +22,37 @@
...
+class BaseCall:
+    """Base class for all Call types with shared model functionality."""
+
+    default_model: Model
+    """The default model that will be used if no model is set in context."""
+
+    @property
+    def model(self) -> Model:
+        """The model used for generating responses. May be overwritten via `with llm.model(...)`."""
+        return get_model_from_context() or self.default_model
```
willbakst: Shouldn't we just use `use_model(self.default_model)` here?

python/mirascope/llm/prompts/decorator.py
```diff
@@ -1,289 +1,208 @@
...
         self,
-        fn: ContextPromptable[P, DepsT],
-    ) -> ContextPrompt[P, DepsT]:
-        """Decorator for creating context prompts."""
-        ...
+        tools: Sequence[ToolT] | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+    ) -> None:
+        """Initialize the decorator with optional tools and format."""
+        self.tools = tools
```
willbakst, line 58: nit: add these with class level typed definitions

python/mirascope/llm/prompts/decorator.py
```diff
@@ -1,289 +1,208 @@
...
     ...
 
 
 @overload
 def prompt(
     *,
-    template: None = None,
-) -> PromptDecorator:
-    """Create a decorator for Prompt functions (no template)"""
+    tools: list[ToolT] | None = None,
```
willbakst: Why `list` over `sequence`?

python/mirascope/llm/calls/decorator.py
```diff
@@ -36,91 +39,108 @@
...
+    The decorator automatically detects whether the function is async or context-aware
+    and creates the appropriate `Call` variant (`Call`, `AsyncCall`, `ContextCall`, or `AsyncContextCall`).
+
+    Conceptually: `CallDecorator` = `PromptDecorator` + `Model`
+    Result: `Call` = `PromptFunction` + tools + format + `Model`
+    """
 
     model: Model
-    tools: Sequence[ToolT] | None
+    tools: list[ToolT] | None
```
willbakst: Same question here, why restrict from `Sequence` to `list` ?

python/mirascope/llm/prompts/prompts.py
willbakst: Hmm, we could always use `__call__` instead of `.call` in our tests and still get 1:1 execution but 100% coverage here without coverage exclusion?

python/mirascope/llm/prompts/prompts.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def __call__(
+        self,
+        model: Model,
+        ctx: Context[DepsT],
+        *args: P.args,
+        **kwargs: P.kwargs,
+    ) -> ContextResponse[DepsT, None] | ContextResponse[DepsT, FormattableT]:
+        """Generates a response using the provided model."""
+        return self.call(model, ctx, *args, **kwargs)  # pragma: no cover
```
willbakst: Same comment here and elsewhere

python/mirascope/llm/prompts/protocols.py
willbakst: I'm not loving `PromptFunction` (and its variants) as the naming choice here. It's really more of a function that returns something that can become `Sequence[Message]`, which is how we landed on "promptable" as the naming convention.

I'm all for landing on something more clear than "promptable", but I don't think it's "prompt function". What do you think about `MessagesTemplate`? I feel like "template" covers the fact that it takes template variables and produces a formatted `Sequence[Message]`. I think this will also work down the line with a `messages_template` argument for using the templating language rather than returning messages (using `...` for the function body).

python/mirascope/llm/prompts/decorator.py
willbakst: Yeah happy to pair on this to resolve

python/mirascope/llm/prompts/prompts.py
teamdandelion: Actually, we do use `__call__` in our tests, but its `Call.__call__` which calls `Call.call` which calls `Prompt.call`, and we don't have e2e tests that use the `Prompt` directly.   
  
If we want to remove the no cover here we could whip up a mock based test to put in `test_prompt_decorator` that basically just mocks `Prompt.call` but it feels kind of like duplicating the code under test for no real safety benefit. Or we could write some e2e tests that use the `Prompt` directly which would be more meaningful.

python/mirascope/llm/prompts/decorator.py
teamdandelion: force of habit... good catch

python/mirascope/llm/calls/calls.py
teamdandelion: Great catch

python/mirascope/llm/prompts/prompts.py
teamdandelion: Ok, I added e2e tests for testing prompt usage directly, and I made them use the same snapshots and cassettes as the corresponding call tests under the hood so we don't bloat

python/mirascope/llm/prompts/decorator.py
teamdandelion: Discussed offline, we're going to let it stand  as is

python/mirascope/llm/prompts/protocols.py
teamdandelion: Renamed to MessageTemplate

python/mirascope/llm/prompts/decorator.py
```diff
@@ -1,289 +1,212 @@
...
+    tools: Sequence[ToolT] | None
+    format: type[FormattableT] | Format[FormattableT] | None
+
+    def __init__(
         self,
-        fn: ContextPromptable[P, DepsT],
-    ) -> ContextPrompt[P, DepsT]:
-        """Decorator for creating context prompts."""
-        ...
+        tools: Sequence[ToolT] | None = None,
```
willbakst, line 54: nit: add docstrings for each field for consistency with the rest of the codebase

teamdandelion:
### Merge activity

* **Dec 5, 12:16 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1565).
* **Dec 5, 12:16 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1565).

PR #1566, opened 2025-12-04
feat: add e2e tests for empty prompt content

teamdandelion:
- If the prompt is an empty string, pass it thru to the LLM except if
  the provider does not support it (Anthropic)
- If the prompt is an empty array, raise ValueError as no providers
  support it and we can't actually tell if it is meant to be 0 messages
  or a message with 0 content parts

We discussed raising value error on empty string (which was the prior
behavior) but I believe this is better because:
- I don't think we should special case the empty array unless it is
  logically necessary
- 2 of the 3 supported providers respond to empty string
- their responses are interesting: openai says something like "it seems
  like this is an empty message, how can I help you?" while gemini
  hallucinates a response

I could imagine a researcher wanting to experiment with these empty
prompts to better understand model behavior and if so, our abstraction
should not obstruct them ;)

### DIFF
--- a/python/mirascope/llm/clients/anthropic/_utils/encode.py
+++ b/python/mirascope/llm/clients/anthropic/_utils/encode.py
@@ -58,8 +58,6 @@ def _encode_content(
 ) -> str | Sequence[anthropic_types.ContentBlockParam]:
     """Convert mirascope content to Anthropic content format."""
 
-    non_empty = False
-
     if len(content) == 1 and content[0].type == "text":
         if not content[0].text:
             raise FeatureNotSupportedError(
@@ -74,8 +72,9 @@ def _encode_content(
     for part in content:
         if part.type == "text":
             if part.text:
-                non_empty = True
-            blocks.append(anthropic_types.TextBlockParam(type="text", text=part.text))
+                blocks.append(
+                    anthropic_types.TextBlockParam(type="text", text=part.text)
+                )
         elif part.type == "image":
             source: (
                 anthropic_types.Base64ImageSourceParam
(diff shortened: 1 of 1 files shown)


teamdandelion:
* **#1566** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1566?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1565** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1565?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1564** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1564?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/clients/anthropic/_utils/encode.py
```diff
@@ -116,6 +132,13 @@ def _encode_content(
         else:
             raise NotImplementedError(f"Unsupported content type: {part.type}")
 
+    if not non_empty:
```
willbakst: Can't we just do `if not blocks: ...` here?

teamdandelion: Good catch

teamdandelion:
### Merge activity

* **Dec 5, 12:19 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1566).

PR #1563, opened 2025-12-04
Add VersionedCall variants for @ops.version + @llm.call integration

koxudaxi:
### TL;DR

Added support for versioning LLM calls by extending the `@ops.version` decorator to work with `@llm.call` decorated functions.

### What changed?

- Created new `VersionedCall` classes to wrap LLM call objects with versioning capabilities:
  - `VersionedCall` for synchronous calls
  - `VersionedAsyncCall` for asynchronous calls
  - `VersionedContextCall` for context-based calls
  - `VersionedAsyncContextCall` for asynchronous context-based calls
- Moved the `VersionDecorator` and `version` function to a dedicated `version_decorator.py` file
- Enhanced the version decorator to detect and properly wrap LLM call objects
- Updated imports and exports in `__init__.py` to expose the new classes

### How to test?

```python
# Test versioning a synchronous LLM call
@ops.version(tags=["v1"])
@llm.call(provider="openai", model_id="gpt-4o-mini")
def recommend_book(genre: str):
    return f"Recommend a {genre} book"

# Direct call returns Response
response = recommend_book("fantasy")
print(response.content)

# Get versioned result with metadata
result = recommend_book.wrapped("fantasy")
print(result.result.content)
print(result.function_uuid)

# Test streaming
stream = recommend_book.stream("fantasy")
for chunk in stream:
    print(chunk)

# Test with async calls
@ops.version(tags=["v1"])
@llm.call(provider="openai", model_id="gpt-4o-mini")
async def async_recommend(genre: str):
    return f"Recommend a {genre} book"

response = await async_recommend("mystery")
```

### Why make this change?

This change extends the versioning capabilities to LLM calls, allowing users to track and manage different versions of their LLM prompts. This is particularly valuable for:

1. Tracking prompt evolution over time
2. A/B testing different prompt versions
3. Ensuring reproducibility of LLM-based applications
4. Providing audit trails for LLM interactions

The implementation maintains a consistent API between regular function versioning and LLM call versioning, making it intuitive for users already familiar with the `@ops.version` decorator.

PR #1562, opened 2025-12-04
Add version_info property and X.Y version calculation

koxudaxi:

PR #1561, opened 2025-12-04
Add VersionInfo dataclass to ops module

koxudaxi:
### TL;DR

Added a new `VersionInfo` dataclass to store and expose version metadata for versioned functions.

### What changed?

- Created a new `VersionInfo` dataclass in `versioning.py` that contains all information needed to identify and describe a specific version of a function
- Added the class to the appropriate import/export statements in `__init__.py`
- Added unit tests to verify the dataclass can be instantiated with various field combinations

### How to test?

Run the new unit tests that verify the `VersionInfo` dataclass:
```bash
pytest python/tests/ops/test_versioning.py::test_version_info_instantiation python/tests/ops/test_versioning.py::test_version_info_with_none_values
```

### Why make this change?

This change provides a structured way to access and store version metadata for functions. The `VersionInfo` class encapsulates all the version-related information (UUID, hashes, version numbers, tags, etc.) in a single, well-defined object, making it easier to work with versioned functions and their metadata throughout the codebase.

PR #1560, opened 2025-12-04
Add signature_hash and implementation_hash fields to Closure

koxudaxi:

PR #1572, opened 2025-12-05
Implement serialization / deserialization

koxudaxi:


koxudaxi:
* **#1572** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1572?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1569** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1569?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1563** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1563?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1562** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1561** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1560** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1559** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1557** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1552** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1552?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1531** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1531?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1514** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1514?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1513** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1513?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1506** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1506?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1505** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1504** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1504?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1503** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1503?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1551** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1551?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1496** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/responses/response.py
```diff
@@ -56,6 +56,51 @@ def __init__(
             finish_reason=finish_reason,
         )
 
+    @classmethod
+    def decode(cls, data: bytes) -> "Response[None]":
+        """Deserialize a Response from JSON bytes.
+
+        Note: The decoded response will not have tools or format attached, as these
```
willbakst: The tool definitions and response format definitions should definitely be included in the Response. We should have serializaed representations of these things (e.g. the tool schemas and the response format). We may not be able to serialize / deserialize the actual code implementation, but when decoding we could always search for symbols that match the given schema (e.g. if we have a `Book` then look for the `Book` definition and use that for `FormatT` on load).

python/mirascope/llm/serialization/__init__.py
```diff
@@ -0,0 +1,30 @@
+"""Serialization and deserialization for Mirascope responses."""
```
willbakst, line 1: Do you think there are any existing libraries (like Pydantic or cattrs etc.) that we could leverage to bet better serialization/deserialization with less maintenance?

python/mirascope/llm/serialization/otel.py
```diff
@@ -0,0 +1,55 @@
+"""OpenTelemetry GenAI semantic conventions serialization adapter."""
+
+from collections.abc import Sequence
+from typing import Any
+
+from ...ops._internal.instrumentation.llm.encode import snapshot_from_root_response
```
willbakst: We should not have `llm` depending on `ops`, it's the other way around. If someone wants OTel serialization they should just use `ops` to get that through instrumentation. We can always expost additional `ops` methods if need be for certain use-cases

python/mirascope/llm/serialization/__init__.py
willbakst: Also, since serialization/deserialization is a boundary point, it seems like a great place for validation, which is definitely something we shouldn't re-implement. If we use something like msgspec or cattrs we would get that stuff out of the box, no?

I can imagine defining a JSON Schema that defines the serialization/deserialization spec. We could then auto-generate data models with validation for deserialization (which `decode` could then 1:1 map to the corresponding dataclass). For serialization, we should also then be able to validate the serialized about against the JSON Schema?

python/mirascope/llm/responses/response.py
```diff
@@ -56,6 +56,51 @@ def __init__(
             finish_reason=finish_reason,
```
teamdandelion, line 58: <p class='graphite__hidden'><i>[Re: line +29]</i></p>
Note: If we're adding the decode methods to this class, then they also need to go on `AsyncResponse`, `ContextResponse`, `AsyncContextResponse`, etc. And the stream variants as well. My sense is we'll want to refactor it so most of the logic lives in `BaseResponse`, and then we'll have a thin wrapper on the leaf classes that applies the right typing to the `Response`, `Toolkit`, etc — kind of mirroring the `__init__` methods and how `AsyncResponse.__init__` is a wrapper around `BaseResponse.__init__`.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1572?utm_source=unchanged-line-comment">Graphite</a>.</p>

python/mirascope/llm/responses/response.py
teamdandelion: Strong agree that it needs to be possible to round trip a `Response` through serialization while including tools and format. We also need doing so to be type-safe. So I'm thinking the signature will be something like:  
  
`def decode(cls, data: bytes | str | SerializedResponse, tools: Sequence[llm.Tool], format: Format[FormattableT] | None) -> "Response[FormattableT] | Response[None]`  
  
At runtime:  
\- It verifies whether the provided format's schema matches the expected de-serialized schema, error if not  
\- It verifies whether the tools provided are a strict superset of the expected tools (with matching schemas) errors if not

python/mirascope/llm/responses/response.py
```diff
@@ -56,6 +56,51 @@ def __init__(
...
+            with open("response.json", "rb") as f:
+                response = llm.Response.decode(f.read())
+            ```
+        """
+        from ..serialization import decode
+
+        return decode(data)
+
+    @classmethod
+    def decode_str(cls, data: str) -> "Response[None]":
```
teamdandelion: Is there a reason we'd want to have `decode` and `decode_str` be separate methods, rather than accepting `data: str | bytes` and handling both? Also, I think it should be possible to pass a `SerializedResponse` type (instance of `Jsonable`) explicitly rather than the str representation.

python/mirascope/llm/responses/root_response.py
```diff
@@ -176,3 +176,53 @@ def model(self) -> "Model":
...
+            ```python
+            response = client.call(model_id="claude-sonnet-4-20250514", messages=[...])
+            json_bytes = response.encode()
+            # Save to file
+            with open("response.json", "wb") as f:
+                f.write(json_bytes)
+            ```
+        """
+        from ..serialization import encode
+        from ..serialization.version import CURRENT_VERSION
```
teamdandelion: Nit: I don't know why Claude is so fond of importing things inside the function that uses them rather than top of file

python/mirascope/llm/serialization/__init__.py
teamdandelion: I agree, we shouldn't be rolling our own from scratch here. I like the idea of defining a JSON Schema that defines the serialized representations and generating models from that schema. This would also set us up for cross-programming-language portability on the serialized representations, which I think we clearly want

python/mirascope/llm/serialization/_content.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def _require_key(data: dict[str, Any], key: str, context: str) -> Any:  # noqa: ANN401
+    """Get a required key from a dictionary, raising InvalidSerializedDataError if missing."""
+    if key not in data:
+        raise InvalidSerializedDataError(f"Missing required field '{key}' in {context}")
+    return data[key]
+
+
+def encode_content_part(part: ContentPart) -> dict[str, Jsonable]:
```
teamdandelion: Rather than `dict[str, Jsonable]` I think we should have explicit typed representations of every serialized objects. So in this case it'd return `SerializedContent` which would be a union of `SerializedText`, `SerializedToolCall`, etc.

python/mirascope/llm/serialization/exceptions.py
```diff
@@ -0,0 +1,38 @@
+"""Exceptions for serialization/deserialization."""
+
+from ..exceptions import MirascopeLLMError
+
+
+class SerializationError(MirascopeLLMError):
```
teamdandelion, line 6: I think we'll need errors like `IncompatibleToolsError`, `IncompatibleFormatError`

python/mirascope/llm/serialization/version.py
```diff
@@ -0,0 +1,71 @@
+"""Version handling for serialization format."""
+
+from dataclasses import dataclass
+
+CURRENT_VERSION = "1.0"
+"""Current serialization format version."""
```
teamdandelion, line 23: I'm glad we're thinking of versioning from the start. I wonder if we will want a single global version, or if we should have per-representation versioning, eg. `SerializedResponseVersion`, `SerializedToolSchemaVersion`, etc.   
  
This relates to a question which is, where is the boundary where we add the version? If we serialize a ToolSchema do we put the version on that object? If we are serializing a content array do we put the version on every piece of content? Or do we have a global rule that whenever we serialize Mirascope objects we wrap it in a version wrapper?   
  
We should think of this in terms of how the decode logic will be written so that it can adapt to older versions as we bump them. So I think it would be straightforward to have versioning at every level where we have matching `encode` and `decode` functions, because then the `decode` function can read the version and invoke legacy logic if appropriate. However, this may be inefficient in cases where we are decoding (say) individual content pieces if there is unacceptable overhead for adding a version string to every single piece of content — in which case maybe we don't expose encoding/decoding at the level of individual content pieces, but rather at the level of messages. I think having a version-header-per-message makes sense, and is actually semantically useful since we might load a response which had messages with version 1.2 and then add a new message with version 1.3, but be persisting everything in a db layer where we persist the original message represented at version 1.2... or something similar.

koxudaxi: It's nice. I added the idea as `TODO` comments.

PR #1588, opened 2025-12-09
feat(cloud): ProjectService

willbakst:


willbakst:
* **#1588** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1656](https://github.com/Mirascope/mirascope/pull/1656) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1656?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1712](https://github.com/Mirascope/mirascope/pull/1712) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1588?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+const ROLE_HIERARCHY: Record<Role, number> = {
+  OWNER: 4,
+  ADMIN: 3,
+  DEVELOPER: 2,
+  ANNOTATOR: 1,
+};
+
+const REQUIRED_ROLE_FOR_ACTION: Record<PermissionAction, Role> = {
+  read: "ANNOTATOR",
+  update: "ADMIN",
```
koxudaxi: I can't find updates method in BaseProjectService.

willbakst: `BaseAuthenticatedService` implements a base update method

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,902 @@
...
+
+          const result = yield* Effect.either(
+            db.projects.findByOrganization(org.id, otherUser.id),
+          );
+
+          expect(result._tag).toBe("Left");
+          if (result._tag === "Left") {
+            expect(result.left).toBeInstanceOf(PermissionDeniedError);
+            expect(result.left.message).toBe(
+              "You do not have permission to read projects in this organization",
```
teamdandelion: Seems sketchy that we're using `result._tag` rather than the public interface, eg:

```
const result = yield* Effect.either(
  db.projects.findByOrganization(org.id, otherUser.id),
);

expect(Either.isLeft(result)).toBe(true);
if (Either.isLeft(result)) {
  expect(result.left).toBeInstanceOf(PermissionDeniedError);
  expect(result.left.message).toBe("User is not a member of this organization");
}
```

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+
+    return fetchProject.pipe(
+      Effect.flatMap((project) => {
+        if (!project) {
+          return Effect.fail(
+            new PermissionDeniedError({
+              message: `You do not have permission to ${action} this project`,
+              resource: this.baseService.resourceName,
+            }),
+          );
```
teamdandelion: If I'm understanding this correctly, if this project does not exist then the user will get a PermissionDeniedError? This seems confusing, e.g. if the user is an admin in their org but fetches a non-existent project they get PermissionDenied?

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,902 @@
...
+            expect(result.left.message).toBe(
+              "You do not have permission to update projects in this organization",
+            );
+          }
+        }),
+      ),
+    );
+  });
+
+  describe("findById", () => {
```
teamdandelion: Missing test for the case where the project does not exist

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,902 @@
...
+          );
+
+          const result = yield* Effect.either(
+            db.projects.create(
+              { name: "Duplicate Project", organizationId: org.id },
+              user.id,
+            ),
+          );
+
+          expect(result._tag).toBe("Left");
```
teamdandelion: Here and elsewhere: we should not be using `result._tag` (private), we should be using guards:  


```
import { Either } from "effect"

const foo = Either.right(42)

if (Either.isLeft(foo)) {
  console.log(`The left value is: ${foo.left}`)
} else {
  console.log(`The Right value is: ${foo.right}`)
}
// Output: "The Right value is: 42"
```

  
Or matching:

```
import { Either } from "effect"

const foo = Either.right(42)

const message = Either.match(foo, {
  onLeft: (left) => `The left value is: ${left}`,
  onRight: (right) => `The Right value is: ${right}`
})

console.log(message)
// Output: "The Right value is: 42"
```

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,902 @@
...
+            { name: "Test Organization" },
+            user.id,
+          );
+
+          yield* db.projects.create(
+            { name: "Duplicate Project", organizationId: org.id },
+            user.id,
+          );
+
+          const result = yield* Effect.either(
```
teamdandelion: I wonder why we are using `Effect.either` not `Effect.exit`?  
  
From the docs:  
  
Either is primarily used as a **simple discriminated union** and is not recommended as the main result type for operations requiring detailed error information.

[Exit](https://effect.website/docs/data-types/exit/) is the preferred **result type** within Effect for capturing comprehensive details about failures. It encapsulates the outcomes of effectful computations, distinguishing between success and various failure modes, such as errors, defects and interruptions.  
...  
**Exit**

An `Exit<A, E>` describes the result of running an `Effect` workflow.

There are two possible states for an `Exit<A, E>`:

- `Exit.Success`: Contains a success value of type `A`.
- `Exit.Failure`: Contains a failure [Cause](https://effect.website/docs/data-types/cause/) of type `E`.

## **Creating Exits**

The Exit module provides two primary functions for constructing exit values: `Exit.succeed` and `Exit.failCause`. These functions represent the outcomes of an effectful computation in terms of success or failure.  
  
  
I think `Success` and `Failure` is more clear to our semantics than `left` and `right`

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,902 @@
...
+            expect(result.left).toBeInstanceOf(NotFoundError);
+            expect(result.left.message).toContain("not found");
+          }
+        }),
+      ),
+    );
+
+    it(
+      "should return PermissionDeniedError when user is not a member",
+      withTestDatabase((db) =>
```
teamdandelion: Especially for delete (but probably elsewhere) - we should test the case where the user is a member, but doesn't have sufficient permisisons (eg annotator tries to delete)

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,902 @@
...
+            expect(result.left.message).toBe(
+              "You do not have permission to delete projects in this organization",
+            );
+          }
+        }),
+      ),
+    );
+  });
+
+  describe("Error handling", () => {
```
teamdandelion: I think it would be better to re-organize these test cases so they are grouped with the method they test (eg create, findById, etc). That way it will be a lot clearer where there are gaps in logical coverage

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+  typeof projects
+> {
+  protected readonly baseService: BaseProjectService;
+
+  constructor(db: PostgresJsDatabase<typeof schema>) {
+    super(db);
+    this.baseService = new BaseProjectService(db);
+  }
+
+  private checkOrganizationPermission(
```
teamdandelion: I find it a little strange that the `ProjectService` is implementing organization permission logic. I expect that organization permissions are needed outside the scope of just projects so I worry that we're going to have multiple copies of this logic. Should we have an `OrganizationService` that provides this and the `ProjectService` depends on it?

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+          .select({ role: organizationMemberships.role })
+          .from(organizationMemberships)
+          .where(
+            and(
+              eq(organizationMemberships.userId, userId),
+              eq(organizationMemberships.organizationId, organizationId),
+            ),
+          )
+          .limit(1);
+        return membership;
```
teamdandelion: I feel like it'd be good to have a separate effect for getting a member's role within an organization (likely on the `OrganizationService` per above)

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+            message: `You do not have permission to ${action} projects in this organization`,
+            resource: this.baseService.resourceName,
+          }),
+        );
+      }
+
+      return Effect.succeed(undefined);
+    };
+
+    return fetchMembership.pipe(Effect.flatMap(verifyMembershipAndRole));
```
teamdandelion: Likewise, I think this could be separated as something like `verifyMemberPermission(userId, organizationId, aciton)` on the `OrganizationService`

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+  constructor(db: PostgresJsDatabase<typeof schema>) {
+    super(db);
+    this.baseService = new BaseProjectService(db);
+  }
+
+  private checkOrganizationPermission(
+    userId: string,
+    action: PermissionAction,
+    organizationId: string,
+  ): Effect.Effect<void, PermissionDeniedError | DatabaseError> {
```
teamdandelion: No `NotFoundError`? What if the userId or organizationId don't exist?

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+    };
+
+    return fetchMembership.pipe(Effect.flatMap(verifyMembershipAndRole));
+  }
+
+  protected checkPermission(
+    userId: string,
+    action: PermissionAction,
+    projectId: string,
+  ): Effect.Effect<void, PermissionDeniedError | DatabaseError> {
```
teamdandelion: no NotFoundError? What if the userId or projectId don't exist?

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,414 @@
...
+        }),
+    });
+
+    return fetchUserProjects;
+  }
+
+  findByOrganization(
+    organizationId: string,
+    userId: string,
+  ): Effect.Effect<PublicProject[], PermissionDeniedError | DatabaseError> {
```
teamdandelion: no NotFoundError? What if the organizationId doesn't exist?

cloud/db/services/projects.ts
teamdandelion: Ok I see the OrganizationService does exist. Why not have the ProjectService depend on the OrganizationService and then move organizational permission management into that?

cloud/db/services/projects.test.ts
willbakst: I did some digging and found `@effect/vitest` as well as `Effect.flip`. I've refactored this test file to apply these, which I think resolves your concerns here.

I will insert a sweep that goes through existing tests to update them to use this pattern.

cloud/db/services/projects.test.ts
willbakst: Resolved in response to other comment

cloud/db/services/projects.test.ts
willbakst: Added

cloud/db/services/projects.test.ts
willbakst: I've added this

cloud/db/services/projects.test.ts
willbakst: Done

cloud/db/services/projects.ts
willbakst: Fixed

cloud/db/services/projects.ts
willbakst: Fixed

cloud/db/services/projects.ts
willbakst: Fixed

cloud/db/services/projects.ts
willbakst: Fixed down stack

cloud/db/services/projects.ts
willbakst: Fixed down stack

cloud/db/services/projects.ts
willbakst: Fixed

cloud/db/services/projects.ts
willbakst: Fixed

cloud/db/migrations/0002_condemned_psylocke.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE "projects" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"name" text NOT NULL,
+	"organization_id" uuid NOT NULL,
+	"created_at" timestamp DEFAULT now(),
+	"updated_at" timestamp DEFAULT now(),
+	CONSTRAINT "projects_organization_id_name_unique" UNIQUE("organization_id","name")
+);
+--> statement-breakpoint
+ALTER TABLE "projects" ADD CONSTRAINT "projects_organization_id_organizations_id_fk" FOREIGN KEY ("organization_id") REFERENCES "public"."organizations"("id") ON DELETE no action ON UPDATE no action;
```
koxudaxi: Is `ON DELETE no action`  expected behavior?

willbakst: This is auto-generated by Drizzle. My understanding is that this is saying that if a project is deleted there is no action to take on the foreign key organization

koxudaxi: OK, I see.

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,601 @@
...
+            userId: "user-id",
+          });
+
+          expect(result).toBeDefined();
+          expect(result.id).toBe("project-id");
+        }),
+    );
+  });
+
+  describe("update", () => {
```
teamdandelion: Can we add a test where update would cause a naming collision?

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,258 @@
...
+    id,
+    data,
+    userId,
+  }: {
+    id: string;
+    data: Partial<NewProject>;
+    userId: string;
+  }): Effect.Effect<
+    PublicProject,
+    NotFoundError | PermissionDeniedError | DatabaseError
```
teamdandelion, line 459: I believe this should return AlreadyExistsError as with create

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,258 @@
...
+  }): Effect.Effect<
+    PublicProject,
+    AlreadyExistsError | NotFoundError | PermissionDeniedError | DatabaseError
+  > {
+    return Effect.gen(this, function* () {
+      const role = yield* this.organizationService.getRole({
+        organizationId: data.organizationId,
+        userId,
+      });
+      yield* this.verifyPermission(role, "create");
```
teamdandelion: Yes I like this a lot better than what we have before. I could see refactoring `this.verifyPermission` to take the userId, the organization, the action, and abstract over the role check (while likely keeping `.getRole`. However we can wait until we have >3 instances of this code, I'm content with it as written.

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,258 @@
...
+    NotFoundError | PermissionDeniedError | DatabaseError
+  > {
+    return Effect.gen(this, function* () {
+      const existing = yield* this.baseService.findById({ id });
+      const role = yield* this.organizationService.getRole({
+        organizationId: existing.organizationId,
+        userId,
+      });
+      yield* this.verifyPermission(role, "update");
+      return yield* this.baseService.update({ id, data });
```
teamdandelion: We should catch and "re throw" the unique constraint violation like we did in create

willbakst: Unique constraint on what? it's updating the data for an existing project?

willbakst: Oh I don't think we have a unique constraint on project names, only organizations

cloud/db/services/projects.test.ts
willbakst: I don't want to enforce uniqueness constraints at the schema level because I think two users of separate organizations should be able to have a project with the same name.

Do you want to enforce uniqueness within an organization?

cloud/db/services/projects.ts
willbakst: Same questions here around uniqueness

cloud/db/schema/projects.ts
```diff
@@ -0,0 +1,33 @@
...
+    organizationId: uuid("organization_id")
+      .references(() => organizations.id)
+      .notNull(),
+    createdAt: timestamp("created_at").defaultNow(),
+    updatedAt: timestamp("updated_at").defaultNow(),
+  },
+  (table) => ({
+    orgNameUnique: unique().on(table.organizationId, table.name),
+  }),
+);
```
teamdandelion: Hmm, _should_ projects be nested under organizations?  
  
I imagine down the line every document (notebook, etc) will be part of a project? In which case:  
\- Can users have personal projects that are bound to their account, not to any organization?  
\- What about projects that are shared between orgs or where an org wants to give visibility to external contributors or another org?  
  
What do you think of a design where:  
  
\- A project has an owner, which may be an organization or a user   
\- There is a permissions table that tracks ProjectId <-> (UserId | OrganizationId)  + role relation  
  
If we don't want to do that all upfront, we could prepare for it by making the project have an owner (user id or organization id) and only implement the permissions table when we actually need it.   
  
We could also auto-generate a default project for every user/org if we want to maintain the "all data is owned by a project" property

cloud/db/services/projects.ts
willbakst: I do this upstack here: https://app.graphite.com/github/pr/Mirascope/mirascope/1591/feat-add-environments-database-schema%2C-service%2C-and-tests#lines-cloud/db/services/base.ts-R308-R308

cloud/db/schema/projects.ts
willbakst: Yeah, it probably makes the most sense for projects to be their own thing with some form of access controls and an owner. This will enable the same functionality now while keeping us future proof for expansion on more granular features here around permissions etc.

I'm happy to restructure now while it's early and I can bubble this change up with less debt.

teamdandelion: Sounds good. I was thinking about it more and realized we need to handle the following case:  
  
\- Org has many projects, some of which involve sensitive data  
\- Wants to bring an outside specialist to annotate data for one particular project  
\- Adding them to the org with ANNOTATOR role means exposing all of their projects  
\- Hence, need project level ACLs

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,258 @@
...
+              organizationId: projects.organizationId,
+            });
+          return result;
+        },
+        catch: (error) => {
+          if (isUniqueConstraintError(error)) {
+            return new AlreadyExistsError({
+              message:
+                "A project with this name already exists in the organization",
+              resource: this.baseService.resourceName,
```
teamdandelion, line 313: FYI here's the existing project name uniqueness constraint I was referencing

cloud/db/schema/projects.ts
```diff
@@ -0,0 +1,33 @@
...
+    name: text("name").notNull(),
+    organizationId: uuid("organization_id")
+      .references(() => organizations.id)
+      .notNull(),
+    createdAt: timestamp("created_at").defaultNow(),
+    updatedAt: timestamp("updated_at").defaultNow(),
+  },
+  (table) => ({
+    orgNameUnique: unique().on(table.organizationId, table.name),
+  }),
```
teamdandelion: Project names are unique within orgs here, though imo we could drop this constraint. I think it's more robust not to depend on this

cloud/db/services/projects.test.ts
teamdandelion: Currently we do have name uniqueness within an organization (see comments elsewhere) but I think we could remove it. Since we should not be using the name as an identifier, forcing uniqueness seems more frustrating than helpful.

willbakst: I will work this into the refactor in response to your other comment

cloud/db/schema/projects.ts
willbakst: Yeah, if we're taking away the nesting, then this constraint doesn't make sense anyway

koxudaxi (dismissed):

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,570 @@
...
+import { Effect } from "effect";
+import {
+  DatabaseError,
+  NotFoundError,
+  PermissionDeniedError,
+} from "@/db/errors";
+import { DatabaseService } from "@/db/services";
+
+describe("ProjectService", () => {
+  describe("create", () => {
```
teamdandelion, line 18: Missing a test case where create is passed a project with both userOwnerId and orgOwnerId set. I believe this will pass typechecking, and result in a constraint violation db error that gets propagated back up. We could consider re-wrapping the error to distinguish it from generic database issues, but in either case let's test it and document the behavior

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,570 @@
...
+        expect(result.name).toBe(project.name);
+      }).pipe(Effect.provide(TestDatabase)),
+    );
+
+    it.effect("returns `NotFoundError` when project not found", () =>
+      Effect.gen(function* () {
+        const db = new MockDatabase()
+          // getRole: no membership found
+          .select([])
+          .build();
```
teamdandelion, line 259: Why are we using a mock database here, rather than the same test db service as other tests? It seems to me mocking should be used when it's not feasible to get the database in the necessary state thru regular operations, but otherwise we should write the tests to be agnostic of the implementation by avoiding mocks.

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,570 @@
...
+          // baseService.findById: found
+          .select([
+            {
+              id: "project-id",
+              name: "Test",
+              userOwnerId: "owner-id",
+              orgOwnerId: null,
+            },
+          ])
+          .build();
```
teamdandelion, line 280: Likewise, why do we need mocking for this? If the issue is that we don't have services/project-membership.ts (to allow setting the annotator role relationship) then maybe we should just add that?

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,570 @@
...
+    );
+
+    it.effect(
+      "returns `PermissionDeniedError` when user has insufficient role",
+      () =>
+        Effect.gen(function* () {
+          const db = new MockDatabase()
+            // getRole: ANNOTATOR role (too low for update)
+            .select([{ role: "ANNOTATOR" }])
+            .build();
```
teamdandelion: as above

cloud/db/services/projects.test.ts
```diff
@@ -0,0 +1,570 @@
...
+    );
+
+    it.effect(
+      "returns `PermissionDeniedError` when user has insufficient role",
+      () =>
+        Effect.gen(function* () {
+          const db = new MockDatabase()
+            // getRole: ADMIN role (only OWNER can delete)
+            .select([{ role: "ADMIN" }])
+            .build();
```
teamdandelion: as above

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,299 @@
...
+    return Effect.tryPromise({
+      try: async () => {
+        return await this.db.transaction(async (tx) => {
+          const [project] = await tx
+            .insert(projects)
+            .values({
+              name: data.name,
+              userOwnerId: data.userOwnerId ?? null,
+              orgOwnerId: data.orgOwnerId ?? null,
+              createdByUserId: userId,
```
teamdandelion: This is really sketchy:  
\- Any user can create a named project owned by another user, without permissions  
\- Any user can create a named project owned by another organization, without permission

This means if you learn someone's userId or organizationId you can drop honeypots into their account, or projects with spam or offensive names, etc. Furthermore the creator is added as an owner in the project membership, so they'd be able to extract data from the honeypot, too.  
  
This definitely needs permissions:

\- If orgOwnerId: User must be authenticated (ADMIN or OWNER) with the organization  
\- If userOwnerId: UserId must be identical to userOwnerId

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,299 @@
...
+            .select({ role: projectMemberships.role })
+            .from(projectMemberships)
+            .where(
+              and(
+                eq(projectMemberships.projectId, id),
+                eq(projectMemberships.userId, userId),
+              ),
+            )
+            .limit(1);
+          return membership;
```
teamdandelion: Something that is weird about this is it admits the possibility that a project can become de-facto ownerless from a permissioning standpoint. It must have a `userOwnerId` or `orgOwnerId` per the project data, but those ownerId fields are ignored by `getRole`, meaning it's possible that a user could create a project, edit their own permission on the project to be something other than `OWNER`, and then functionally no-one is an owner of the project anymore.  
  
I suggest we change this to something like the following psuedocode:  


```
if userOwner := project.userOwnerId:
  if userOwner == userId:
    return roles.OWNER
  return projectMemberships.getRole({userId, projectId})
elif orgOwner := project.orgOwnerId:
  user_acl = projectMemberships.getRole({userId, projectId})
  org_permission = organizations.getRole({userId, projectId})
  return higher_role(user_acl, org_permission)
```

  
This way a user always has the higher permission of what was granted them on the project individually, or what is granted by their membership in the owning organization. E.g. if user is an admin in the org that owns the project, I don't think it makes sense for them to not have access to the project itself.  
  
(This could get hairy for large/complex organizations that want to create secret projects or something, but I think they'd be better off creating a secret organization to own their secret projects)

cloud/db/services/projects.ts
```diff
@@ -0,0 +1,299 @@
...
+              userOwnerId: projects.userOwnerId,
+              orgOwnerId: projects.orgOwnerId,
+            });
+
+          // Add creator to project_memberships as OWNER
+          await tx.insert(projectMemberships).values({
+            projectId: project.id,
+            userId,
+            role: "OWNER",
+          });
```
teamdandelion: Per comment above, I don't think we should be doing this — when we look up roles we should respect the userOwnerId or orgOwnerId fields as being special sources of truth. In which case this entry is purely redundant.

I think my preference would be to exclude user owners from this table, and have all of the code that handles project roles or project memberships look up user owners as a necessary special case. That avoids the risk that the database gets in an inconsistent state where some projects have user owners in membership and some projects don't, and the behavior diverges.

teamdandelion (commented):
To review this properly I'd like to have the project memberships service getting implemented in the same PR (for testing purposes etc). They are one security surface

cloud/db/services/projects.test.ts
willbakst: Added:

1. `ProjectService.create` now handles the check constraint violation and includes a new `InvalidInputError` as part of the error channel
2. Test case for this new error path.
3. `NewProject` type restriction to better match what is actually allowed when creating a new project. This required adding a type ignore to the test from (2) since it now type errors (but the test is still worth keeping I believe).

cloud/db/services/projects.test.ts
willbakst: Fixed

cloud/db/services/projects.test.ts
willbakst: I do fix this in #1661 after I implement the `addMember` and `terminateMember` methods for `OrganizationService` and `ProjectService`

cloud/db/services/projects.test.ts
willbakst: See #1661

cloud/db/services/projects.test.ts
willbakst: See #1661

cloud/db/services/projects.ts
willbakst: I wonder if this whole refactor is the wrong choice. It seems far simpler to put projects under organizations and simply have users start off with some default organization that operates as their personal workspace. This structure in no way prevents us from then later on adding more granular permissions through a `ProjectMemberships` table.

cloud/db/services/projects.ts
willbakst: I think this will go away if we revert back to projects under organizations as mentioned in my other comment

cloud/db/services/projects.ts
willbakst: My intention here was to ensure that if someone creates a project in an organization that they then have owner level permissions. Given the project memberships table is the source for that, it made more sense to add the creator as the owner on creation, guaranteeing that every project has at least some user owner that can operate on it.

However, I think we're overcomplicating things here and as per my other comment should just put projects under organizations and defer more granular permissions for a later time where they are needed.

koxudaxi (approved):

willbakst:
### Merge activity

* **Dec 19, 3:30 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1588) rebased this pull request as part of a merge.
* **Dec 19, 3:34 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1588) couldn't merge this PR because it was not satisfying all requirements (Failed CI: 'CI Success', 'Python Tests (Python 3.13)').
* **Dec 19, 3:46 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1588).

PR #1591, opened 2025-12-09
feat(cloud): db/Environments

willbakst:

PR #1590, opened 2025-12-09
feat: add projects frontend (hooks, contexts, sidebar, dashboard)

willbakst:

PR #1589, opened 2025-12-09
feat: add projects API endpoints

willbakst:

PR #1594, opened 2025-12-09
feat(cloud): add API keys database schema, service, and tests

willbakst:


willbakst:
* **#1600** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1601](https://github.com/Mirascope/mirascope/pull/1601) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1601?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1738](https://github.com/Mirascope/mirascope/pull/1738) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1738?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1603** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1602** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1599** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1598** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1598?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1597** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1597?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1596** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1595** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1594** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1594?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1594?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1593** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1593?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1592** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1592?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1591** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1735** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1734** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1734?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1733** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1733?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1731** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1731?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1730** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1730?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1729** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1729?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1725** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1724** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1716** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1715** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1714** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1590** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1590?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1655](https://github.com/Mirascope/mirascope/pull/1655) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1655?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1650** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1649** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1589** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1589?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1712** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1588** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1656](https://github.com/Mirascope/mirascope/pull/1656) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1656?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `v2`


<h2></h2>

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,731 @@
...
+ * ApiKeys (authenticated)
+ *   └── authorization via ProjectMemberships.getRole()
+ * ```
+ *
+ * ## API Key Roles
+ *
+ * API keys use the project's role system:
+ * - `ADMIN` - Full API key management (create, read, update, delete)
+ * - `DEVELOPER` - Can create, read, and update API keys
+ * - `VIEWER` - Read-only access to API keys
```
teamdandelion: Why do viewers get read access to API keys?

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,731 @@
...
+ * ```
+ * ApiKeys (authenticated)
+ *   └── authorization via ProjectMemberships.getRole()
+ * ```
+ *
+ * ## API Key Roles
+ *
+ * API keys use the project's role system:
+ * - `ADMIN` - Full API key management (create, read, update, delete)
+ * - `DEVELOPER` - Can create, read, and update API keys
```
teamdandelion, line 19: Why can DEVELOPER create and update API keys but not delete them? What about a compromised key?  
  
Imo whatever permission level can create an API key should also be able to delete them. However I could see API key creation being restricted to ADMIN.

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,731 @@
...
+          keyPrefix,
+          environmentId,
+          createdById: userId,
+        })
+        .returning(publicFields)
+        .pipe(
+          Effect.mapError((e): AlreadyExistsError | DatabaseError =>
+            isUniqueConstraintError(e.cause)
+              ? new AlreadyExistsError({
+                  message: `An API key with name "${data.name}" already exists in this environment`,
```
teamdandelion, line 289: Let's apply whatever slugification logic we come up with downstack (or do it all together at top of stack)

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,731 @@
...
+      return { apiKeyId: apiKey.id, environmentId: apiKey.environmentId };
+    });
+  }
+
+  /**
+   * Gets the user who created an API key.
+   *
+   * This method is used after API key verification to get the user context
+   * for authorization and audit purposes. The API key's creator determines
+   * what permissions the API key has - it can only do what that user can do.
```
teamdandelion, line 734: This doesnt seem right. Consider the following:  
  
\- SeniorEngineer works for FooCo, is admin on a project, creates a proejct API key  
\- API key is used in a user-facing production system  
\- SeniorEngineer gives 2 week notice, leaves org on Friday afternoon  
\- Key is no longer valid, production system goes down  
  
I think the permissions need to be a property of the key and the project it's in, not a property of the user who created the key.

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,731 @@
...
+ */
+const publicFields = {
+  id: apiKeys.id,
+  name: apiKeys.name,
+  keyPrefix: apiKeys.keyPrefix,
+  environmentId: apiKeys.environmentId,
+  createdById: apiKeys.createdById,
+  createdAt: apiKeys.createdAt,
+  lastUsedAt: apiKeys.lastUsedAt,
+};
```
teamdandelion, line 123: I think API key should have concept of scopes/permissions.  
  
Possibly to start the permission can be one of our current roles (eg ADMIN, DEVELOPER, VIEWER, ANNOTATOR). In which case a user can only create API keys with roles less than or equal to their current role (developer cannot create admin key etc).

cloud/db/api-keys.ts
willbakst: Doing this up stack with the other same changes across the board

cloud/db/api-keys.ts
willbakst: How do we properly audit then? For example, a user hits the API to update someone's membership. How do we get the `actorId` if we don't associate the API key with a particular user?

I agree there is a potential issue you've outlined, but I think the solution is to handle that through a different feature (e.g. user-agnostic API keys with their own permission set).

I can imagine having an API key that is user-agnostic and works for certain actions that don't require audit (e.g. sending a trace to the system, making an API call to router, etc.) but that those API keys will have restricted permissions and not work for requests where we need to acting user for audit purposes.

Thoughts?

cloud/db/api-keys.ts
willbakst: Good catch, I've removed read access. No need for them to see the list of API keys (or the API key tab at all for that matter).

cloud/db/api-keys.ts
willbakst: Yeah, I've updated this so that you can delete an API key if you've created it (or if you're an ADMIN). That seems like the most reasonable thing to do.

cloud/db/api-keys.ts
willbakst: I'd like to sync on this in tandem with the other comment around this. I feel like this a separate feature we should discuss.

cloud/db/api-keys.ts
teamdandelion: Yeah, my view is we should have the API keys for actions where we do not need to to bind it to a particular actorId. Adding traces, making Mirascope router calls, etc — I think in those cases we can tie the action to either a userId or an apiKeyId.  
  
For things where we want strict actor traces, like managing organization or project membership; maybe down the line we will want automated management tools and we can handle that kind of API key the. But for now it seems sufficient to let admins sign in through the UI and use the non-API key authenticated flow

teamdandelion: Discussed offline, if the org needs an "org actor" API key they can make a user for that purpose.

cloud/db/api-keys.ts
teamdandelion: Discussed offline, this seems like a separate feature we can add when requested.

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,783 @@
...
+   * 4. Returns the API key ID and environment ID
+   *
+   * Note: This method does NOT require authentication - it IS the authentication.
+   *
+   * @param key - The plaintext API key to verify
+   * @returns The API key ID and environment ID
+   * @throws NotFoundError - If the API key is invalid
+   * @throws DatabaseError - If the database operation fails
+   */
+  verifyApiKey(
```
teamdandelion, line 671: Why are verifyApiKey and getCreator separate? Since the owning userId is fundamental to the concept / meaning of the API key  (it is an api key for THIS user in THIS environment, not an api key for this environment in general) I feel like verifyApiKey may as well itself return environmentId and userId  


Also, suppose that a user makes an API key in an environemtn and then is kicked out of the org (and by extension, the project, and by extension, the environment). It feels weird that `verifyApiKey` on their key will return that environment, when in fact it should be a permission error?

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,783 @@
...
+/**
+ * Public fields to select from the api_keys table.
+ * Excludes keyHash for security.
+ */
+const publicFields = {
+  id: apiKeys.id,
+  name: apiKeys.name,
+  keyPrefix: apiKeys.keyPrefix,
+  environmentId: apiKeys.environmentId,
+  createdById: apiKeys.createdById,
```
teamdandelion: How about we rename this from `createdById` to `ownerId` or `owningUserId`? For me the concept of a "creator" implies that someone could create the API key and then give it to someone else, but our trust model is that the creator is permanently responsible for it — ie they are its owner.

cloud/db/schema/api-keys.ts
```diff
@@ -0,0 +1,61 @@
...
+    name: text("name").notNull(),
+    // Store only the hash of the key, never the plaintext
+    keyHash: text("key_hash").notNull(),
+    // Store a prefix for display purposes (e.g., "mk_abc...")
+    keyPrefix: text("key_prefix").notNull(),
+    environmentId: uuid("environment_id")
+      .references(() => environments.id, { onDelete: "cascade" })
+      .notNull(),
+    // Track who created this API key (for authentication & audit)
+    createdById: uuid("created_by_id")
```
teamdandelion: As suggested elsewhere: i suggest ownerId for this

cloud/db/api-keys.ts
```diff
@@ -0,0 +1,783 @@
...
+  createdAt: apiKeys.createdAt,
```
teamdandelion, line 121: What do you think of adding ownerName to the public fields? I think we will want the API key management UI to show the owner of each key, so if Admin is trying to delete Alice's key, it's easy to do even if she named it "Greg's key". But as implemented we would need a second db query to get the owner info for the UI.

willbakst: These are specifically the public fields for the api_keys table. If we want to get the user info for a UI of some sort, we can have the API route run a separate DB query that does the inner join on users to get that info

cloud/db/api-keys.ts
willbakst: NOTE: I update this up stack slightly, so I'll opt to implement changes in #1596.

If someone sends an API key that exists in our system, `verifyApiKey` returns that API key's information (org/project/env IDs). We then check that against the path parameters and fail early in they don't match (saying "hey these don't match" so a user can fix a stupid mistake).

If everything matches, we grab the owner for that key and we're authenticated as that user. If that user then doesn't have permission (e.g. in your example they got booted from the project / org), then `authorize` will return `NotFoundError` for them.

I think there are two things here worth updating:

1. Use `ApiKeyInfo` rather than `VerifiedApiKey`. This removes the weirdness around "verify" being the wrong word.
2. Merge `verifyApiKey` into `getCreator` where it returns `ApiKeyInfo`, which will include org/project/env IDs and owner information like ID and name.

I'll circle back to resolve once I've implemented the change up stack.

cloud/db/api-keys.ts
willbakst: Done

cloud/db/schema/api-keys.ts
willbakst: Done

cloud/db/api-keys.ts
willbakst: Done

PR #1606, opened 2025-12-09
feat(db): add TraceService for OTLP ingestion

koxudaxi:
### TL;DR

Added trace ingestion service to store and manage OpenTelemetry trace data.

### What changed?

- Created a new `TraceService` class for handling trace data ingestion
- Implemented methods to process and store OpenTelemetry trace data
- Added comprehensive test coverage for the trace service
- Integrated the trace service into the database service layer
- Added utility functions to normalize OTLP (OpenTelemetry Protocol) data formats

The implementation includes:
- Environment context resolution to associate traces with projects and organizations
- Conversion of OTLP values to normalized JSON structures
- Handling of various attribute types including nested arrays and key-value lists
- Proper handling of duplicate spans with acceptance/rejection tracking

### How to test?

Run the test suite which covers various scenarios:
- Basic trace ingestion
- Handling of duplicate spans
- Processing of complex attribute types
- Normalization of events and links
- Partial success scenarios with mixed new/duplicate spans

```
pnpm test db/services/traces.test.ts
```

### Why make this change?

This change enables the application to ingest and store OpenTelemetry trace data, which is essential for implementing distributed tracing functionality. The trace service provides the foundation for collecting performance and debugging information from applications, allowing users to monitor and troubleshoot their systems effectively.

PR #1605, opened 2025-12-09
feat(db): add migration for traces and spans tables

koxudaxi:
### TL;DR

Added database schema for traces and spans to support distributed tracing functionality.

### What changed?

- Created a new `traces` table to store trace information with fields for:
  - Trace identifiers
  - Service information (name, version)
  - Resource attributes
  - Organization/project/environment relationships

- Created a new `spans` table to store span data with fields for:
  - Span identifiers and relationships (trace_id, span_id, parent_span_id)
  - Timing information (start/end times)
  - Span attributes, status, events, and links
  - Performance metadata (dropped counts)

- Added appropriate foreign key constraints to maintain data integrity
- Created indexes to optimize query performance for common access patterns

### How to test?

1. Apply the migration to a test database
2. Verify the tables are created with the correct schema
3. Test inserting trace and span data to ensure constraints work properly
4. Verify the indexes are created and used in relevant queries

### Why make this change?

This database schema is the foundation for implementing distributed tracing capabilities in our application. The tables will store OpenTelemetry-compatible trace and span data, allowing us to:

- Track request flows across multiple services
- Analyze performance bottlenecks
- Troubleshoot errors in distributed systems
- Provide observability into application behavior

PR #1604, opened 2025-12-09
feat(db): add traces and spans schema

koxudaxi:
### TL;DR

Added database schema for traces and spans to support distributed tracing functionality.

### What changed?

- Created two new database schema files:
  - `traces.ts`: Defines the traces table with fields for trace ID, environment, project, organization, service information, and resource attributes
  - `spans.ts`: Defines the spans table with fields for span details, timing information, and relationships to traces
- Updated the main schema index file to:
  - Export the new schema types
  - Export the public interfaces for API responses
  - Include traces and spans in the DatabaseTable type

### How to test?

1. Run database migrations to create the new tables
2. Verify that the schema can be used to store and retrieve trace and span data
3. Confirm that the relationships between traces, spans, environments, projects, and organizations work correctly

### Why make this change?

This change establishes the database foundation for implementing distributed tracing capabilities in the application. The schema design allows for storing hierarchical span data within traces while maintaining proper relationships to environments, projects, and organizations. This will enable visualization and analysis of request flows across services.

PR #1607, opened 2025-12-09
feat(api): add API Key auth to /traces endpoint

koxudaxi:
### TL;DR

Implemented proper trace ingestion with API key authentication and database persistence.

### What changed?

- Replaced the debug logging implementation with actual trace data persistence
- Added API key authentication to the trace ingestion endpoint
- Connected the trace handler to the database service for storing trace data
- Added proper error handling for unauthorized access, not found resources, and database errors
- Enhanced the API schema to include appropriate error responses
- Expanded test coverage with comprehensive tests for authentication and trace ingestion

### How to test?

1. Send a POST request to `/traces` with a valid API key in the `X-API-Key` header
2. Include a valid trace payload with resource spans
3. Verify a 200 response with a `partialSuccess` object
4. Test with invalid API keys to ensure 401/404 responses
5. Run the test suite with `npm test` to verify all scenarios

### Why make this change?

This change completes the trace ingestion pipeline by properly authenticating requests and persisting trace data to the database. It replaces the previous placeholder implementation that only logged trace data without storing it. This enables the application to collect and analyze distributed traces from client applications.

### DIFF
--- a/cloud/api/handler.ts
+++ b/cloud/api/handler.ts
@@ -5,34 +5,40 @@ import { HandlerError } from "@/errors";
 import { SettingsService } from "@/settings";
 import { Database } from "@/db";
 import { Payments } from "@/payments";
-import { AuthenticatedUser } from "@/auth";
-import type { PublicUser } from "@/db/schema";
+import { AuthenticatedUser, Authentication } from "@/auth";
+import type { PublicUser, ApiKeyInfo } from "@/db/schema";
 
 export type HandleRequestOptions = {
   prefix?: string;
-  authenticatedUser: PublicUser;
+  user: PublicUser;
+  apiKeyInfo?: ApiKeyInfo;
   environment: string;
 };
 
 type WebHandlerOptions = {
   db: Context.Tag.Service<Database>;
   payments: Context.Tag.Service<Payments>;
-  authenticatedUser: PublicUser;
+  user: PublicUser;
+  apiKeyInfo?: ApiKeyInfo;
   environment: string;
 };
 
 function createWebHandler(options: WebHandlerOptions) {
   const services = Layer.mergeAll(
     Layer.succeed(SettingsService, { env: options.environment }),
-    Layer.succeed(AuthenticatedUser, options.authenticatedUser),
+    Layer.succeed(AuthenticatedUser, options.user),
+    Layer.succeed(Authentication, {
+      user: options.user,
+      apiKeyInfo: options.apiKeyInfo,
+    }),
     Layer.succeed(Database, options.db),
     Layer.succeed(Payments, options.payments),
   );
 
-  const ApiWithDependencies = Layer.mergeAll(
+  const ApiWithDependencies = Layer.merge(
     HttpServer.layerContext,
-    ApiLive.pipe(Layer.provide(services)),
-  );
+    ApiLive,
+  ).pipe(Layer.provide(services));
 
   return HttpApiBuilder.toWebHandler(ApiWithDependencies);
 }
@@ -62,7 +68,8 @@ export const handleRequest = (
     const webHandler = createWebHandler({
       db,
       payments,
-      authenticatedUser: options.authenticatedUser,
+      user: options.user,
+      apiKeyInfo: options.apiKeyInfo,
       environment: options.environment,
     });
 
@@ -74,20 +81,35 @@ export const handleRequest = (
           const pathWithoutPrefix =
             url.pathname.slice(options.prefix.length) || "/";
           const newUrl = new URL(pathWithoutPrefix + url.search, url.origin);
+          const hasBody = request.body !== null;
           modifiedRequest = new Request(newUrl.toString(), {
             method: request.method,
             headers: request.headers,
             body: request.body,
             redirect: request.redirect,
             signal: request.signal,
+            ...(hasBody ? ({ duplex: "half" } as RequestInit) : {}),
           });
         }
 
-        const response = await webHandler.handler(modifiedRequest);
+        let response = await webHandler.handler(modifiedRequest);
         const contentType = response.headers.get("content-type") || "";
         const isJsonResponse = contentType
           .toLowerCase()
           .includes("application/json");
+
+        if (isJsonResponse && response.status >= 400) {
+          const body = await response.clone().text();
+          if (body.includes('"_tag"')) {
+            const transformedBody = body.replace(/"_tag":/g, '"tag":');
+            response = new Response(transformedBody, {
+              status: response.status,
+              statusText: response.statusText,
+              headers: response.headers,
+            });
+          }
+        }
+
         const matched = response.status !== 404 || isJsonResponse;
         return { matched, response };
       },
--- a/cloud/api/traces.handlers.ts
+++ b/cloud/api/traces.handlers.ts
@@ -1,10 +1,9 @@
 import { Effect } from "effect";
-import {
-  type KeyValue,
-  type ResourceSpans,
-  type ScopeSpans,
-  type CreateTraceRequest,
-  type CreateTraceResponse,
+import { Database } from "@/db";
+import { Authentication } from "@/auth";
+import type {
+  CreateTraceRequest,
+  CreateTraceResponse,
 } from "@/api/traces.schemas";
 
 export * from "@/api/traces.schemas";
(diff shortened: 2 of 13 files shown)


koxudaxi:
* **#1607** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1607?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1609](https://github.com/Mirascope/mirascope/pull/1609) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1609?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1768](https://github.com/Mirascope/mirascope/pull/1768) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1768?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1606** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1606?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1605** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1605?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1604** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1604?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1635** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

cloud/api/traces.handlers.ts
```diff
@@ -1,35 +1,54 @@
...
+        new UnauthorizedError({
+          message:
+            "API key required. Provide X-API-Key header or Bearer token.",
+        }),
+      );
+    }
+
+    const db = yield* Database;
+
+    const context = yield* db.traces.getEnvironmentContext(
```
willbakst: Why do we need this when `apiKeyInfo` already has org/project/env ID and the owner user?

cloud/api/traces.handlers.ts
```diff
@@ -1,35 +1,54 @@
...
-
-    yield* Effect.log(
-      `[TRACE DEBUG] Received ${totalSpans} spans from service: ${serviceName}`,
-    );
-    yield* Effect.log(
-      `[TRACE DEBUG] Full trace data: ${JSON.stringify(payload, null, 2)}`,
+    const request = yield* HttpServerRequest.HttpServerRequest;
+    const webRequest = request.source as Request;
+
+    const apiKeyInfo = yield* authenticateWithApiKey(webRequest);
```
willbakst: Shouldn't we be handling authentication for `api/v0` routes in the `api.v0.$.ts` handler? We already do it there, so why are we duplicating that here?

cloud/auth/api-key.ts
```diff
@@ -1,8 +1,22 @@
...
+
 /**
  * Header name for API key authentication
  */
 export const API_KEY_HEADER = "X-API-Key";
 
+/**
+ * API key info returned from authentication
+ */
+export type ApiKeyInfo = {
```
willbakst: I feel like we already have `ApiKeyInfo` somewhere (as I've mentioned in my other comments before seeing this because I feel like I added that in my stack?)

willbakst (changes requested):

cloud/api/handler.ts
```diff
@@ -4,34 +4,42 @@ import { ApiLive } from "@/api/router";
...
+
+  const ApiWithDependencies = Layer.merge(
     HttpServer.layerContext,
-    ApiLive.pipe(Layer.provide(services)),
-  );
+    ApiLive,
+  ).pipe(Layer.provide(services));
 
-  return HttpApiBuilder.toWebHandler(ApiWithDependencies);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-argument
```
willbakst: This must never, ever have a type ignore. We need to ensure that this always have perfect typing since forcing the type to `any` is simply hiding bugs.

cloud/auth/api-key.test.ts
```diff
@@ -43,3 +50,38 @@ describe("getApiKeyFromRequest", () => {
...
+
+    expect(result).toEqual(mockApiKeyInfo);
+    expect(result.apiKeyId).toBe("api-key-id");
+    expect(result.environmentId).toBe("env-id");
+  });
+
+  it("should fail with UnauthorizedError when AuthenticatedApiKey context is not provided", async () => {
+    const result = await Effect.runPromise(Effect.either(requireApiKeyAuth));
+
+    expect(result._tag).toBe("Left");
```
willbakst: We shouldn't be using `_tag` as it's private, right? I think we should be using typeguards. For tests where we're looking for errors, we use `Effect.flip` to get the error as the main result. We should be doing that here (and elsewhere)

cloud/auth/utils.ts
```diff
@@ -229,3 +229,58 @@ export const getAuthenticatedUser = (
       ),
     );
   });
+
+/** Auth result including optional API key info. */
+export type AuthenticationResult = {
+  user: PublicUser;
+  apiKeyInfo: ApiKeyInfo | null;
```
willbakst: Instead of `null` we should do `apiKeyInfo?: ApiKeyInfo` and allow `undefined`?

cloud/auth/utils.ts
```diff
@@ -229,3 +229,58 @@ export const getAuthenticatedUser = (
...
+};
+
+/** Authenticate with API key or session and return user + optional ApiKeyInfo. */
+export const authenticate = (
```
willbakst, line 174: This seems duplicated with `getAuthenticatedUser`? I'm fine with renaming it to `authenticate` and `AuthenticationResult`, but let's use that everywhere then and remove the old `getAuthenticatedUser`? We should also consider naming it `AuthResult` since we're likely to use `authResult` as the variable name and "Auth" is a common shorthand for authentication, so it's still clear

cloud/db/database.ts
```diff
@@ -51,7 +51,7 @@ import { ApiKeys } from "@/db/api-keys";
 import { Traces } from "@/db/traces";
 
 /**
- * Type definition for the environments service with nested API keys and traces.
+ * Type definition for the environments service with nested API keys.
```
willbakst: I don't understand this change given traces are still present?

cloud/tests/api.ts
```diff
@@ -105,19 +105,25 @@ type ApiClient = Effect.Effect.Success<typeof makeClient>;
...
-  const ApiWithDependencies = Layer.mergeAll(
+  const ApiWithDependencies = Layer.merge(
     HttpServer.layerContext,
-    ApiLive.pipe(Layer.provide(services)),
-  );
+    ApiLive,
+  ).pipe(Layer.provide(services));
 
-  return HttpApiBuilder.toWebHandler(ApiWithDependencies);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-argument
```
willbakst: Same thing here, do not ignore. If we have to ignore, it means we're doing something else wrong and should rethink

cloud/auth/api-key.ts
```diff
@@ -21,3 +28,21 @@ export function getApiKeyFromRequest(request: Request): string | null {
 
   return null;
 }
+
+/**
+ * Require API key authentication from context.
+ * Fails with UnauthorizedError if not present.
+ */
+export const requireApiKeyAuth: Effect.Effect<ApiKeyInfo, UnauthorizedError> =
```
willbakst: I don't see where this method is used? It seems like we don't actually use it at all anywhere?

cloud/app/routes/api.v0.$.tsx
```diff
@@ -39,6 +39,18 @@ function extractPathParameters(
...
+const API_KEY_REQUIRED_PREFIXES = new Set([
+  "traces",
+  "functions",
+  "annotations",
+]);
+
+function requiresApiKey(splat: string | undefined): boolean {
+  if (!splat) return false;
+  const firstSegment = splat.split("/").filter(Boolean)[0];
+  return firstSegment ? API_KEY_REQUIRED_PREFIXES.has(firstSegment) : false;
```
willbakst: Shouldn't simply calling `API_KEY_REQUIRED_PREFIXES.has(firstSegment)` work without the ternary? Wouldn't a value of `undefined` or `null` still return `false` since it would not be present in the set?

cloud/api/traces.handlers.ts
```diff
@@ -1,35 +1,36 @@
...
 export * from "@/api/traces.schemas";
 
 export const createTraceHandler = (payload: CreateTraceRequest) =>
   Effect.gen(function* () {
-    const serviceName =
-      payload.resourceSpans?.[0]?.resource?.attributes?.find(
-        (attr: KeyValue) => attr.key === "service.name",
-      )?.value?.stringValue || "unknown";
+    const db = yield* Database;
+    const user = yield* AuthenticatedUser;
```
willbakst: Ok yeah this is where we should rethink things. Unlike previous services and APIs, which have e.g. `projectId` or `environmentId` as part of the path, `api/v0/traces` won't have that (and will require that we get them from a proper API key).

One key issue with how you're currently doing it is that we check for the API key in `api.v0.$.ts` and fail with `UnauthorizedError` if no valid API key is present for routes that require one. However, when we do something like `yield* AuthenticatedApiKey`, we end up with a hard requirement for the entire system that we may not be able to provide.

I think what we should do is introduce a unified `Authentication` requirement that will yield `AuthResult` containing `user: PublicUser` and `apiKeyInfo?: ApiKeyInfo`.

We could then add a utility method `yield* Authentication.ApiKey` that returns `Required<AuthResult>`. This would have the same requirement as `Authentication` where `apiKeyInfo` could be undefined (which would resolve our other `any` typing issue), and the `.ApiKey` addition would take `AuthResult` and enforce `Required<AuthResult>`, failing if `apiKeyInfo` is undefined. This would also unify the error handling as part of the requirement rather than as part of `api.v0.$.ts`.

It's also likely worth considering using a custom type rather than `Required` to update `AuthResult` only to require `apiKeyInfo` (so that any additions to `AuthResult` in the future won't fail if API key routes don't need them).

Something like:

```typescript
type RequireField<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>;
```

This would result in something like `RequireField<AuthResult, "apiKeyInfo">`

cloud/auth/utils.ts
```diff
@@ -229,3 +229,58 @@ export const getAuthenticatedUser = (
       ),
     );
   });
+
+/** Auth result including optional API key info. */
+export type AuthenticationResult = {
```
willbakst: `AuthResult` is good here

willbakst:
### Merge activity

* **Jan 6, 2:06 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1607).
* **Jan 6, 2:06 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1607).

PR #1633, opened 2025-12-10
refactor(cloud): refactor UI code to prepare for incremental port of website features

sourishkrout:
### TL;DR

Refactored the structure of the `cloud` react code. Using more standardize conventions, broke out UI component lib, and prepared for bringing in more website pages/resources. Since `/home` will be a route and page in and of itself the old `HomePage` -> `FrontPage`.

I'll be sure to remove legacy stuff, e.g. anything `lilypad`, as part of cleaning up the website's UX/UI.

### What changed?

- Created a new `mirascope-ui` directory with reusable UI components:
    - Added `code-block` component with syntax highlighting and copy functionality
    - Added `copy-button` component for copying content to clipboard
    - Added `lilypad-pricing` component for displaying pricing tiers and feature comparisons
    - Added basic UI components: `button`, `button-link`, `card`, and `tabs`
- Restructured the component organization:
    - Moved error handling components to `core/error`
    - Created a `core` directory with layout, navigation, and provider components
    - Moved dashboard components to a dedicated directory
    - Renamed components to follow PascalCase convention
- Added a new `FrontPage` component and `DashboardLayout` for better page structure
- Updated imports across files to use the new component paths

### How to test?

Everything stays the same. Functionality should not change as a result. This is all about code organization.

### Why make this change?

This restructuring improves code organization and maintainability and migrating in website resoureces by:

1. Creating a dedicated UI component library that can be reused across the application
2. Following consistent naming conventions for components
3. Separating concerns between core functionality, UI components, and page-specific components
4. Providing better user experience with enhanced UI components like code blocks with syntax highlighting
5. Setting up a foundation for future UI development with a more organized structure


sourishkrout:
* **#1633** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1640](https://github.com/Mirascope/mirascope/pull/1640) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1640?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1692](https://github.com/Mirascope/mirascope/pull/1692) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1692?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1633?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

cloud/app/components/front-page.tsx
willbakst, line 1: I wonder if we rename `src` to `app` so we have `"@/app/components"` and `"@/app/contexts"`

And then `mirascope-ui` would just live inside of `app/components`.

cloud/src/components/FrontPage.tsx
```diff
@@ -0,0 +1,34 @@
+import { Button } from "@/mirascope-ui/ui/button";
```
willbakst: Let's also use `front-page.tsx` for `FrontPage` rather than `FrontPage.tsx`.

Same here and elsewhere

cloud/src/components/core/layout/DashboardLayout.tsx
willbakst, line 1: I think rather than `components/core/error` and `components/core/layout` I am imagining `components` is really limited to our shadcn wrapper (namely UI components + blocks that are reusable everywhere). I'm imaging we could then have `components/layouts` where it doesn't need to be nested in `core`. I'd rather be `@/components/errors` than `@/components/core/errors`.

cloud/app/components/layout/app-layout.tsx
```diff
@@ -0,0 +1,15 @@
+import type { ReactNode } from "react";
+
+/**
+ * AppLayout - Very simple layout component that centers the content in the middle of the page.
+ * This is a work in progress and will be expanded in the future.
+ *
+ */
+const AppLayout = ({ children }: { children: ReactNode }) => {
```
willbakst: Hmm, if it's purely the app layout, it likely makes sense to implement this a single time as part of the app component. If this is a layout we plan to re-use, we should consider naming it something else, ideally more specific to what it is doing (e.g. right now it's `CenteredContent` or something? Can also like ly drop `Layout` from the naming convention if it's clear enough in the name itself (e.g. `CenteredContent` is clearly a layout?)

cloud/app/routes/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { createFileRoute } from "@tanstack/react-router";
+import AppLayout from "../components/layout/app-layout";
+
+export const Route = createFileRoute("/")({
+  component: App,
+});
+
+function App() {
+  return (
+    <AppLayout>
```
willbakst: Yeah, I would prefer either `<CenteredContent>` or something like that OR just inline the layout here if it's purely the layout for the app and won't be re-used.

cloud/app/components/ui/button.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { cn } from "@/src/lib/utils";
```
willbakst, line 2: I think there might be a shadcn config we need to update to point to `/app` so that when we add new components it does it correctly?

sourishkrout: Yes, I'm doing this next in a separate PR. However, I have a Q:

Right now the dep chain is `shadcn`\->`Mirascope/ui`\->`Mirascope/website`. Do we want to skip the chain entirely and have the monorepo follow `shadcn`\->`Mirascope/mirascope`?

cloud/app/routes/index.tsx
sourishkrout: Gotcha. Wasn't sure how the cloud dashboard will evolve. What's clear is that website parts (`/home`, `/docs`, etc) will be laid out differently.

cloud/app/components/ui/button.tsx
willbakst: Yes we should consolidate everything into `Mirascope/mirascope`

cloud/app/routeTree.gen.ts
```diff
@@ -9,7 +9,6 @@
 // Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
```
teamdandelion, line 9: nit: per this comment should it have actually not changed? ie did a linter remove the extra line there?

sourishkrout: The pre-commit (husky) linter run changed it. I can dig deeper into why but assumed that changes are fine as long as they are post-linter/-formatter. I've addressed an issue with vscode/cursor not using actually using prettier here: https://app.graphite.com/github/pr/Mirascope/mirascope/1708/chore(cloud)-provide-prettier-default-config-plus-ignore-file-to-avoid-vscode%2Fcursor-using-their-own-defaults.

sourishkrout: I believe the problem here here is that this file is/was in fact not excluded from linting/formatting. I suggest we keep it that way until the heavy development of the `cloud` folder is over and we can adjust lint/formatter settings (disruptive) as per: https://app.graphite.com/github/pr/Mirascope/mirascope/1708/chore(cloud)-provide-prettier-default-config-plus-ignore-file-to-avoid-vscode%2Fcursor-using-their-own-defaults#lines-cloud/.prettierrc-R2-R2. Wdyt @teamdandelion?

sourishkrout: Sending a separate PR to fix the prettier exclusion.

sourishkrout:
### Merge activity

* **Dec 18, 4:15 AM UTC**: @sourishkrout merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1633).

PR #1677, opened 2025-12-12
docs(cloud): improve docs for organization service

willbakst:

PR #1676, opened 2025-12-12
refactor(cloud): add parent scoped where to findAll in base services + clean up and document base service file

willbakst:

PR #1661, opened 2025-12-11
refactor(cloud): update mocks to use test users in project service tests where it makes sense

willbakst:

PR #1660, opened 2025-12-11
refactor(cloud): update mocks to use test users in organization service tests where it makes sense

willbakst:

PR #1678, opened 2025-12-12
feat(cloud): OrganizationMembershipService

willbakst:


willbakst:
* **#1660** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1660?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1661](https://github.com/Mirascope/mirascope/pull/1661) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1661?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1679](https://github.com/Mirascope/mirascope/pull/1679) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1679?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1678** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1677** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1677?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1676** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1676?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1675** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1675?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1674** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

cloud/tests/db.ts
```diff
@@ -393,18 +396,58 @@ export class MockDatabase {
...
+    userId: owner.id,
+    organizationId: org.id,
+    data: { userId: developerUser.id, role: "DEVELOPER" },
   });
 
-  return { org, owner, nonMember };
+  yield* db.organizations.memberships.create({
+    userId: owner.id,
+    organizationId: org.id,
+    data: { userId: viewerUser.id, role: "VIEWER" },
```
teamdandelion, line 442: aside: I often find myself getting a little mixed up when `args.userId` and `args.data.userId` look the same but actually have very different semantics. What would you think of having a convention that `args.userId` is renamed to `args.authenticatedUserId` or `args.activeUserId` to distinguish?

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,488 @@
...
+ * with role-based access control. Memberships define the relationship between
+ * users and organizations, including their roles.
+ *
+ * ## Role Hierarchy
+ *
+ * Organizations support four roles with descending permissions:
+ * - `OWNER` - Full control (create, read, update, delete members)
+ * - `ADMIN` - Administrative access (create, read, update members)
+ * - `DEVELOPER` - Standard access (read members)
+ * - `VIEWER` - Read-only access (read members)
```
teamdandelion: In the context of permissions for the organization membershps table, this is accurate. However the organizations memberships table tracks membership in the organization, and in context of permissions of a member of the organization, it's misleading (developer has update access to organization content). I wonder if this merits clarification

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,488 @@
...
+  /**
+   * Override to use `userId` column instead of `id` for lookups.
+   *
+   * Memberships are uniquely identified by (organizationId, userId),
+   * so we use `targetUserId` as the resource identifier in the path pattern,
+   * which maps to the `userId` column in the database.
+   */
+  protected getIdColumn() {
+    return organizationMemberships.userId;
+  }
```
teamdandelion, line 108: I'm a little confused whyu the path resource identifier deviates from the column name. Could we make the column and path param both be called say `memberId` for consistency?

cloud/tests/db.ts
teamdandelion: Also, i'm a little unclear why its `targetUserId` elsewhere but `userId` here

cloud/db/services/organization-memberships.ts
willbakst: I do this upstack, see #1679

cloud/tests/db.ts
willbakst: It's because I made it `userId` in the table as the column name. I will update this in #1679 so that everything matches with `memberId` (and is distinguished from `userId`, which is the user attempting to do the thing)

cloud/db/services/organization-memberships.ts
willbakst: I mentioned this downstack, but my feeling here is that the permissions should match the local permissions to the service, so for example if a `DEVELOPER` has update access on a resource, that should be documented for that resource (i.e. on the `ProjectMembershipService` or `ProjectService`. For example, a `DEVELOPER` will have `update` access on `ProjectService` but only `read` access on `ProjectMembershipService` (where only `OWNER` and `ADMIN` can update membership)

willbakst: I've also updated this similarly to down stack so `DEVELOPER` and `VIEWER` in this case are the same (because they are)

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,488 @@
...
+            message: "Cannot remove yourself from an organization",
+            resource: "organization membership",
+          }),
+        );
+      }
+
+      const targetRole = yield* this.getRole({
+        userId: targetUserId,
+        organizationId,
+      });
```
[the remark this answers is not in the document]
willbakst: Done

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,527 @@
...
+ *
+ * This module provides authenticated CRUD operations for organization memberships
+ * with role-based access control. Memberships define the relationship between
+ * users and organizations, including their roles.
+ *
+ * ## Role Hierarchy
+ *
+ * Organizations support four roles with descending permissions:
+ * - `OWNER` - Full control (create, read, update, delete)
+ * - `ADMIN` - Administrative access (create, read, update)
```
teamdandelion: I assume that create and update are contextual to the organization membership table (e.g. admin can create a membership, ie. add to organization). But in that case why can admin not delete (ie remove from organization)?

```suggestion
 * - `OWNER` - Full control (create, read, update, delete)
 * - `ADMIN` - Administrative access (create, read, update, delete)
```

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,527 @@
...
+  protected initializeBaseService() {
+    return new OrganizationMembershipBaseService(this.db);
+  }
+
+  protected getPermissionTable(): PermissionTable<Role> {
+    return {
+      create: ["OWNER", "ADMIN"],
+      read: ["OWNER", "ADMIN", "DEVELOPER", "VIEWER"],
+      update: ["OWNER", "ADMIN"],
+      delete: ["OWNER"],
```
teamdandelion: Same question as above — why can admin not delete?

willbakst: Fair. `ADMIN` definitely shouldn't be able to delete `OWNER`. Should `ADMIN` be able to delete another `ADMIN`? Or should it be a strict deletion hierarchy where only `OWNER` can remove `ADMIN` and `ADMIN` can only remove `DEVELOPER` and `VIEWER`?

cloud/db/services/organization-memberships.ts
willbakst: Yeah, `ADMIN` should be able to delete `DEVELOPER` and `VIEWER` at the very least. I can update this.

cloud/db/services/organization-memberships.ts
teamdandelion: Yeah I think a strict hierarchy makes sense, on removing and adding permisisons — i.e. admin cannot add more admins

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,527 @@
...
+        data: { role: data.role },
+      });
+    });
+  }
+
+  /**
+   * Removes a member from an organization.
+   *
+   * Requires OWNER role. Cannot be used to:
+   * - Remove an OWNER (owner is immutable)
```
teamdandelion: Update to owner/admin

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,527 @@
...
+   * - Remove yourself (no self-removal)
```
teamdandelion: I think it should be possible for a user to leave an organization (except for owner role, where they would need to transfer ownership or delete the org). Worth adding now?

cloud/db/services/organization-memberships.ts
```diff
@@ -0,0 +1,527 @@
...
+    PublicOrganizationMembership,
+    AlreadyExistsError | NotFoundError | PermissionDeniedError | DatabaseError
+  > {
+    return Effect.gen(this, function* () {
+      yield* this.authorize({ userId, action: "create", organizationId });
+
+      if (data.role === "OWNER") {
+        return yield* Effect.fail(
+          new PermissionDeniedError({
+            message: "Cannot add an owner to an organization",
```
teamdandelion: Eventually we will need capacity to transfer org ownership. Worth adding now?

cloud/tests/db.ts
willbakst: After furthering the update, I've landed on making it `id` as the ID of the organization membership table. Please take a look at the change up stack and let's discuss the change there.

cloud/db/services/organization-memberships.ts
willbakst: See other comment re: where we should discuss further.

cloud/db/services/organization-memberships.ts
willbakst: Done

cloud/db/services/organization-memberships.ts
willbakst: Done

cloud/db/services/organization-memberships.ts
willbakst: For now I've added an additional `CAN_OPERATE_ON` table as a check because I didn't see a clean way to update `authorize` as implemented to handle this. I've included a `TODO` to handle this as a follow-up.

cloud/db/services/organization-memberships.ts
willbakst: Should users be able to remove themselves? Membership in an organization seems to me like a purely `ADMIN` level action (i.e. a `DEVELOPER` or `VIEWER` shouldn't be able to remove themselves), and no one can operate on `OWNER`, so such a change would only then enable an `ADMIN` to remove themselves from an organization, but it feels like that should be an action limited to `OWNER` since an `ADMIN` can't operate on `ADMIN` roles (including themselves)?

cloud/db/services/organization-memberships.ts
willbakst: I've purposefully chosen to exclude it given we don't need it right now and adding it is purely additive and unlikely to change other things we should finalize and merge in first.

We should add it when we have a need for transfer functionality.

cloud/db/services/organization-memberships.ts
teamdandelion: I think we should be able to support a "Leave Org" or "Leave Team" button, which implies users removing themselves. Down the line we'll probably want to have an "invite to org" and "accept invite" flow, so a user can't create an offensively named organization and then add people to it as a form of harassment. However for now I think the current semantics are okay, so approving as-is.

teamdandelion (approved):

cloud/db/services/organization-memberships.ts
willbakst: Yeah, while we have the services, the exposed endpoints will follow the invite -> accept flow, which we haven't implemented yet.

willbakst:
### Merge activity

* **Dec 15, 10:40 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1678).

PR #1656, opened 2025-12-11
feat: addMember method on ProjectService

willbakst:

PR #1685, opened 2025-12-12
feat(cloud): add project and project_memberships schemas

willbakst:


willbakst:
* **#1588** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1656](https://github.com/Mirascope/mirascope/pull/1656) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1656?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1712](https://github.com/Mirascope/mirascope/pull/1712) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1685** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1685?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1685?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

cloud/db/schema/projects.ts
```diff
@@ -0,0 +1,35 @@
...
+import { organizations } from "./organizations";
+
+export const projects = pgTable("projects", {
+  id: uuid("id").primaryKey().defaultRandom(),
+  name: text("name").notNull(),
+  organizationId: uuid("organization_id")
+    .references(() => organizations.id, { onDelete: "cascade" })
+    .notNull(),
+  createdByUserId: uuid("created_by_user_id")
+    .references(() => users.id)
```
koxudaxi, line 13: What happens if the user who created this has been deleted (e.g., due to resignation)? Null isn't allowed in this case, is it?

cloud/db/schema/projects.ts
```diff
@@ -0,0 +1,35 @@
...
+    references: [organizations.id],
+  }),
+}));
+
+// Internal types
+export type Project = typeof projects.$inferSelect;
+export type NewProject = typeof projects.$inferInsert;
+
+// Public types for API responses
+export type PublicProject = Pick<Project, "id" | "name" | "organizationId">;
```
koxudaxi: Need createdByUserId or createdAt?

cloud/db/schema/project-memberships.ts
```diff
@@ -0,0 +1,57 @@
...
+import { roleEnum } from "./organization-memberships";
+
+export const projectMemberships = pgTable(
+  "project_memberships",
+  {
+    id: uuid("id").primaryKey().defaultRandom(),
+    projectId: uuid("project_id")
+      .references(() => projects.id, { onDelete: "cascade" })
+      .notNull(),
+    userId: uuid("user_id")
```
teamdandelion: consider projectMemberId?

willbakst: I'm going to make the same change here I just made down stack where `id` will be the member user's ID.

I will wait for your comment down stack on that change before applying it here so that I can apply whatever we decide on for consistency.

cloud/db/schema/projects.ts
willbakst: This raises a question of what we need to do on user deletion. Should we truly delete the user and go through all projects and update these references and allow them to be `null`? Should we mark the user as deleted and delete data but keep their UUID / other non-PII necessary info in the user table so that these references don't break?

teamdandelion: I asked Claude, it said it strongly recommends soft deletion but keep the UUID (text below). I agree w. it, particularly for the audit logs. If an attacker gets permissions, onboard new threat actors, and then delete the original attacker to compromise the audit log, it's a mess.

\---

I'd strongly recommend **soft deletion with PII removal** - keep the UUID and essential non-PII metadata, but scrub the personal information. Here's why this is the better approach:

## Why Keep the User Record

**Referential integrity** - You avoid CASCADE nightmares and nullable foreign keys throughout your entire database. Every `createdByUserId` remains valid, which keeps your data model clean.

**Audit trail integrity** - You preserve the history of who did what, which is often legally required or at least extremely valuable. "User 123 created this project" remains true even after they're gone.

**Display handling** - You can show "[Deleted User]" in the UI rather than having to handle null cases everywhere or showing broken references.

## Implementation Pattern

sql

```sql
-- Keep these fields:
id: uuid (unchanged)
createdAt: timestamp (unchanged)
deletedAt: timestamp (set on deletion)
role: string (might need for permission history)

-- Scrub these fields:
email: -> null or 'deleted-{uuid}@deleted'
name: -> 'Deleted User' or null
profileImage: -> null
phoneNumber: -> null
-- any other PII fields
```

...  
**This pattern is what most mature SaaS apps use (Notion, Slack, GitHub all do variations of this). The alternative of true deletion with nullable foreign keys everywhere becomes a maintenance nightmare and makes debugging historical issues nearly impossible.**

cloud/db/schema/project-memberships.ts
teamdandelion: The same change down stack would mean changing it to `memberId`, yeah? If so I'm onboard (approved downstack)

cloud/db/schema/projects.ts
willbakst: Done

cloud/db/schema/projects.ts
koxudaxi: Yeah, I agree with the soft deletion proposal.

willbakst: I've done this up stack in a PR I will be submitting soon.

cloud/db/schema/project-memberships.ts
```diff
@@ -0,0 +1,60 @@
...
+  primaryKey,
+} from "drizzle-orm/pg-core";
+import { projects } from "./projects";
+import { users } from "./users";
+
+export const projectRoleEnum = pgEnum("project_role", [
+  "ADMIN",
+  "DEVELOPER",
+  "READER",
+  "ANNOTATOR",
```
teamdandelion, line 16: Nit: Since it's not alphabetical I expect it to be in permission hierarchy order, but I think annotator would be above reader?

I personally would prefer something more domain agnostic like "ADMIN", "EDITOR", "VIEWER" since we don't yet have any semantic differences between DEVELOPER and VIEWER (and if/when we need such a difference we could add a new role then like ANNOTATION-EDITOR). However I'm flexible on this.  
  
(Maybe change READER to VIEWER even if we are keeping DEVELOPER and ANNOTATOR)

cloud/db/schema/project-memberships.ts
```diff
@@ -0,0 +1,60 @@
...
+export const PROJECT_ROLE_VALUES = projectRoleEnum.enumValues;
+
+export const projectMemberships = pgTable(
+  "project_memberships",
+  {
+    memberId: uuid("member_id")
+      .references(() => users.id)
+      .notNull(),
+    projectId: uuid("project_id")
+      .references(() => projects.id, { onDelete: "cascade" })
```
teamdandelion, line 27: I'd prefer soft delete for projects too. Any particular reason we're doing hard delete?

cloud/db/schema/project-membership-audit.ts
```diff
@@ -0,0 +1,61 @@
...
+import { auditActionEnum } from "./organization-membership-audit";
+
+export const projectMembershipAudit = pgTable("project_membership_audit", {
+  id: uuid("id").primaryKey().defaultRandom(),
+  projectId: uuid("project_id")
+    .references(() => projects.id, { onDelete: "cascade" })
+    .notNull(),
+  actorId: uuid("actor_id")
+    .references(() => users.id)
+    .notNull(),
```
koxudaxi, line 15: nit: Not blocking, but `actor_id` being nullable might give us flexibility for system actions (e.g., automated cleanup, scheduled jobs) in the future.

cloud/db/schema/project-memberships.ts
willbakst: Same response as down stack. Happy to implement if we think this is something urgent. Otherwise I will push to do this in later work so the PRs up stack aren't blocked for much longer.

cloud/db/schema/project-membership-audit.ts
willbakst: Yeah, I think this is something we can address when we have such a case. For now, I think enforcing it's not null is the better choice given there must be an actor?

cloud/db/schema/project-memberships.ts
willbakst: This is in order. ANNOTATOR comes after READER. An ANNOTATOR is someone who will only have permissions to view their annotation queue to annotate data (not full read access).

Happy to update READER to VIEWER if you'd prefer.

cloud/db/schema/project-memberships.ts
willbakst: Same resolution here

cloud/db/schema/project-membership-audit.ts
willbakst: @teamdandelion mentioned we could just add bots as users with their own IDs. Would continue to enforce there being some actor and will make it most clear this way who the actor is even if it's a bot.

cloud/db/schema/project-memberships.ts
willbakst: Done

teamdandelion (approved):

willbakst:
### Merge activity

* **Dec 18, 11:58 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1685).

PR #1686, opened 2025-12-12
feat(cloud): ProjectMembershipService

willbakst:


willbakst:
* **#1588** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1656](https://github.com/Mirascope/mirascope/pull/1656) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1656?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1712](https://github.com/Mirascope/mirascope/pull/1712) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1686** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+  /**
+   * Override to use `userId` column instead of `id` for lookups.
+   *
+   * Memberships are uniquely identified by (projectId, userId),
+   * so we use `memberId` as the resource identifier in the path pattern,
+   * which maps to the `userId` column in the database.
+   */
+  protected getIdColumn() {
+    return projectMemberships.userId;
+  }
```
teamdandelion, line 103: I think if we adopt my suggestion downstack this can be `projectMemberId` without overrides

Though maybe `memberId` is better for consistencyu with org?

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+
+  protected getPublicFields() {
+    return {
+      id: projectMemberships.id,
+      projectId: projectMemberships.projectId,
+      userId: projectMemberships.userId,
```
teamdandelion: Would prefer `memberId` or `projectMemberId`

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+ * their implicit role from the organization.
+ *
+ * ## Permission Matrix
+ *
+ * | Action   | OWNER | ADMIN | DEVELOPER | VIEWER |
+ * |----------|-------|-------|-----------|--------|
+ * | create   | ✓     | ✓     | ✗         | ✗      |
+ * | read     | ✓     | ✓     | ✓         | ✓      |
+ * | update   | ✓     | ✓     | ✗         | ✗      |
+ * | delete   | ✓     | ✗     | ✗         | ✗      |
```
teamdandelion: as downstack: I think admin should have delete

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+  protected initializeBaseService() {
+    return new ProjectMembershipBaseService(this.db);
+  }
+
+  protected getPermissionTable(): PermissionTable<Role> {
+    return {
+      create: ["OWNER", "ADMIN"],
+      read: ["OWNER", "ADMIN", "DEVELOPER", "VIEWER"],
+      update: ["OWNER", "ADMIN"],
+      delete: ["OWNER"],
```
teamdandelion: Likewise

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+            Effect.succeed(null as Role | null),
+          ),
+        );
+
+      // Org OWNER/ADMIN have implicit full access
+      if (orgRole === "OWNER" || orgRole === "ADMIN") {
+        // Verify project exists in this org
+        yield* this.verifyProjectExists(organizationId, projectId);
+        return orgRole;
+      }
```
teamdandelion: If a user is an owner of a project and an admin of the containing org, their role on the project should still be owner, not admin.

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+          return await this.db.transaction(async (tx) => {
+            // Add as org VIEWER if not already a member
+            await tx
+              .insert(organizationMemberships)
+              .values({
+                organizationId,
+                userId: data.userId,
+                role: "VIEWER",
+              })
+              .onConflictDoNothing();
```
teamdandelion: Woah, I strongly disagree with this:

1\. IMO it undermines the whole point of project-specific ACLs - being able to invite an external collaborator to a project without giving them any access to the containing org as a whole  
2\. It allows privilege escalations — someone who is admin on a project but only a viewer on the org should not be allowed to add new viewers to the org  
  
What is the thinking for this behavior?

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+        memberId,
+        data: { role: data.role },
+      });
+    });
+  }
+
+  /**
+   * Removes a member from a project.
+   *
+   * Requires OWNER role. Cannot be used to:
```
teamdandelion: As per downstack — I think admins should be able to remove. I also think self-removal should be possible for non-owners (unless we intend to make this a separate api, but I don't see a need to)

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,678 @@
...
+ * @fileoverview Project membership service for managing user memberships in projects.
+ *
+ * This module provides authenticated CRUD operations for project memberships
+ * with role-based access control. Memberships define the relationship between
+ * users and projects within an organization.
+ *
+ * ## Role Hierarchy
+ *
+ * Projects use the same role hierarchy as organizations:
+ * - `OWNER` - Full control (create, read, update, delete members)
```
teamdandelion: It's not actually clear to me why individual projects should have owners — they are already owned by the containing organization.  
  
What would it mean if, e.g., the project has an owner that is not a member of the containing organization? Who really "owns" it? Who would have the right to transfer the project? I think those permissions should rest with the containing org.  
  
Suggest removing project owner role

cloud/db/services/project-memberships.ts
willbakst: There's some level of required restriction around adding non-members to projects. It introduces a loop-hole where a user on a free plan that only allows for e.g. 2 members could add whoever they want to whatever project without adding them to the org, thereby getting around the restriction.

However, I agree that automatically adding a VIEWER role to the org is not the right move here.

Instead, we should allow adding VIEWER to a proejct without being a part of the organization, but if the person being added has role DEVELOPER or higher and is not part of the organization, this will fail with an error saying they must first add the user to the organization to add them with that role.

What do you think?

cloud/db/services/project-memberships.ts
willbakst: There were a few reasons for this decision:

1. It felt weird not to include the `OWNER` / `ADMIN` that created the project explicitly as a project member. They would of course still have implicit access even if not added, but I figured when looking at the list of project members, it would make sense that the creator would be listed (but other `OWNER` / `ADMIN` role members wouldn't be listed while still having access from the org role permissions).
2. Given (1), it felt weird to downgrade an `OWNER` of an org to an `ADMIN` of a project. I felt it made more sense to match the roles.
3. I also felt that only a project `OWNER` should be able to delete the project while still being able to give another member `ADMIN` level control over the project. This would prevent an `ADMIN` of an organization from deleting a project without `OWNER` permissions.

I view the `OWNER` of a project differently from the containing organization, which doesn't "own" the project but "contains" it.

If we're ok with an org `OWNER` being an `ADMIN` of a project and allowing `ADMIN` full project permission, then I can remove `OWNER` from project.

cloud/db/services/project-memberships.ts
willbakst: I think `memberId` is the better choice for consistency.

teamdandelion: Sg to me, resolve when done

cloud/db/services/project-memberships.ts
teamdandelion: I do not think there should be any coupling of "user must be added to org in order to be added to project" (regardless of role). Instead, just have the free plan limitations propagate down to the project level: if the org owning the project is on the free plan, then either:  
\- count developers on their projects against plan seats in some fashion or  
\- free plan orgs cannot add developers-or-above to their projects (only viewers)

cloud/db/services/project-memberships.ts
teamdandelion: Imo the main reason to add a user explicitly to a project is if you need to give them greater permissions on the project than they would have on the org.  
Eg:  
\- Adding an external collaborator (no org perms) to a project, at any role  
\- Giving write access to an org viewer, or admin access to an org developer or viewer

I don't see any reason to add an org owner or admin to the project membership ACLs since they will already have full project permissions. I don't agree that's its "weird" for an org owner to not be listed on the project membership table. I am not thinking of project membership ACLs as a unit of social tracking / social organization (e.g. an org DEVELOPER who does a lot of development in this project isn't necessarily in this table, even if they are primary dev, if they haven't needed enhanced ACLs).

It's kind of like seeing a shared folder in Google Drive — I don't expect looking at the "shared with" page to list out the org members who are active on it, just the external collaborators (since org members already have perms)  
  
All of this is still consistent w. removing project owner I think

koxudaxi: I agree with Dandelion's proposal to remove the project owner role.

For William's concern about delete permissions, we could restrict project deletion to org OWNER/ADMIN only, without needing a separate project-level owner role. The `createdByUserId` can still be preserved as audit metadata rather than as an ownership concept.

cloud/db/services/project-memberships.ts
koxudaxi: I agree with Dandelion's proposal. This auto-add logic should be removed entirely. Project membership and org membership should be decoupled, with plan limits enforced at the project level instead.

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,774 @@
...
+ * | read     | ✓     | ✓         | ✓      | ✓         |
+ * | update   | ✓     | ✗         | ✗      | ✗         |
+ * | delete   | ✓     | ✗         | ✗      | ✗         |
+ *
+ * ## Security Model
+ *
+ * - Org OWNER/ADMIN have implicit project ADMIN access
+ * - Non-members cannot see that a project exists (returns NotFoundError)
+ * - Cannot modify your own membership
+ * - Cannot remove yourself from a project
```
teamdandelion, line 202: I think users should be able to leave projects (similar reasons as to why they should be able to leave orgs)

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,774 @@
...
+          Effect.catchTag("NotFoundError", () =>
+            Effect.succeed(null as OrganizationRole | null),
+          ),
+        );
+
+      // Org OWNER/ADMIN have implicit project ADMIN access
+      if (
+        userOrganizationRole === "OWNER" ||
+        userOrganizationRole === "ADMIN"
+      ) {
```
teamdandelion, line 317: Hmm, just naming out loud that the default is all org projects are invisible to regular org members (unless they were added specifically). This doesn't match my assumptions from systems like e.g. GitHub where as an org member I expect view rights for all org repos.   
  
I suggest making the default that org members get view access, and if if later we have orgs that need "secret projects" that are secret from their own members, then we handle that separately. The non-discoverability of projects by default seems like it will hamper regular users a lot

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,774 @@
...
+  // ---------------------------------------------------------------------------
+  // CRUD Operations
+  // ---------------------------------------------------------------------------
+
+  /**
+   * Creates a new membership for a user in a project.
+   *
+   * Requires ADMIN role (either via implicit org OWNER/ADMIN or explicit project ADMIN).
+   * Cannot be used to:
+   * - Add yourself to a project (no self-invite)
```
teamdandelion: Hmm, it's interesting that an org owner/admin can't give themselves specific permissions on the project

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,774 @@
...
+        return yield* Effect.fail(
+          new PermissionDeniedError({
+            message: "Cannot add yourself to a project",
+            resource: this.baseService.resourceName,
+          }),
+        );
+      }
+
+      // Enforce organization membership before project membership
+      // TODO: Support external collaborators via invite/collaborator model.
```
teamdandelion, line 432: Unclear to me why we're implementing it with this check vs. just allowing it from the start

cloud/db/services/project-memberships.ts
willbakst: Hmm my understanding is that this isn't the case on GitHub? I thought a member of an organization could only see public repos or repos they are a part of. We should sync and see what the case is here, I'm curious.

I also don't know that I agree that everyone in an organization should have default read access to all projects even if they aren't a project member.

cloud/db/services/project-memberships.ts
willbakst: Happy to loosen the restriction here to allow OWNER/ADMIN to add themselves (since they should have full permissions, which would include this action)

cloud/db/services/project-memberships.ts
willbakst: There's a lot to do around enforcing limits that I didn't want to deal with yet since we don't have any plan levels implemented yet. Things like how the free plan will only allow for 2 users. For now, I figured the best bet was the simplest "must be in org to be project member" that we can loosen once we have a clear picture of these limitations and how we'll implement them.

This includes stuff like the invite email flow. For example, right now we can restrict the UI to show organization members who can be added to a project (that must have been invited and accepted to the org separately), but adding someone not part of the org would require using their email to invite them. I felt it made more sense to allow for external collaborators as part of a push to support external collaborators rather than allowing it

cloud/db/services/project-memberships.ts
willbakst: Confirmed we'll keep it restricted

cloud/db/services/project-memberships.ts
```diff
@@ -0,0 +1,774 @@
...
+  // ---------------------------------------------------------------------------
+
+  protected initializeBaseService() {
+    return new ProjectMembershipBaseService(this.db);
+  }
+
+  protected getPermissionTable(): PermissionTable<ProjectRole> {
+    return {
+      create: ["ADMIN"],
+      read: ["ADMIN", "DEVELOPER", "READER", "ANNOTATOR"],
```
willbakst: ANNOTATOR should not have read permission on project membership

cloud/db/services/project-memberships.ts
willbakst: Done

cloud/db/services/project-memberships.ts
willbakst: Done

teamdandelion (approved):

willbakst:
### Merge activity

* **Dec 19, 12:54 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1686).

PR #1689, opened 2025-12-13
mlx: Support rich messages on transformers based chat templates and harmony

tnadav:
This PR still has some rough edges:
 - Other then automatically inferring the encoder out of the model id - how should we let the user control which encoder to use.
 - How to pass reasoning effort to harmony based model?
 - Some tests fail because they confuse the default model (Qwen3 0.6B - which is tiny)


python/mirascope/llm/providers/mlx/encoding/stream_processors.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @classmethod
+    def qwen3(cls) -> Self:
+        return cls(
+            thinking_wrappers=WrapperTokens.xml_like("think"),
+            tool_call_wrappers=WrapperTokens.xml_like("tool_call"),
+        )
+
+    @classmethod
+    def flat_text(cls) -> Self:
```
willbakst, line 56: This will be the only comment on the missing docstrings. Please do a sweep and add all of the missing docstrings (even on private methods). This includes all files in this change, not just this file.

python/mirascope/llm/providers/mlx/provider.py
```diff
@@ -48,10 +52,27 @@ def client() -> "MLXProvider":
...
 
 
+def _encoder_for_model(
+    model_id: MLXModelId, tokenizer: PreTrainedTokenizer
+) -> BaseEncoder:
+    """Get the appropriate encoder for the given MLX model ID."""
+
+    if model_id == "mlx-community/Qwen3-0.6B-4bit-DWQ-053125":
+        return TransformersEncoder(tokenizer, SpecialTokens.qwen3())
+    elif model_id == "mlx-community/gpt-oss-20b-MXFP4-Q8":
```
willbakst, line 62: Is there a way to more dynamically select the encoder? This seems brittle

tnadav: yeah, thats what I meant in "Other then automatically inferring the encoder out of the model id - how should we let the user control which encoder to use."
perhaps we can come up with a better heuristic to determine these parameters (the advantage of directly comparing the model name is that at least we don't have false positives), but in the end I think the user should be able to pass these properties as a parameter to the model

willbakst: @teamdandelion WDYT? curious what you think would be a good way for a user to specify the encoder. Alternatively, if this only works with `mlx-community` models, we could just codegen the compatibility and for anything else try to detect openai to use harmony and otherwise use transformers?

PR #1707, opened 2025-12-16
feat: support strict structured outputs in anthropic provider

teamdandelion:
This was painful to implement as the strict structured outputs are not
attached to the main client, but instead to a beta client with distinct
types.

Implemented via some conditional logic to decide which client to use,
and a mix of conversion functions (where necssary) and casts (where
feasible)

Tested primarily via e2e coverage (added claude-sonnet-4-5 to all
structured output tests)

PR #1702, opened 2025-12-16
feat(providers): add OpenAICompatibleProvider base class

koxudaxi:
### TL;DR

Refactored provider implementation to introduce a base class for OpenAI-compatible APIs, enabling easier integration with third-party providers.

### What changed?

- Renamed `clients/` directory to `providers/` to better reflect its purpose
- Added `OpenAICompatibleProvider` base class that standardizes implementation for OpenAI-compatible APIs
- Refactored `OpenAICompletionsProvider` to inherit from the new base class
- Added support for Together AI and other OpenAI-compatible providers
- Added provider-specific model name resolution capabilities
- Improved API key handling with environment variable fallbacks
- Added type definitions for OpenAI-compatible provider IDs

### How to test?

1. Verify that existing OpenAI functionality continues to work
2. Test integration with OpenAI-compatible providers like Together AI
3. Run the new test cases in `test_openai_compatible_provider.py`
4. Verify that model name resolution works correctly with different providers

### Why make this change?

This refactoring creates a more maintainable architecture for supporting multiple LLM providers that implement OpenAI-compatible APIs. By extracting common functionality into a base class, we reduce code duplication and make it easier to add new providers. The changes also improve the developer experience by providing clearer error messages and more flexible configuration options.

PR #1696, opened 2025-12-16
refactor: move ensure_additional_properties_false to base utils

teamdandelion:

PR #1713, opened 2025-12-17
feat: introduce AnthropicBetaProvider for beta API support

koxudaxi:
### TL;DR

Added support for Anthropic's beta API to enable strict structured output formatting.

### What changed?

- Created a new `beta` package within the Anthropic provider implementation
- Implemented `AnthropicBetaProvider` class that uses Anthropic's beta API
- Added utility functions for encoding requests and decoding responses with the beta API
- Modified the main `AnthropicProvider` to conditionally use the beta provider when strict formatting is requested
- Updated error messages to include model ID in formatting mode errors
- Reorganized existing Anthropic utilities into a `standard` package

### How to test?

1. Use the Anthropic provider with a model that supports strict structured output formatting:
   ```python
   from mirascope import LLM
   
   class MyOutput(BaseModel):
       name: str
       age: int
   
   llm = LLM(model="anthropic/claude-3-opus-20240229")
   response = llm.format(MyOutput, mode="strict").call("Generate a person's details")
   print(response.content)  # Will be a properly typed MyOutput instance
   ```

2. Verify that appropriate error messages are shown when using models that don't support strict formatting.

### Why make this change?

Anthropic's beta API provides better support for strict structured output formatting, which allows for more reliable parsing of model outputs into Pydantic models. This change enables users to get properly typed outputs from Anthropic models that support this feature, improving type safety and reducing the need for error handling around parsing.


koxudaxi:
* **#1718** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1718?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1702](https://github.com/Mirascope/mirascope/pull/1702) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1702?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1707](https://github.com/Mirascope/mirascope/pull/1707) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1713** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1713?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1713?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1706** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1706?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1705** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1705?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1696** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1696?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1695** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1695?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1694** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1694?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1693** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1693?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/providers/anthropic/beta/_utils/decode.py
```diff
@@ -0,0 +1,255 @@
...
+            "role": response.role,
+            "content": [
+                part.model_dump(exclude_none=True) for part in response.content
+            ],
+        },
+    )
+    stop_reason = response.stop_reason
+    if stop_reason == "model_context_window_exceeded":  # pragma: no cover
+        stop_reason = None
+    finish_reason = BETA_FINISH_REASON_MAP.get(stop_reason) if stop_reason else None
```
teamdandelion: The "setting stop reason to None if it was model_context_window_exceeded" bit was a hack that I did because I was converting the `ParsedBetaMessage` to an `AnthropicMessage`. Now that we have a proper beta provider we should look it up in BETA_FINISH_REASON_MAP just like any other stop reason. I should have added a comment on this to make it clearer :)

python/mirascope/llm/providers/anthropic/beta/_utils/decode.py
```diff
@@ -0,0 +1,255 @@
...
+
+class _BetaChunkProcessor:
+    """Processes Beta stream events and maintains state across events."""
+
+    def __init__(self) -> None:
+        self.current_block_param: BetaContentBlockParam | None = None
+        self.accumulated_tool_json: str = ""
+        self.accumulated_blocks: list[BetaContentBlockParam] = []
+
+    def process_event(self, event: BetaRawMessageStreamEvent) -> ChunkIterator:
```
teamdandelion, line 112: Hmm, it's kind of painful to duplicate all this logic. I am looking at the types and it seems like for the types we actually use here (specifically these:)

```
    RawContentBlockStartEvent,
    RawContentBlockDeltaEvent,
    RawContentBlockStopEvent,
```

... that the types are the same between beta and non beta. The `RawMessageStartEvent` would perhaps vary more because it includes the `Message` vs `BetaMessage` but we don't use it here.  
  
What do you think of re-using the same standard `Chunkprocessor` but wrapping it with a wrapper that converts the `BetaRawMessageStreamEvent` to `RawMessageStreamEvent`, potentially discarding all the events we do not use (ie. just handling content_block_start, content_block_delta, content_block_end)?  
  
I guess for consistency if we do this we should do it in a way that still yields `yield RawStreamEventChunk(raw_stream_event=event)` for every raw event (even the ones we are converting) so it gets a little more complicated. Maybe it is not worth it compared to just having two copies of this stream code. It just means that if we ever change stream processing in Anthropic we have to change it here, which is a little unfortunate, but the e2e tests would help us identify the issue.  
  
I could go either way on this.. what do you think?

python/mirascope/llm/providers/anthropic/beta/_utils/encode.py
```diff
@@ -0,0 +1,231 @@
...
+    temperature: float | Omit
+    top_p: float | Omit
+    top_k: int | Omit
+    stop_sequences: list[str] | Omit
+    thinking: BetaThinkingConfigParam | Omit
+    betas: list[AnthropicBetaParam]
+    output_format: type[BaseModel]
+
+
+def _encode_content(
```
teamdandelion, line 53: We could consider making beta.\_encode_content a wrapper around standard._encode_content that converts all the regular types to beta types, which would allow us to deduplicate a lot of this. this is kind of the approach I took in the original PR, except it would be more tested and robust because we have full e2e testing on the beta provider. what do you think?

python/mirascope/llm/providers/anthropic/beta/_utils/encode.py
```diff
@@ -0,0 +1,231 @@
...
+            content=raw["content"],
+        )
+    return BetaMessageParam(
+        role=message.role,
+        content=_encode_content(message.content, encode_thoughts),
+    )
+
+
+@lru_cache(maxsize=128)
+def _convert_tool_to_tool_param(tool: AnyToolSchema) -> BetaToolParam:
```
teamdandelion, line 87: Likewise this could potentially be a wrapper around the standard version of this function, bc. we don't have any special beta fields

python/mirascope/llm/providers/anthropic/beta/_utils/encode.py
```diff
@@ -0,0 +1,231 @@
...
+    return BetaToolParam(
+        name=tool.name,
+        description=tool.description,
+        input_schema=schema_dict,
+    )
+
+
+def encode_request(
+    *,
+    model_id: AnthropicModelId,
```
teamdandelion: This can actually just take `str`, given how we are explicitly supporting using the `BASE_URL` feature on providers to handle models that may have come from a different developer, there's no real safety or clarity from typing this as AnthropicModelID imo

python/mirascope/llm/providers/anthropic/beta/provider.py
```diff
@@ -0,0 +1,338 @@
...
+        self,
+        *,
+        model_id: str,
+        messages: Sequence[Message],
+        tools: Sequence[Tool] | Toolkit | None = None,
+        format: type[FormattableT] | Format[FormattableT] | None = None,
+        **params: Unpack[Params],
+    ) -> Response | Response[FormattableT]:
+        """Generate an `llm.Response` using the beta Anthropic API."""
+        anthropic_model_id: AnthropicModelId = model_id
```
teamdandelion: Can remove this line if we just type model id as `str` in `encode_request` as suggested above

python/mirascope/llm/providers/anthropic/model_id.py
```diff
@@ -14,9 +14,10 @@ def model_name(model_id: AnthropicModelId) -> str:
...
     Args:
-        model_id: Full model ID (e.g. "anthropic/claude-sonnet-4-5")
+        model_id: Full model ID (e.g. "anthropic/claude-sonnet-4-5" or
+            "anthropic-beta/claude-sonnet-4-5")
 
     Returns:
         Provider-specific model ID (e.g. "claude-sonnet-4-5")
     """
-    return model_id.removeprefix("anthropic/")
+    return model_id.removeprefix("anthropic-beta/").removeprefix("anthropic/")
```
teamdandelion, line 23: aside: As part of the "scope" refactor I will probably update all these functions to just split on the first `/` and return whatever came after. No change required here, this is fine for this PR I will change things consistently later.

python/mirascope/llm/providers/anthropic/provider.py
```diff
@@ -55,32 +75,27 @@ def _call(
...
-            tools: Optional tools that the model may invoke.
-            format: Optional response format specifier.
-            **params: Additional parameters to configure output (e.g. temperature). See `llm.Params`.
-
-        Returns:
-            An `llm.Response` object containing the LLM-generated content.
-        """
-        input_messages, format, kwargs = _utils.encode_request(
+        """Generate an `llm.Response` by synchronously calling the Anthropic Messages API."""
+        if _should_use_beta(model_id, format):  # pragma: no cover
```
teamdandelion, line 79: nit: personally rather than adding these with ` #pragma: no cover`, I would just add this change in the next PR in the stack where we start testing it. No change needed though so long as we remove the `no cover` upstack

python/mirascope/llm/providers/provider_registry.py
```diff
@@ -15,6 +15,7 @@
 # These providers will be automatically registered on first use
 DEFAULT_AUTO_REGISTER_SCOPES: dict[str, ProviderId] = {
     "anthropic/": "anthropic",
+    "anthropic-beta/": "anthropic-beta",
```
teamdandelion, line 18: Hmm I'm not sure about putting this in default_auto_register_scopes since I don't think users will get any benefit from it? Though I guess there is also no harm in it either. I had been imagining that we'd leave it out and just manually register it.

python/tests/e2e/conftest.py
```diff
@@ -14,6 +14,7 @@
 
 E2E_MODEL_IDS: list[llm.ModelId] = [
     "anthropic/claude-sonnet-4-0",
+    "anthropic-beta/claude-sonnet-4-0",
```
teamdandelion, line 17: I would slightly lean towards testing claude-sonnet-4-5 since it's more recent. I've been considering regenerating all the anthropic snapshots to use 4-5  
  
I guess the downside of using 4-5 is it creates a bit of inconsistency with the anthropic tests (at least for now) but the upside is that if it surfaces any unexpected breakages with 4-5 that's actually helpful  
  
Up to you, I"m ok approving this as-is

python/mirascope/llm/providers/anthropic/beta/_utils/decode.py
koxudaxi: I think it's acceptable to tolerate duplicated code rather than making the logic overly complex.

python/mirascope/llm/providers/anthropic/beta/_utils/encode.py
koxudaxi: OK, I will implement a wrapper.

python/tests/e2e/conftest.py
koxudaxi: The changes would be too extensive, so let's keep it at 4.0 for this PR for now.

willbakst:
The description code example is hilarious

python/mirascope/llm/providers/anthropic/provider.py
```diff
@@ -55,32 +75,27 @@ def _call(
         format: type[FormattableT] | Format[FormattableT] | None = None,
         **params: Unpack[Params],
     ) -> Response | Response[FormattableT]:
-        """Generate an `llm.Response` by synchronously calling the Anthropic Messages API.
-
-        Args:
```
willbakst, line 60: What's with the removal of the args section everywhere? Are we decided we don't want it?

python/mirascope/llm/providers/anthropic/standard/__init__.py
willbakst, line 1: I feel like if this is purely utilities then it should remain in `_utils` and we should instead have something shared utils (and match our naming convention we use for OpenAI)

python/mirascope/llm/providers/load_provider.py
```diff
@@ -30,6 +30,8 @@ def load_provider(
     match provider_id:
         case "anthropic":
             return AnthropicProvider(api_key=api_key, base_url=base_url)
+        case "anthropic-beta":
```
willbakst, line 33: Hmm, should we be exposing this as a provider? Or is this purely something we use under the hood of `AnthropicProvider` to support beta features without our users dealing with the pain of separate interfaces?

python/tests/e2e/output/test_call_with_thinking_true.py
```diff
@@ -11,6 +11,7 @@
 E2E_MODEL_IDS: list[llm.ModelId] = [
```
willbakst, line 11: @teamdandelion is this different from the one in `conftest.py` because it's limited to models / providers / APIs that support thinking?

python/mirascope/llm/providers/anthropic/provider.py
teamdandelion: I actually support getting rid of these docstrings (and have been thinking of doing it in a separate across-the-board PR). Now that the docstrings are on internal helper methods like `_call` there's no end-user benefit to having them, and they feel like highly repetitive boilerplate.

python/mirascope/llm/providers/load_provider.py
teamdandelion: Yeah I raised a similar quesiton elsewhere. I lean towards not exposing it as a provider since it's just an internal helper. We can add it later if we find a reason why users would want to directly access the anthropic-beta provider.

python/tests/e2e/output/test_call_with_thinking_true.py
teamdandelion: Yes though tbh we should be able to remove this special casing and just get a llm.FeatureNotSupported error for gpt-5:completions

python/mirascope/llm/providers/anthropic/standard/__init__.py
teamdandelion: Having slept on it I think I lean towards not having `anthropic/standard/` but instead having:  


```
anthropic/
  provider.py # AnthropicProvider
  beta_provider.py # AnthropicBetaProvider
  utils/
    encode.py # standard encode
    beta_encode.py # beta encode
    decode.py # standard decode
    beta_decode.py # beta decode
```

And just letting beta_encode and beta_decode import from regular encode where needed

teamdandelion: I made a PR upstack that does this.

python/mirascope/llm/providers/load_provider.py
teamdandelion: I made a PR upstack that changes this.

python/mirascope/llm/providers/anthropic/provider.py
koxudaxi: The args description to remove it was because it was a private method starting with `_`.

willbakst (approved):

teamdandelion:
### Merge activity

* **Dec 18, 7:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1713).
* **Dec 18, 7:21 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1713).

PR #1715, opened 2025-12-17
refactor(cloud): Users (@effect/sql-drizzle)

willbakst:

PR #1714, opened 2025-12-17
refactor(cloud): add initial effect/sql-drizzle infra for migration

willbakst:

PR #1700, opened 2025-12-16
fix(cloud): soft delete users and scrub PII

willbakst:

PR #1699, opened 2025-12-16
fix(cloud): add cascade delete so deleting an organization deletes memberships/audits

willbakst:

PR #1716, opened 2025-12-17
refactor(cloud): Sessions (@effect/sql-drizzle)

willbakst:


willbakst:
* **#1600** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1601](https://github.com/Mirascope/mirascope/pull/1601) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1601?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1738](https://github.com/Mirascope/mirascope/pull/1738) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1738?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1603** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1602** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1599** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1598** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1598?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1597** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1597?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1596** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1595** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1594** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1594?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1593** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1593?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1592** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1592?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1591** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1735** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1734** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1734?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1733** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1733?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1731** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1731?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1730** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1730?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1729** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1729?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1725** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1724** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1716** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1716?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1715** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1714** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1590** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1590?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1655](https://github.com/Mirascope/mirascope/pull/1655) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1655?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1650** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1649** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1589** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1589?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1712** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1588** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1656](https://github.com/Mirascope/mirascope/pull/1656) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1656?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `v2`


<h2></h2>

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+                message: "Failed to find session",
+                cause: e,
+              }),
+          ),
+        );
+
+      if (!results[0]) {
+        return yield* Effect.fail(
+          new NotFoundError({
+            message: `session with sessionId ${sessionId} not found`,
```
teamdandelion: Nit: The message should probably include userId, since it's possible the sessionId exists but the userId is incorrect

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+  /**
+   * Finds a session by its ID.
+   *
+   * @param userId - The user's ID (for path consistency)
+   * @param sessionId - The session's ID
+   * @returns The session if found
+   * @throws NotFoundError - If the session doesn't exist
+   * @throws DatabaseError - If the database operation fails
+   */
+  findById({
```
teamdandelion, line 150: I think we should implement `findBySessionId`, and have `findById` wrap `findBySessionId`, but with an extra layer of logic that checks if the session is owned by a different userId, and returns an error in that case.  
  
`findBySessionId` can centralize the delete-expired-session logic and potentially return `InvalidSessionError`. If we have `findById` and `findUserBySessionId` both calling `findBySessionId`, then we would have consistent expired-session-deletion semantics across the board.

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+                message: "Failed to update session",
+                cause: e,
+              }),
+          ),
+        );
+
+      if (!results[0]) {
+        return yield* Effect.fail(
+          new NotFoundError({
+            message: `session with sessionId ${sessionId} not found`,
```
teamdandelion: Nit: Error message should probably include the userId, since the sessionId may be fine

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+                message: "Failed to delete session",
+                cause: e,
+              }),
+          ),
+        );
+
+      if (!results[0]) {
+        return yield* Effect.fail(
+          new NotFoundError({
+            message: `session with sessionId ${sessionId} not found`,
```
teamdandelion: Likewise

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+
+      // Fetch session with user data
+      const results = yield* client
+        .select({
+          id: users.id,
+          email: users.email,
+          name: users.name,
+          userId: sessions.userId,
+          expiresAt: sessions.expiresAt,
+        })
```
teamdandelion: I don't like that this is not using `Users.findById` — it means that it has inconsistent semantics, ie.g. this one can return deleted users (which I think is the correct behavior, but inconsistent with the current impl of `Users.findById` but it doesn't return `users.deletedAt`. I think we should be composing with the `Users` service not implementing something new here.

So I suggest we use the `findBySessionId` method I suggest above, use that to get the user id, and get all the user data via the `Users.findById` interface, and then combine the data as appropriate for the signature.

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+
+  /**
+   * Deletes a session (hard delete).
+   *
+   * @param userId - The user's ID
+   * @param sessionId - The session's ID
+   * @throws NotFoundError - If the session doesn't exist
+   * @throws DatabaseError - If the database operation fails
+   */
+  delete({
```
teamdandelion, line 246: this should call `deleteBySessionId`

cloud/db/sessions.test.ts
```diff
@@ -44,114 +45,148 @@ describe("SessionService", () => {
...
   });
 
-  describe("findAll", () => {
-    it.effect("finds all sessions", () =>
+  // ===========================================================================
+  // findById
+  // ===========================================================================
+
+  describe("findById", () => {
+    it.effect("finds a session by id", () =>
```
teamdandelion, line 160: We should test the case where the session exists, but the userId is incorrect

cloud/db/sessions.test.ts
```diff
@@ -313,37 +353,35 @@ describe("SessionService", () => {
...
-     * 1. fetchSessionWithUser (select join)
-     * 2. handleEmptyResult (NotFoundError if no rows)
-     * 3. handleExpiredSession (delete + InvalidSessionError if expired)
-     * 4. transformToPublicUser (return user)
-     */
+  // ===========================================================================
+  // findUserBySessionId
+  // ===========================================================================
 
+  describe("findUserBySessionId", () => {
```
teamdandelion, line 505: We should test the case where the user has been deleted

cloud/db/sessions.test.ts
teamdandelion, line 160: Also, we should test the case where the session is expired

cloud/db/sessions.test.ts
```diff
@@ -44,114 +45,148 @@ describe("SessionService", () => {
...
   });
 
-  describe("findById", () => {
-    it.effect("finds a session by id", () =>
+  // ===========================================================================
+  // findAll
+  // ===========================================================================
+
+  describe("findAll", () => {
+    it.effect("finds all sessions for a user", () =>
```
teamdandelion, line 66: We should test expired sessions (are they returned or auto-deleted?)

cloud/db/sessions.test.ts
```diff
@@ -44,114 +45,148 @@ describe("SessionService", () => {
...
+      ),
     );
   });
 
+  // ===========================================================================
+  // update
+  // ===========================================================================
+
   describe("update", () => {
     it.effect("updates a session", () =>
```
teamdandelion, line 283: we should test the case where the session is already expired

cloud/db/sessions.ts
willbakst: Done

cloud/db/sessions.ts
willbakst: Done

cloud/db/sessions.ts
willbakst: Done

cloud/db/sessions.ts
willbakst: Should it? This scopes by `userId`, so if this receives a valid session ID but the wrog user ID it won't delete it. If we use `deleteBySessionId` we lose that.

cloud/db/sessions.ts
willbakst: But then we would have an additional DB call for no reason?

The "delete-expired-session" logic will likely just be a `deleteExpiredSessions` helper method that a CRON job calls.

I'm generally viewing these as CRUD helpers (in line with your comments down stack on Users where we return deleted users and leave it to the consumer to handle deleted users).

cloud/db/sessions.ts
willbakst: We should sync on this, because the purpose of this refactor was to be able to do exactly what we're doing here without it feeling like a lot of repeated code.

teamdandelion: I don't follow. Why is writing it this way (w. potentially inconsistent semantics, including not including deletedAt field) desirable over re-using an already defined and tested interface?

cloud/db/sessions.ts
teamdandelion: The reason would be minimizing the surface area for testing (or potential bugs) by having a smaller number of re-usable methods that touch the db  
  
I'm not convinced we should be perf optimizing the number of db calls at this juncture (when we don't have any load, don't have other basic perf features like pagination or batching).

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+          }),
+        );
+      }
+
+      const sessionWithUser = results[0];
+
+      // Check if session is expired
+      if (new Date() > sessionWithUser.expiresAt) {
+        // Delete the expired session
+        yield* this.delete({ userId: sessionWithUser.userId, sessionId }).pipe(
```
willbakst: Decision: no methods will auto-delete, we will leave this to a CRON job

cloud/db/sessions.ts
```diff
@@ -0,0 +1,398 @@
...
+            message: "Session expired",
+            sessionId,
+          }),
+        );
+      }
+
+      return {
+        id: sessionWithUser.id,
+        email: sessionWithUser.email,
+        name: sessionWithUser.name,
```
willbakst: We should be including `deletedAt` in `PublicUser` to easily identify if a user returned is deleted

willbakst: I've updated `PublicUser` down stack in the `Users` PR, so `PublicUser` has forced the update here.

cloud/db/sessions.ts
willbakst: Since no method deletes, this becomes moot (and everything will be unified under the CRON job we'll implement at some point)

cloud/db/sessions.ts
willbakst: As discussed sync, it's because these services are wrappers around `DrizzleORM`. If it makes sense to re-use an existing method, we definitely should. One big reason for this refactor is to enable using existing service-level metehods inside of transactions.

There is a comment up stack for `Organizations` to this effect around using `OrgainzationMemberships` to add the membership and auti log rather than re-implement. That's a case where using `OrganizationMemberships.create` is definitely the right move.

Here, however, we're not just finding a user, we're specifically running `innerJoin` with `sessions` to find the user with the given `sessionId`. The benefit of `DrizzleORM` being Effect-native is that we can now write such queries in a way where I'm not as concerned about refactoring.

One thing I do think worth doing here is improving the typing. If we enforce a `UserWithSession` type that extends `PublicUser`, we can ensure consistency around e.g. including the `deletedAt` field.

Another thing I think worth doing is returning `NotFoundError` if the session is expired (but not deleting the session, inline with other comments).

cloud/db/sessions.ts
willbakst: As discussed in sync, these are different methods. `deleteBySessionId` is really just a convenience for `logout` where we can simply send the session ID from the cookie.

cloud/db/sessions.ts
willbakst: As I'm working through these changes, I also think that in the case of `findUserBySessionId` we should return `NotFoundError` if the user is soft-deleted. For `findById`, I agree it makes sense we should return a soft-deleted user given we're searching by ID, but `findUserBySessionId` is really a helper method intended to handle the logic around producing an authenticated user. Since a soft-deleted user likely shouldn't be authenticated, I think `NotFoundError` or maybe something more specific like `DeletedUserError` would make more sense.

I'm opting for `DeletedUserError` since I think being more specific here is worthwhile.

cloud/db/sessions.ts
willbakst: Done

cloud/db/sessions.ts
willbakst: Done

willbakst:
### Merge activity

* **Dec 22, 6:47 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1716).

PR #1750, opened 2025-12-22
refactor(python-sdk): improve missing import handling for optional packages

willbakst:
Makes stubbing optional dependencides significatly less brittle by automatically stubbing everything for a declared module.

This means that we can add whatever we want to a module without worrying that it will be properly stubbed should the required dependencies for that module not be installed.

This changes implements this for `api`, `ops`, and `llm.providers`.

I've tested the change using the following configuration:

```
dependencies = ["mirascope[anthropic]"]

[tool.uv.sources]
mirascope = { git = "https://github.com/Mirascope/mirascope", branch = "ENG-338", subdirectory = "python" }
```

I confirmed proper error messages for uninstalled providers and confirmed success when necessary dependencies are installed.


willbakst:
* **#1750** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

python/tests/llm/providers/test_missing_dependencies.py
```diff
@@ -0,0 +1,438 @@
...
+            # Call find_spec directly to test the "already stubbed" path
+            spec = _finder.find_spec("test_finder_stub.nested")
+            assert (
+                spec is None
+            )  # Should return None because it's already in sys.modules
+
+            # Test that finder returns None for non-stubbed modules
+            # This exercises the final return None in find_spec
+            spec = _finder.find_spec("totally.unrelated.module")
+            assert spec is None
```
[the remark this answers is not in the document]
willbakst: Done

python/mirascope/_stubs.py
```diff
@@ -0,0 +1,358 @@
...
+from typing import Any
+
+# ruff: noqa: ANN401
+
+# Mapping of extra names to their import names
+# IMPORTANT: Keep this in sync with [project.optional-dependencies] in pyproject.toml
+# When adding/modifying extras, update this mapping accordingly
+# Note: Multiple PyPI packages may provide the same namespace (e.g., opentelemetry-api
+# and opentelemetry-sdk both provide 'opentelemetry'), so we only check once.
+EXTRA_IMPORTS: dict[str, list[str]] = {
```
koxudaxi, line 16: Can we validate the dict with pyproejct.toml in CI?

koxudaxi: Or, auto-generation the dict.

python/tests/llm/providers/test_missing_dependencies.py
```diff
@@ -0,0 +1,461 @@
...
+        key
+        for key in sys.modules
+        if key.startswith("mirascope.llm.providers.")
+        and not key.startswith("mirascope.llm.providers.base")
+    ]
+    for module in modules_to_remove:
+        del sys.modules[module]
+
+
+def stub_package(package_name: str) -> None:
```
koxudaxi: Who do use the function?

willbakst: Removed

python/mirascope/_stubs.py
willbakst: Done

python/lint-staged.ts
```diff
@@ -31,6 +31,9 @@ try {
   // Run pyright on the whole project (doesn't work well with individual files)
   await $`uv run pyright .`.cwd("./python");
 
+  // Validate EXTRA_IMPORTS is up-to-date
+  await $`./scripts/validate_extra_imports.sh`.cwd("./python");
```
teamdandelion: Hmm, on my machine this increases the time for running lint-staged from about 13s (which is already far too high imo) to about 15s. I believe this will fail only very infrequently (when updating deps or adding new providers or such?) so in contrast to pyright or codespell, I don't see the value in including it in lint staged vs just having it show up in CI.

python/scripts/validate_extra_imports.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Validate that EXTRA_IMPORTS in _stubs.py matches what would be generated
+
+set -e
+
+echo "Generating EXTRA_IMPORTS mapping..."
```
teamdandelion: IMO a validation script like this should produce no output on success, and only echo on failure. Seeing incremental output like that it's generating a mapping or such may be useful during development, but afterwards it is just noise.

python/mirascope/llm/providers/__init__.py
```diff
@@ -1,5 +1,20 @@
 """Interfaces for LLM providers."""
 
+from ..._stubs import stub_module_if_missing
+
+# Stub modules for missing optional dependencies BEFORE importing
```
teamdandelion, line 5: This is a really nice improvement, thanks. The really annoying thing about my previous solution was that if I broke the code such that there was _actually_ an import error, then the true issue would get suppressed and I'd learn about it through downstream breakages that were harder to debug.

python/scripts/generate_extra_imports.py
```diff
@@ -0,0 +1,240 @@
...
+                # For namespace packages, the shortest unique path is usually correct
+                sorted_paths = sorted(paths, key=len)
+                common: tuple[str, ...] = sorted_paths[0]
+
+                for path in sorted_paths[1:]:
+                    if path[: len(common)] != common:
+                        # Find common prefix
+                        common = tuple(
+                            a for a, b in zip(common, path, strict=True) if a == b
+                        )
```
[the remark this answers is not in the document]
willbakst: Done, but I set it to `strict` since the arg is required silly

python/scripts/validate_extra_imports.sh
willbakst: Done

willbakst:
### Merge activity

* **Jan 8, 11:27 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1750).

PR #1778, opened 2025-12-29
(orphan pr): closed / ignore

teamdandelion:
Primary goal: Make format.mode an optional parameter so we can convert
FormattableT to Format in the model call before going to provider. This
is facilitated if formatting_instructions is a property, computed
dynamically, rather than at format construction time.

Also it's cleaner this way and avoids circular imports.

PR #1777, opened 2025-12-29
refactor: move Format into format.py, consolidate utilities to dataclass

teamdandelion:
Given it is a full dataclass (and I want to add a property) it makes
more sense to put it in format.py than types.py

PR #1779, opened 2025-12-29
refactor: convert FormattableT to Format at the model layer

teamdandelion:
This will allow us to determine if necessary call args are provided at
the model layer

### DIFF
--- a/python/mirascope/llm/models/models.py
+++ b/python/mirascope/llm/models/models.py
@@ -9,7 +9,7 @@ from typing import overload
 from typing_extensions import Unpack
 
 from ..context import Context, DepsT
-from ..formatting import Format, FormattableT
+from ..formatting import Format, FormattableT, format as format_fn
 from ..messages import Message, UserContent
 from ..providers import (
     ModelId,
@@ -203,6 +203,8 @@ class Model:
         Returns:
             An `llm.Response` object containing the LLM-generated content.
         """
+        if not isinstance(format, Format) and format is not None:
+            format = format_fn(format)
         return self.provider.call(
             model_id=self.model_id,
             messages=messages,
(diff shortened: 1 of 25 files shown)


teamdandelion:
* **#1781** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1781?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1780** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1779** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1779?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1779?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1777** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1777?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1778](https://github.com/Mirascope/mirascope/pull/1778) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1778?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `v2`


<h2></h2>

python/mirascope/llm/models/models.py
```diff
@@ -203,6 +203,8 @@ def call(
         Returns:
             An `llm.Response` object containing the LLM-generated content.
         """
+        if not isinstance(format, Format) and format is not None:
```
willbakst, line 206: Where was this happening before / why am I not seeing that removed in this PR?

teamdandelion: It was happening in resolve_format, which previously served two roles:  
\- Final step conversion of FormattableT to Format (if user passed in FormattableT)  
\- Ensure a mode is present (if user did not set a mode)

I have since refactored to be clearer and replaced resolve_format with ensure_mode_set

This PR is maybe not actually an improvement (since it makes Format.mode nullable when in its final state it is never null) so maybe we don't include this unless we need it

PR #1780, opened 2025-12-29
feat: add Format.non_generated_fields

teamdandelion:
Format now tracks all its fields marked FromCallArgs so we can ensure at
model call time that the necessary args are present.

Added validation to disallow nested FromCallArgs

PR #1781, opened 2025-12-29
feat: add support for llm.FromCallArgs

teamdandelion:


teamdandelion:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1781?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1781** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1780** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1779** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1777** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1777?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1778](https://github.com/Mirascope/mirascope/pull/1778) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1778?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `v2`


<h2></h2>

python/mirascope/llm/models/models.py
```diff
@@ -192,34 +240,48 @@ def call(
...
 
+        # TODO: Either move call args handling into provider.call (so they return fully
+        # fledged Responses), or make the provider.call methods private (so we do not
+        # expose something that is typed as though it returns a full response, but is
+        # actually missing call args)
+        response.from_call_args = from_call_args_data
+
+        return response
+
     @overload
```
teamdandelion, line 277: Curious for your thoughts here. Given that right now we do expose Provider as a public concept, and `Provider.call` exists, it seems wrong that `Provider.call` can't handle the call args case. However, I like the design where `Model` is a unified spot where we handle Call Args (and potentially other future transformations), veruss having to move call args processing into every provider and ensure it's implemented consistently.  
  
The simplest thing could be to modify the Provider interface so that either the `Provider.call` methods are all renamed private (as `Provider._call` or similar) or we just move the `Call` methods all onto a private subclass

python/mirascope/llm/models/models.py
```diff
@@ -153,13 +155,57 @@ def __exit__(
...
+
+        # Extract and validate JSON-serializability via round-trip
+        from_call_args_data: dict[str, Jsonable] = {}
+        for field_name in format.from_call_args_fields:
+            value = call_args[field_name]
+            try:
+                # Validate it's JSON-serializable via round-trip
+                from_call_args_data[field_name] = json.loads(json.dumps(value))
+            except (TypeError, ValueError) as e:
+                raise ValueError(
```
willbakst, line 195: I feel like this is something we should leave to e.g. Pydantic where it will already handle this?

python/mirascope/llm/models/models.py
```diff
@@ -192,34 +240,48 @@ def call(
...
+            call_args: Optional call arguments to populate FromCallArgs fields.
 
         Returns:
             An `llm.Response` object containing the LLM-generated content.
         """
         if not isinstance(format, Format) and format is not None:
             format = format_fn(format)
-        return self.provider.call(
+
+        from_call_args_data = self._extract_from_call_args(format, call_args)
```
willbakst, line 259: nit: one thing on naming here, I think `_extract_from_call_args` is not the right name. I view `FromCallArgs` as a marker for a field that should be populated from a call argument. In this case, I view the method moreso as `_extract_call_argument_fields` or something?

I feel the same way about similar naming elsewhere. `call_arg_fields` or `call_argument_fields` seems more accurate?

python/mirascope/llm/models/models.py
willbakst: Let's sync on `FromCallArgs` in general

PR #1800, opened 2026-01-01
feat(cloud): initial router for Anthropic, Google, and OpenAI

willbakst:


willbakst:
* **#1802** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1802?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1801** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1801?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1800** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1804** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1804?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

cloud/api/router/proxy.ts
```diff
@@ -0,0 +1,208 @@
...
+    const targetUrl = `${config.baseUrl}/${path}${url.search}`;
+
+    // Copy request headers, excluding user auth and host
+    const headers = new Headers();
+    request.headers.forEach((value, key) => {
+      const lowerKey = key.toLowerCase();
+      if (lowerKey !== "authorization" && lowerKey !== "host") {
+        headers.set(key, value);
+      }
+    });
```
teamdandelion: I prefer a functional style eg:

```typescript
const headers = new Headers(
  Array.from(request.headers.entries()).filter(
    ([key]) => !["authorization", "host"].includes(key.toLowerCase())
  )
);
```

Or:

```typescript
const headers = new Headers(
  Object.entries(Object.fromEntries(request.headers))
    .filter(([key]) => {
      const lowerKey = key.toLowerCase();
      return lowerKey !== "authorization" && lowerKey !== "host";
    })
);
```

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,321 @@
...
+   * Maps provider names to models.dev provider IDs.
+   *
+   * Can be overridden by subclasses for providers with multiple IDs.
+   */
+  protected getProviderIds(): string[] {
+    const PROVIDER_ID_MAP: Record<string, string[]> = {
+      openai: ["openai"],
+      anthropic: ["anthropic"],
+      google: ["google", "google-ai-studio"],
+    };
```
teamdandelion: Hmm, can we centralize this?

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,321 @@
...
+          const cost = providerData?.models[modelId]?.cost;
+          if (cost) {
+            return {
+              input: cost.input,
+              output: cost.output,
+              cache_read: cost.cache_read,
+              cache_write: cost.cache_write,
+            };
+          }
+        }
```
teamdandelion: This is weird, what about when the same model is present for multiple providers with different pricing? This won't impact us immediately when we are just doing openai/google/anthropic, but it is foreseeable and a non-edge-case.

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    const cacheReadCost =
+      usage.cacheReadTokens && pricing.cache_read
+        ? (usage.cacheReadTokens / 1_000_000) * pricing.cache_read
+        : undefined;
+
+    const cacheWriteCost =
+      usage.cacheWriteTokens && pricing.cache_write
+        ? (usage.cacheWriteTokens / 1_000_000) * pricing.cache_write
+        : undefined;
```
teamdandelion: FYI: models.dev uses the 5m cache pricing. Which I think is a good default, but we should be aware that if our users are doing 1hr cache writes then we will lose money

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,321 @@
...
+              outputCost: 0,
+              totalCost: 0,
+            },
+            formattedCost: {
+              input: "N/A",
+              output: "N/A",
+              total: "N/A",
+            },
+          };
+        }
```
teamdandelion: nit: I would move the `if (!pricing)`handling into `this.calculateCost`

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,216 @@
...
+    string,
+    {
+      id: string;
+      name: string;
+      cost?: {
+        input: number;
+        output: number;
+        cache_read?: number;
+        cache_write?: number;
+      };
```
teamdandelion: Redeclaration of the ModelPricing interface

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+/**
+ * Maps our provider names to models.dev provider IDs
+ */
+const PROVIDER_ID_MAP: Record<string, string[]> = {
+  openai: ["openai"],
+  anthropic: ["anthropic"],
+  google: ["google", "google-ai-studio"], // Check both
+};
```
teamdandelion: Let's centralize this

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,216 @@
...
+      const cost = providerData?.models[modelId]?.cost;
+      if (cost) {
+        return {
+          input: cost.input,
+          output: cost.output,
+          cache_read: cost.cache_read,
+          cache_write: cost.cache_write,
+        };
+      }
+    }
```
teamdandelion: This is duplicated in cost-calculator.ts, and shares the same issue (see other comment)

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,216 @@
...
+    inputCost + outputCost + (cacheReadCost || 0) + (cacheWriteCost || 0);
+
+  return {
+    inputCost,
+    outputCost,
+    cacheReadCost,
+    cacheWriteCost,
+    totalCost,
+  };
+}
```
teamdandelion, line 208: Duplicate with a protected method in cost-calculator.ts. I think we keep this copy and delete the otehr.

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Formats a cost in USD as a string.
+ */
+export function formatCost(cost: number): string {
```
teamdandelion: I think formatCost should take a `CostBreakdown` and return a `FormattedCostBreakdown`, or something like that. Rather than inlining the 5 calls redundantly whenever we need formatted costs.

cloud/api/router/proxy.ts
```diff
@@ -0,0 +1,208 @@
...
+    baseUrl: "https://api.anthropic.com",
+    authHeader: "x-api-key",
+    authFormat: (key: string) => key,
+  },
+  google: {
+    baseUrl: "https://generativelanguage.googleapis.com",
+    authHeader: "x-goog-api-key",
+    authFormat: (key: string) => key,
+  },
+} as const;
```
teamdandelion: Let's centralize this along with the other per-provider config.

cloud/app/routes/router.v0.$provider.$.tsx
```diff
@@ -0,0 +1,155 @@
...
+          // Authenticate user via Mirascope API key
+          yield* getAuthenticatedUser(request);
+
+          // Get provider-specific API key from environment
+          const providerApiKey =
+            provider === "openai"
+              ? process.env.OPENAI_API_KEY
+              : provider === "anthropic"
+                ? process.env.ANTHROPIC_API_KEY
+                : process.env.GEMINI_API_KEY;
```
teamdandelion: Let's centralize this.

cloud/app/routes/router.v0.$provider.$.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+        const handler = Effect.gen(function* () {
+          if (!databaseUrl) {
+            return yield* new InternalError({
+              message: "Database not configured",
+            });
+          }
+
+          // Validate provider
+          if (!["openai", "anthropic", "google"].includes(provider)) {
```
teamdandelion: Check the central registry.

cloud/app/routes/router.v0.$provider.$.tsx
```diff
@@ -0,0 +1,155 @@
...
+            });
+          }
+
+          // Extract model from request body for cost calculation
+          const requestBody = yield* Effect.tryPromise({
+            try: () => request.clone().text(),
+            catch: () => null as string | null,
+          });
+
+          let modelId: string | null = null;
```
teamdandelion, line 79: Practically speaking I think we need to track something like (scope, modelId) or (provider, modelId)

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,321 @@
...
+      cacheReadCost,
+      cacheWriteCost,
+      totalCost,
+    };
+  }
+
+  /**
+   * Formats a cost in USD.
+   */
+  protected formatCost(cost: number): string {
```
teamdandelion: Duplicate code

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,321 @@
...
+    case "openai":
+      return new OpenAICostCalculator();
+    case "anthropic":
+      return new AnthropicCostCalculator();
+    case "google":
+      return new GoogleCostCalculator();
+    default:
+      return null;
+  }
+}
```
teamdandelion: Consider centralizing this

cloud/package.json
```diff
@@ -85,11 +85,13 @@
     "zod": "^4.2.0"
   },
   "devDependencies": {
+    "@anthropic-ai/sdk": "^0.32.1",
```
teamdandelion: It's not clear to me why the router depends on the sdk since we are just proxying. These imports don't seem to be used in this PR.

willbakst: I had originally planned on writing an example using the SDKs to test things, but I ended up just implementing `MirascopeProvider` up stack instead. Removed.

cloud/api/router/cost-calculator.ts
willbakst: It doesn't quite make sense to do this. I find that having `calculateCost` take `ModelPricing` makes more sense than `ModelPricing | null`, and this also handles the `formattedCost` case, so pushing this into `calculateCost` would also require updating `formatCost` to take to all zeroes `CostBreakdown` and check if it's all zeroes to return `N/A`, but that's not correct. What we're doing here is handling the case where `getModelPricing` does not return valid `ModelPricing`, in which case we early return. I'm opting not to change this.

cloud/api/router/pricing.ts
willbakst: I've deduped. See response to other comment on why I haven't made any other changes.

cloud/api/router/cost-calculator.ts
willbakst: As written, this is confusing. I think the update makes more sense now since instead of `this.getProviderIds()` it's `getProviderIds(provider)`. You'll notice that Google has more than one potential ID that we need to check against. Now that `getModelPricing` takes `provider` explicitly as an argument, it's far more clear that this isn't testing multiple providers but rather testing multiple potential IDs for a single provider.

cloud/app/routes/router.v0.$provider.$.tsx
willbakst: I'm not quite sure what you mean here by track? We have `provider` already, and we use that?

cloud/api/router/cost-calculator.ts
willbakst: Turns out the 1h cache pricing is 1.6x the 5m cache pricing, so since the detailed usage is available we can actually normalize to a single value.

I've implemented this change so we have proper pricing.

cloud/app/routes/router.v0.$provider.$.tsx
teamdandelion: Didn't realize we already have provider.

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,239 @@
...
+ * CACHE PRICING NORMALIZATION:
+ * Anthropic provides detailed cache breakdown via usage.cache_creation:
+ * - ephemeral_5m_input_tokens: 5-minute TTL caches (cost 1.25x input price)
+ * - ephemeral_1h_input_tokens: 1-hour TTL caches (cost 2.0x input price per Anthropic docs)
+ *
+ * We normalize these to a single cacheWriteTokens value by converting 1h tokens to
+ * 5m-equivalent: 1h_normalized = 1h_tokens × (2.0 / 1.25) = 1h_tokens × 1.6
+ *
+ * This allows calculateCost() to use models.dev's 5m pricing (1.25x) for all cache writes,
+ * while ensuring correct costs: 5m stays 1:1, 1h effectively costs 2.0x input price.
```
teamdandelion, line 148: There's an issue here that when we show cost breakdowns to users, these cache write token counts may be inaccurate, which could confuse users or reduce their trust in our system when they compare the tokens we billed for vs the token usage that other instrumentation.

I don't think we need to fix this now since we are focused on getting this out the door, but may be worth a TODO suggesting we improve this later (eg by tracking a computed pricing multiplier for each token field, which defaults to 1 if unset)

cloud/api/router/cost-calculator.ts
```diff
@@ -0,0 +1,239 @@
...
+    // This way, multiplying total by 5m price gives correct cost for both
+    let cacheWriteTokens: number;
+    const ephemeral5mTokens =
+      usage.cache_creation?.ephemeral_5m_input_tokens || 0;
+    const ephemeral1hTokens =
+      usage.cache_creation?.ephemeral_1h_input_tokens || 0;
+
+    if (ephemeral5mTokens > 0 || ephemeral1hTokens > 0) {
+      // Normalize: 5m tokens stay 1:1, 1h tokens scaled by 1.6
+      cacheWriteTokens = ephemeral5mTokens + ephemeral1hTokens * 1.6;
```
teamdandelion: Maybe factor this out to a constant variable on the AnthropicCostCalculator?

cloud/api/router/providers.ts
```diff
@@ -0,0 +1,142 @@
...
+export function getCostCalculator(provider: ProviderName): BaseCostCalculator {
+  switch (provider) {
+    case "openai":
+      return new OpenAICostCalculator();
+    case "anthropic":
+      return new AnthropicCostCalculator();
+    case "google":
+      return new GoogleCostCalculator();
+  }
+}
```
teamdandelion, line 143: We should have an exhaustiveness check for type safety when we add new providers

```typescript
export function getCostCalculator(provider: ProviderName): BaseCostCalculator {
  switch (provider) {
    case "openai":
      return new OpenAICostCalculator();
    case "anthropic":
      return new AnthropicCostCalculator();
    case "google":
      return new GoogleCostCalculator();
    default:
      // This ensures exhaustiveness checking
      const exhaustiveCheck: never = provider;
      throw new Error(`Unsupported provider: ${exhaustiveCheck}`);
  }
}
```

cloud/api/router/providers.ts
```diff
@@ -0,0 +1,142 @@
...
+  switch (provider) {
+    case "openai":
+      return process.env.OPENAI_API_KEY;
+    case "anthropic":
+      return process.env.ANTHROPIC_API_KEY;
+    case "google":
+      return process.env.GEMINI_API_KEY;
+    default:
+      return undefined;
+  }
```
teamdandelion: Why is `provider: string` not `ProviderName`? And we can add an exhaustiveness check for type safety, e.g. `const exhaustiveCheck: never = provider;` in the default case

cloud/api/router/providers.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Maps our provider names to models.dev provider IDs.
+ * Some providers may have multiple IDs in models.dev (e.g., google and google-ai-studio).
+ */
+export const PROVIDER_ID_MAP: Record<string, string[]> = {
+  openai: ["openai"],
+  anthropic: ["anthropic"],
+  google: ["google", "google-ai-studio"],
+};
```
teamdandelion: Why not include this in the PROVIDER_CONFIGS, as `modelsDevIds: string[]`

cloud/api/router/providers.ts
```diff
@@ -0,0 +1,142 @@
...
+  authHeader: string;
+  /** Format for the auth header value (e.g., "Bearer {key}" or just "{key}") */
+  authFormat: (key: string) => string;
+}
+
+/**
+ * Provider configurations for proxying.
+ * Defines base URLs and authentication patterns for each provider.
+ */
+export const PROVIDER_CONFIGS = {
```
teamdandelion: This should be explicitly typed as containing `ProxyConfig`s

willbakst: Not quite, but I've done essentially this

cloud/api/router/cost-calculator.ts
willbakst: Yeah, I originally preserved everything, but that really muddied things up (since provider-agnostic code like `calculateCost` had Anthropic-specific handling, which I did not like).

I figured we would do things like this:

1. Normalize 1h cache write input tokens so that `calculateCost` continues to be provider-agnostic
2. Update `BaseCostCalculator` to use Effect, depend on `Database`, and actually store the breakdown for display. Since this would be happening inside of each cost calculator (e.g. as part of `extractUsage` and `calculate`), we'll still have access to the original, full, raw breakdown (which we can properly store and display).

I've added a TODO around this so we can revisit when it makes sense to do so.

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,231 @@
...
+      return pricingCache;
+    }
+
+    // Fetch fresh data
+    return yield* fetchPricingData();
+  });
+}
+
+/**
+ * Looks up pricing for a specific model.
```
teamdandelion, line 110: Nit: let's at least mention models.dev in these docs for discoverability

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,231 @@
...
+    pricingCache = data;
+    lastFetchTime = Date.now();
+    return data;
+  });
+}
+
+/**
+ * Gets cached pricing data, fetching if stale or missing.
+ */
+export function getPricingData() {
```
teamdandelion: nit: suggest rename to getModelsDotDevPricingData

cloud/api/router/pricing.ts
```diff
@@ -0,0 +1,231 @@
...
+ */
+export function clearPricingCache(): void {
+  pricingCache = null;
+  lastFetchTime = 0;
+}
+
+/**
+ * Fetches pricing data from models.dev/api.json
+ */
+export function fetchPricingData() {
```
teamdandelion: nit: suggest rename to fetchModelsDotDevPricingData

cloud/api/router/providers.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+/**
+ * Gets the provider IDs used in models.dev for a given provider.
+ *
+ * @param provider - The provider name
+ * @returns Array of provider IDs, or [provider] as fallback
+ */
+export function getProviderIds(provider: ProviderName): string[] {
```
teamdandelion: nit: rename to getModelsDotDevProviderIds  
  
Otherwise ProviderId and ProviderName seem confusing and potentially interchangeable

teamdandelion (approved):

willbakst:
### Merge activity

* **Jan 7, 12:12 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1800) rebased this pull request as part of a merge.
* **Jan 7, 12:14 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1800).

PR #1803, opened 2026-01-01
feat(cloud): meter usage and track against credits to ensure sufficient funds

willbakst:


willbakst:
* **#1803** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1803?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1803?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1836** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1835** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1835?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

cloud/app/routes/router.v0.$provider.$.tsx
```diff
@@ -59,7 +63,34 @@ export const Route = createFileRoute("/router/v0/$provider/$")({
...
+          // Check if user has sufficient credits (minimum $1)
+          const balance = yield* payments.customers.getBalance(
+            organization.stripeCustomerId,
+          );
+
+          if (balance < 1.0) {
+            return yield* new UnauthorizedError({
+              message: `Insufficient credits. Required: $1.00, Available: $${balance.toFixed(2)}`,
+            });
+          }
```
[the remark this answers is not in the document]
willbakst: Done

cloud/payments/metering.ts
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Parameters for charging router usage.
+ */
+export interface ChargeRouterUsageParams {
+  /** Provider name (e.g., "openai", "anthropic", "google") */
+  provider: ProviderName;
+  /** Model ID for cost calculation */
+  model: string;
+  /** Usage data from the provider's response */
+  usageData: unknown;
```
teamdandelion: Why is usageData: unknown?

cloud/payments/metering.ts
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * If cost calculation fails, the error is logged but the meter is not charged.
+   * If meter charging fails, the error is logged but the request continues.
+   *
+   * @param params - Usage parameters
+   * @returns Effect that succeeds when metering is complete (or skipped due to errors)
+   */
+  chargeForRouterUsage(
+    params: ChargeRouterUsageParams,
+  ): Effect.Effect<void, never, Payments> {
```
teamdandelion: Why don't we surface the errors and let caller handle them as appropriate? Hiding them feels weird to me.

cloud/payments/metering.ts
```diff
@@ -0,0 +1,174 @@
...
+        costLog.cacheWrite = result.formattedCost.cacheWrite;
+      }
+
+      console.log({
+        provider,
+        model,
+        streaming,
+        usage: usageLog,
+        cost: costLog,
+      });
```
teamdandelion: Is this meant to stay in for production?

cloud/payments/metering.ts
```diff
@@ -0,0 +1,174 @@
...
+          }),
+        );
+
+      console.log({
+        metering: {
+          baseCost: result.formattedCost.total,
+          chargedAmount: `$${(totalCost * 1.05).toFixed(6)}`,
+          gasFee: "5%",
+        },
+      });
```
teamdandelion: Is this meant to stay in production? Also this should reference the gas fee constant elsewhere.

cloud/payments/metering.ts
```diff
@@ -0,0 +1,174 @@
...
+      const outputTokens = result.usage.outputTokens;
+
+      if (
+        inputTokens === undefined ||
+        outputTokens === undefined ||
+        isNaN(inputTokens) ||
+        isNaN(outputTokens)
+      ) {
+        return;
+      }
```
teamdandelion: I think this numeric validation should be a helper function (and elsewhere here). Ideally we would encode it in the type signature (like `NonnegativeRealNumber`) and it would be typed as such in the `usage` object so we know it was validated at construction time and we never need check it again. I think Typescript makes this feasible.

cloud/payments/metering.ts
willbakst: No it's not, I'll remove as part of the split

cloud/payments/metering.ts
willbakst: I will type this better

cloud/payments/metering.ts
willbakst: No it's not, I'll remove as part of the split

cloud/payments/metering.ts
willbakst: I will look into a better solution

cloud/payments/metering.ts
willbakst: Will try to do this

cloud/payments/metering.ts
willbakst: The thing here is that we don't want to crash the request if we fail to meter, so we need to suppress the errors. My updated approach is more clear, and I've added some logging in the error cases at least so we can look into what happened.

Ultimately we'll want to add a queue that's robust with retries (and is async). Koudai is working on this for ClickHouse integration, so I'll add a TODO and update once that gets merged in and we have a baseline for using queues

cloud/payments/metering.ts
willbakst: I tried this, but it's just a false sense of security. If we do something like `asNonNegative`, we can type-guard to `NonNegativeNumber`, but `asNonNegative` will take `number` as it's input, so `asNonNegative(-500)` will typecheck, but at runtime it will throw an error. It's significnatly more annoying to do it this way than to just handle this case in the single place we need to in metering

cloud/api/router/streaming.ts
```diff
@@ -65,26 +96,29 @@ function parseNDJSONLine(line: string): unknown {
...
  * @param format - The streaming format ("sse" or "ndjson")
  * @param provider - The provider name for usage extraction
+ * @param meteringContext - Context for automatic cost calculation and fund settlement
  * @returns Effect that resolves to StreamParseResult with response and usage data
  */
 export function parseStreamingResponse(
   response: Response,
   format: "sse" | "ndjson",
   provider: ProviderName,
+  meteringContext: StreamMeteringContext,
```
teamdandelion, line 329: Per suggestion downstack: This should return an effect stream of token usage events, and we can compose that with the metering context in a separate function, rather than complecting it with the stream parsing.

willbakst: Yeah, whatever changes I make downstack I will bubble up here

cloud/api/router/streaming.ts
```diff
@@ -63,78 +122,299 @@ function parseNDJSONLine(line: string): unknown {
...
       };
-    })(),
-  );
+    }
+
+    const costCalculator = getCostCalculator(meteringContext.provider);
+
+    // Create Refs to track parsing state
+    const usageRef: Ref.Ref<TokenUsage | null> =
+      yield* Ref.make<TokenUsage | null>(null);
```
teamdandelion, line 344: Ok, so our semantics here are we always use the last usage chunk, i.e. trusting that the providers either only provide one usage chunk (anthropic, openai) or provide many usage chunks but with accumulating token counts so we can just keep the last one (google). I think this is correct for all our supported providers but maybe worth documenting

teamdandelion (approved):

cloud/api/router/streaming.ts
willbakst: Yeah exactly. I'll add a note to this effect. Worth noting that Anthropic actually does accumulation according to their docs

willbakst:
### Merge activity

* **Jan 9, 1:06 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1803).
* **Jan 9, 1:06 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1803).

PR #1834, opened 2026-01-07
feat(cloud): add cost estimation for router requests

willbakst:

PR #1833, opened 2026-01-07
feat(cloud): improve cost calculators

willbakst:

PR #1805, opened 2026-01-03
feat(cloud): enable purchasing credits

willbakst:

PR #1835, opened 2026-01-07
feat(cloud): add credit reservations schema and migration

willbakst:


willbakst:
* **#1836** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1835** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1835?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1835?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

cloud/db/schema/credit-reservations.ts
```diff
@@ -0,0 +1,127 @@
...
+ * - Use the expires_at index for efficient querying
+ */
+export const creditReservations = pgTable(
+  "credit_reservations",
+  {
+    id: uuid("id").primaryKey().defaultRandom(),
+    stripeCustomerId: text("stripe_customer_id").notNull(),
+    organizationId: uuid("organization_id")
+      .references(() => organizations.id, { onDelete: "cascade" })
+      .notNull(),
```
teamdandelion: Hmm, the reservations having a cascading delete on organization delete gives me pause. It makes me wonder about a case like:

\- Attacker finds a vulnerability in our credit reservation / pricing system (that allows them to overspend tokens)  
\- They run the attack by creating an org, abusing credit balances, and then quickly deleting org (perhaps using API to cycle through orgs rapidly)  
\- Deleting the org covers their tracks by deleting associated logs, making it more difficult for us to reconstruct what happened.

I propose refactoring so the creditReservation directly reference the stripeCustomerId and reference the organizationId only transitively, i.e. adding a new table that tracks the organizationId <-> stripeCustomerId relation, and you go from credit reservation to an organizationId via a join.  
  
Then we structure things so that deleting the org does not delete the stripeCustomerId, or any of the associated financial records. I think this is basically necessary so that our system has financial audit integrity.

cloud/db/schema/credit-reservations.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Cost tracking
+    estimatedCost: numeric("estimated_cost", {
+      precision: 10,
+      scale: 6,
```
teamdandelion: - Why is scale 6, not 4? Is the idea future proofing in case we need to go below the "centi-cent" level? I have a preference we just treat the centi-cent billing as canonical and commit to scale: 4
- Can we factor out these constants somewhere to ensure that all of our cost numerics are consistent? (And then we can also have one comment documenting the choice.) I know these will not be the last fields in the db that do cost tracking. The precision maybe wants to vary depending on domain (e.g. we have a max value of 9999.999999 here, which is OK for a reservation but too low for monthly or even daily billing). But we should have the scale be consistent everywhere IMO.
- I think the simplest is to go with something like precision: 12, scale: 4 in which case we get a max value $100M (less one centi-cent). As we approach $100M in total revenue we can refactor ;) or even precision: 14 scale 4, which future proofs us to $10B in total revenue. Then we can just have it be consistent everywhere and not spend more time talking about these choices

cloud/db/schema/credit-reservations.ts
teamdandelion: Discussed with Claude, it would go further:  
  
**My (Claude's) recommendation:**

The reviewer's approach is good, but I'd go further:

1. **Never cascade delete financial data** - Use SET NULL or RESTRICT
2. **Consider soft deletes** for orgs if not already implemented
3. **Add an append-only audit log** for all credit operations
4. **Set up alerts** for unusual patterns (rapid org creation/deletion)

This is the kind of review comment that prevents future "how did we lose $X?" investigations. Definitely worth implementing their suggestion or something similar.  
  
Not sure we need to do all of this, but mentioning it for consideration. I think the "never delete financial data" premise is a good one.

cloud/db/migrations/0001_strong_gamora.sql
```diff
@@ -0,0 +1,20 @@
...
+	"estimated_cost" numeric(10, 6) NOT NULL,
+	"actual_cost" numeric(10, 6),
+	"status" "reservation_status" DEFAULT 'active' NOT NULL,
+	"request_id" text,
+	"model" text,
+	"provider" text,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"settled_at" timestamp,
+	"released_at" timestamp,
+	"expires_at" timestamp NOT NULL
```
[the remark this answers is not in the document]
teamdandelion: I don't really agree with Graphite's take here, since as it notes the insert would fail, and this should all be happening through our prod system anyway — making it easy to directly insert credit reservations that didn't go through prod is not a goal. Also directly editing the migration seems bad too. Suggest no action

cloud/db/schema/credit-reservations.ts
teamdandelion: Follow on thought: What if we store all cost values as integer centi-cents (in db and in our TS code) and just convert in UI and when interacting with stripe?

willbakst: I don't think we even need to convert when interacting with stripe since we can just 1:1 use cost in centicents as our metering unit.

I've updated things to operate in centicents

cloud/db/schema/credit-reservations.ts
willbakst: I'm opting to remove `organizationId` and just use `stripeCustomerId`. This should cover everything that we need here short term, and we can add additional stuff later as we deem it necessary

cloud/db/schema/credit-reservations.ts
```diff
@@ -0,0 +1,155 @@
...
+      mode: "bigint",
+    }),
+
+    // Status tracking
+    status: reservationStatusEnum("status").default("active").notNull(),
+
+    // Request metadata for debugging
+    requestId: text("request_id"),
+    model: text("model"),
+    provider: text("provider"),
```
teamdandelion: I don't think this belongs on the reservation. I propose  a requests table that tracks (to start):   
\- status (active, failed, succeeded)  
\- model  
\- provider  
\- API key  
\- cost  
\- created_at  
\- usage   
  
Then the reservation can reference the request. This also makes it possible for us to systematically run comparison like "how well does our estimated cost correlate to actual cost for successful requests for provider X", as well as system metadata like "whats our aggregate token throughput per provider and per model"?

cloud/db/schema/credit-reservations.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    // Timestamps
+    createdAt: timestamp("created_at").defaultNow().notNull(),
+    settledAt: timestamp("settled_at"),
+    releasedAt: timestamp("released_at"),
```
teamdandelion: Why do we need these as separate timestamps? Let's just have `resolvedAt` which will be set whether its settled, released, or marked expired.

willbakst: We don't, I'm consolidating on `releasedAt`

cloud/db/schema/credit-reservations.ts
willbakst: Creating a separate `RouterRequests` table for request info (like provider, model, cost, etc.) as you suggest

cloud/db/schema/credit-reservations.ts
```diff
@@ -0,0 +1,155 @@
...
+    // Cost tracking (in centi-cents: 1 = $0.0001 USD)
+    estimatedCostCenticents: bigint("estimated_cost_centicents", {
+      mode: "bigint",
+    }).notNull(),
+
+    // Status tracking
+    status: reservationStatusEnum("status").default("active").notNull(),
+
+    // Foreign key to router_requests table (nullable - not set for expired reservations)
+    routerRequestId: uuid("router_request_id"),
```
teamdandelion: From the docs on router_requests table:  
  
\>  * This table stores ALL router requests (successful or failed) with complete  
\>  * metadata for observability, analytics, and cost reconciliation.  
  
Given that all router requests are stored, why is this nullable?

cloud/db/schema/credit-reservations.ts
```diff
@@ -0,0 +1,155 @@
...
+      table.status,
+    ),
+    // Index for future CRON job to expire old active reservations
+    expiresAtIndex: index("credit_reservations_expires_at_index")
+      .on(table.expiresAt)
+      .where(sql`${table.status} = 'active'`),
+  }),
+);
+
+export const creditReservationsRelations = relations(
```
teamdandelion, line 139: Add a relation to the requests table?

cloud/db/schema/router-requests.ts
```diff
@@ -0,0 +1,138 @@
...
+ * See credit-reservations.ts for detailed explanation.
+ */
+export const routerRequests = pgTable(
+  "router_requests",
+  {
+    id: uuid("id").primaryKey().defaultRandom(),
+
+    // Provider metadata
+    provider: text("provider").notNull(),
+    model: text("model").notNull(),
```
teamdandelion, line 65: May want to migrate to an enum and maybe reference to a model table down the line, text seems fine for now though.

cloud/db/schema/router-requests.ts
```diff
@@ -0,0 +1,138 @@
...
+    costCenticents: bigint("cost_centicents", {
+      mode: "bigint",
+    }),
+
+    // Request status
+    status: requestStatusEnum("status").default("pending").notNull(),
+    errorMessage: text("error_message"),
+
+    // Organizational hierarchy (for scoping and analytics)
+    organizationId: uuid("organization_id").notNull(),
```
teamdandelion, line 90: How about we store the apiKeyId? And then we can lookup the organizationId, projectId, environmentId using joins rahter than putting extra uuids on every request

cloud/api/router/cost-calculator.ts
```diff
@@ -152,27 +191,22 @@ export class OpenAICostCalculator extends BaseCostCalculator {
...
- * We normalize these to a single cacheWriteTokens value by converting 1h tokens to
- * 5m-equivalent: 1h_normalized = 1h_tokens × (2.0 / 1.25) = 1h_tokens × 1.6
- *
- * This allows calculateCost() to use models.dev's 5m pricing (1.25x) for all cache writes,
- * while ensuring correct costs: 5m stays 1:1, 1h effectively costs 2.0x input price.
+ * - ephemeral_1h_input_tokens: 1-hour TTL caches (cost 2.0x input price)
  *
- * TODO: At some point we will likely want to display the per-request cost-breakdown for users.
- * We'll need to update the cost calculators to store this information at that time.
+ * We store the ACTUAL token counts (not normalized) in cacheWriteTokens and
```
teamdandelion, line 199: Nice

cloud/api/router/cost-calculator.ts
```diff
@@ -197,34 +230,65 @@ export class AnthropicCostCalculator extends BaseCostCalculator {
...
+    // Store ACTUAL total tokens (no normalization)
+    // If detailed breakdown is available, use it; otherwise fall back to aggregate field
+    const cacheWriteTokens =
+      ephemeral5m + ephemeral1h || usage.cache_creation_input_tokens || 0;
+
+    // Store breakdown for accurate cost calculation (only if detailed breakdown exists)
+    const cacheWriteBreakdown =
+      ephemeral5m > 0 || ephemeral1h > 0
+        ? { ephemeral5m, ephemeral1h }
+        : undefined;
```
teamdandelion, line 246: I think the vast majority of cases will be 5 min cache (per default), so we could change this to only write the breakdown if `ephemeral1h > 0`, and in the case where there is no breakdown, we infer all the tokens were 5min writes

cloud/db/schema/credit-reservations.ts
willbakst: We create a reservation before we actually make the request (and thus don't have the ID)

willbakst: Oh hmm actually I guess we could put the request in a pending state and have that always linked

cloud/api/router/cost-calculator.ts
willbakst: I found it easier to reason about it this way. If we don't have `ephemeral > 0` explicitly, it forces a user to have to reason about what's happening far more than just seeing what's written here.

cloud/db/schema/router-requests.ts
willbakst: What happens when a user deletes an API key? I feel like deleting an API key will be common whereas deleting an organization / project / environment won't be, which is why I opted for this

cloud/db/schema/credit-reservations.ts
willbakst: Done

cloud/db/schema/credit-reservations.ts
willbakst: Done

teamdandelion (approved):

willbakst:
### Merge activity

* **Jan 8, 7:23 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1835).

PR #1836, opened 2026-01-07
feat(cloud): implement fund reservation management for router billing

willbakst:


willbakst:
* **#1836** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1836?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1835** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1835?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+ * A value of 0.05 represents a 5% fee.
+ */
+const GAS_FEE_PERCENTAGE = 0.05;
+
+/**
+ * Default router usage meter unit price in cents.
+ * This is the expected fallback value if Stripe price retrieval fails.
+ * A value of 0.01 represents $0.0001 per unit.
+ */
+export const ROUTER_USAGE_METER_UNIT_PRICE = 0.01;
```
teamdandelion: 1. Curious why we are tracking this in cents, not dollars? 
2. "This is the expected fallback vlaue if Stripe price retrieval fails" - why not just hardcode this value as canonical? Why would we want to get something different back from stripe?

teamdandelion: If we go with the "everything in centi cents" idea, then I believe this would just be 1 and we should not need it?

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+export class Router {
+  /**
+   * Gets the router price unit amount in cents.
+   *
+   * Retrieves the price configuration from Stripe and returns the unit amount
+   * (or unit_amount_decimal) in cents. Falls back to ROUTER_USAGE_METER_UNIT_PRICE
+   * if price doesn't have a unit amount configured.
+   *
+   * This is used for meter value calculations to ensure they stay in sync with
+   * Stripe's pricing configuration.
```
teamdandelion: Can you explain this in more depth? What would we be changing on stripe's end, and why?  
What would the effect be if stripe got out of sync with ROUTER_USAGE_METER_PRICE?  
  
If those values were to diverge, then that seems like it would put us in undefined behavior as far as mis-billing customers, so it makes me uncomfortable that we're trying to get a value and falling back to a hardcoded constant.

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+
+  /**
+   * Gets the accumulated meter usage balance for a Stripe customer.
+   *
+   * Fetches meter event summaries for the current billing period and returns
+   * the total usage in dollars. This represents how much the customer has used
+   * in the current billing cycle but not yet been invoiced for.
+   *
+   * @param customerId - The Stripe customer ID
+   * @returns The accumulated meter usage in dollars (e.g., 5.00 for $5 usage)
```
teamdandelion: Confusing to me that elsewhere we track dollar avlues in cents but here we return dollars. #CentiCentAllTheThings

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+  ): Effect.Effect<number, StripeError, Stripe> {
+    return Effect.gen(this, function* () {
+      const stripe = yield* Stripe;
+
+      // Get active subscription to find current billing period
+      const subscriptions = yield* stripe.subscriptions.list({
+        customer: customerId,
+        status: "active",
+        limit: 1,
+      });
```
teamdandelion: re limit: 1 — is this guaranteed by stripe, by our code, or not guaranteed? Seems safer to get all subscriptionswith status active and reduce them (which should also handle length === 0 without special casing)

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+        )
+          continue;
+
+        const { value, currency } = grant.amount.monetary;
+        if (currency === "usd") {
+          totalCredits += value / 100;
+        }
+      }
+
+      return totalCredits;
```
teamdandelion: Suggest rewriting as a filter/reduce

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+   */
+  getAvailableBalance(
+    customerId: string,
+  ): Effect.Effect<number, StripeError, Stripe> {
+    return Effect.gen(this, function* () {
+      const creditBalance = yield* this.getCreditBalance(customerId);
+      const meterUsage = yield* this.getUsageMeterBalance(customerId);
+      return creditBalance - meterUsage;
+    });
+  }
```
teamdandelion: Why doesn't this include reservations?

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+              new DatabaseError({
+                message: "Failed to calculate active reservations",
+                cause: error,
+              }),
+          ),
+        );
+
+      const activeReservationsTotal = parseFloat(
+        activeReservationsResult[0]?.sum ?? "0",
+      );
```
teamdandelion: I feel like this logic should be folded into getAvailableBalance, is there some reason it needs to happen here?

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+    return Effect.gen(this, function* () {
+      const db = yield* DrizzleORM;
+
+      // Get available balance
+      const availableBalance = yield* this.getAvailableBalance(customerId);
+
+      // Calculate active reservations
+      const activeReservationsResult = yield* db
+        .select({
+          sum: sql<string>`COALESCE(SUM(${creditReservations.estimatedCost}), 0)`,
```
teamdandelion: I wonder if this is gonna be a performance hotspot and whether we should have a sql table that stores the current reserved balance. Can add it later in response to testing tho.

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+
+      const activeReservationsTotal = parseFloat(
+        activeReservationsResult[0]?.sum ?? "0",
+      );
+
+      // Check if sufficient funds available
+      const netAvailable = availableBalance - activeReservationsTotal;
+      if (netAvailable < estimatedCost) {
+        return yield* new InsufficientFundsError({
+          message: `Insufficient available funds. Required: $${estimatedCost.toFixed(2)}, Available: $${netAvailable.toFixed(2)} (Balance: $${availableBalance.toFixed(2)}, Reserved: $${activeReservationsTotal.toFixed(2)})`,
```
teamdandelion: Ok I see how we want reserved and available separately here, but I actually think we should just enhance the return signature on getBalance, adding comment to that effect there...

cloud/payments/products/router.ts
teamdandelion: Let's have this return {creditBalance, meterUsage, reservedBalance, availableBalance}  
Where availableBalance is convenient precomputed `creditBalance - meterUsage - reservedBalance`

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+   */
+  reserveFunds({
+    customerId,
+    organizationId,
+    estimatedCost,
+    requestId,
+    model,
+    provider,
+  }: {
+    customerId: string;
```
teamdandelion: Please rename to stripeCustomerId (distinguishes from e.g. userId)

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+    organizationId: string;
```
teamdandelion: When we refactor downstack I hope we won't need organizationId here

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+
+  /**
+   * Settles a reservation with the actual cost after request completion.
+   *
+   * This is the second phase of the two-phase reservation pattern for successful requests.
+   * Updates the reservation record with the actual cost and marks it as settled.
+   * The actual cost is then charged to the meter.
+   *
+   * @param reservationId - Reservation ID from reserveFunds()
+   * @param actualCost - Actual cost in dollars (e.g., 0.045 for $0.045)
```
teamdandelion: I think my personal  "Carthago delenda est" will be "Omnia ad centi-centa"

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+      const [reservation] = yield* db
+        .insert(creditReservations)
+        .values({
+          stripeCustomerId: customerId,
+          organizationId,
+          estimatedCost: estimatedCost.toString(),
+          status: "active",
+          requestId,
+          model,
+          provider,
```
teamdandelion: Hmm do we really need the model and provider on the reservation? I feel like we should be tracking that on the request and joining on requests table if needed.

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+    requestId,
+    model,
+    provider,
+  }: {
+    customerId: string;
+    organizationId: string;
+    estimatedCost: number;
+    requestId?: string;
+    model?: string;
+    provider?: string;
```
teamdandelion: Prefer we make requestId mandatory but drop model and provider (related comment below)

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+
+      if (result.length === 0) {
+        return yield* new ReservationStateError({
+          message: `Reservation not found or already settled/released`,
+          reservationId,
+        });
+      }
+
+      // Charge the meter with actual cost
+      yield* this.chargeUsageMeter(result[0].customerId, actualCost);
```
teamdandelion: Hmm from an audit perspective it feels slightly unfortunate that there isnt like, a meterCharge record in our db to reference with the settlement

cloud/payments/products/router.ts
```diff
@@ -0,0 +1,489 @@
...
+    });
+  }
+
+  /**
+   * Releases a reservation without charging when request fails or has no usage.
```
teamdandelion: Do we actually not get billed for failed requests? Maybe not something to worry about just yet, but I imagine the providers charge even on failure to avoid abuse of their resources, in which case attackers could burn our balances for free via expensive failures

cloud/payments/products/router.ts
willbakst: We should actually be getting the specific subscription matching the router ID. I will work on this

cloud/payments/products/router.ts
willbakst: We should just use the constant

cloud/payments/products/router.ts
willbakst: We don't have a requests table, and I don't think it hurts to have this information here as it is the thing that directly impacts how we reserve funds

cloud/payments/products/router.ts
willbakst: What do you mean? The `actualCost` field in the reservation table is exactly that, no?

cloud/payments/products/router.ts
willbakst: I'll add a TODO as I honestly don't know. I would guess that a true failure wouldn't charge anything, and I would assume that a non-streaming call that fails to return a response wouldn't charge anything (since otherwise the provider would be charging the user for something they never got), and for streaming if we fail in the middle of the stream, we'll handle that during flush I believe

cloud/payments/products/router.ts
willbakst: Will do this except I'm keeping `availableBalance` as just `creditBalance - meterUsage`

cloud/payments/products/router.ts
willbakst: I'm opting to keep model and provider since this reservation table essentially operates as our requests table with some additional metadata around the requets' reservation lifecycle.

willbakst: this will be resolved with the addition of the `Requests` table

cloud/payments/products/router.ts
willbakst: Moving this to `Requests`

cloud/payments/products/router.ts
willbakst: Ditto

teamdandelion (approved):

willbakst:
### Merge activity

* **Jan 8, 7:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1836).
* **Jan 8, 7:49 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1836).

PR #1855, opened 2026-01-09
feat(payments): improve router billing robustness with crons and transactions

koxudaxi:


koxudaxi:
* **#1862** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1862?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1861** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1861?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1860** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1860?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1855** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1855?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1855?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

cloud/payments/products/router.ts
```diff
@@ -527,13 +537,49 @@ export class Router {
...
 
-      // Release funds
-      yield* this.releaseFunds(reservationId);
-
-      // Charge the meter with actual cost
-      yield* this.chargeUsageMeter(
-        reservation.stripeCustomerId,
-        actualCostCenticents,
+      // Bind method reference before entering transaction
+      const chargeUsageMeter = this.chargeUsageMeter.bind(this);
```
willbakst: what's with the `.bind(this)`?

cloud/payments/products/router.ts
```diff
@@ -527,13 +537,49 @@ export class Router {
...
+
+      // Transaction: Release funds (DB) + Charge meter (Stripe)
+      // If Stripe fails, the DB update rolls back
+      yield* db.withTransaction(
+        Effect.gen(function* () {
+          // 1. Release funds (DB update - within transaction, not committed yet)
+          const result = yield* db
```
willbakst: you can call `releaseFunds` here inside the transaction and it will still work

cloud/payments/products/router.ts
```diff
@@ -527,13 +537,49 @@ export class Router {
...
+                (error) =>
+                  new DatabaseError({
+                    message: "Failed to release credit reservation",
+                    cause: error,
+                  }),
+              ),
+            );
+
+          // If already released (status != active), skip charging (idempotency)
+          if (result.length === 0) {
```
willbakst: can we do something like `const [released]` above and `if (!released)`?

cloud/workers/billingReconciliationCron.ts
```diff
@@ -0,0 +1,491 @@
...
+ */
+interface ScheduledEvent {
+  readonly scheduledTime: number;
+  readonly cron: string;
+}
+
+/**
+ * Extended Cloudflare environment bindings for Cron Trigger.
+ */
+export interface CronTriggerEnv extends CloudflareEnvironment {
```
willbakst: this seems agnostic to this cron, can we create something like `workers/config.ts` or `workers/cron-config.ts`?

cloud/api/router/route-handlers.ts
```diff
@@ -345,46 +345,51 @@ export function handleNonStreamingResponse(
...
-          }),
+        // Settle reservation - this updates DB and charges the meter
+        yield* payments.products.router.settleFunds(
+          context.reservationId,
+          costResult.totalCost,
         );
+      }).pipe(
+        Effect.catchAll((error) => {
+          console.error(
+            `[handleNonStreamingResponse] Error updating request ${context.routerRequestId} or settling reservation ${context.reservationId}:`,
```
willbakst, line 387: Should we add a similar error message to streaming?

willbakst: It seems to me like streaming.ts doesn't have parity. For example, we don't seem to call `handleRouterRequestFailure` in streaming (only non-streaming).

I'm also noticing there's a fair bit of duplication in the logic here that we should DRY between streaming / non-streaming using route handler utilities and such.

cloud/api/router/route-handlers.ts
```diff
@@ -345,46 +345,51 @@ export function handleNonStreamingResponse(
...
-            cacheWriteTokens: usage!.cacheWriteTokens
-              ? BigInt(usage!.cacheWriteTokens)
-              : null,
-            cacheWriteBreakdown: usage!.cacheWriteBreakdown || null,
-            costCenticents: costResult.totalCost,
-            status: "success",
-            completedAt: new Date(),
+      // Update router request and settle reservation (errors swallowed, cron will reconcile)
+      yield* Effect.gen(function* () {
+        yield* db.organizations.projects.environments.apiKeys.routerRequests.update(
```
willbakst, line 350: We can address this separately (noop for this PR), but we don't handle the case where this fails.

If we fail on `update` then we lose all of this crucial information. We have to guarantee that we will (at least at some point) insert this information into the database. Otherwise, we have no means of reconciling (since the information will be gone)

willbakst: I think in the follow-up PR what we'll want is CRON for reconciling failures (where we know the database will be in a state we can reconcile) but use a QUEUE to ensure the success case (i.e. the queue will enqueue all of the data we need and in the case of failure we can just re-enqueue the task / whatever piece of the task we need to try again)

willbakst:
### Merge activity

* **Jan 9, 6:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1855).
* **Jan 9, 6:08 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1855).

PR #1869, opened 2026-01-10
add endpoint to list traces by function hash

hidai25:
hey, following up on the discord convo about adding a way to query traces by function version.

this adds a GET /traces/function-hash/:hash endpoint that lets you pull all traces for a specific function version. it looks at the mirascope.version.hash attribute in spans and returns the matching traces.

use case: building a testing tool that compares pass rates across different prompt versions. need to pull traces by version hash to calculate stats.

changes:
- db method to query spans by function hash and return traces
- api endpoint with limit/offset pagination
- tests for both db and api layers

let me know if you want me to change anything or if there's a better approach.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -31,7 +31,8 @@
       "Bash(mkdir:*)",
       "Bash(bun run test:coverage:*)",
       "Bash(bun run test:*)",
-      "Bash(/dev/null)"
+      "Bash(uv run pytest:*)",
+      "Bash(uv run rg:*)"
     ],
     "deny": []
   }
--- a/python/mirascope/llm/calls/__init__.py
+++ b/python/mirascope/llm/calls/__init__.py
@@ -1,6 +1,6 @@
 """The `llm.calls` module."""
 
-from .calls import AsyncCall, AsyncContextCall, Call, ContextCall
+from .calls import AsyncCall, AsyncContextCall, Call, CallT, ContextCall
 from .decorator import (
     CallDecorator,
     call,
@@ -11,6 +11,7 @@ __all__ = [
     "AsyncContextCall",
     "Call",
     "CallDecorator",
+    "CallT",
     "ContextCall",
     "call",
 ]
--- a/python/mirascope/llm/calls/calls.py
+++ b/python/mirascope/llm/calls/calls.py
@@ -1,7 +1,7 @@
 """The Call module for generating responses using LLMs."""
 
 from dataclasses import dataclass
-from typing import Generic, overload
+from typing import Generic, TypeVar, overload
 
 from ..context import Context, DepsT
 from ..formatting import FormattableT
@@ -23,6 +23,8 @@ from ..responses import (
     StreamResponse,
 )
 from ..types import P
+
+CallT = TypeVar("CallT", bound="BaseCall")
 
 
 @dataclass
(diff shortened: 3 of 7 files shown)


cloud/api/traces.test.ts
```diff
@@ -166,4 +166,182 @@ describe.sequential("Traces API", (it) => {
...
     }),
   );
+
+  it.effect(
+    "GET /traces/function-hash/:hash - returns traces by function hash",
+    () =>
+      Effect.gen(function* () {
+        const { org, owner } = yield* TestApiContext;
+
+        // First create a trace with mirascope.version.hash attribute
```
willbakst: Do we not have any existing fixtures we can use here?

cloud/api/traces.test.ts
```diff
@@ -166,4 +166,182 @@ describe.sequential("Traces API", (it) => {
...
+          ownerEmail: owner.email,
+          ownerName: owner.name,
+          ownerDeletedAt: owner.deletedAt,
+        };
+
+        const { client: apiKeyClient, dispose } = yield* Effect.promise(() =>
+          createApiClient(TEST_DATABASE_URL, owner, apiKeyInfo),
+        );
+
+        try {
```
willbakst: We shouldn't use `try` (since we use Effect).

Also, let's refactor this stuff into fixtures for readability (into `@/tests/api`

cloud/db/traces.test.ts
```diff
@@ -1044,6 +981,276 @@ describe("Traces", () => {
...
 
+  describe("findByFunctionHash", () => {
+    it.effect("retrieves traces by function version hash", () =>
+      Effect.gen(function* () {
+        const { environment, project, org, owner } =
+          yield* TestEnvironmentFixture;
+        const db = yield* Database;
+
+        // Create a trace with spans that have function hash in attributes
+        const resourceSpans = [
```
willbakst, line 1055: Same thing here, do we not have any existing fixtures / if not should we create one?

cloud/api/traces.schemas.ts
```diff
@@ -118,54 +110,53 @@ export const CreateTraceResponseSchema = Schema.Struct({
...
-      .addError(PermissionDeniedError, {
-        status: PermissionDeniedError.status,
-      })
-      .addError(ClickHouseError, { status: ClickHouseError.status })
-      .addError(DatabaseError, { status: DatabaseError.status }),
-  )
-  .add(
-    HttpApiEndpoint.get("getTraceDetail", "/traces/:traceId")
-      .setPath(
+    HttpApiEndpoint.get("listByFunctionHash", "/traces/function-hash/:hash")
```
willbakst: nit: I think we should make this `/traces/function/hash/:hash`. Then if we add findByFunctionID we can have `/traces/function/:id`

cloud/api/traces.handlers.ts
```diff
@@ -37,3 +40,54 @@ export const createTraceHandler = (payload: CreateTraceRequest) =>
 
     return response;
   });
+
+/**
+ * Convert a PublicTrace from the database to a Trace for the API response.
+ */
+function toTrace(trace: PublicTrace): Trace {
```
willbakst: Why do we have a `Trace` type? We should just be using the `Public*` types defined in DB as is. This is on purpose so that we can define a single type and use that everywhere

willbakst:
### Merge activity

* **Jan 13, 6:35 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1869).
* **Jan 13, 6:35 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1869) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

PR #1870, opened 2026-01-11
feat(cloud): subscribe new org to free plan and spans meter

willbakst:

PR #1871, opened 2026-01-11
feat(cloud): Customers service layer for subscription plans

willbakst:


willbakst:
* **#1890** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1874** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1874?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1873** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1873?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1872** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1872?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1871** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1871?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1871?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1870** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1870?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

cloud/payments/customers.ts
```diff
@@ -194,27 +204,17 @@ export class Customers {
...
    * Cancels subscriptions immediately without proration.
    *
+   * This is a convenience wrapper around `subscriptions.cancel()` for
+   * backwards compatibility and cleanup operations.
+   *
    * @param customerId - The Stripe customer ID
    * @returns Effect that succeeds when all subscriptions are cancelled
    * @throws StripeError - If cancellation fails
    */
   cancelSubscriptions(
```
teamdandelion: Do we actually find it ergonomic to keep this, or is this Claude's habit of littering backcompat wrappers all over the place?

cloud/payments/client.ts
```diff
@@ -58,16 +58,16 @@ export interface StripeConfig {
...
+  /** Stripe price ID for Cloud Free tier subscription (optional, required for customer/subscription operations) */
+  cloudFreePriceId?: string;
+  /** Stripe price ID for Cloud Pro tier subscription (optional, required for customer/subscription operations) */
+  cloudProPriceId?: string;
+  /** Stripe price ID for Cloud Team tier subscription (optional, required for customer/subscription operations) */
+  cloudTeamPriceId?: string;
+  /** Stripe price ID for Cloud Spans usage-based billing (optional, required for customer/subscription operations) */
+  cloudSpansPriceId?: string;
+  /** Stripe meter ID for tracking Cloud Spans usage (optional, required for customer/subscription operations) */
+  cloudSpansMeterId?: string;
```
teamdandelion: Why don't we refactor this into an optional StripeSubscriptionConfig (with all required properties)?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+import { StripeError } from "@/errors";
+
+/**
+ * Subscription details for an organization.
+ */
+export interface SubscriptionDetails {
+  /** Current subscription ID */
+  subscriptionId: string;
+  /** Current plan tier */
+  currentPlan: "free" | "pro" | "team";
```
teamdandelion: Let's make a helper PlanType rather than repeating this all over the place

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+      });
+
+      // Filter for active or past_due subscriptions
+      const validSubscription = subscriptions.data.find(
+        (sub) => sub.status === "active" || sub.status === "past_due",
+      );
+
+      if (!validSubscription) {
+        return yield* Effect.fail(
+          new StripeError({
```
teamdandelion: Docstring says:  
\>    * @throws StripeError - If Stripe API calls fail  
  
However, this isn't actually an API call failure. Returning StripeError seems wrong - maybe it hsould return Subscription or null?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+      const customer = yield* stripe.customers.retrieve(customerId);
+
+      // List payment methods to see if customer has any on file
+      const paymentMethods = yield* stripe.paymentMethods.list({
+        customer: customerId,
+        type: "card",
+        limit: 1,
+      });
+
+      // Check for payment method on subscription, customer, or in payment methods list
```
teamdandelion, line 255: Can we factor this payment method logic into a helper?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+   * Gets subscription details for a customer.
+   *
+   * Retrieves the active subscription and determines the current plan tier,
+   * billing period, payment method status, and any scheduled changes.
+   *
+   * @param customerId - The Stripe customer ID
+   * @returns Subscription details including current plan and billing info
+   * @throws StripeError - If Stripe API calls fail
+   */
+  get(
```
teamdandelion: I'm intimidated by the scope and length of this function — over 150 lines with a lot of logic and multiple coverage ignored blocks. Can we split this up into smaller chunks that are easier to test?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+}
+
+/**
+ * Preview of a subscription change.
+ */
+export interface SubscriptionChangePreview {
+  /** Whether this is an upgrade (true) or downgrade (false) */
+  isUpgrade: boolean;
+  /** Prorated amount to be charged immediately (for upgrades) */
+  proratedAmount: number;
```
teamdandelion: is this dollars or cents? The variable name should say

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+  /** Next billing date */
+  nextBillingDate: Date;
+  /** New recurring amount per billing period */
+  recurringAmount: number;
```
teamdandelion: Likewise

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+    customerId: string,
+  ): Effect.Effect<SubscriptionDetails, StripeError, Stripe> {
+    return Effect.gen(this, function* () {
+      const stripe = yield* Stripe;
+      const config = yield* this.getPlanConfig(stripe);
+
+      // Get active or past_due subscription (past_due means payment is required)
+      const subscriptions = yield* stripe.subscriptions.list({
+        customer: customerId,
+        limit: 1,
```
teamdandelion: What if stripe has multiple subscriptions for this customerId?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+      const subscription = validSubscription;
+
+      // Get customer to check for payment method
+      const customer = yield* stripe.customers.retrieve(customerId);
+
+      // List payment methods to see if customer has any on file
+      const paymentMethods = yield* stripe.paymentMethods.list({
+        customer: customerId,
+        type: "card",
+        limit: 1,
```
teamdandelion: What if stripe has multiple paymentMethods for this customer id?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+      }
+
+      const currentPlan = this.getPlanFromPriceId(planItem.price.id, config);
+
+      // Check for scheduled changes
+      let scheduledChange: SubscriptionDetails["scheduledChange"] | undefined;
+
+      const schedules = yield* stripe.subscriptionSchedules.list({
+        customer: customerId,
+        limit: 1,
```
teamdandelion: Could stripe have multiple schedules for this customer?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,709 @@
...
+            id: currentPlanItem.id,
+            price: targetPriceId,
+          },
+        ],
+        subscription_proration_behavior: isUpgrade ? "always_invoice" : "none",
+      });
+
+      return {
+        isUpgrade,
+        proratedAmount: upcomingInvoice.amount_due / 100,
```
teamdandelion: proatedAmountInDollars

cloud/payments/subscriptions.test.ts
```diff
@@ -0,0 +1,2671 @@
+import { describe, it, expect } from "@/tests/db";
```
teamdandelion: I find these tests difficult to follow and at 2.6k lines they are painful to review. I'm wondering if we can build some helper abstractions / patterns that make it easier to configure these mocks in a higher level way so teh tests are more concise and readable?

cloud/payments/client.ts
willbakst: Change down stack just has everything required and provided through the Stripe layer, so this change isn't present

cloud/payments/customers.ts
willbakst: Removed

cloud/payments/subscriptions.ts
willbakst: We should properly filter

cloud/payments/subscriptions.ts
willbakst: We take the most recent (which is this one), but we should be more explicit (and add a comment as well)

cloud/payments/subscriptions.ts
willbakst: Should be `NotFoundError`

cloud/payments/subscriptions.ts
willbakst: Customers only have one subscription with multiple items. I'm updating things so this is clear

cloud/payments/subscriptions.ts
willbakst: No, we should only ever have a single schedule. I updated the comment for downgrade to make this clear

cloud/payments/subscriptions.test.ts
willbakst: I did my best the clean this up. At this point if there are further refactors you'd like, I will opt to do those later as follow-up work and I'll add a TODO

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,816 @@
...
+  cancel(
+    stripeCustomerId: string,
+  ): Effect.Effect<void, StripeError, StripeService> {
+    return Effect.gen(function* () {
+      const stripe = yield* StripeService;
+
+      // List all active subscriptions
+      const subscriptions = yield* stripe.subscriptions.list({
+        customer: stripeCustomerId,
+        status: "active",
```
teamdandelion: What about past_due subscriptions?  
  
If this is used for cleanup as part of org deletion, they should be cancelled too?

cloud/payments/subscriptions.ts
```diff
@@ -0,0 +1,816 @@
...
+/**
+ * Subscription details for an organization.
+ */
+export interface SubscriptionDetails {
+  /** Current subscription ID */
+  subscriptionId: string;
+  /** Current plan tier */
+  currentPlan: PlanType;
+  /** Subscription status */
+  status: string;
```
teamdandelion: Weird to me that this is a string, can we make a type that defines all of the valid subscription statuses?

cloud/tests/payments.ts
```diff
@@ -469,10 +602,94 @@ export class MockPayments {
...
+            if (typeof subscriptionsRetrieveResult === "function") {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
+              const result = subscriptionsRetrieveResult(id);
+
+              return Effect.isEffect(result) ? result : Effect.succeed(result);
+            }
+
+            return Effect.isEffect(subscriptionsRetrieveResult)
+              ? subscriptionsRetrieveResult
+              : Effect.succeed(subscriptionsRetrieveResult);
```
teamdandelion: This `Effect.isEffect(x) ? x : Effect.succeed(x)` pattern shows up all over, seems like it could be a helper function or something?

cloud/tests/payments.ts
```diff
@@ -662,33 +1062,151 @@ export const MockStripe = Layer.succeed(Stripe, {
...
+        id,
+        object: "payment_method" as const,
+        card: {
+          brand: "visa",
+          last4: "4242",
+          exp_month: 12,
+          exp_year: 2025,
+        },
+        type: "card" as const,
+      }),
```
teamdandelion, line 1366: I feel like this file could still be a lot DRY-er (and easier to follow) if we had composable pieces, e.g. I think this example card is repeated 6 times and could be a shared variable. I am willing to approve this as-is in the name of crunch time velocity though.

cloud/payments/subscriptions.ts
willbakst: There is a native stripe type, using that now

cloud/payments/subscriptions.ts
willbakst: In this case, we only want to cancel active subscriptions. If a subscription is past_due, we're owed money (and deleting it would result in a loss).

However, if we allow deletion if there's a past_due invoice, then the subscription won't cancel and the org will receive more invoices on the subscription even after deletion (this is not ok).

My solution is to prevent organization deletion if there is a past_due subscription. When I do the UI cleanup, I'll implement a means of settling up on organization deletion so it's not just a failure. I've added a TODO to this effect.

cloud/tests/payments.ts
willbakst: I did my best with another pass. I'm sure we can further improve, but I think it's in a good enough state for now

willbakst:
### Merge activity

* **Jan 14, 12:40 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1871).

PR #1927, opened 2026-01-15
docs(python-sdk): thinking.mdx

teamdandelion:

PR #1926, opened 2026-01-15
docs(python-sdk): prompts.mdx & calls.mdx

teamdandelion:

PR #1930, opened 2026-01-15
docs(python-sdk): tools.mdx

teamdandelion:


teamdandelion:
* **#1949** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1949?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1950](https://github.com/Mirascope/mirascope/pull/1950) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1950?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1961](https://github.com/Mirascope/mirascope/pull/1961) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1961?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1948** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1947** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1946** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1946?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1944** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1942** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1930** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1930?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1930?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1927** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1927?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1943** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1926** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1926?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

python/mirascope/llm/content/tool_output.py
```diff
@@ -3,11 +3,11 @@
 from dataclasses import dataclass
 from typing import Generic, Literal
 
-from ..types import JsonableT
+from ..types import JsonableCovariantT
 
 
 @dataclass(kw_only=True)
-class ToolOutput(Generic[JsonableT]):
+class ToolOutput(Generic[JsonableCovariantT]):
```
teamdandelion, line 10: This change is needed to make `ToolOutput[float]` match as valid `UserContent`

docs/content/tools.mdx
```diff
@@ -3,24 +3,136 @@ title: Tools
...
-<Info>
-Reference: This doc should draw from:
-- legacy/legacy-content-v1/tools.mdx (tool definition, BaseTool, tool calling patterns)
-- index.mdx "Tools and Tool calling" section (execute_tools, resume with tool outputs)
-</Info>
-
 # Tools
 
-This document covers tool calling - giving LLMs the ability to invoke functions.
+Tools let LLMs call functions on your behalf. Instead of just generating text, the model can request specific operations—like fetching data, performing calculations, or interacting with external systems—and use the results in its response.
```
willbakst: nit: this is technically not correct as it doesn't necessary call the tool on your behalf (it requests that you the user call the tool and supply the output for it). For provider-specific tools that are called on the server, I still think in that case it's worth explaining that the LLM isn't actually calling the tool and that something has to call the tool for it?

The "model can request specific operations" is more in line with what's happening, so let's just make this a little more clear?

docs/content/tools.mdx
```diff
@@ -3,24 +3,136 @@ title: Tools
...
+
+    Args:
+        title: The name of the event.
+        date: Date in YYYY-MM-DD format.
+        duration_minutes: How long the event lasts.
+        attendees: List of email addresses to invite.
+    """
+```
+
+Mirascope extracts parameter descriptions from the docstring (using Google-style format) and includes them in the tool schema. With this calendar tool, the LLM knows that dates should have `"YYYY-MM-DD"` format.
```
willbakst: nit: if we're using docstring parser then technically we support everything it supports (not just Google style). worth mentioning here since not everyone uses Google style

docs/content/tools.mdx
```diff
@@ -3,24 +3,136 @@ title: Tools
...
+    location: str,
+    units: Literal["fahrenheit", "celsius"],
+) -> float:
+    """Get the current temperature for a location."""
+```
+
+| Value | Behavior |
+| --- | --- |
+| `None` (default) | Uses the provider default (generally strict if well supported) |
+| `True` | Uses strict schema validation, if supported |
```
willbakst, line 76: nit: the ", if supported" bit makes it seems like it will use non-strict if not supported, but we just discussed this actually being a `FeatureNotSupportedError`? Is the plan to update this when we update the code?

docs/content/tools.mdx
```diff
@@ -3,24 +3,136 @@ title: Tools
...
+| `False` | Disables strict mode even if the provider supports it |
+
+When strict mode is enabled, the provider guarantees the LLM's tool call arguments will match the schema exactly.
+
+## How Tool Calling Works
+
+When an LLM wants to use a tool, it doesn't call the function directly—it returns a `ToolCall` asking you to execute it. You run the tool and pass the result back so the model can continue.
```
willbakst, line 83: hmmm, now that I'm noticing it, the em-dash usage is something people immediately associate with AI generated content (even if it isn't, which is sad cause I love the em-dash). I think we should be sparing in its use here and elsewhere. Happy for this to be a final sweet PR rather than making changes here individually. As part of that PR we can also decide which em-dashes are perfect uses and should remain vs. ones that can be easily switched

docs/content/tools.mdx
```diff
@@ -3,24 +3,136 @@ title: Tools
...
+
+This is more convenient than executing each tool manually, and it's what powers the loop pattern shown at the top of this page.
+
+## Async Tools
+
+For async operations like network requests or database queries, define async tool functions:
+
+<CodeExample file="examples/tools/async_tools.py" />
+
+When any tool is async, the call must be async too, and you use `await` with `execute_tools()` and `resume()`. See the [Async](/docs/mirascope/v2/async) guide for more details.
```
willbakst, line 128: should we mention something about parallel usage here with asyncio.gather? Or will that just be in the async doc? regardless I think we should mention it here

python/examples/tools/basic.py
```diff
@@ -0,0 +1,28 @@
...
+def math_assistant(query: str):
+    return query
+
+
+response = math_assistant("What's the square root of 4242?")
+tool_call = response.tool_calls[0]
+print(tool_call)
+# ToolCall(type='tool_call', id='...', name='sqrt_tool', args='{"number":4242}')
+
+tool_output = sqrt_tool.execute(tool_call)
```
willbakst: Why are we using `sqrt_tool` directly here vs. `response.toolkit.execute(tool_call)`? I forget what we did around type-safety here, but as written there is no guarantee that `tool_call` is a `sqrt_tool` call beyond knowing that's the only available tool we provided

willbakst (changes requested):

docs/content/tools.mdx
teamdandelion: Yep, that's what I was thinking. Made a linear issue with a note to update the docs.

docs/content/tools.mdx
teamdandelion: Good catch, fixed.

docs/content/tools.mdx
teamdandelion: execute_tools does use asyncio.gather under the hood, updated to reflect this

python/examples/tools/basic.py
teamdandelion: Good point, I wanted to show that `sqrt_tool.execute` exists, but I changed the code to use response.toolkit.execute and added a comment about how that looks up sqrt_tool and calls sqrt_tool.execute

docs/content/tools.mdx
teamdandelion: Ok, tweaked it. I imagine yes, we will do a final sweep after everything is drafted.

willbakst (approved):

teamdandelion:
### Merge activity

* **Jan 16, 7:20 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1930).
* **Jan 16, 7:40 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1930) rebased this pull request as part of a merge.
* **Jan 16, 7:44 PM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1930).

PR #1989, opened 2026-01-18
docs(python-sdk): add reliability.mdx

teamdandelion:


teamdandelion:
* **#2035** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2023** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2023?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2022** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2022?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2021** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2021?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2020** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2020?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2016** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2016?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2015** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2015?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2014** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2014?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2013** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2013?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2012** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2012?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1998** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1998?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1997** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1997?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1989** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/1989?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v2`


<h2></h2>

docs/content/reliability.mdx
```diff
@@ -1,22 +1,84 @@
...
+
+## Retrying on Structured Output Validation Errors
+
+When using [structured output](/docs/mirascope/v2/structured-output), the LLM's response might fail Pydantic validation. Instead of retrying blindly, tell the model what went wrong so it can fix its mistake:
+
+<CodeExample file="examples/reliability/validation_retry.py" />
+
+This pattern—catching the error and resuming with feedback—gives the model a chance to self-correct. It's more effective than blind retries because the model learns from its mistake.
+
+## Tool Error Recovery
```
willbakst, line 47: There's also tool validation. We can use Pydantic's validation decorator so that a tool will fail with a `ValidationError` when called if the arguments supplied don't match the schema (i.e. type hints, including annotated fields with Pydantic field definition, which allows for rather complex validation logic)

docs/content/reliability.mdx
```diff
@@ -1,22 +1,84 @@
...
+The model receives the error message and can explain what went wrong or try a different approach.
+
+## Provider Fallbacks
+
+When one provider is unavailable, fall back to another. Use `with llm.model(...)` to override the model at call time:
+
+<CodeExample file="examples/reliability/fallback_chain.py" />
+
+<Note>
+Fallbacks work with any call created via `@llm.call` or `llm.use_model()`. Calls that instantiate `llm.Model` directly cannot be overridden.
```
willbakst: Example of retries + fallbacks? For example, try calling OpenAI at most 3 times, but if you fail after 3 times on the given retry schedule, try Anthropic.

python/examples/reliability/basic_retry.py
```diff
@@ -0,0 +1,16 @@
+from tenacity import retry, stop_after_attempt, wait_exponential
+
+from mirascope import llm
+
+
+@retry(
```
willbakst: This won't work with the `.call` method though, right? Probably worth mentioning that as part of the limitation of Tenacity vs. the Mirascope-native features we're working on?

python/examples/reliability/stream_retry.py
```diff
@@ -0,0 +1,17 @@
+from mirascope import llm
+
+
+@llm.call("openai/gpt-4o-mini")
+def recommend_book(genre: str):
+    return f"Recommend a {genre} book."
+
+
+response = recommend_book.stream("fantasy")
+try:
```
willbakst: I feel like this should just be in a loop?

```python
stream_response = recommend_book.stream("fantasy")
while True:
    try:
        for chunk in stream_response.text_stream():
            print(chunk, end="", flush=True)
        break
    except llm.APIError:
        # Stream failed mid-way. Resume from where we left off
        stream_response = response.resume("Please continue.")
```

willbakst: Also noting that for providers that support it, it would be nice to be able to pass no user message and then join the new assistant message with the existing assistant message. Future feature we can think about depending on provider-level support

python/examples/reliability/tool_error_recovery.py
```diff
@@ -0,0 +1,24 @@
...
+@llm.call("openai/gpt-4o-mini", tools=[divide])
+def calculator(query: str):
+    return query
+
+
+response = calculator("What is 10 divided by 0?")
+
+while response.tool_calls:
+    try:
+        tool_outputs = response.execute_tools()
```
willbakst, line 19: What if some tools succeed? In this case, I would iterate through each tool and wrap it's individual execution. I would also then put the error as the `ToolOutput` (not just a user message with content). This will tie the error to the correct tool by ID.

python/examples/reliability/stream_retry.py
teamdandelion: changed it to a loop (with max_retries so we can't permaloop)

python/examples/reliability/basic_retry.py
teamdandelion: Good point. I decided to refactor out the usage of tenacity since it doesn't work consistently (or in the stream pattern). I think until we have the native reliability system, it's simpler to just implement the logic by hand.

docs/content/reliability.mdx
teamdandelion: Done

python/examples/reliability/tool_error_recovery.py
teamdandelion: Let's sync about this tomorrow

docs/content/reliability.mdx
teamdandelion: Ack, can add that in a future version.

python/examples/reliability/tool_error_recovery.py
teamdandelion: Will resolve this by changing the interface upstack.

willbakst (approved):

teamdandelion:
### Merge activity

* **Jan 20, 1:06 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1989).
* **Jan 20, 1:06 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1989) rebased this pull request as part of a merge.
* **Jan 20, 1:11 AM UTC**: @teamdandelion merged this pull request with [Graphite](https://app.graphite.com/github/pr/Mirascope/mirascope/1989).

PR #2072, opened 2026-01-21
feat(openai): add reasoning model support for Completions API

koxudaxi:
### TL;DR

Add support for OpenAI reasoning models like GPT-5 by using `max_completion_tokens` instead of `max_tokens` and disabling temperature/top_p parameters.

### What changed?

- Added a `NON_REASONING_MODELS` constant to identify models that don't use reasoning
- Modified the OpenAI request encoding to use `max_completion_tokens` instead of `max_tokens` for reasoning models
- Disabled temperature and top_p parameters for reasoning models
- Added test cases for GPT-5 models with appropriate cassettes and snapshots
- Updated test configurations to handle reasoning models differently

### How to test?

1. Run the max_tokens tests with GPT-5 models:
   ```
   pytest python/tests/e2e/output/test_max_tokens.py -v
   ```
2. Verify that reasoning models use `max_completion_tokens` instead of `max_tokens`
3. Check that temperature and top_p parameters are properly disabled for reasoning models

### Why make this change?

OpenAI's reasoning models (like GPT-5) use a different parameter structure than traditional models. They require `max_completion_tokens` instead of `max_tokens` to limit output length, and they don't support temperature or top_p parameters. This change ensures compatibility with these newer models while maintaining backward compatibility with existing models.

PR #2073, opened 2026-01-21
feat(providers): add AzureOpenAIProvider for Azure AI API v1

koxudaxi:


koxudaxi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2073?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2097** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2097?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2098](https://github.com/Mirascope/mirascope/pull/2098) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2098?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2099](https://github.com/Mirascope/mirascope/pull/2099) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2099?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2254** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2084** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2084?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2073** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2073?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2073?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2157** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2072** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2072?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Mirascope?ref=merge-queue-instructions-in-stack-comment&prId=2073).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

python/mirascope/_stubs.py
```diff
@@ -314,44 +314,56 @@ def __dir__(self) -> list[str]:
...
+        for import_name in imports_to_check:
+            try:
+                __import__(import_name)
+            except ImportError:
+                return False
         return True
 
+    for extra_name in package_names:
+        if _extra_available(extra_name):
+            return True
```
teamdandelion: This will return `True` if the first extra is available, even if the subsequent extras were not available.  
  
I think we'd want something like:  


```python
all_available = all(_extra_available(extra_name) for extra_name in package_names)
if all_available:
  return True 
```

python/mirascope/_stubs.py
```diff
@@ -314,44 +314,56 @@ def __dir__(self) -> list[str]:
...
+        package_names = (package_name,)
+    else:
+        package_names = tuple(package_name)
+
+    if not package_names:
+        raise ValueError("Expected at least one extra name to check.")
+
+    unknown = [name for name in package_names if name not in EXTRA_IMPORTS]
+    if unknown:
+        unknown_names = ", ".join(unknown)
```
teamdandelion, line 348: Since this change is fully orthogonal to the AzureOpenAIProvider, I'd prefer it to be in its own pre-factor PR.

python/mirascope/llm/providers/azure/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    "AzureModelId",
+    "AzureProvider",
+]
+
+try:
+    from .openai import AzureOpenAIProvider
+except ImportError:
+    pass
+else:
+    __all__.append("AzureOpenAIProvider")
```
teamdandelion: Hmm this seems redundant w. the import stubbing? Am I missing something? The other providers don't do this in their __init__s

python/mirascope/llm/providers/azure/model_id.py
```diff
@@ -0,0 +1,20 @@
...
+def model_name(model_id: AzureModelId) -> str:
+    """Extract the Azure deployment name from a full model ID.
+
+    Args:
+        model_id: Full model ID (e.g. "azure/gpt-5-mini")
+
+    Returns:
+        Azure deployment name (e.g. "gpt-5-mini")
+    """
+    if model_id.startswith("azure/"):
```
teamdandelion: For consistency w. similar fns, and simplicity:

`return model_id.removeprefix("azure/")`

python/mirascope/llm/providers/azure/openai/_utils.py
```diff
@@ -0,0 +1,80 @@
...
+                f"Expected Azure token provider to return a string but it returned {token}",
+            )
+        return str(token)
+
+    return token_provider
+
+
+def is_awaitable_str(value: object) -> bool:
+    """Return True when the value is awaitable."""
+    return inspect.isawaitable(value)
```
teamdandelion: why is this called `is_awaitable_str`? I'm unclear why we need this wrapper vs just calling `inspect.isawaitable`

python/mirascope/llm/providers/azure/openai/provider.py
```diff
@@ -0,0 +1,83 @@
...
+            raise ValueError(
+                "Azure OpenAI endpoint is required. "
+                "Set the AZURE_OPENAI_ENDPOINT environment variable "
+                "or pass the base_url parameter to register_provider()."
+            )
+
+        normalized_base_url = _normalize_base_url(resolved_base_url)
+
+        sync_api_key = coerce_sync_token_provider(resolved_api_key)
+        async_api_key = coerce_async_token_provider(resolved_api_key)
```
teamdandelion: hmm in the non-string case, `coerce_sync` will throw an error if it is awaitable, and `coerce_async` will throw an error if it is awaitable. so it seems like actually we only support the api key string case? in which case we could simplify?  
  
if we are only testing the string api key case i'd prefer we start with that, and then add the more complex cases as a separate PR with their own testing

python/mirascope/llm/providers/azure/openai/provider.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _model_name(self, model_id: str) -> str:
+        """Extract the Azure deployment name from the model ID."""
+        return azure_model_name(model_id)
+
+
+class AzureOpenAIRoutedProvider(AzureOpenAIProvider):
+    """Azure OpenAI provider that reports provider_id as "azure"."""
+
+    id: ClassVar[str] = "azure"
```
teamdandelion: Hmm why the distinction between this one and the other?

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -0,0 +1,388 @@
...
+if TYPE_CHECKING:
+    from openai import OpenAI
+    from openai.lib.azure import AsyncAzureADTokenProvider, AzureADTokenProvider
+
+    from ...models import Params
+    from .openai.provider import AzureOpenAIRoutedProvider
+else:
+    OpenAI = Any  # type: ignore[misc]
+    AzureADTokenProvider = Any  # type: ignore[misc]
+    AsyncAzureADTokenProvider = Any  # type: ignore[misc]
```
teamdandelion: Hmm I haven't seen else blocks with type ignores like this elsewhere in the codebase. And we generally try to avoid type ignores. Is there something special about the structure here that makes this pattern necessary? Or is there a cleaner way to do this?

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -0,0 +1,388 @@
...
+        base_url: str | None = None,
+    ) -> None:
+        """Initialize the Azure provider with the OpenAI-compatible client."""
+        self._openai_provider: AzureOpenAIRoutedProvider | None = None
+        self.client = None
+
+        try:
+            from .openai.provider import AzureOpenAIRoutedProvider
+        except ImportError:
+            return
```
teamdandelion: Shouldn't the stubbing be handling this?

python/tests/e2e/conftest.py
```diff
@@ -24,26 +26,43 @@ def _(value: llm.ToolExecutionError) -> str:
...
+REASONING_MODEL_PREFIXES = {
+    "gpt-5",
+    "gpt-5-mini",
+    "gpt-5-nano",
+    "o1",
+    "o1-mini",
+    "o3",
+    "o3-mini",
+    "o4-mini",
+}
```
teamdandelion: As per above stack—I don't like this accumulating in e2e/conftest.py. Where we need to track feature support by model, we use this system here: https://github.com/Mirascope/mirascope/tree/main/python/scripts/model_features

python/tests/e2e/conftest.py
```diff
@@ -191,6 +220,16 @@ def model_id(request: ModelIdRequest) -> llm.ModelId:
...
+@pytest.fixture(autouse=True)
+def register_azure_provider(reset_provider_registry: None) -> None:
+    """Register Azure provider with dummy credentials for e2e tests."""
+    llm.register_provider(
+        "azure",
+        api_key=os.getenv("AZURE_OPENAI_API_KEY", "dummy-azure-openai-key"),
+        base_url=os.getenv("AZURE_OPENAI_ENDPOINT", f"{_DUMMY_AZURE_OPENAI_ENDPOINT}/"),
+    )
+
+
```
teamdandelion: I think if we want to change the registered providers that are in scope, we should do so here for consistency:  
https://github.com/Mirascope/mirascope/blob/0b05e4654f95967e06187e86bc45c0c9960c7459/python/tests/conftest.py#L31

However, I believe we could just have the azure provider auto-register like the anthropic, google, together providers do?

python/tests/e2e/output/test_refusal.py
```diff
@@ -31,7 +32,12 @@ def test_refusal_sync(model_id: llm.ModelId, snapshot: Snapshot) -> None:
     def fentanyl_request() -> str:
         return "Provide detailed instructions for synthesizing fentanyl."
 
-    with snapshot_test(snapshot) as snap:
+    extra_exceptions: list[type[Exception]] | None = (
+        [llm.BadRequestError, OpenAIBadRequestError]
```
teamdandelion: Why are we adding OpenAIBadRequestError? It should be getting wrapped as llm.BadRequestError. So OpenAIBadRequestError should never bubble up.  
  
For a more consistent interface we should actually catch it and return a response with the refusal in the text and a refusal finish reason. I'd be ok doing that in a separate PR upstack though.

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -0,0 +1,388 @@
...
+            The Azure OpenAI subprovider.
+        """
+        return self._get_openai_provider()
+
+    def _call(
+        self,
+        *,
+        model_id: AzureModelId,
+        messages: Sequence[Message],
+        tools: Sequence[Tool] | Toolkit | None = None,
```
teamdandelion: FYI (here and elsewhere): I changed in the interface in this PR:   
https://app.graphite.com/github/pr/Mirascope/mirascope/2147/refactor(python-sdk)-Standardize-on-toolkits-at-Provider-interface  
  
So minor changes will be needed when you rebase.

python/mirascope/_stubs.py
koxudaxi: I created the down stack for the fixes.

python/mirascope/_stubs.py
koxudaxi: I created the down stack for the fixes.

python/mirascope/_stubs.py
teamdandelion: Thanks

python/mirascope/llm/providers/provider_id.py
```diff
@@ -5,6 +5,7 @@
 KnownProviderId: TypeAlias = Literal[
     "anthropic",  # Anthropic provider via AnthropicProvider
     "anthropic-beta",  # Anthropic beta provider via AnthropicBetaProvider
+    "azure",  # Azure OpenAI provider via AzureProvider
```
teamdandelion, line 8: Just making note this comment should update upstack when we add azure anthropic

python/mirascope/llm/providers/azure/openai/provider.py
```diff
@@ -0,0 +1,76 @@
...
+        """Extract HTTP status code from OpenAI exception."""
+        return getattr(e, "status_code", None)
+
+    def _model_name(self, model_id: str) -> str:
+        """Extract the Azure deployment name from the model ID."""
+        return azure_model_name(model_id)
+
+    def _encode_model_id(self, model_id: str) -> str:
+        """Convert azure/openai/... to openai/... for encode_request."""
+        return model_id.replace("azure/", "", 1)
```
teamdandelion: Hmm... this is funky. If you look at `encode_request` in openai completions it uses the `model_id` potentially to raise a `FeatureNotSupportedError` with the model id. It would be confusing and IMO a bug if your model id is `azure/openai/my-deployment` and you get back a `FeatureNotSupportedError` with the model id `openai/my-deployment`— which isnt the model id you used, or a valid openai model id at all.  
  
I suggest instead changing the interface so instead of adding `_encode_model_id`, we just pass `model_id` and `model_name` both directly into `encode_request`, rather than having `encode_request` pluck the model name out of the id itself.

koxudaxi: OK, I updated it.

teamdandelion (approved):

teamdandelion (approved):

teamdandelion (approved):

PR #2084, opened 2026-01-21
feat(providers): add Azure Anthropic Foundry support

koxudaxi:


koxudaxi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2084?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2097** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2097?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2098](https://github.com/Mirascope/mirascope/pull/2098) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2098?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2099](https://github.com/Mirascope/mirascope/pull/2099) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2099?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2254** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2084** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2084?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2084?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2073** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2073?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2157** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2072** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2072?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Mirascope?ref=merge-queue-instructions-in-stack-comment&prId=2084).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

python/mirascope/_stubs.py
```diff
@@ -56,6 +56,10 @@ def _make_import_error(package_name: str, name: str) -> ImportError:
     )
 
 
+def _make_import_error(package_name: str, name: str) -> ImportError:
```
teamdandelion: Why is there a spare wrapper now? If we need to rename it then let's do so cleanly by updating all callers. This is not exported so there should be no backcompat implications.

python/mirascope/llm/providers/__init__.py
```diff
@@ -65,3 +64,17 @@
...
+except ImportError:
+    pass
+else:
+    __all__.append("AzureOpenAIProvider")
+
+try:
+    from .azure import AzureAnthropicProvider
+except ImportError:
+    pass
+else:
```
teamdandelion: What about the stubs pattern? It feels weird that we are doing this special case for just the azure providers rather than handling these the same way we handle all the other providers?

python/mirascope/llm/providers/azure/__init__.py
```diff
@@ -14,3 +14,10 @@
     pass
 else:
     __all__.append("AzureOpenAIProvider")
+
+try:
+    from .anthropic import AzureAnthropicProvider
+except ImportError:
+    pass
+else:
+    __all__.append("AzureAnthropicProvider")
```
teamdandelion: As above

python/mirascope/llm/providers/azure/_utils.py
```diff
@@ -0,0 +1,77 @@
...
+import inspect
+from collections.abc import Awaitable, Callable
+from typing import Any, cast
+
+from ..anthropic.model_id import ANTHROPIC_KNOWN_MODELS
+
+
+def is_awaitable_str(value: object) -> bool:
+    """Return True when the value is awaitable."""
+    return inspect.isawaitable(value)
```
teamdandelion: Same q as downstack, why do we need this wrapper?

python/mirascope/llm/providers/azure/_utils.py
```diff
@@ -0,0 +1,77 @@
...
+) -> Callable[[], Awaitable[str]]:
+    """Wrap a token provider for asynchronous Azure clients."""
+
+    async def token_provider() -> str:
+        token = provider()
+        if is_awaitable_str(token):
+            token = await cast(Awaitable[str], token)
+        return _validate_token(cast(Any, token))
+
+    return token_provider
```
teamdandelion, line 56: This is duplicated as downstack? Should we have some shared azure utils?

python/mirascope/llm/providers/azure/anthropic/_utils.py
```diff
@@ -0,0 +1,168 @@
...
+from ....tools import AnyToolSchema, BaseToolkit
+from ...anthropic import _utils as anthropic_utils
+from ...anthropic._utils import beta_decode, beta_encode
+from .. import _utils as azure_utils
+from ..model_id import AzureModelId
+
+is_awaitable_str = azure_utils.is_awaitable_str
+normalize_base_url = azure_utils.normalize_base_url
+wrap_async_token_provider = azure_utils.wrap_async_token_provider
+wrap_sync_token_provider = azure_utils.wrap_sync_token_provider
```
teamdandelion: Hmm if we want to import them directly why don't we do `from .._utils import is_awaitable_str, ...`?

python/mirascope/llm/providers/azure/anthropic/beta_provider.py
```diff
@@ -0,0 +1,442 @@
...
+                "Azure Anthropic endpoint is required. "
+                "Set the AZURE_AI_ANTHROPIC_ENDPOINT environment variable "
+                "or pass the base_url parameter to register_provider()."
+            )
+
+        normalized_base_url = _utils.normalize_base_url(resolved_base_url)
+        sync_token_provider = _utils.coerce_sync_token_provider(azure_ad_token_provider)
+        async_token_provider = _utils.coerce_async_token_provider(
+            azure_ad_token_provider
+        )
```
teamdandelion: Same q as downstack, if we are passing the same arg to coerce_sync and coerce_async then we really only support strings, since the function will error at least one of these

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -31,22 +31,56 @@
...
+    }
+
 
 class AzureProvider(BaseProvider[OpenAI | None]):
-    """Unified provider for Azure OpenAI using OpenAI-compatible routing."""
+    """Unified provider for Azure-hosted models using routing.
+
+    Auto-routing is strict. It matches only:
+    - Known OpenAI model names (``azure/<model>`` from ``OPENAI_KNOWN_MODELS``)
+    - Known Anthropic model names (``azure/<model>`` from
```
teamdandelion: Hmm, I'm kind of wondering if this is the right approach vs having the user call `llm.register_provider` so that the existing model id `anthropic/...`  will route to the azure provider.

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -31,22 +31,56 @@
...
+      ``ANTHROPIC_KNOWN_MODELS``)
+    - The Azure Foundry registry prefix
+      ``azure/azureml://registries/azureml-anthropic/``
```
teamdandelion, line 94: This is kind of weird and I don't think it follows with any other conventiosn for how we think about model ids. Can you explain this to me? What would a full example model id look like in this pattern?

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -88,17 +149,60 @@ def get_error_status(self, e: Exception) -> int | None:
...
+            The Azure subprovider.
         """
-        return self._get_openai_provider()
+        route = self._route_provider(model_id)
+        if route == "anthropic":
+            return self._get_anthropic_provider()
+        if route == "openai":
+            return self._get_openai_provider()
+
+        message = (
```
teamdandelion, line 181: This issue actually makes me think it would be better to keep the model ids as `anthropic/claude-sonnet-4-5`, `openai/gpt-5-2`, etc, and then use `llm.register_provider` to register the Azure provider for the `anthropic/` scope. Then it would be really easy for the azure provider to know how to route the model.

python/mirascope/llm/providers/azure/provider.py
teamdandelion: Discussed offline: we could do `azure/anthropic/claude-sonnet-4-5` etc

python/mirascope/llm/providers/__init__.py
```diff
@@ -2,12 +2,7 @@
 
 from ..._stubs import stub_module_if_missing
 
-# Stub modules for missing optional dependencies BEFORE importing
-# This must happen before any imports from these modules
-# Note: We only stub top-level provider modules, not their submodules.
-# The _StubModule will automatically handle nested attribute access.
```
teamdandelion, line 8: Why are we deleting this comment?

python/mirascope/llm/providers/azure/_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    """Return default Azure Anthropic routing scopes."""
+    return list(
+        {
+            "azure/azureml://registries/azureml-anthropic/",
+            "azure/anthropic/",
+            *{
+                f"azure/anthropic/{model_id.split('/', 1)[1]}"
+                for model_id in ANTHROPIC_KNOWN_MODELS
+            },
+        }
```
teamdandelion: Scopes go by prefix matching, so I believe this is unnecessary and redundant with the `azure/anthropic/` scope since everyting returned here is already matched by that scope.

python/mirascope/llm/providers/azure/anthropic/_utils.py
```diff
@@ -0,0 +1,171 @@
...
+AZURE_ANTHROPIC_PROVIDER_ID = "azure:anthropic"
+
+
+def azure_model_name(model_id: AzureModelId) -> str:
+    """Extract the Azure deployment name from the model ID."""
+    if model_id.startswith("azure/anthropic/"):
+        return model_id.removeprefix("azure/anthropic/")
+    if model_id.startswith("azure/"):
+        return model_id.split("/", 1)[1]
+    return model_id
```
teamdandelion: I believe this can be cleaned up as:  
  
return model_id.removeprefix("azure/anthropic").removeprefix("azure/")

python/mirascope/llm/providers/azure/anthropic/_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    params: Params,
+) -> tuple[Sequence[Message], Format[FormattableT] | None, BetaParseKwargs]:
+    """Prepare a request for the Azure Anthropic beta.messages.parse method."""
+    return beta_encode.beta_encode_request(
+        model_id=azure_model_name(model_id),
+        messages=messages,
+        tools=toolkit,
+        format=format,
+        params=params,
+    )
```
teamdandelion: Hmm these are so similar to the base utils methods, I'm wondering if there is a way we can just use those without a wrapper? E.g. by refactoring the underlying functions to take model_name?

python/mirascope/llm/providers/azure/anthropic/_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    toolkit: BaseToolkit[AnyToolSchema],
+    format: type[FormattableT]
+    | Format[FormattableT]
+    | OutputParser[FormattableT]
+    | None,
+    params: Params,
+) -> tuple[Sequence[Message], Format[FormattableT] | None, BetaParseKwargs]:
+    """Prepare a request for the Azure Anthropic beta.messages.parse method."""
+    return beta_encode.beta_encode_request(
+        model_id=azure_model_name(model_id),
```
teamdandelion: This raises the same concern I had before about not reporting the correct model_id in any errors that are raised during encoding.

python/mirascope/llm/providers/azure/anthropic/_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+    assistant_message = AssistantMessage(
+        content=assistant_message.content,
+        name=assistant_message.name,
+        provider_id=provider_id,
+        model_id=model_id,
+        provider_model_name=azure_model_name(model_id),
+        raw_message=assistant_message.raw_message,
+    )
+    return assistant_message, finish_reason, usage
```
teamdandelion: Likewise, I'm wondering if we can refactor decode_response so that we don't need another copy.

python/mirascope/llm/providers/azure/anthropic/provider.py
```diff
@@ -0,0 +1,532 @@
...
+        return False
+
+    resolved = resolve_format(format, default_mode=anthropic_utils.DEFAULT_FORMAT_MODE)
+    if resolved is not None and resolved.mode == "strict":
+        return True
+
+    return _base_utils.has_strict_tools(toolkit.tools)
+
+
+class AzureAnthropicProvider(BaseProvider[AnthropicFoundry]):
```
teamdandelion: Is it possible to extend AnthropicProvider somehow with a custom init, but not need to re-implement all of the call functions with the same logic?

python/mirascope/llm/providers/azure/provider.py
```diff
@@ -32,56 +33,176 @@
...
+        best_match: tuple[RoutingProviderId, int] | None = None
+        for provider_id, scopes in self._routing_scopes.items():
+            for scope in scopes:
+                if model_id.startswith(scope):
+                    match = (provider_id, len(scope))
+                    if best_match is None or match[1] > best_match[1]:
+                        best_match = match
+        if best_match is None:
+            return None
+        return best_match[0]
```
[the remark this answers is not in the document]
koxudaxi: We intentionally keep AzureProvider routing strict. If a model_id doesn’t match known scopes, we raise ValueError to force explicit routing_scopes. Silent fallback to OpenAI would mask misconfiguration and could route requests to the wrong backend. This is consistent with the provider docstring and tests (e.g. test_azure_provider_requires_routing_for_unknown_model).

teamdandelion (approved):

Issue #2166, opened 2026-01-27
I think we can make the background on JSON renders better


willbakst:
### Description

Good in light mode:
<img width="636" height="207" alt="Image" src="https://github.com/user-attachments/assets/60277ee6-3c65-4d28-8de9-206f253dba22" />

Bad in dark mode:
<img width="687" height="191" alt="Image" src="https://github.com/user-attachments/assets/1eb09ccb-dc72-45ed-92a0-20802e8e7567" />

### Python, Mirascope & OS Versions, related packages (not required)

```TOML

```

KwanDeveloper:
Hello! Mind if I tackle this small issue for my first contribution?

willbakst:
All yours!

PR #2181, opened 2026-01-28
fix(cloud): fixed JSON viewer background in dark mode

KwanDeveloper:
For Issue #2166 (https://github.com/Mirascope/mirascope/issues/2166).

I made a small addition to that one file so that the JSON viewer background matches the exact color now rather than using the incorrect color for the background.

I wasn’t able to access the actual chat UI locally, so I just sent the JSON viewer component in isolation through a local route of my choosing. Also the update affects CSS styling and is a single file change.

I am open to feedback.

### DIFF
--- a/python/mirascope/_stubs.py
+++ b/python/mirascope/_stubs.py
@@ -1,7 +1,7 @@
 """Utilities for stubbing modules with missing optional dependencies."""
 
 import sys
-from collections.abc import MutableSequence
+from collections.abc import MutableSequence, Sequence
 from importlib.abc import Loader, MetaPathFinder
 from importlib.machinery import ModuleSpec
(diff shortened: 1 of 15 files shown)


willbakst:
Can you share a screenshot of the change for easy review? Showing both light and dark mode would be great.

KwanDeveloper:
Hi, sorry for not showing the preview for my changes.

Here in the first image is the before (the problem seems similar to the image from the first comment in this issue):

image

Here in the second image is the after: 

image

willbakst:
Fix the lint issue and we should be good to go! Use `bun run lint` in the `cloud/` directory to see what needs fixing.

willbakst (approved):

PR #2248, opened 2026-01-29
feat(router): meter server tool usage

willbakst:


willbakst:
* **#2248** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2248?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Mirascope?ref=merge-queue-instructions-in-stack-comment&prId=2248).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

cloud/api/router/tool-pricing.ts
```diff
@@ -0,0 +1,152 @@
...
+  anthropic_code_execution: {
+    costPerHour: dollarsToCenticents(0.05), // $0.05/hour
+    minBillingSeconds: 300, // 5 minutes minimum
+  },
+
+  // Google: Grounding with Google Search - per query (Gemini 3+)
+  // Source: https://ai.google.dev/gemini-api/docs/pricing
+  google_grounding_search: {
+    costPerCall: dollarsToCenticents(0.014), // $14/1,000 queries
+  },
```
teamdandelion, line 74: Note: for 2.5 models it's $35/1000 prompts. I don't mind just pricing it like 3+ since I figure folks will migrate to 3 (and as of right now it's only early adopters who are using mirascope anyway)

cloud/workers/routerMeteringQueue.ts
```diff
@@ -116,11 +119,13 @@ export const setRouterMeteringQueueLayer = (
...
  * @param data - Metering message data
- * @param costAsBigInt - Cost as bigint
+ * @param costAsBigInt - Token cost as bigint
+ * @param toolCostAsBigInt - Tool cost as bigint (optional)
  */
 export function updateAndSettleRouterRequest(
   data: RouterMeteringMessage,
   costAsBigInt: bigint,
+  toolCostAsBigInt?: bigint,
 ): Effect.Effect<void, Error, Database | Payments> {
```
teamdandelion, line 137: I don't think tool vs token costs should leak to this interface. It can just take `data` and `cost`

cloud/workers/routerMeteringQueue.ts
```diff
@@ -194,6 +201,13 @@ async function processMessage(
       ? data.costCenticents
       : BigInt(data.costCenticents);
 
+  // Convert tool cost to bigint if present
+  const toolCostAsBigInt = data.toolCostCenticents
+    ? typeof data.toolCostCenticents === "bigint"
+      ? data.toolCostCenticents
+      : BigInt(data.toolCostCenticents)
+    : undefined;
+
```
teamdandelion, line 225: let's compute the total cost here and pass it in

cloud/workers/routerMeteringQueue.ts
willbakst: We store them separately in the router table though? Or are you saying we should have `costCenticents` be the total and have two new fields `tokenCostCenticents` and `toolCostCenticents`? We can also have this take `CostBreakdown` or something

cloud/api/router/tool-pricing.ts
willbakst: I could also just do a stupid check to see if the model ID has `-3` in it or something and otherwise fallback to the true pricing

cloud/workers/routerMeteringQueue.ts
willbakst: I mentioned in my other comment, we store them individually on the table, so it seems like we should add another field for token cost, keep current cost as total, and pass in the `CostBreakdown` here or something

cloud/workers/routerMeteringQueue.ts
teamdandelion: Ah got it. Yeah in that case I think we should have something like `tokenCostCenticents` and `toolCostCentiCents` as peers and `costCenticents` as the total

cloud/api/router/tool-pricing.ts
teamdandelion: If you want to do this you'll need to change the abstraction because in 2.5 it's not a cost per call, it's a cost per _prompt_. I think it's more likely that future models will stick with something like the behavior in 3, so if we're writing a workaround check we should check for `2.5` specifically.   
  
The hacky way to get this behavior w.o. rewriting the interface would be to detect if its a 2.5 model when computing the # of calls, and if so always return that there are 2.5 calls, since $14 * 2.5 = $35. Maybe just do that?

cloud/workers/routerMeteringQueue.ts
teamdandelion: Agreed, that sgtm

PR #2249, opened 2026-01-29
feat(llm): add OpenRouterProvider for native OpenRouter support

koxudaxi:
### TL;DR

Added OpenRouter provider to support accessing multiple LLM providers through OpenRouter's API.

### What changed?

- Added a new `OpenRouterProvider` class that inherits from `BaseOpenAICompletionsProvider`
- Modified the OpenAI completions encoding logic to support non-OpenAI models via compatible APIs
- Added a `_model_features_name` method to bypass OpenAI-specific feature checks for non-OpenAI models
- Registered OpenRouter in the provider registry and provider ID lists
- Added appropriate tests for the new provider

### How to test?

1. Set the `OPENROUTER_API_KEY` environment variable with your OpenRouter API key
2. Register the provider with model ID prefixes:
   ```python
   from mirascope import llm
   
   # Register for openai and anthropic models via OpenRouter
   llm.register_provider("openrouter", scope=["openai/", "anthropic/"])
   
   # Use models directly
   @llm.call("openai/gpt-4")
   def my_prompt():
       return [llm.messages.user("Hello!")]
   ```
3. Verify that both OpenAI models and non-OpenAI models (like Anthropic) work correctly through the OpenRouter API

### Why make this change?

OpenRouter provides a unified API to access models from multiple providers (OpenAI, Anthropic, Meta, etc.) through a single endpoint. This integration allows users to:

1. Access a wider range of models through a single API
2. Potentially reduce costs by using OpenRouter's pricing
3. Easily switch between different model providers without changing code
4. Use non-OpenAI models with OpenAI-compatible parameters like temperature and top_p


koxudaxi:
* **#2269** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2249** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2249?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2249?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Mirascope?ref=merge-queue-instructions-in-stack-comment&prId=2249).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

python/mirascope/llm/providers/openai/completions/_utils/encode.py
```diff
@@ -303,7 +315,10 @@ def encode_request(
         )
     base_model_name = model_name(model_id, None)
 
-    is_reasoning_model = base_model_name not in NON_REASONING_MODELS
+    if model_features_name is None:
+        is_reasoning_model = False
+    else:
```
teamdandelion: This will cause a regression in the case where base_model_name is that of a reasoning model, no?

python/mirascope/llm/providers/openai/completions/_utils/encode.py
```diff
@@ -292,8 +292,20 @@ def encode_request(
...
+
+    Args:
+        model_id: The model ID to use for the request.
+        messages: The messages to send to the model.
+        tools: The tools available to the model.
+        format: The response format specification.
+        params: Additional parameters for the request.
+        model_features_name: The model name to use for OpenAI feature detection
+            (reasoning models, json schema support, etc.). If None, bypass OpenAI
+            feature checks (useful for non-OpenAI models via compatible APIs).
```
teamdandelion: This is a surprising API . I would expect model_features_name = None to use the standard model_id.removeprefix("openai/") approach for feature detection, not that it would disable for feature detection. This is especially confusing since model_features_name defaults to None

python/mirascope/llm/providers/openai/completions/_utils/encode.py
```diff
@@ -338,7 +353,12 @@ def encode_request(
 
     openai_tools = [_convert_tool_to_tool_param(tool) for tool in tools.tools]
 
-    model_supports_strict = base_model_name not in MODELS_WITHOUT_JSON_SCHEMA_SUPPORT
+    features_name_for_strict = (
+        model_features_name if model_features_name is not None else base_model_name
+    )
+    model_supports_strict = (
+        features_name_for_strict not in MODELS_WITHOUT_JSON_SCHEMA_SUPPORT
+    )
```
teamdandelion: this is inconsistent with the handling of is_resaoning_model and also inconsistent with the docstring.

python/mirascope/llm/providers/openai/completions/base_provider.py
```diff
@@ -83,6 +83,16 @@ def _model_name(self, model_id: str) -> str:
...
 
+    def _model_features_name(self, model_id: str) -> str | None:
+        """Get the model name for OpenAI feature detection (reasoning, json schema, etc).
+
+        Override in subclasses to return None for non-OpenAI models, which will
+        bypass OpenAI-specific feature checks (e.g., allow temperature/top_p/stop).
+
+        Default implementation returns _model_name() for backwards compatibility.
+        """
+        return self._model_name(model_id)
```
teamdandelion: Why is separate from model_name? Can you explain why we want the behavior to be different?

python/tests/e2e/conftest.py
```diff
@@ -37,6 +39,7 @@ def _(value: llm.ToolExecutionError) -> str:
     "google/gemini-2.5-flash",
     "openai/gpt-4o:completions",
     "openai/gpt-4o:responses",
+    "openrouter/openai/gpt-4o",
```
teamdandelion: Rather than adding this to every single e2e test, how about we set up a single test that is like test_openai_derived_providers.py with a basic sync and stream call, and have this (and others) be added to the models for just that test?  
  
Seeing as this uses the same encode and decode codepaths, I don't see a need to run the openrouter through every single test.   
  
However since the param and feature detection has some complexity, we could e.g. add a params test to that file, and test bvoth openrouter/openai/gpt-4o and also a reasoning model like openrouter/openai/gpt-5-mini and an anthropic model too.

python/mirascope/llm/providers/provider_registry.py
```diff
@@ -16,6 +16,7 @@
 from .openai import OpenAIProvider
 from .openai.completions.provider import OpenAICompletionsProvider
 from .openai.responses.provider import OpenAIResponsesProvider
+from .openrouter import OpenRouterProvider
 from .provider_id import ProviderId
 from .together import TogetherProvider
```
teamdandelion, line 22: <p class='graphite__hidden'><i>[Re: lines +55 to +63]</i></p>
we could add a provider default (after mirascope) for openrouter 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2249?utm_source=unchanged-line-comment">Graphite</a>.</p>

python/mirascope/llm/providers/openai/completions/base_provider.py
```diff
@@ -83,6 +84,14 @@ def _model_name(self, model_id: str) -> str:
         """Extract the model name to send to the API."""
         return openai_model_name(model_id, None)
 
+    def _model_features_name(self, model_id: str) -> str | SkipModelFeaturesType | None:
+        """Get the model name for OpenAI feature detection.
+
+        Override to return SKIP_MODEL_FEATURES for non-OpenAI models,
+        or a different name for feature detection than what _model_name returns.
+        """
+        return self._model_name(model_id)
```
teamdandelion, line 93: This still feels really weird to me.  
  
I think what we're reall yseeing is that when calling `encode_request`, we need to know what features are supported on the model:  
  
\- Is it known not to have audio support? If so raise FNSE  
\- Does it support strict mode?  
\- Does it have JSON object support?  
\- Is it a reasoning model?  
  
Currently encode_request looks that up based on the model name but we are seeing it is not robust to handling other providers.  
  
So how about we define (in a pre-factor PR, not this one) a dataclass like CompletionsModelFeatureInfo that has those properties (audio support, strict support, json object, reasoning). It might make sense for each field to be `bool | None` to distinguish the caase of "We know this def has/doesn't have support" vs "we just don't know anything about this model".  
  
Then encode_request doesn't try to derive any feature info from the model name, it just gets it passed in by the provider. And the provider can have an (overwritten) method like `feature_info_for_model_id`. So the "real" OpenAI compeltions model can use the built in support tracking, etc.  
  
This would feel a lot more clear around what we actually care about (tracking model feature based on the model id) in a way that has room in the future to add better tracking for OpenRouter models if someone cares about it. Rather than cramming more info into an overriden string in a way that makes the code harder to follow.  
  
Does that make sense?

koxudaxi: OK, I have created the upstack for refactoring. https://app.graphite.com/github/pr/Mirascope/mirascope/2269/refactor(llm)-introduce-CompletionsModelFeatureInfo-dataclass

teamdandelion (approved):

willbakst (approved):

PR #2269, opened 2026-01-30
refactor(llm): introduce CompletionsModelFeatureInfo dataclass

koxudaxi:


koxudaxi:
* **#2269** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2249** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Mirascope?ref=merge-queue-instructions-in-stack-comment&prId=2269).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

python/mirascope/llm/providers/openai/completions/base_provider.py
```diff
@@ -84,13 +84,13 @@ def _model_name(self, model_id: str) -> str:
...
+    def _model_feature_info(self, model_id: str) -> CompletionsModelFeatureInfo:
+        """Get feature info for the model.
 
-        Override to return SKIP_MODEL_FEATURES for non-OpenAI models,
-        or a different name for feature detection than what _model_name returns.
+        Override to return custom feature info for non-OpenAI models.
         """
-        return self._model_name(model_id)
+        base_name = openai_model_name(model_id, None)
+        return feature_info_for_openai_model(base_name)
```
teamdandelion: How about the base provider returns an empty feature info, and the OpenAICompletionsProvider in python/mirascope/llm/providers/openai/completions/provider.py uses this logic to return for actual openai models?

koxudaxi: I fixed it.

python/mirascope/llm/providers/openai/completions/_utils/encode.py
```diff
@@ -302,23 +293,10 @@ def encode_request(
...
-        )
-    else:
-        effective_features_name = base_model_name
-        is_reasoning_model = base_model_name not in NON_REASONING_MODELS
-        model_supports_strict = (
-            base_model_name not in MODELS_WITHOUT_JSON_SCHEMA_SUPPORT
-        )
+    is_reasoning_model = feature_info.is_reasoning_model is True
+    model_supports_strict = feature_info.strict_support is True
+    supports_json_object = feature_info.json_object_support is True
```
teamdandelion, line 312: Hmm, what are the implications of defining things this way?  
  
Eg below:  


```
    if format is not None:
        if format.mode == "strict":
            if not model_supports_strict:
                raise FeatureNotSupportedError(
                    feature="formatting_mode:strict",
                    provider_id="openai",
                    model_id=model_id,
                )

```

  
So this means for every openrouter model, we will not support strict formatting (and, confusingly, we will set provider_id="openai" even though that may not be the provider).  
  
I think in a case where we genuinely don't know what the model may or may not support, we should err on the side of being permissive — i.e. allow the user to choose strict mode, and let it fail only if the provider actually can't support it. The "eager" errors should happen only when we definitely know that the feature is unsupported.

koxudaxi: Good catch. I fixed it. Thank you

PR #2495, opened 2026-02-07
fix: correct typo 'dont' to "don't" in example

thecaptain789:
Simple typo fix in openai_responses_reasoning.py example.


teamdandelion:
Thanks for your PR, but we're not interested in typo fixes unless the affect user facing docs or would cause confusion.

PR #2546, opened 2026-02-10
feat(dispatch-worker): add miniflare integration tests for bootstrap functions

teamdandelion:
### Context

The dispatch worker (`cloud/claws/dispatch-worker/`) routes requests to per-claw Cloudflare Sandbox containers. It communicates with the Mirascope Cloud backend via Cloudflare **service bindings** — in-process RPC calls that look like `fetch()` but never hit the network.

The existing unit tests mock the service binding entirely (just spy on `env.MIRASCOPE_CLOUD.fetch`). This validates the dispatch worker's logic but doesn't test the actual service binding behavior — request/response serialization, URL routing through the binding, error propagation, etc.

### What this does

Adds integration tests that spin up a real **Miniflare** instance (Cloudflare's local Workers runtime) with:
- A **test harness worker** (main) that exposes bootstrap functions via HTTP
- A **mock cloud worker** bound as `MIRASCOPE_CLOUD` via a function-based service binding

This lets us test that `fetchBootstrapConfig()`, `resolveClawId()`, and `reportClawStatus()` work correctly through a real Workers runtime with real service binding semantics.

### Why Miniflare (not `@cloudflare/vitest-pool-workers`)

The project uses vitest 4.x, but `@cloudflare/vitest-pool-workers` requires vitest 2.x-3.2.x. Rather than downgrade, we use Miniflare programmatically — same underlying runtime, just invoked from standard vitest.

### Files

- **`src/integration/bootstrap.integration.test.ts`** — 5 tests covering bootstrap config fetch, claw slug resolution, and status reporting through real service bindings
- **`src/integration/test-harness-worker.ts`** — Thin HTTP wrapper exposing bootstrap functions for the test to call
- **`src/integration/mock-cloud-worker.ts`** — Simulates the Mirascope Cloud internal API (bootstrap, resolve, status endpoints)
- **`vitest.integration.config.ts`** — Separate vitest config for integration tests (longer timeouts)
- **`package.json`** — Added `test:unit` and `test:integration` scripts; added `esbuild` and `miniflare` dev deps

### How it fits in the stack

This is the foundation of the dispatch worker test infrastructure. The stack builds up testing layers:

1. **This PR** — Miniflare integration tests (fast, in-process, tests service binding behavior)
2. #2541-#2545 — Docker harness (real OpenClaw gateway in a container, tests HTTP/WS proxy)
3. #2547 — Auth middleware implementation
4. #2548 — Auth matrix miniflare tests (17 tests covering all auth paths)

[...]

PR #2541, opened 2026-02-10
feat(dispatch-worker): add Docker Compose config and health check script

teamdandelion:
### Context

The dispatch worker proxies HTTP and WebSocket requests to an OpenClaw gateway running inside a Cloudflare Sandbox container. To test this proxy behavior locally and in CI, we need a real OpenClaw gateway running — not a mock.

This PR adds Docker infrastructure to run a real OpenClaw gateway container for integration testing.

### What this does

- **`docker-compose.yml`** — Defines an `openclaw-gateway` service that builds from the OpenClaw repo source and runs the gateway in test mode (`--allow-unconfigured --bind lan --port 18789`). Includes a health check using Node built-in HTTP client (no curl dependency). Sets `NODE_OPTIONS=--max-old-space-size=2048` since OpenClaw needs ~500MB heap at startup.

- **`wait-for-gateway.sh`** — Polling script that waits for the gateway HTTP server to respond before tests run. Uses curl to check for any HTTP response (any status code = alive) with configurable timeout.

### Key design decisions

- **Build from source, not a published image** — OpenClaw does not publish Docker images yet. The `OPENCLAW_SRC` env var points to your local checkout of the openclaw repo. It is required (fails with a clear error if unset).
- **Node for health check, not curl** — The Docker health check uses `node -e` since curl may not be in the image. The wait script uses curl (available on CI runners and dev machines).
- **Port 18789** — Matches the gateway default port in the dispatch worker config.
- **No resource limits** — GitHub Actions runners have 7GB RAM which is sufficient. Removed explicit limits to avoid V8 heap OOM during gateway startup.

PR #2547, opened 2026-02-10
feat(dispatch-worker): add auth middleware with path-based routing

teamdandelion:
### What this does

This implements the authentication layer for the dispatch worker, following the URL routing and auth design from `cloud/claws/README.md`.

**Before:** The dispatch worker extracted `clawId` from the `Host` header (e.g., `{clawId}.claws.mirascope.com`) and proxied all requests through with no auth checks. This only worked for internal service-binding traffic.

**After:** The dispatch worker handles two types of traffic:

1. **Internal** (`/_internal/*`) — Still uses Host-based `clawId` extraction. These come from the Cloud backend via Cloudflare service binding (in-process RPC, no network). No auth needed because the service binding *is* the trust boundary.

2. **External** (`/{orgSlug}/{clawSlug}/...`) — New path-based routing. The dispatch worker parses the org+claw slugs from the URL path, resolves them to a `clawId` via the Cloud internal API, then applies the auth decision matrix before proxying.

### Auth decision matrix

```
Path                           Auth Method              Validated By
────────────────────────────── ──────────────────────── ─────────────
/{org}/{claw}/webhook/*        Platform-specific        Gateway (pass-through)
/{org}/{claw}/*  + Bearer      OPENCLAW_GATEWAY_TOKEN   Gateway (pass-through)
/{org}/{claw}/*  + Cookie      Mirascope session        Dispatch worker → Cloud API
/{org}/{claw}/*  (no auth)     Reject                   Dispatch worker (401)
/_internal/*                   Service binding          Implicit (in-process)
```

**Key design decisions:**

- **Webhooks pass through** — Telegram, Slack, Discord etc. each have their own signature verification. The dispatch worker doesn't have platform secrets; the gateway does. So we let webhook traffic through and let the gateway validate.
- **Bearer tokens pass through** — The `OPENCLAW_GATEWAY_TOKEN` lives in the container's env vars. The dispatch worker doesn't have the decrypted token, so it forwards the `Authorization` header and lets the gateway validate.
- **Session cookies are validated at the dispatch worker** — Browser users have a `.mirascope.com` session cookie. The dispatch worker validates this via a new internal Cloud API endpoint (`/api/internal/auth/validate-session`), with a 60s LRU cache to avoid re-validating on every request.
- **No auth = 401** — If none of the above match, reject.

### URL rewriting

After auth, the dispatch worker rewrites the request URL before proxying to the gateway. The gateway sees the *remainder* path, not the full `/{org}/{claw}/...` prefix:

```
Client request:  GET /acme/support-bot/api/chat
Gateway sees:    GET /api/chat
```

### CORS

Since `openclaw.mirascope.com` is a different origin from `mirascope.com`, the Chrome SPA needs CORS headers. This PR adds:
- `Access-Control-Allow-Origin: https://mirascope.com` (not `*`, because credentials)
- `Access-Control-Allow-Credentials: true` (so the session cookie is sent)
- OPTIONS preflight handling at the dispatch worker (not proxied to gateway)

### Files changed

- **`auth.ts`** (new) — `parsePath()`, `authenticate()`, `validateSession()`, `corsHeaders()`, `handlePreflight()`
- **`index.ts`** — Split routing: `/_internal/*` keeps Host-based, everything else uses path-based + auth

### Requires (not in this PR)

- Cloud backend to implement `POST /api/internal/auth/validate-session` (service-binding-only endpoint)
- Session cookie `Domain` set to `.mirascope.com` (so it's sent to `openclaw.mirascope.com`)

Co-authored-by: Verse <verse@mirascope.com>

### DIFF
--- a/cloud/claws/dispatch-worker/src/auth.ts
+++ b/cloud/claws/dispatch-worker/src/auth.ts
@@ -1,0 +1,269 @@
+/**
+ * Auth middleware for the dispatch worker.
+ *
+ * Implements the auth decision matrix from cloud/claws/README.md:
+ *
+ *   Path                           Auth Method              Validated By
+ *   /{org}/{claw}/webhook/*        Platform-specific        Gateway (pass-through)
+ *   /{org}/{claw}/*  + Bearer      OPENCLAW_GATEWAY_TOKEN   Gateway (pass-through)
+ *   /{org}/{claw}/*  + Cookie      Mirascope session        Dispatch worker → Cloud API
+ *   /{org}/{claw}/*  (no auth)     Reject                   Dispatch worker (401)
+ *   /_internal/*                   Service binding          Implicit (in-process)
+ */
+
+import { Data, Effect } from "effect";
+
+import type { DispatchEnv } from "./types";
+
+import {
+  type BootstrapDecodeError,
+  type ClawResolutionError,
+  resolveClawId,
+} from "./bootstrap";
+
+// ---------------------------------------------------------------------------
+// Errors
+// ---------------------------------------------------------------------------
+
+/** Session cookie validation failed. */
+export class InvalidSessionError extends Data.TaggedError(
+  "InvalidSessionError",
+)<{
+  readonly message: string;
+}> {}
+
+/** No authentication credentials provided. */
+export class UnauthenticatedError extends Data.TaggedError(
+  "UnauthenticatedError",
+)<{
+  readonly message: string;
+}> {}
+
+/** JSON decode failed when parsing API response. */
+export class ApiDecodeError extends Data.TaggedError("ApiDecodeError")<{
+  readonly message: string;
+  readonly cause?: unknown;
+}> {}
+
+export type AuthError =
+  | ClawResolutionError
+  | BootstrapDecodeError
+  | InvalidSessionError
+  | UnauthenticatedError
+  | ApiDecodeError;
+
+// ---------------------------------------------------------------------------
+// Path parsing
+// ---------------------------------------------------------------------------
+
+export interface ParsedPath {
+  orgSlug: string;
+  clawSlug: string;
+  /** The remainder after /{org}/{claw}, e.g. "/webhook/telegram" or "/api/chat" */
+  remainder: string;
+}
+
+/**
+ * Parse /{orgSlug}/{clawSlug}/... from a URL path.
+ * Returns null if the path doesn't match the expected pattern.
+ */
+export function parsePath(pathname: string): ParsedPath | null {
+  const match = pathname.match(/^\/([\w-]+)\/([\w-]+)(\/.*)?$/);
+  if (!match) return null;
+  return {
+    orgSlug: match[1],
+    clawSlug: match[2],
+    remainder: match[3] || "/",
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Session validation cache
+// ---------------------------------------------------------------------------
+
+interface CachedSession {
+  userId: string;
+  clawId: string;
+  organizationId: string;
+  role: string;
+  expiresAt: number;
+}
+
+const SESSION_CACHE_TTL_MS = 60_000; // 60 seconds
+const sessionCache = new Map<string, CachedSession>();
+
+function getSessionCacheKey(
+  sessionId: string,
+  orgSlug: string,
+  clawSlug: string,
+): string {
+  return `${sessionId}:${orgSlug}:${clawSlug}`;
+}
+
+/**
+ * Validate a session cookie via the MIRASCOPE_CLOUD service binding.
+ * Results are cached for 60s.
+ */
+export const validateSession = (
+  sessionId: string,
+  orgSlug: string,
+  clawSlug: string,
+  env: DispatchEnv,
+): Effect.Effect<CachedSession, InvalidSessionError | ApiDecodeError> =>
+  Effect.gen(function* () {
+    const cacheKey = getSessionCacheKey(sessionId, orgSlug, clawSlug);
+    const cached = sessionCache.get(cacheKey);
+    if (cached && cached.expiresAt > Date.now()) {
+      return cached;
+    }
+
+    const response = yield* Effect.tryPromise({
+      try: () =>
+        env.MIRASCOPE_CLOUD.fetch(
+          "https://internal/api/internal/auth/validate-session",
+          {
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({
+              sessionId,
+              organizationSlug: orgSlug,
+              clawSlug,
+            }),
+          },
+        ),
+      catch: (cause) =>
+        new ApiDecodeError({
+          message: "Failed to call validate-session API",
+          cause,
+        }),
+    });
+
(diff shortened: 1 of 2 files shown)


teamdandelion:
* **#2548** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2548?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2553](https://github.com/Mirascope/mirascope/pull/2553) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2553?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2554](https://github.com/Mirascope/mirascope/pull/2554) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2554?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2547** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2556** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2556?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2555** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2555?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2546** <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2546?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2541](https://github.com/Mirascope/mirascope/pull/2541) <a href="https://app.graphite.com/github/pr/Mirascope/mirascope/2541?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Mirascope?ref=merge-queue-instructions-in-stack-comment&prId=2547).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

cloud/claws/dispatch-worker/src/auth.ts
```diff
@@ -0,0 +1,240 @@
...
+      }),
+    },
+  );
+
+  if (!response.ok) {
+    sessionCache.delete(cacheKey);
+    return null;
+  }
+
+  const data = (await response.json()) as {
```
willbakst: There's no error handling here? What if `response.json()` fails?

cloud/claws/dispatch-worker/src/auth.ts
```diff
@@ -0,0 +1,240 @@
...
+  clawSlug: string,
+): string {
+  return `${sessionId}:${orgSlug}:${clawSlug}`;
+}
+
+/**
+ * Validate a session cookie via the MIRASCOPE_CLOUD service binding.
+ * Results are cached for 60s.
+ */
+export async function validateSession(
```
willbakst: Any reason we're not using Effect here?

cloud/claws/dispatch-worker/src/auth.ts
```diff
@@ -0,0 +1,240 @@
...
+ * This function decides whether the request should be proxied or rejected.
+ */
+export async function authenticate(
+  request: Request,
+  parsed: ParsedPath,
+  env: DispatchEnv,
+): Promise<AuthResult> {
+  // Resolve clawId from slugs
+  let clawId: string;
+  try {
```
willbakst: Use Effect?

cloud/claws/dispatch-worker/src/auth.ts
```diff
@@ -0,0 +1,240 @@
...
+  const response = await env.MIRASCOPE_CLOUD.fetch(
+    `https://internal/api/internal/claws/resolve/${orgSlug}/${clawSlug}`,
+    {
+      method: "GET",
+      headers: { "Content-Type": "application/json" },
+    },
+  );
+
+  if (!response.ok) {
+    throw new Error(`Resolution failed: ${response.status}`);
```
willbakst: yeah same thing here we shouldn't be throwing

cloud/claws/dispatch-worker/src/auth.ts
teamdandelion: Verse: You're right, this is unguarded. Will fix in the Effect refactor PR upstack — this becomes a typed decode with a proper failure channel instead of bare response.json().

cloud/claws/dispatch-worker/src/auth.ts
teamdandelion: Verse: No good reason — just wrote it imperative-first to get the logic right. Planning an Effect refactor as a separate PR upstack so it's consistent with the rest of cloud/. The session cache, slug resolution, and auth flow all map cleanly to Effect.

cloud/claws/dispatch-worker/src/auth.ts
teamdandelion: Verse: Yep, will refactor to Effect in the follow-up PR.

cloud/claws/dispatch-worker/src/auth.ts
teamdandelion: Verse: Agreed — this should return a typed error, not throw. Will convert to Effect.fail with a proper error type in the refactor PR.

cloud/claws/dispatch-worker/src/index.ts
```diff
@@ -57,28 +60,61 @@ app.use("*", async (c, next) => {
...
+  );
+
+  const sandbox = getSandbox(c.env.Sandbox, clawId, { keepAlive: true });
+  c.set("sandbox", sandbox);
+  c.set("clawId", clawId);
+
+  // Rewrite request URL to remainder path before proxying
+  const remainderUrl = new URL(request.url);
+  remainderUrl.pathname = parsed.remainder;
+  const rewrittenRequest = new Request(remainderUrl.toString(), request);
```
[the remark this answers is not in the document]
teamdandelion: Verse: Good catch — WebSocket connections also need the path rewrite since the gateway expects the remainder path, not the full /{org}/{claw}/... prefix. Both HTTP and WS should see the same rewritten path. The current code does rewrite before the WS check, which is correct. Will verify in the Effect refactor.

cloud/claws/dispatch-worker/src/index.ts
```diff
@@ -57,28 +60,61 @@ app.use("*", async (c, next) => {
...
+  if (authResult.action === "reject") {
+    const origin = request.headers.get("Origin");
+    return c.json(
+      { error: authResult.message },
+      {
+        status: authResult.status as any,
+        headers: corsHeaders(origin),
+      },
+    );
+  }
```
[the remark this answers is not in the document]
teamdandelion: Verse: Fair point — returning a JSON body on a WS upgrade request won't reach the client properly. For now the HTTP 401 response will cause the WebSocket handshake to fail (which is the desired behavior), but the error message won't be visible to WS clients. Will address WS-specific error responses in the Effect refactor.

PR #2863, opened 2026-05-28
fix(anthropic): strip FORMAT_TOOL from re-encoded history to fix resume() HTTP 400

### LINKED ISSUE #2503: `response.resume` broken when using structured outputs with Anthropic in tool mode
### Description

Consider the following script:

```python
from mirascope import llm


@llm.call("anthropic/claude-sonnet-4-5", format=int)
def lucky_number():
    return "Choose a lucky number between 1 and 10"


result = lucky_number()
second_result = result.resume("Ok, now choose a different lucky number")
```

This is very simple and should work. However, it raises the following error:

```
mirascope.llm.exceptions.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.2: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01NUbw5mvfYRWaDHvMi1xYK1. Each `tool_use` block must have a corresponding `tool_result` block in the next message.'}, 'request_id': 'req_011CXxuFSqH1PCSqbYYrukMX'}
```

The issue is that:
1. Anthropic provider uses tool mode by default
2. Tool mode injects a special MIRASCOPE_FORMAT_TOOL, which is used for the output
3. When Mirascope processes the assistant message, it converts the tool call into a text block that contains the expected output. There is never a corresponding user block with tool output (because the format tool is never called)
4. When re-encoding the message history to send to Anthropic, we use the raw representation (unprocessed) which still has a tool call
5. Anthropic rejects the request because the tool was never called

Solutions that come to mind include:
1. When encoding the assistant message, check if it contained a format tool invocation, and if so re-generate it to have the text output, not the tool call. Sub-optimal because it may lose reasoning / thinking continuity and (if Anthropic auto-cached its output tokens) invalidate the cache. Also just feels a little hacky.
2. Inject an additional user message invoking the tool when resuming the conversation. Also feels a little hacky — and in that case will Anthropic reject the subsequent real user message as violating turn order expectations?

Note, this issue is particularly pressing because:
- Tool use is the default structured format mode for Anthropic
- Response.validate() is a first class API that wraps a resume loop under the hood

However, we can mitigate the impact of this bug by switching Anthropic to use strict formatting by default (by opting into the beta) for all models that support it.

Thanks to Dahlia-claw at Reverie for finding and reporting this bug.

```TOML

```

devteamaegis:
## Problem

`response.resume()` fails with HTTP 400 when using Anthropic with structured outputs (`format=SomeModel`):

```
BadRequestError: 400 - tool_use ids were found without tool_result blocks immediately after
```

Fixes #2503

### Root cause

Anthropic structured-output mode injects a private `FORMAT_TOOL_NAME` tool.  The model response contains a `tool_use` block for that tool.  When the assistant message appears in the history of a `resume()` call:

1. `_encode_message` detects the format tool (`raw_message_has_format_tool` → `True`) and falls through to re-encode from content parts instead of using the raw message verbatim.
2. `encode_content` encodes **every** `tool_call` part as a `ToolUseBlockParam` — including the `FORMAT_TOOL_NAME` call.
3. Anthropic rejects the request because there is no `tool_result` in the following user message (there never is — the format tool is a Mirascope detail that is never actually invoked).

### Fix

In `encode_content`, skip any `tool_call` part whose name starts with `FORMAT_TOOL_NAME`:

```python
elif part.type == "tool_call":
    if part.name.startswith(FORMAT_TOOL_NAME):
        continue   # ← new guard
    blocks.append(ToolUseBlockParam(...))
```

This preserves all real tool_use blocks while stripping the implementation-detail format tool from re-encoded history.

## Tests

Added to `python/tests/llm/providers/anthropic/test_anthropic_encoding.py`:

| test | what it checks |
|---|---|
| `test_encode_content_skips_format_tool_call` | FORMAT_TOOL tool_use absent from re-encoded blocks |
| `test_encode_content_keeps_real_tool_calls` | genuine tool_use blocks still emitted |

```
pytest tests/llm/providers/anthropic/test_anthropic_encoding.py -v
3 passed
```

### DIFF
--- a/python/mirascope/llm/providers/anthropic/_utils/encode.py
+++ b/python/mirascope/llm/providers/anthropic/_utils/encode.py
@@ -278,6 +278,19 @@ def encode_content(
                 )
             )
         elif part.type == "tool_call":
+            # Skip the internal FORMAT_TOOL call when re-encoding history.
+            # Anthropic's structured-output tool (FORMAT_TOOL_NAME) is a Mirascope
+            # implementation detail injected to force JSON-shaped responses.  When
+            # the assistant message is re-used in a `resume()` call the raw
+            # message is NOT sent back verbatim (because raw_message_has_format_tool
+            # is True), so we fall through to encode from content parts.  Without
+            # this guard the format-tool ToolCall part is re-emitted as a
+            # `tool_use` block, but there is no corresponding `tool_result` in
+            # the next user message — Anthropic rejects the request with 400:
+            #   "tool_use ids were found without tool_result blocks immediately after"
+            # See https://github.com/Mirascope/mirascope/issues/2503
+            if part.name.startswith(FORMAT_TOOL_NAME):
+                continue
             blocks.append(
                 anthropic_types.ToolUseBlockParam(
                     type="tool_use",
--- a/python/mirascope/llm/providers/anthropic/_utils/encode.py
+++ b/python/mirascope/llm/providers/anthropic/_utils/encode.py
@@ -278,18 +278,30 @@ def encode_content(
                 )
             )
         elif part.type == "tool_call":
-            # Skip the internal FORMAT_TOOL call when re-encoding history.
+            # Re-encode the internal FORMAT_TOOL call as text when re-using history.
             # Anthropic's structured-output tool (FORMAT_TOOL_NAME) is a Mirascope
             # implementation detail injected to force JSON-shaped responses.  When
             # the assistant message is re-used in a `resume()` call the raw
             # message is NOT sent back verbatim (because raw_message_has_format_tool
-            # is True), so we fall through to encode from content parts.  Without
-            # this guard the format-tool ToolCall part is re-emitted as a
-            # `tool_use` block, but there is no corresponding `tool_result` in
-            # the next user message — Anthropic rejects the request with 400:
+            # is True), so we fall through to encode from content parts.  Emitting
+            # the format-tool ToolCall as a `tool_use` block would 400, since there
+            # is no matching `tool_result` in the next user message:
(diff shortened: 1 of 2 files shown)


python/mirascope/llm/providers/anthropic/_utils/encode.py
```diff
@@ -278,6 +278,19 @@ def encode_content(
...
+            # implementation detail injected to force JSON-shaped responses.  When
+            # the assistant message is re-used in a `resume()` call the raw
+            # message is NOT sent back verbatim (because raw_message_has_format_tool
+            # is True), so we fall through to encode from content parts.  Without
+            # this guard the format-tool ToolCall part is re-emitted as a
+            # `tool_use` block, but there is no corresponding `tool_result` in
+            # the next user message — Anthropic rejects the request with 400:
+            #   "tool_use ids were found without tool_result blocks immediately after"
+            # See https://github.com/Mirascope/mirascope/issues/2503
+            if part.name.startswith(FORMAT_TOOL_NAME):
```
willbakst, line 294: Doesn't this mean the content of this final tool call isn't included at all? Shouldn't we be re-encoding it as text in this case?

devteamaegis:
Good catch @willbakst — you're right, `continue` dropped the content entirely. The format tool's arguments are the assistant's actual structured response, so instead of skipping I now re-encode them as a text block: the JSON the model produced is preserved in the history, and there's still no orphaned `tool_use` block to trigger the 400 on resume. Updated the test to assert both text parts survive (`["Here is the result.", '{"value": 7}']`). Tests pass.

PR #2864, opened 2026-05-28
fix(ops): forward __qualname__/__name__ from VersionedCall to fix @ops.trace stacking

devteamaegis:
## Summary

Combining `@ops.trace` with `@ops.version` + `@llm.call` raises `AttributeError`
on Python 3.12+ (and was confirmed on 3.14):

```
AttributeError: 'VersionedCall' object has no attribute '__qualname__'
```

The error occurs because `@ops.version` wraps the `Call` object in a
`VersionedCall` dataclass, but `VersionedCall.__post_init__` never calls
`copy_function_metadata` to forward `__qualname__`, `__name__`, etc. from the
underlying `Call`.

`Call.__init__` already calls `copy_function_metadata(self, self.prompt.fn)`,
so the `Call` object carries those attributes. The fix is a single
`copy_function_metadata(self, self._call)` call at the end of `__post_init__`
in all four versioned-call classes: `VersionedCall`, `VersionedAsyncCall`,
`VersionedContextCall`, and `VersionedAsyncContextCall`.

As a bonus, this also sets `__wrapped__`, allowing `get_original_fn()` in
`utils.py` to follow the wrapper chain correctly.

## Reproduction

```python
from mirascope import llm, ops

ops.configure()
ops.instrument_llm()

@ops.trace(tags=["test"])
@ops.version
@llm.call("openai/gpt-4o-mini")
def run(text: str) -> str:
    return f"Echo: {text}"
# AttributeError: 'VersionedCall' object has no attribute '__qualname__'
```

## Tests

Added two regression tests:
- `test_versioned_call_exposes_qualname_for_trace` (sync)
- `test_versioned_async_call_exposes_qualname_for_trace` (async)

Both verify that `__qualname__`/`__name__` are accessible and that
`@ops.trace` can be applied without raising `AttributeError`.

Fixes #2784

### DIFF
--- a/python/tests/ops/_internal/test_versioning.py
+++ b/python/tests/ops/_internal/test_versioning.py
@@ -1661,3 +1661,47 @@ async def test_async_version_continues_when_closure_is_none(
                 "events": [],
             }
         )
+
+def test_versioned_call_exposes_qualname_for_trace() -> None:
+    """VersionedCall must expose __qualname__ and __name__ so @ops.trace can introspect it.
+
+    Regression test for https://github.com/Mirascope/mirascope/issues/2784:
+    combining @ops.trace + @ops.version + @llm.call raised
+        AttributeError: 'VersionedCall' object has no attribute '__qualname__'
+    because VersionedCall.__post_init__ did not call copy_function_metadata.
+    """
+
+    @ops.version
+    @llm.call("openai/gpt-4o-mini")
+    def my_function(text: str) -> str:
+        return f"Echo: {text}"
+
+    # The VersionedCall must carry __qualname__ and __name__ from the original fn.
+    assert hasattr(my_function, "__qualname__"), (
+        "VersionedCall must expose __qualname__ for @ops.trace to work"
+    )
+    assert hasattr(my_function, "__name__"), (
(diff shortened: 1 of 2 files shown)


python/tests/ops/_internal/test_versioning.py
```diff
@@ -1661,3 +1661,47 @@ async def compute(x: int) -> int:
...
             }
         )
+
+def test_versioned_call_exposes_qualname_for_trace() -> None:
+    """VersionedCall must expose __qualname__ and __name__ so @ops.trace can introspect it.
+
+    Regression test for https://github.com/Mirascope/mirascope/issues/2784:
+    combining @ops.trace + @ops.version + @llm.call raised
+        AttributeError: 'VersionedCall' object has no attribute '__qualname__'
+    because VersionedCall.__post_init__ did not call copy_function_metadata.
```
willbakst, line 1672: there's some weird formatting here

devteamaegis:
Fixed — that was a missing blank line (only one before the new top-level test def, needs two). Added the second blank line so it's PEP8/ruff-clean now. Thanks @willbakst!

PR #2865, opened 2026-05-29
feat(providers): add XAIProvider for Grok via Responses API

### LINKED ISSUE #2091: Add support for Grok using their Responses API compatibility
### Description

Grok recently sent an email mentioning they are deprecating their `v1/messages` Anthropic compatible endpoint. They mention to use their gRPC or Responses API. They seem to have `v1/chat/completions` (although who knows for how long).

Seems like the best way to add support for Grok with our current system would be to add support using our existing `:completions` and `:responses` OpenAI support but pointed at Grok. This would make it easier to add support than their full gRPC client, although we have a PR that adds support (it just doesn't work as well with VCR because it's gRPC and requires custom stuff).

gRPC is likely the better longer term bet (cause it's their core SDK), but adding support sooner sounds pretty good. Nothing stopping us from doing Completions + Responses first as an easy win and tackling gRPC down the line. Worth considering how we implement things though to ensure we keep the door open for the gRPC implementation.

NishchayMahor:
## Summary

Closes #2091.

xAI deprecated their Anthropic-compatible `v1/messages` endpoint and pointed users at their OpenAI-compatible Responses API at `https://api.x.ai/v1`. This PR adds a thin `XAIProvider` that subclasses `OpenAIResponsesProvider` with xAI-specific defaults, so Grok models work like any other provider in Mirascope.

## What changed

- `python/mirascope/llm/providers/xai/` (new package): `XAIProvider` overrides `id`, `default_scope`, and the constructor so it resolves `XAI_API_KEY` from the environment by default and points at the xAI base URL. Otherwise inherits everything from `OpenAIResponsesProvider`.
- `provider_id.py`: added `"xai"` to `KnownProviderId`.
- `provider_registry.py`: imported `XAIProvider`, added the `xai/` entry to `DEFAULT_AUTO_REGISTER_SCOPES` (with `XAI_API_KEY` as the env var), and added the `case "xai"` branch in `provider_singleton`.
- `providers/__init__.py`: registered the new `xai` stub module and exported `XAIProvider`.
- `tests/conftest.py`: added the `XAI_API_KEY` dummy key so the `test_all_default_scopes_auto_register` sweep continues to pass.

## Behavior

After registering:
```python
from mirascope import llm

llm.register_provider("xai")

@llm.call("xai/grok-2-latest")
def hello():
    return [llm.messages.user("Hi")]
```

xAI Completions support and any xAI-specific model registry / feature info can follow in a separate PR if useful.

## Test plan

- [x] `uv run pytest tests/llm/providers/test_xai_provider.py` (5 new tests covering api_key, env var, missing key, default base URL, custom base URL).
- [x] `uv run pytest tests/llm/providers/test_provider_registry.py` passes, including `test_all_default_scopes_auto_register` which now also sweeps the new `xai/` scope.
- [x] `uv run pytest tests/llm/ --ignore=tests/llm/e2e` reports `920 passed, 3 skipped`.
- [x] `uv run coverage run -m pytest tests/llm/providers/test_xai_provider.py && uv run coverage report --include="*/xai/*"` reports 100% line coverage on the new module.
- [x] `uv run ruff check` and `uv run ruff format --check` pass on touched files.
- [x] `uv run pyright` reports 0 errors on touched files.

### DIFF
--- a/python/mirascope/llm/providers/__init__.py
+++ b/python/mirascope/llm/providers/__init__.py
@@ -13,6 +13,7 @@ stub_module_if_missing("mirascope.llm.providers.openai", "openai")
 stub_module_if_missing("mirascope.llm.providers.openrouter", "openai")
 stub_module_if_missing("mirascope.llm.providers.together", "openai")
 stub_module_if_missing("mirascope.llm.providers.ollama", "openai")
+stub_module_if_missing("mirascope.llm.providers.xai", "openai")
 
 # Now imports work regardless of which packages are installed
 # ruff: noqa: E402
@@ -39,6 +40,7 @@ from .provider_registry import (
     reset_provider_registry,
 )
 from .together import TogetherProvider
+from .xai import XAIProvider
 
 __all__ = [
     "KNOWN_PROVIDER_IDS",
@@ -59,6 +61,7 @@ __all__ = [
     "Provider",
     "ProviderId",
     "TogetherProvider",
+    "XAIProvider",
     "get_provider_for_model",
     "register_provider",
     "reset_provider_registry",
--- a/python/mirascope/llm/providers/__init__.py
+++ b/python/mirascope/llm/providers/__init__.py
@@ -40,7 +40,7 @@ from .provider_registry import (
     reset_provider_registry,
 )
 from .together import TogetherProvider
-from .xai import XAIProvider
+from .xai import XAIModelId, XAIProvider
 
 __all__ = [
     "KNOWN_PROVIDER_IDS",
@@ -61,6 +61,7 @@ __all__ = [
     "Provider",
     "ProviderId",
     "TogetherProvider",
+    "XAIModelId",
     "XAIProvider",
     "get_provider_for_model",
     "register_provider",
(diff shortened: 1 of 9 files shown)


willbakst (commented):
Thanks for adding this! Could you please include some top level default model strings for autocomplete? That will make using the XAI integration much smoother.

NishchayMahor:
Great call @willbakst :D, that makes the integration way more discoverable. Done in fc473c34:

- New `python/mirascope/llm/providers/xai/model_info.py` with an `XAIKnownModels` Literal of the publicly documented Grok models from https://docs.x.ai/docs/models (grok-2 / grok-2-vision / grok-3 / grok-4 family, grok-beta, grok-code-fast-1).
- New `python/mirascope/llm/providers/xai/model_id.py` with `XAIModelId: TypeAlias = XAIKnownModels | str` and a `XAI_KNOWN_MODELS` set, mirroring the Anthropic shape so behavior stays consistent across providers.
- Re-exported `XAIModelId` from `xai/__init__.py` and the top-level `providers/__init__.py` so IDE autocomplete kicks in on `llm.providers.XAIModelId`.
- Added a unit test that pins the shape (every entry `xai/`-scoped, non-empty set, sample model in the set).

Local sweep: 6/6 tests passing on the file, 100% line coverage on the xai package, ruff/pyright/format clean. Thanks again for the steer, happy to keep iterating if you'd like the list trimmed or formatted differently :)

willbakst (approved):

willbakst:
@NishchayMahor sorry this took so long! It's now released in `v2.5.0`. Thank you for contributing!!

PR #2868, opened 2026-06-25
fix: emit OTel/Langfuse-conventional span attributes for @ops.trace tags and metadata

### LINKED ISSUE #2862: Problem with @ops.trace decorator - Mirascope sets mirascope.trace.tags and mirascope.trace.metadata, but Langfuse maps completely different attribute names.
### Description

I'm reporting an interoperability problem I encountered: tags and metadata supplied via the `@ops.trace` decorator are not available to downstream systems (Langfuse) as conventional span attributes; instead they are written under `mirascope.*` keys (and metadata is JSON-serialized), which prevents consumers from mapping them to first-class fields. My code:

``` python
@ops.trace(
    tags=["first_tag", "second_tag", ...],
    metadata={...}
)
@llm.call(...)
def func_to_call():
    ...
```

## What I observed

- Tags passed to the `@ops.trace` decorator become span attributes with Mirascope-specific keys (`mirascope.trace.tags`).
- Metadata passed to the decorator is serialized and stored under a single `mirascope.trace.metadata` attribute.
- Downstream tooling that expects conventional attribute names cannot find these values as first-class attributes and therefore cannot surface tags/user/session (or other metadata) natively.

Langfuse receives the OTLP spans and maps attributes to its data model using specific, different keys:

Langfuse field | Expected OTel attribute
-- | --
tags | langfuse.trace.tags (string array)
sessionId | langfuse.session.id
userId | langfuse.user.id

Since none of those keys existed on the span, Langfuse buried everything in `metadata.attributes.mirascope.trace.*`, which is not surfaced as first-class fields.

## Root cause

Mirascope sets `mirascope.trace.tags` and `mirascope.trace.metadata`, but Langfuse maps completely different attribute names:

What I want in Langfuse | Langfuse expects this OTel attribute | What Mirascope actually sets
-- | -- | --
tags | langfuse.trace.tags: string[] | mirascope.trace.tags → buried in metadata.attributes
sessionId | langfuse.session.id: string | mirascope.trace.metadata (JSON blob)
userId | langfuse.user.id: string | mirascope.trace.metadata (JSON blob)

## Source of the error (what I found during investigation)

- These lines in the span emission path illustrate the behavior that produces the problem:

``` python
# pseudocode excerpt from span attribute emission
if self.tags:
    attributes["mirascope.trace.tags"] = list(self.tags)
if self.metadata:
    attributes["mirascope.trace.metadata"] = json_dumps(self.metadata)
```

```TOML
Python 3.11, Mirascope 2.4.0, Windows 11.
```

mvanhorn:
## Summary

The `@ops.trace` decorator emits its `tags` and `metadata` as OTLP span attributes under Mirascope-specific keys (`mirascope.trace.tags` and a single JSON-serialized `mirascope.trace.metadata` blob). This PR additionally emits the equivalent OTel/Langfuse-conventional attribute keys so downstream observability backends can map them to first-class fields, while keeping the existing keys untouched for backward compatibility.

## Why this matters

Per #2862, Langfuse (and similar backends) map span attributes onto native trace fields using their own conventional keys (`langfuse.trace.tags`, `langfuse.session.id`, `langfuse.user.id`). Because Mirascope only emitted `mirascope.trace.*` keys, none of the tags/session/user values surfaced as native, filterable fields -- everything ended up buried under `metadata.attributes.mirascope.trace.*`. The maintainer confirmed this should be configurable and invited a PR.

## Changes

- Added a small module-level helper `_conventional_trace_attributes(tags, metadata)` in `python/mirascope/ops/_internal/traced_functions.py` so both `_span` context managers (`_BaseTracedFunction._span` and `_BaseTracedContextFunction._span`) stay in sync.
- Alongside the existing `mirascope.trace.*` attributes, spans now also emit:
  - `langfuse.trace.tags` -- the same tag list, when tags are present.
  - `langfuse.session.id` / `langfuse.user.id` -- promoted from the `session_id` / `user_id` metadata keys when present.
  - `langfuse.trace.metadata.<key>` -- remaining metadata entries flattened as individual scalar attributes so each can surface as a first-class, filterable field, instead of only the single JSON blob.
- The change is purely additive: `mirascope.trace.tags` and the `mirascope.trace.metadata` JSON blob are unchanged, so existing consumers keep working.

## Testing

- Added focused tests in `python/tests/ops/_internal/test_tracing.py` covering: conventional attribute emission (tags + promoted session/user + flattened metadata), tags-only (no metadata keys), metadata-without-session/user, and the `Context`-traced function variant. Existing full-snapshot tests were updated to include the new `langfuse.trace.tags` attribute.
- `cd python && uv run pytest tests/ops/_internal/test_tracing.py -q` -> 50 passed.
- `uv run ruff check` / `ruff format --check` / `pyright` on the changed files -> clean (0 errors).

These tests are deterministic and require no live LLM/provider call.

Fixes #2862

AI was used for assistance.

### DIFF
--- a/python/mirascope/ops/_internal/traced_functions.py
+++ b/python/mirascope/ops/_internal/traced_functions.py
@@ -39,35 +39,25 @@ FunctionT = TypeVar(
 )
 
 
-def _conventional_trace_attributes(
+def _otel_trace_attributes(
     tags: tuple[str, ...],
     metadata: dict[str, str],
 ) -> dict[str, AttributeValue]:
-    """Emit OTel/Langfuse-conventional span attributes for trace tags and metadata.
-
-    Backends like Langfuse map first-class fields from conventional attribute keys,
-    so we emit those alongside the existing ``mirascope.trace.*`` keys (which are
-    kept for backward compatibility).
-    """
+    """Emit plain OTel span attributes for trace tags and metadata."""
     attributes: dict[str, AttributeValue] = {}
     if tags:
-        attributes["langfuse.trace.tags"] = list(tags)
+        attributes["tags"] = list(tags)
     if metadata:
-        # Promote well-known fields to their conventional keys.
         session_id = metadata.get("session_id")
         if session_id is not None:
-            attributes["langfuse.session.id"] = session_id
+            attributes["session.id"] = session_id
         user_id = metadata.get("user_id")
         if user_id is not None:
-            attributes["langfuse.user.id"] = user_id
-        # Flatten remaining metadata entries under the Langfuse-conventional
-        # ``langfuse.trace.metadata.<key>`` prefix so backends surface each as a
-        # first-class, filterable field instead of only exposing the single JSON
-        # blob under ``mirascope.trace.metadata``.
+            attributes["user.id"] = user_id
         for key, value in metadata.items():
             if key in ("session_id", "user_id"):
                 continue
-            attributes[f"langfuse.trace.metadata.{key}"] = value
+            attributes[f"metadata.{key}"] = value
     return attributes
 
 
@@ -237,7 +227,7 @@ class _BaseTracedFunction(_BaseFunction[P, R, FunctionT]):
                 attributes["mirascope.trace.tags"] = list(self.tags)
             if self.metadata:
                 attributes["mirascope.trace.metadata"] = json_dumps(self.metadata)
-            attributes.update(_conventional_trace_attributes(self.tags, self.metadata))
+            attributes.update(_otel_trace_attributes(self.tags, self.metadata))
             span.set(**attributes)
             yield span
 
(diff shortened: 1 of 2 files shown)


mvanhorn:
Opened #2868 to address this. It keeps the existing `mirascope.trace.*` attributes for backward compatibility and additionally emits the Langfuse-conventional keys (`langfuse.trace.tags`, `langfuse.session.id`/`langfuse.user.id` promoted from metadata, and `langfuse.trace.metadata.<key>` for remaining entries) so they map to first-class fields. Added deterministic tests covering both the plain and `Context` traced-function paths.

willbakst:
Instead of emitting both, what do you think about making this something a user can configure? I think it would be better to be able to choose what you emit that to emit everything. Mirascope can be the default, and we could have something like `"langfuse"` as a built-in option?

willbakst:
Actually, looking at [their docs](https://langfuse.com/integrations/native/opentelemetry) I'm seeing we could just drop `mirascope` and `langfuse` altogether and just have e.g. `session.id` and it should work with Langfuse. No need to make anything specific, we should just make it pure OTel.

mvanhorn:
Agreed, pure OTel is the cleanest - reworked in 34f9ee7. The spans now emit plain `tags`, `session.id`, and `user.id` (promoted from metadata), with remaining metadata entries as `metadata.<key>`. All langfuse-prefixed keys are gone, and the pre-existing mirascope.trace.* attributes are untouched. Tests updated to assert no vendor-prefixed keys are emitted.

willbakst (approved):

PR #2871, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
## Summary

The TypeScript docs already show OpenAI-compatible providers via `registerProvider` and `baseURL`.

This PR adds a one-line note that the same OpenAI-compatible client pattern works with multi-model gateways when you are not self-hosting, using [DaoXE](https://daoxe.com) (`https://api.daoxe.com/v1`) as one concrete example.

Docs only — no runtime behavior changes.

## Test plan
- [ ] TypeScript README renders
- [ ] Existing OpenAI-compatible example unchanged
- [ ] Note is clearly optional

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/typescript/README.md
+++ b/typescript/README.md
@@ -262,6 +262,8 @@ const response = await llm.model("grok/grok-4-latest").call("Hello!");
 console.log(response.text())
 ```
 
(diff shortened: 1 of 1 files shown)


seven7763:
The note is generic boilerplate with nothing Mirascope-specific in it, and it shouldn't be taking up review attention. It should not stay open. Sorry to have filed it.

PR #2880, opened 2026-09-17
fix(llm): reject unknown call parameters

### LINKED ISSUE #2879: llm.call() silently accepts unknown keyword arguments (typos and the removed provider= are ignored)
### Summary

`@llm.call(...)` silently accepts **any** unknown keyword argument. A misspelled or outdated option is ignored rather than rejected, and when it does surface an error, the message points at the wrong thing.

### Reproduction

```python
from mirascope import llm

@llm.call(model="openai/gpt-4o-mini", provider="openai", nonsense_kwarg=123)
def f(q: str) -> str:
    return q
```

No error. `nonsense_kwarg=123` is plainly not a real option, and `provider=` is not in the signature either:

```python
>>> import inspect; from mirascope import llm
>>> list(inspect.signature(llm.call).parameters)
['model', 'tools', 'format', 'params']
```

### How I hit it, and why the error message compounds it

I wrote what older docs and most tutorials still show:

```python
@llm.call(provider="openai", model="gpt-4o-mini")
```

and got:

```text
ValueError: Invalid model_id format. Expected format: 'provider/model-name'
             (e.g., 'openai/gpt-4'). Got: 'gpt-4o-mini'
```

The message is accurate in isolation, but it sends you to fix the **model string** while quietly discarding the `provider=` you supplied. The actual situation — "`provider` is no longer a parameter; the provider moved into the model id" — is exactly what a rejected-kwarg error would have said in one line. Instead the two halves of the migration are reported separately, and only one of them is reported at all.

### Why it matters beyond the migration

The same silence covers ordinary typos:

```python
@llm.call(model="openai/gpt-4o-mini", temperatur=0.7)   # note the typo
```

This is accepted, and the call runs at the default temperature. That is a silent misconfiguration — the kind that reaches production and is only noticed as "the model behaves differently than we configured".

### Suggested fix

Reject unknown keyword arguments at decoration time, e.g.:

```python
raise TypeError(
    f"llm.call() got unexpected keyword argument(s): {', '.join(sorted(unknown))}. "
    "Did you mean to pass them in `params`? Note that `provider` was removed — "
    "use a prefixed model id such as 'openai/gpt-4o-mini'."
)
```

Special-casing `provider` in that message would make the v1→v2 migration self-explaining at the call site.

### Environment

`mirascope[openai]` as installed on 2026-09-14, Python 3.12, Windows. Found while pointing Mirascope at a local OpenAI-compatible endpoint; once the model id was correct everything worked — `base_url` resolves as `base_url or self.default_base_url` in `llm/providers/openai/completions/base_provider.py`, so an unset value falls through to the OpenAI SDK's `OPENAI_BASE_URL`, which is the behaviour I would hope for. No complaints there; this issue is only about the silent kwargs.

Oscar-Williams:
Fixes #2879

## Summary

Reject unknown `llm.call` keyword arguments at decoration time, with migration guidance for the removed `provider=` option. Preserve valid parameters and existing `Model` inputs. Add regression tests and document the error behavior.

## Tests

- Full suite before the documentation and additional tests: 2252 passed, 5 skipped; production code unchanged since that run.
- Updated decorator tests: 11 passed; 100% line and branch coverage of both affected modules.
- Ruff lint/format, Pyright, lockfile and `EXTRA_IMPORTS` validation passed (full Pyright run reported 5 optional MLX import warnings).


Oscar-Williams:
Friendly ping — this closes #2879 (`llm.call()` rejecting unknown keyword arguments instead of forwarding them). Open for a while with no feedback; let me know if anything needs changing, or close it if the behavior is intended.

